PR #3, opened 2016-07-28
Use a temp file for the signature

brianmcgillion:
On a multi user machine reuse of the /tmp/hash file
by multiple users will cause the build to fail as they
do not have the permission to replace a file created by
another user.

Signed-off-by: Brian McGillion <brian.mcgillion@intel.com>


chiache:
Thanks for reporting this issue.

In fact, the use of temp file is not necessary here. I've removed the lines in the master branch.

PR #21, opened 2016-12-05
Replace realpath with Python code

adombeck:
I replaced the use of `realpath` in `LibOS/shim/test/Makefile` with some equivalent Python code, because `realpath` is not installed by default in Ubuntu 14.04 *and* the package that is available in 14.04 does not support the used `--relative-to` option.

I also fixed the `ls.manifest.template` and minor stuff in the readme.


chiache:
Thanks for the patch!

adombeck:
You're welcome :)

adombeck:
Your usage of realpath suggests you are running Graphene on a system different from Ubuntu 14.04. Could you tell me what system you use? I would love to use something newer without breaking SGX (which is currently only supported on Ubuntu 14.04).

chiache:
I am currently developing on 16.04. However, the Graphene library OS is
supposed to be backward-compatible to older Ubuntu versions. We certainly
appreciate any bug reports on other Ubuntu.

Tsai, Chia-che (Jerry) <chiache.tsai@gmail.com>

adombeck:
Do the SGX SDK and driver work out of the box on 16.04? That would be great!

chiache:
It does: https://github.com/01org/linux-sgx/issues/11

Tsai, Chia-che (Jerry) <chiache.tsai@gmail.com>

adombeck:
Great, thanks!

PR #37, opened 2017-01-17
Fix ls.manifest.template and check for NULL after malloc

adombeck:


donporter:
Looks good to me.  Thanks!

adombeck:
I think you forgot to merge this.

PR #48, opened 2017-03-01
fix for large files within an SGX enclave

adilahmad17:
This pull request fixes the problem with Graphene for larger files (using SGX). Previously, since 'int' and 'unsigned long' were used, the enclave application could not create larger files. Also, the file size was also wrong if you tried opening a larger files (again because of the same issue).


chiache:
Moved to #136.

PR #49, opened 2017-03-23
Bugfix to prevent infinite hanging.

jovanbulck:
Latest commit #3cf3bd2 commented out memset in _DkVirtualMemoryAlloc, not sure why, but this leaves Graphene hanging infinitely on our SGX machine..


donporter:
I can reproduce the issue; thanks for reporting it.
I think we should eventually get rid of this memset, but not without figuring out what is using uninitialized memory first.

PR #72, opened 2017-08-03
Catching up with the Intel SGX driver and SDK

chiache:
The latest Intel SGX driver allows creating enclave at lower addresses.
Thus, we can deprecate most functionality of the Graphene-SGX driver,
except setting the CR4 register to enable WRFSGSBASE.

The new driver is backward-compatible to SGX driver 1.6 and 1.7, but
requires driver version to to be specified at compilation.

The new driver will interrupt the compilation of whole Graphene-SGX PAL
at the first time, to ask for the SGX driver path and version.

Also changed the interaction with AESMD through a new UNIX socket path
(but old domain name is till tried).


chiache:
Closed for now. Will create another pull request.

PR #73, opened 2017-08-03
Catching up with the Intel SGX driver and SDK

chiache:
The latest Intel SGX driver allows creating enclave at lower addresses.
Thus, we can deprecate most functionality of the Graphene-SGX driver,
except setting the CR4 register to enable WRFSGSBASE.

The new driver is backward-compatible to SGX driver 1.6 and 1.7, but
requires driver version to be specified at compilation.

The new driver will interrupt the compilation of whole Graphene-SGX PAL
at the first time, to ask for the SGX driver path and version.

Also changed the interaction with AESMD through a new UNIX socket path
(but old domain name is till tried).


donporter:
In general, this seems like an improvement.  It is good to use the newer versions, and to prune back the graphene wrapper.

My only concern is that it seems like we are losing the multi-page adding feature, and instead doing a page-at-a-time add to the enclave.  Are we sure this won't make things slower?  It might be good to have a couple of quick performance measurements.

Alternatively, could we do a PR for this feature in the Intel driver?

chiache:
I benchmarked the multi-page adding vs single-page adding, and the results have no difference at all. The fact is, adding a page into an enclave takes ~22-30 microsecond in average. The average overhead for making a system call is  ~0.05 microsecond, so mostly have no impact on enclave creation time.

My guess is that the enclave creation time would be more effectively optimized by reducing memory copy inside the kernel. Batching up page adding requests does not make too much difference.

Issue #53, opened 2017-04-04
Linux-SGX file_read() TOCTTOU vulnerability


adombeck:
I think the `file_read()` implementation in [Linux-SGX/db_files.c#L86](https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/db_files.c#L86) has a time-of-check-to-time-of-use vulnerability:

- In [L108](https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/db_files.c#L108) the target file is mapped to untrusted memory. 
- In `verify_trusted_file()` in [L114](https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/db_files.c#L114), the memory is verified using a CMAC. 
- If the verification was successful, the memory is copied in [L124](https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/db_files.c#L124).

Because the file is mapped to untrusted memory, the untrusted OS can modify the content of the memory after the verification and before the memory is copied.

Maybe I got this wrong, because I don't really understand your file verification process in the first place. First you try to verify the file with the SHA256 checksum in `load_trusted_file()` called by `file_open()`, generate a CMAC in `load_trusted_file()`, and then try to verify the CMAC in `file_read()`. Why the CMAC? 
Also, IIUC, `load_trusted_file()` operates on untrusted files. Why perform a check in `load_trusted_file()` in the first place, if the untrusted OS can modify the file afterwards?

adombeck:
> The correct behavior is: load_trusted_file() is supposed to check the SHA256 checksum of the file and cache the chesksums of each block to prevent rehashing the whole file. Each time when a block is loaded into the enclave, its checksum must always be checked.

I don't see any checksum caching. I think you just calculate the AES CMAC over the whole memory in both `file_open()` and `file_read()`. I'm not sure if this makes sense if the only reason is performance.

chiache:
1. We are no longer using AES CMAC to generate the file checksums. The initial checksum of a trusted file is a SHA256 digest. When the file is opened for the first time, each 16k block is loaded and hashed into a SHA512 digest. The SHA512 digests are cached in the enclave and checked against the content whenever reading from the file.

2. Besides the TOCTOU issue you described, I don't see why this strategy is insecure or meaningless. Would you care to elaborate why you think this design is problematic?

adombeck:
[...]

PR #79, opened 2017-08-10
Fix access being denied to trusted empty files

adombeck:
If a trusted file is empty the `ocall_map_untrusted` fails and as a result the access is denied. Commit 	e4f31f3 fixes this by skipping `ocall_map_untrusted` and the subsequent hash calculation if the file has size 0.

Note that `load_trusted_file` is still vulnerable to the TOCTOU attack I reported in #53.


chiache:
After merging with the master branch, this patch is no longer necessary.

Issue #80, opened 2017-08-13
os.fork() in Python causes segmentation fault


chiache:
This bug is reproduced with the "developing" branch, on both Ubuntu 16.04 and 14.04.

When running the following code:
```
import os
os.fork()
```
The program is terminated **quietly** by `SIGSEGV`. The segmentation fault can be observed in the log when turning on "debug_type=inline".  The log is as follows:
```
[    2] ---- shim_getpid() = 2
[    2] ---- shim_set_robust_list(0xba1de60,24) = 0
[    2] memory fault at 0x0 (IP = 0x4b610d)
[P4006] set tcb to 0xffd9a20
[    2] SIGSEGV handled
[P4006] ipc helper thread started
[    2] run signal handler 0xbac172e (11, 0xba7ccb0, 0xba7cd38)
[    2] killed by SIGSEGV
[    2] ipc broadcast: IPC_CLD_EXIT(1, 2, 0)
```

The segfault is trigger by accessing a variable called `_PyThreadState_Current` in python. In Graphene, this variable was not migrated when forking. The variable is 0 in the child thread as initially assigned in the Python binary.

donporter:
Looks like this issue can be closed.

PR #81, opened 2017-08-13
A bugfix for issue #80 (os.fork() sigfaults in Python)

chiache:
A bugfix for issue #80. The migration is not complete when a file-backed VMA is larger than the file size, but the file size is not page-aligned. In Linux, if the file size is smaller than the mapped memory, accessing the remaining memory that is not backed by the file will trigger a SIGBUS. However, it is fine to access the remaining memory within the last page that still overlaps with the file, and won't trigger a SIGBUS. This area is commonly used in ELF binary to store uninitialized, static variables. To improve fork latency, Graphene chooses to truncate the migration size as the file size, but missed the memory which belongs to the last file-backed page. The migration size should be aligned up to the page size.

### DIFF
--- a/LibOS/shim/src/bookkeep/shim_vma.c
+++ b/LibOS/shim/src/bookkeep/shim_vma.c
@@ -1124,6 +1124,21 @@ BEGIN_CP_FUNC(vma)
         bool protected = false;
 
         if (vma->file) {
+            /*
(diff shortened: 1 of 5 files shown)


donporter:
The change looks fine, but it might be good to put this explanation in a comment in the code too.

donporter:
Test this please

PR #83, opened 2017-09-03
Replace the atomics implementation

donporter:
This should be the last LGPL PR, and should be applied last.  I'll address the merge issue with master once earlier PRs are merged.

-  Consolidate the atomics into one file; some PAL handle definition changes needed to make this work.

- Use the atomics code from musl (MIT license) as a starting point. Keep the volatile-in-a-struct pattern.

- Drop atomic xchg; used only in a few ways in ipc helper state management that seem obviated by a lock (and perhaps belie a misunderstanding of atomic_read and atomic_set.

-  Drop non-binary semaphores from PAL - only binary semaphores are used in the code, and this simplifies the implementation considerably.

- Adjust the reference counting on PAL handles to start at 1; fix some initialization errors where things worked only because the initial value happened to be zero; drop atomic_dec_and_test_nonnegative.

- Further prune unused/dropped atomic APIs

-   Reimplement mutexes in PAL, to reduce complexity around packing two fields into one word (and remove need for xchg).


donporter:
Re-run with tip of this branch (12/7) - just to be sure that no intervening churn broke anything.
Only going to test SGX, as Jenkins now tests the Linux host.

---
  - [x] **Code version:** branch atomics
  - [x] **Ubuntu version:** 16.04.03 LTS
  - [x] **SGX SDK version:** 01org/linux-sgx@cdf563
  - [x] **SGX driver version:** 01org/linux-sgx-driver@e3e42d4
---
- [x] **PAL**
  - [x] **Linux+SGX host**
    - [x] **Compilation:** passed
    - [ ] **Regression tests:** 
---
- [x] **Library OS**
    - [x] **Compilation:** passed
    - [x] **Linux-SGX host**
      - [ ] **Regression tests:** 
---
- [ ] **Applications**
    - [x] **Linux-SGX host**
      - [ ] **LMBench:** lat_proc shell gets a segmentation fault.  lat_tcp, bw_tcp 127.0.0.1 gets permission denied messages in connect; bw_pipe has an occasional failure with the signal queue overflowing (same problems observed on master), an occasional vma assertion blows.  - generally, pretty flaky on SGX, but not necessarily different than master.
      - [ ] **GCC:** 
      - [ ] **Apache:**
      - [ ] **Python:**

PR #87, opened 2017-10-03
Fix the limitation on the number of trusted files

hanjae:
A bug fix for issue #9. Total key size of the trusted files was limited by CONFIG_MAX. This commit fixes it by dynamically allocate the config buffer.


hanjae:
Hi, I made changes you suggested.
I removed `size` field from `int get_config_entries()` , because now we can create a buffer which fits a key size.

donporter:
As far as I can tell, this code does address @chiache 's requests.  I would like to add one more request before we approve and merge: can you add a unit test for this case to PAL/regression that breaks on current master and succeeds on this branch?

neilbd:

hanjae:
I added a unit test which fails when a lot of `loader.env` items exist in a manifest.

hanjae:
---
  - [x] **Code version:** branch hanjae/master
  - [x] **Ubuntu version:** 16.04.3 LTS
  - [ ] **SGX SDK version:** 
  - [ ] **SGX driver version:** 
---
- [x] **PAL**
  - [x] **Linux host**
    - [x] **Compilation:** passed
    - [x] **Regression tests:** all passed
  - [x] **Linux+SGX host**
    - [x] **Compilation:** passed
    - [x] **Regression tests:** Same as master
---
- [ ] **Library OS**
  - [ ] **Compilation:** passed
  - [ ] **Linux host**
    - [ ] **Regression tests:**
    - [ ] **LTP tests:**
  - [ ] **Linux-SGX host**
    - [ ] **Regression tests:**
---
- [ ] **Applications**
  - [ ] **Linux host**
    - [ ] **LMBench:**
    - [ ] **GCC:**
    - [ ] **Lighttpd:**
    - [ ] **Apache:**
    - [ ] **Python:**
  - [ ] **Linux-SGX host**
    - [ ] **LMBench:**
    - [ ] **GCC:**
    - [ ] **Lighttpd:**
    - [ ] **Apache:**
    - [ ] **Python:**

donporter:
ok to test

donporter:
Hi @hanjae in general, this looks good, except that: 
1) The unit test you wrote generates a bus error on SGX, and 
2) This seems to also cause apache to get Bus Errors on SGX.  

Can you reproduce this issue?  If so, it might be nice if you could try to debug.  It is possible this is tickling an unrelated bug.  Perhaps @chiache can also take a look at this one when he gets a chance.

hanjae:
I got exactly the same error on SGX.
I changed first __alloca on cfgbuf to malloc because the bus error was raised on second __alloca. Which is line 141 of Pal/src/db_main.c
`cfgbuf = __alloca(sizeof(char) * CONFIG_MAX);`
I'm not sure why this occurred. Possibly memory allocated by __alloca was deallocated before function end.

Currently, apache is not working for me with master branch code even without SGX. So I couldn't test it.

donporter:
That seemed to fix it.  Perhaps the issue is that the alloca was overflowing the stack on SGX.

Anyway, I think the fix is good to merge now.  The remaining CI issues are known flakiness with timeouts on travis.

PR #88, opened 2017-10-13
Fix signess bug in the syscall dispatcher

mkow:
This code generates a SIGSEGV when user's code executes a syscall with a negative index (such syscall is invalid, but shouldn't cause a SIGSEGV).


donporter:
Seems reasonable to me.  Thanks for the fix.

PR #89, opened 2017-10-13
Check architecture in seccomp filters

mkow:
Without the checks a malicious application can circumvent the filter by
issuing a 32-bit syscall using either "int 0x80" instruction or by switching cs
segment to a 32-bit one and using "sysenter". This works because syscall
numbers differ between 32- and 64-bit x86.

### DIFF
--- a/Pal/src/security/Linux/filter.c
+++ b/Pal/src/security/Linux/filter.c
@@ -9,6 +9,7 @@ typedef __builtin_va_list __gnuc_va_list;
 #include "graphene-ipc.h"
 #include "graphene.h"
 
+#include <linux/audit.h>
 #include <linux/types.h>
 #include <linux/filter.h>
 #include <linux/seccomp.h>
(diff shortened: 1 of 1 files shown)


chiache:
Thanks for the patch! I believe this fix could actually be quite helpful. We are currently working on a new version of the SECCOMP filter and reference monitor (see https://github.com/oscarlab/graphene/blob/rm-module/Pal/src/security/Linux/filter.c). It would be a good idea to merge this patch.

donporter:
Do I understand this change correctly that it basically just squashes any 32-bit syscalls?  That seems reasonable, since we currently don't support 32-bit code on Graphene.

mkow:
> Do I understand this change correctly that it basically just squashes any 32-bit syscalls?

Yes, it kills processes which try to use them. You may change this behavior (if you prefer to) by modifying this line:
https://github.com/oscarlab/graphene/pull/89/files#diff-c2295390e22f1f1b8924d37febb025fcR38

@chiache: Yup, you should also patch the new code, it has the same problem.

chiache:
Merged into #139.

PR #91, opened 2017-10-17
Jenkins

neilbd:
Master Jenkinsfile for running tests for Lighttpd, Apache, GCC, Python, and LMBench


chiache:
The change in Pal/src/Makefile seems to be unnecessary but causing the build to fail. I suggest purging the changes in Pal/src/Makefile.

neilbd:
Fixed and working when I run a Jenkins build!

PR #113, opened 2017-11-21
Removed 'input' line when running the tests

neilbd:


donporter:
Am I right that we don't need this change anymore?

PR #116, opened 2017-11-24
Remove unnecessary locks used with xchg

mkow:


neilbd:

donporter:
Ok to test.

donporter:
Hi @mkow .  Thanks for the patch.  Good eye.  

I will note that we are (probably) planning to retire xchg - see PR #83 .  xchg is only used in a few places, and could be replaced with other atomics.

Let's see how the code review on 83 goes, but if xchg survives. I do think your patch makes sense.

mkow:
Sure, sounds like a good plan :)

donporter:
I think this PR has been obviated.  But thanks for the help anyway.

PR #118, opened 2017-11-25
Add lacking .PHONY to Makefiles

mkow:
There seems to be many build targets which lack .PHONY marker. Please check if I marked all of them correctly ;)


neilbd:

mkow:
btw, just out of curiosity: What's the purpose of the bot which commented above? You should get an email about all PRs regardless of it.

donporter:
Ok to test

donporter:
I had the same question.  It looks like a Jenkins security feature (we are working on migrating from Travis to Jenkins), so that an unknown developer can't effectively upload and run code on the Jenkins system.  A maintainer has to at least give it a cursory look and make sure it doesn't look like obvious malware.

Issue #129, opened 2017-12-04
Off-by-one in __hex2str()


omeg:
The last output character is omitted in most cases.
Pal\src\host\Linux-SGX\pal_linux.h

PR #130, opened 2017-12-04
Fix off-by-one error in __hex2str()

omeg:
Fixes oscarlab/graphene#129

### DIFF
--- a/Pal/src/host/Linux-SGX/pal_linux.h
+++ b/Pal/src/host/Linux-SGX/pal_linux.h
@@ -158,7 +158,7 @@ static inline __attribute__((always_inline))
 char * __hex2str(void * hex, int size)
 {
     static char * ch = "0123456789abcdef";
-    char * str = __alloca(size * 2);
+    char * str = __alloca(size * 2 + 1);
 
     for (int i = 0 ; i < size ; i++) {
         unsigned char h = ((unsigned char *) hex)[i];
@@ -166,7 +166,7 @@ char * __hex2str(void * hex, int size)
         str[i * 2 + 1] = ch[h % 16];
     }
 
-    str[size * 2 - 1] = 0;
+    str[size * 2] = 0;
     return str;
 }
 


neilbd:

neilbd:

Pal/src/host/Linux-SGX/pal_linux.h
```diff
@@ -158,15 +158,15 @@ static inline __attribute__((always_inline))
...
 
     for (int i = 0 ; i < size ; i++) {
         unsigned char h = ((unsigned char *) hex)[i];
         str[i * 2] = ch[h / 16];
         str[i * 2 + 1] = ch[h % 16];
     }
 
-    str[size * 2 - 1] = 0;
+    str[size * 2] = 0;
     return str;
```
mkow: Huh, this function returns a buffer from alloca...

donporter:
Ok to test

Pal/src/host/Linux-SGX/pal_linux.h
donporter: I agree that this is not good; this may be ok because it is inlined, but still seems that the buffer should be alloca'ed in the parent function to be safe.  This function is not widely used (just looks like a few debugging output cases), but I would still prefer to fix this while we are "in the pool".

donporter:
Good eye, @omeg .  This fix looks reasonable to me.  Bonus points for writing a unit test for any fixed issues (although there isn't as much infrastructure for helper/library functions, so I can take this on this time).

donporter:
Re-run with tip of this branch (12/5)
Only going to test SGX, as Jenkins now tests the Linux host.

---
  - [x] **Code version:** branch omeg/master
  - [x] **Ubuntu version:** 16.04.03 LTS
  - [x] **SGX SDK version:** 01org/linux-sgx@cdf563
  - [x] **SGX driver version:** 01org/linux-sgx-driver@e3e42d4
---
- [x] **PAL**
  - [x] **Linux+SGX host**
    - [ ] **Compilation:** passed
    - [ ] **Regression tests:** 
---
- [x] **Library OS**
    - [x] **Compilation:** passed
    - [x] **Linux-SGX host**
      - [ ] **Regression tests:** 
---
- [ ] **Applications**
    - [x] **Linux-SGX host**
      - [ ] **LMBench:** lat_proc shell gets a segmentation fault.  lat_tcp, bw_tcp 127.0.0.1 gets permission denied messages in connect; bw_pipe has an occasional failure with the signal queue overflowing (same problems observed on master), an occasional vma assertion blows.  - generally, pretty flaky on SGX, but not necessarily different than master.
      - [ ] **GCC:** 
      - [ ] **Apache:**
      - [ ] **Python:**

donporter:
Test this please

neilbd:
test this please

donporter:
Test this please

PR #131, opened 2017-12-04
Address a TOCTTOU vulnerability in SGX read/map

donporter:
The current SGX PAL checks a file based on the contents in an untrusted buffer.  If the hash passes, the contents are copied in.  This creates a risk of a TOCTTOU race, where the untrusted runtime changes the contents between the hash check and the copy.

This patch changes this behavior, where all contents are hashed from in-enclave buffers.  As an optimization, we try to use the final destination when possible.   Because the hashes are calculated in 16K chunks, any sub-blocks require hashing in (in-enclave) scratch space, along with the rest of the data in the 16K chunk, even though the rest of the data is not returned to the user.  This code is also complicated by the fact that for read(), the user provides a buffer, and for map() you are changing the virtual address space, and may have immovable neighbors.

This code does pass all of the unit tests on SGX, so I'm reasonably confident it is right, but some of the math is subtle enough that it could use more review, and suggestions for how to simplify this further are also welcome.


donporter:
Test this please

donporter:
Test this please

chiache:
Test this please

chiache:
Test this please

chiache:
Test this please

chiache:
Test this please.

Pal/src/host/Linux-SGX/enclave_framework.c
```diff
@@ -244,20 +244,49 @@ int load_trusted_file (PAL_HANDLE file, sgx_stub_t ** stubptr,
         goto failed;
 
     for (; offset < tf->size ; offset += TRUSTED_STUB_SIZE, s++) {
-        uint64_t mapping_size = tf->size - offset;
-        if (mapping_size > TRUSTED_STUB_SIZE)
-            mapping_size = TRUSTED_STUB_SIZE;
+        uint64_t mapping_size = MIN(tf->size - offset, TRUSTED_STUB_SIZE);
+        LIB_AESCMAC_CONTEXT aes_cmac;
+        ret = lib_AESCMACInit(&aes_cmac, (uint8_t *) &enclave_key,
```
donporter: Should the init/finalize be done outside of this loop?  If this is correct, a few more comments on how cmac checking and stubs work would not be amiss here.

Pal/src/host/Linux-SGX/enclave_framework.c
```diff
@@ -244,20 +244,49 @@ int load_trusted_file (PAL_HANDLE file, sgx_stub_t ** stubptr,
...
+
+        for (; chunk_offset < mapping_size; chunk_offset += FILE_CHUNK_SIZE) {
+            uint64_t chunk_size = MIN(mapping_size - chunk_offset, FILE_CHUNK_SIZE);
+
+            /* Any file content needs to be copied into the enclave before
+             * checking and re-hashing */
+            memcpy(small_chunk, umem + chunk_offset, chunk_size);
+
+            /* Update the file checksum */
+            ret = lib_SHA256Update(&sha, small_chunk, chunk_size);
```
donporter: Why are we doing both SHA and AES CMAC?

Pal/lib/crypto/adapters/mbedtls_adapter.c
```diff
@@ -101,6 +101,58 @@ int lib_AESCMAC(const uint8_t *key, uint64_t key_len, const uint8_t *input,
                                input, input_len, mac);
 }
 
+int lib_AESCMACInit(LIB_AESCMAC_CONTEXT * context,
```
donporter: It might be good to add a comment here that one lib_AESCMAC is equal to an init+update*+finish sequence.

Pal/src/host/Linux-SGX/enclave_framework.c
```diff
@@ -310,33 +342,142 @@ int load_trusted_file (PAL_HANDLE file, sgx_stub_t ** stubptr,
...
-    uint64_t checking = offset;
+    /* Check that the untrusted mapping is aligned to TRUSTED_STUB_SIZE
+     * and includes the range for copying into the buffer */
+    assert(umem_start % TRUSTED_STUB_SIZE == 0);
+    assert(offset >= umem_start && offset + size <= umem_end);
+
+    /* Start copying and checking at umem_start. The checked content may or
+     * may not be copied into the file content, depending on the offset of
+     * the content within the file. */
+    uint64_t checking = umem_start;
```
donporter: I think this code looks ok.  The only thing I might recommend is adding a comment reminding the reader about the layout of the stubs.  If I understand correctly, the way to think about this is a packed array of AES CMAC's of TRUSTED_STUB_SIZE units of the file, correct?

Pal/src/host/Linux-SGX/enclave_framework.c
chiache: For performance. Technically we can use any secure hash algorithm for the Merkel tree, but AES-CMAC is much faster and can use hardware acceleration.

donporter: But why both?

chiache: SHA256 is a more suitable algorithm for public-known secure digests, and is strong enough against brute-force. AES-CMAC needs a secret key unknown to the attackers, and such a secret key cannot be sent without a secured channel.

Overall the challenge here is that we need to officially define a hash algorithm as part of the manifest specification, but the hash algorithm used for the internal Merkel Trees can be anything. After experimenting with a few different algorithms, I found using AES-CMAC to be the fastest.

donporter: I see.  I think the main remaining to-do is a bit of documentation here on how files are signed, and then we should be good to merge.

PR #134, opened 2017-12-05
Hex2str clean up and unit testing

donporter:
This PR moves hex2str to the library directory, adds a unit test, and fixes the fact that it returns a buffer allocated via alloca().  It also causes the build to stop upon an error in the Pal code.


Pal/lib/hex.h
```diff
@@ -0,0 +1,52 @@
...
+ * size is the number of bytes pointed to by hex.
+ * str is the caller-provided buffer, len is the length of the buffer.
+ * The len must be at least (size * 2)+1.
+ * 
+ * Note that it does not normalize for endianness, and pads to the
+ * size the compiler things the string is.
+ */
+#include <assert.h>
+static inline __attribute__((always_inline))
+char * __hex2str(void * hex, int size, char *str, int len)
```
mkow: Why not `size_t` for array sizes?

Pal/regression/Hex.c
```diff
@@ -0,0 +1,16 @@
...
+
+#include "pal.h"
+#include "pal_debug.h"
+#include "api.h"
+#include "hex.h"
+
+int main() {
+  char buf[17];
+  uint32_t x = 0xefbeadde;
+  pal_printf("Hex test 1 is %s\n", hex2str(&x, buf, 17));
```
mkow: This is UB in C, this test will fail when porting to any non-little-endian architecture or when compiler optimize the code too aggressively. Fix:
```C
char x[] = {0xde, 0xad, 0xbe, 0xef};
```
Same for the code below.

Also: I'd be better to use `sizeof(buf)` instead of the hardcoded 17.

Pal/lib/hex.h
```diff
@@ -0,0 +1,52 @@
...
+        unsigned char h = ((unsigned char *) hex)[i];
+        str[i * 2] = ch[h / 16];
+        str[i * 2 + 1] = ch[h % 16];
+    }
+
+    str[size * 2] = 0;
+    return str;
+}
+
+#define hex2str(array, str, len) __hex2str(array, sizeof(*array), str, len)
```
mkow: Why this version prints only the first element of the array?
Also, IMO it'd be better to ensure that `array` is in fact an array, not a plain pointer.

Pal/lib/hex.h
mkow: Also: why this function is called hex2str? It converts binary data to hex, not a hex string to data.

Pal/lib/hex.h
mkow: `array` is a macro argument, so it should be wrapped it parentheses, otherwise a call like:
```C
hex2str(array_begin+5, out, out_len)
```
Will quietly lead to very unexpected results (i.e. calculating `sizeof(*array_begin+5)`).

Pal/lib/hex.h
donporter: Fair enough.  How about hex_str?

Pal/lib/hex.h
donporter: Agreed.  Most actual uses are for printing a hash or a mac, where it is some number of bytes.  I'm not sure how to check this without some fancier typed wrappers.

Pal/lib/hex.h
mkow: Hmm, `hex_str` doesn't sound clear to me ;) Maybe `bytes2hex` or `data2hex`?

Pal/regression/Hex.c
```diff
@@ -7,10 +7,10 @@
...
 
 int main() {
-  char buf[17];
-  uint32_t x = 0xefbeadde;
-  pal_printf("Hex test 1 is %s\n", hex2str(&x, buf, 17));
-  uint64_t y = 0xcdcdcdcdcdcdcdcd;
-  pal_printf("Hex test 2 is %s\n", hex2str(&y, buf, 17));
-  return 0;
+    char x[] = {0xde, 0xad, 0xbe, 0xef};
+    uint64_t y = 0xcdcdcdcdcdcdcdcd;
```
mkow: Could you please fix `y` also? ;) It will work, but it's still an UB according to the standard and may fail someday (e.g. after a compiler update).

Pal/lib/hex.h
mkow: The problem is that for `char arr[10]` `sizeof(*arr)` is 1, not 10, so it will print only the first byte. So, it should probably be `sizeof(arr)` + a static assert that `arr` is in fact an array. For a solution for the latter you can look e.g. here: https://stackoverflow.com/questions/16794900/validate-an-argument-is-array-type-in-c-c-pre-processing-macro-on-compile-time.

donporter: How is the latest push?

mkow: Almost good ;) You should wrap `arg` in both `IS_ARRAY` and `IS_INDEXABLE` in parentheses, otherwise a similar problem like the one mentioned before may happen.

PR #94, opened 2017-11-01
Fix a bug introduced by commit 3a40af399d1d85

donporter:
The handling of heap_min broke a PAL unit test.  The failure wasn't obvious, so I also changed the unit tests to fail in a more obvious way when something that should work stops working.

This is largely fixed by pulling in a few commits from @chiache 's development branch.

This change does make intermittent LTP/CI failures more apparent, but the correct thing to do is stabilize these tests.

PR #138, opened 2017-12-07
Automatic library searching for SGX signing

chiache:
This is a usability feature.

Introduce two SGX-specific manifest syntaxes:

- sgx.search_trusted_libs.[id] = [url]

Specifying where to search libraries as dependencies of the trusted
executable and libraries.

- sgx.trusted_libs.[id] = [url]

A library file which should be trusted, and any dependencies of
the library should be included in searching.

sgx.trusted_libs.xxx and sgx.trusted_files.xxx are similar, except
sgx.trusted_libs.xxx will be searched for libraries, whereas
sgx.trusted_files.xxx are treated as regular files.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/138)


donporter:
Add to whitelist

donporter:
CI is failing a fairly often because of the recvfrom01 test added in #145 .  Do you think we should move it to the FLAKY list, or do you want to debug it?

preadv02 is also failing intermittently.  This one is my fault (#94).  But I'll ask the same question.

LibOS/shim/test/apps/lighttpd/lighttpd.manifest.template
```diff
@@ -37,19 +37,23 @@ sys.stack.size = 256K
 sys.brk.size = 4M
 glibc.heap_size = 16M
 
-sgx.thread_num = 28
+sgx.thread_num = 32
 
+<<<<<<< HEAD
```
donporter: Merge booboo?

chiache:
@donporter I am not sure I interpreted your comment correctly. I guess what you mean is with automatic file searching, users no longer control the list of trusted files and may be misled into loading the wrong libraries. The intention of this feature is to save users the step of manually finding out dependencies (and they are likely to be using ldd anyway). This feature does not try to obscure the signing process. Users can always review the trusted file list in *.manifest.sgx files after the signing.

I assume you are also asking the searching order among these directories. For that, I think right now the behavior is a little ambiguous. We can go with the alphabetical order of the keys but adding an extra syntax would improve the clarity.

I do agree that lack of documentation is a problem. My original plan is to add the explanation in the wiki after the patch got merged. I can also add the same explanation in the signing script.

chiache:
@donporter After rereading your comment, I think you are talking about automatically adding these default library locations (e.g., /usr/lib, /lib) to the searchable directories. That does make the search results uncertain. I think the right rules is that the signing scripts should only search from directories listed in the manifest. If users want to search in /usr/lib, for example, they must add the path to the manifest.

donporter:
Right.  To clarify, there are two points here:
1) I think users should specify the directories they trust as part of signing, like in LD_LIBRARY_PATH.
2) Documentation in the wiki and script about what the expected behavior is here.

I think we are on the same page about what needs to be done now, though.

chiache:
Close this PR for now. Need to discuss new features.

PR #145, opened 2017-12-13
[Urgent!!] Bugs caused by merging #87

chiache:
- Fix a severe bug caused by merging #87. The patch does not check the return values from get_config_entries_size(), while the return values can easily be negative (-PAL_ERROR_INVAL). Any usage of this function should carefully check the return values and also pass the size into get_config_entries() to prevent buffer overflow.

- Fix a minor bug in handling IPC disconnection. The callback ipc_child_exit() never receives a "term_signal" parameter. The callback is called when the pipe to a child process unexpectedly terminates, which can only mean the child process has crashed. I believe the proper signal to send is SIGKILL.

- Remove a compilation warning caused by type mismatch of __malloc().

**[Notice!]** Adding three bugfixes:

- Fixing socket implementation to accept ANY address (e.g., 0.0.0.0:0)

- Enabling LD_PRELOAD in ld.so (previously removed by glibc-2.19.patch)

- Disable IO buffering in LTP tests; several LTP programs panic in the middle of execution and do not print out passed tests. This fix is based on stdbuf in coreutils: Using LD_PRELOAD to load libstdbuf.so, with _STDBUF_O=L given in the environment variables.

### DIFF
--- a/LibOS/glibc-2.19.patch
+++ b/LibOS/glibc-2.19.patch
@@ -664,41 +664,10 @@ index 6dcbabc..c87c773 100644
    LIBC_PROBE (init_start, 2, LM_ID_BASE, r);
  
    /* Auditing checkpoint: we are ready to signal that the initial map
-@@ -1605,140 +1654,11 @@ ERROR: ld.so: object '%s' cannot be loaded as audit interface: %s; ignored.\n",
+@@ -1636,89 +1654,6 @@ ERROR: ld.so: object '%s' cannot be loaded as audit interface: %s; ignored.\n",
  	}
      }
  
--  /* We have two ways to specify objects to preload: via environment
--     variable and via the file /etc/ld.so.preload.  The latter can also
--     be used when security is enabled.  */
--  assert (*first_preload == NULL);
--  struct link_map **preloads = NULL;
--  unsigned int npreloads = 0;
--
--  if (__builtin_expect (preloadlist != NULL, 0))
--    {
--      /* The LD_PRELOAD environment variable gives list of libraries
--	 separated by white space or colons that are loaded before the
--	 executable's dependencies and prepended to the global scope
--	 list.  If the binary is running setuid all elements
--	 containing a '/' are ignored since it is insecure.  */
--      char *list = strdupa (preloadlist);
--      char *p;
--
--      HP_TIMING_NOW (start);
--
--      /* Prevent optimizing strsep.  Speed is not important here.  */
--      while ((p = (strsep) (&list, " :")) != NULL)
--	if (p[0] != '\0'
--	    && (__builtin_expect (! INTUSE(__libc_enable_secure), 1)
--		|| strchr (p, '/') == NULL))
--	  npreloads += do_preload (p, main_map, "LD_PRELOAD");
--
--      HP_TIMING_NOW (stop);
--      HP_TIMING_DIFF (diff, start, stop);
--      HP_TIMING_ACCUM_NT (load_time, diff);
--    }
--
 -  /* There usually is no ld.so.preload file, it should only be used
 -     for emergencies and testing.  So the open call etc should usually
 -     fail.  Using access() on a non-existing file is faster than using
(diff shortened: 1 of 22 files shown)


chiache:
Re-run with tip of this branch (11/24)

---
  - [x] **Code version:** branch config-bugfixes
  - [x] **Linux kernel version:** 4.10.0-37-generic
  - [x] **Ubuntu version:** 16.04.02 LTS
  - [x] **SGX SDK version:** 01org/linux-sgx@8383b8c
  - [x] **SGX driver version:** 01org/linux-sgx-driver@3abcf82
---
- [x] **PAL**
  - [x] **Linux host**
    - [x] **Compilation:** passed
    - [x] **Regression tests:** passed
  - [x] **Linux+SGX host**
    - [x] **Compilation:** passed
    - [x] **Regression tests:** passed
---
- [ ] **Library OS**
    - [ ] **Compilation:** passed
    - [ ] **Linux host** 
      - [ ] **Regression tests:** 
      - [x] **LTP tests:** passed
    - [ ] **Linux-SGX host**
      - [ ] **Regression tests:** 
---
- [ ] **Applications**
    - [x] **Linux host**
      - [ ] **LMBench:** 
      - [ ] **GCC:** 
      - [ ] **Lighttpd:** 
      - [ ] **Apache:** 
      - [ ] **Python:** 
    - [ ] **Linux-SGX host**
      - [ ] **LMBench:** 
      - [ ] **GCC:** 
      - [ ] **Apache:**
      - [ ] **Python:**

chiache:
The regression test is still pretty flaky, especially this kill03 test case. I cannot figure out how it happens or how to reproduce it.

chiache:
The LTP tests passed now.

LibOS/glibc-2.19.patch
```diff
@@ -664,41 +664,10 @@ index 6dcbabc..c87c773 100644
    LIBC_PROBE (init_start, 2, LM_ID_BASE, r);
  
    /* Auditing checkpoint: we are ready to signal that the initial map
-@@ -1605,140 +1654,11 @@ ERROR: ld.so: object '%s' cannot be loaded as audit interface: %s; ignored.\n",
```
chiache: This patch used to disable LD_PRELOAD in glibc. There is no strong reason that we should do this, and having LD_PRELOAD can be helpful in many cases.

LibOS/shim/include/shim_ipc.h
```diff
@@ -600,9 +600,9 @@ int do_ipc_duplex (struct shim_ipc_msg_obj * msg,
                    void * private_data);
 
 void ipc_parent_exit  (struct shim_ipc_port * port, IDTYPE vmid,
-                       unsigned int exitcode, unsigned int term_signal);
+                       unsigned int exitcode);
 void ipc_child_exit   (struct shim_ipc_port * port, IDTYPE vmid,
-                       unsigned int exitcode, unsigned int term_signal);
+                       unsigned int exitcode);
```
chiache: These IPC callbacks have a specific prototype, and "term_signal" is never passed into the functions.

Pal/src/host/FreeBSD/db_sockets.c
```diff
@@ -279,54 +279,56 @@ PAL_HANDLE socket_create_handle (int type, int fd, int options,
     return hdl;
 }
```
chiache: ANY address (0.0.0.0:0) was not accepted in the PAL before. But now we are allowing it to enable some use cases (especially for LTP).

donporter:
Test this please

donporter:
The Jenkins job looks like it passed, so I am not sure why the status didn't update.  Kicking off one more run for funzies.

LibOS/shim/src/shim_async.c
```diff
@@ -108,7 +108,7 @@ int install_async_event (PAL_HANDLE object, unsigned long time,
     
     unlock(async_helper_lock);
```
donporter: This line should go away - otherwise we are double-unlocking the async_helper_lock at line 114.  Perhaps this is a merge error?

Seems likely.

Pal/lib/graphene/config.c
```diff
@@ -147,7 +147,7 @@ int get_config (struct config_store * store, const char * key,
 }
 
 int get_config_entries (struct config_store * store, const char * key,
-                        char * key_buf)
+                        char * key_buf, int key_bufsize)
```
donporter: Also size_t

Pal/src/db_main.c
```diff
@@ -97,10 +98,13 @@ static void read_environments (const char *** envpp)
     if (!pal_state.root_config)
         return;
 
-    cfgbuf = malloc(get_config_entries_size(pal_state.root_config,
-                                            "loader.env"));
-    nsetenvs = get_config_entries(pal_state.root_config, "loader.env",
-                                  cfgbuf);
+    int cfgsize = get_config_entries_size(store, "loader.env");
```
donporter: size_t

Pal/src/host/FreeBSD/db_sockets.c
```diff
@@ -279,54 +279,56 @@ PAL_HANDLE socket_create_handle (int type, int fd, int options,
...
 
-static int check_zero (void * mem, int size)
+#if ALLOW_BIND_ANY == 0
+static bool check_zero (void * mem, int size)
```
donporter: size_t

LibOS/shim/src/shim_async.c
chiache: You are right, it is a merge error. I will fix it right away.

Pal/src/host/Linux-SGX/enclave_framework.c
```diff
@@ -433,7 +433,9 @@ static int init_trusted_file (const char * key, const char * uri)
 
 int init_trusted_files (void)
 {
-    char *cfgbuf;
+    struct config_store * store = pal_state.root_config;
+    char * cfgbuf;
+    int cfgsize, nuris;
```
donporter: size_t

Pal/src/host/Linux-SGX/enclave_framework.c
```diff
@@ -443,8 +445,7 @@ int init_trusted_files (void)
     }
 
     cfgbuf = __alloca(CONFIG_MAX);
-    int len = get_config(pal_state.root_config, "loader.preload",
-                         cfgbuf, CONFIG_MAX);
+    int len = get_config(store, "loader.preload", cfgbuf, CONFIG_MAX);
```
donporter: Yet more size_t, you get the idea.

chiache:
I have a small doubt about changing all `int`s to `size_t`. To be honest, the return values of `get_config_entries_size()` and `get_entry()` should be `ssize_t`, but `ssize_t` is not a standard type for C. You can only get the declaration of `ssize_t` from including "`sys/types.h`", but that's something I have tried to avoid.

It seems like the standard C type for signed size values is `ptrdiff_t`.

donporter:
I'm ok with ssize_t, including defining it ourselves in our own header if you prefer.  ptrdiff is less common, and less intuitive.

PR #139, opened 2017-12-07
Graphene Sandboxing Design and Static Binary Support

chiache:
This patch primarily contains three changes:
- Complete rework of the Graphene reference monitor (Linux only) as a kernel module
- Improvement of Secommp filtering. Allowing child processes to inherit PC-related rules by loading PAL at the same address as inside of the parent process. 
- Supporting static binaries and system calls from inline assembly code.

PR #150, opened 2017-12-16
Completely deprecating customized Linux kernel

chiache:
Merge this after #139.

With the new reference monitor module design, we no longer need a customized Linux kernel.


chiache:
No need to test because this patch contains no code change.

donporter:
No objection in principle to this idea, but I think #139 is a bit stalled.  Poke me again once 139 lands.

chiache:
Closing this PR for now because it needs more work.

PR #164, opened 2018-01-13
Fixed the ssl config file

zavalyshyn:
With the current lighttpd-ssl.conf file the Lighttpd will start but will not listen on 443 port. All connections to this server will thus be refused. 
I've added a config parameter that specifies a port for a Lighttpd server to listen on and few additional parameters for encrypted connection.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/164)


zavalyshyn:
This is all for the case when we start lighttpd app with a command:
`make start-ssl-graphene-server`

donporter:
Ok to test

donporter:
Thanks for the fixes, and sorry for the delay!  My main questions about this patch:

1) Are there any licensing issues in including the app code here?  It might make sense to put this as a submodule.  Thoughts?
2) Assuming this works otherwise, I'd like to incorporate this app into our routine CI.  Would you just run apache bench (as with baseline lighttpd), but pointed to port 443?

Thanks!

LibOS/shim/test/apps/lighttpd-python/Makefile
```diff
@@ -0,0 +1,146 @@
...
+target = build-lighttpd $(conf_files) test-data
+exec_target = $(manifests)
+clean-extra = clean-others
+
+extra_rules = -e 's:\$$(HOST):$(HOST):g' -e 's:\$$(PORT):$(PORT):g'
+
+level = ../../
+include ../../Makefile
+
+build-lighttpd: build/sbin/lighttpd
```
mkow: Could you please add `.PHONY` to all targets which need it?

zavalyshyn:
Hi,

@donporter, I'm not quite sure about the licensing issues. The code of lighttpd v1.4, which Graphene is currently using, is available on Github under BSD 3-Clause license. As far as I understand, if we keep the copyright notice file, there should not be any problem. 

For the CI, that's a valid point. I will modify my code and make a new pull request with proper benchmarking for HTTPS too. 

@mkow, that "holod" thing is something that was not supposed to go to the pull request :) Sorry. I will make a new pull request without this. I will also add .PHONY to all targets as requested.

donporter:
BSD license is fine with me.

I think the CI part should be as simple as copy/pasting the current lighttpd part of the script and changing it slightly to test port 443 and use the proposed working directory.

donporter:
Should dhparam.pem be statically put in the code repo, or generated by the makefile if it doesn't exist?

zavalyshyn:
I modified the makefile to generate it automatically during compilation. I
also changed its "hardness" to 2048 instead of 4096, cause it takes way too
much time. 2048 is still secure.

zavalyshyn:
More commits coming

zavalyshyn:
@mkow I'm not quite sure why Jenkins fails on .phony commit. Did I miss something?

mkow:
I don't have access to Jenkins, but it's probably due to general instability of the tests and Graphene itself. Your .PHONY declarations look good. It may be also caused by the two previous commits, but they look fine to me.

@donporter Could you please check what's failing?

zavalyshyn:
Apparently, Apache benchmark doesn't support self-signed SSL certificates. And we do you use one. All the attempts to connect to a server with a self-signed ssl certificate end up with a "SSL handshake failed" error. I'm still checking if there are some workarounds here.

zavalyshyn:
@donporter So, there seems to be no way to use apache bench (ab) for testing HTTPS server with self-signed certificate, as we do with HTTP server.  As I said previously, you get the SSL error in this case. 

Two possible things we can do:
- leave HTTPS part of the lighttpd server without regular CI check
- use alternative benchmarks ([like these](https://stackoverflow.com/questions/189993/how-do-i-fix-ssl-handshake-failed-with-apachebench)) for this specific type of connection

donporter:
Hi @zavalyshyn - sorry for the delay.  I'm ok with using an alternative benchmark or test.  I don't have any specific allegiance to ab.  The simple script on stackoverflow seems fine to me - we just want to make sure that this feature continues working with future changes.

Would you still be willing to add this change?

zavalyshyn:
Yeah, I'll take a look at it.

zavalyshyn:
@mkow done

donporter:
I think we should go ahead and (at long last), land this change.  I'd still like a unit test, but I'd be ok doing this myself in a separate PR.

@zavalyshyn would you be willing to squash this into a single commit (and drop the merge commits), and then rebase to current master?   If you aren't comfortable with this, I can do this on a separate branch. 

Thanks!

zavalyshyn:
@donporter my current repo branch is 16 commits ahead of oscarlab:master. As I understand there is no need to do any rebase from my side. These 16 commits are from this pull request. 

I'm not quite sure what I need to do in this case, so if you have time please do it from your side. It seems like you just need to squash all my commits into one and merge this PR with the master branch.

About the CI test of HTTPS, I'm sorry for not finishing this. It turned out not that straightforward like the HTTP one.

Issue #82, opened 2017-08-30
Cannot allow the files in current directory to be accessed by Graphene


rainfld:
Taken python in Graphene-SGX for example. The original version of running hello world is as follows:

SGX=1 ./python.manifest.sgx scripts/helloworld.py

As scripts was added as an allowed subfolder: "**sgx.allowed_files.scripts = file:scripts**"

However, if I tried "SGX=1 ./python.manifest.sgx helloworld.py", even I put the current directory in manifest as "sgx.allowed_files.curr = $(PWD)", it cannot work.

The reason is that Graphene is judge if the file can be access or not by matching the prefix of allowed folder paths. However, for current directory, the file does not have path, therefore, prefix matching always return false.

pwmarcz:
As discussed in #176, this is intentionally not allowed, and `sgx.allow_file_creation` solves the reporter's use-case. Closing.

PR #176, opened 2018-02-09
Allowing application to access the files under current directory without specifying filenames in "allowed_files"

rainfld:
The PR addresses the issue reported at https://github.com/oscarlab/graphene/issues/82

It is to add a feature that allows the application can access the files under the current working directory without specifying the file name with “allowed_files”. This is to handle the cases where some of the applications are dumping logs and other data to the newly generated files under the current directory. There is no way to know their names before the application is running. 

The user could allow the application to access all the files under the current directory by adding the following line in the manifest 

file:sgx.allowed_files.conf = file:.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/176)

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_framework.c
+++ b/Pal/src/host/Linux-SGX/enclave_framework.c
@@ -128,6 +128,7 @@ int load_trusted_file (PAL_HANDLE file, sgx_stub_t ** stubptr,
     char uri[URI_MAX];
     char normpath[URI_MAX];
     int ret, fd = file->file.fd, uri_len, len;
+    bool file_in_cwd = 0;
 
     if (!(HANDLE_HDR(file)->flags & RFD(0))) 
         return -PAL_ERROR_DENIED;
@@ -149,6 +150,9 @@ int load_trusted_file (PAL_HANDLE file, sgx_stub_t ** stubptr,
     len = get_norm_path(uri + 5, normpath + 5, 0, URI_MAX);
     uri_len = len + 5;
 
+    if (uri_len > 5 && normpath[5] != '/')
+	file_in_cwd = 1;
+
     _DkSpinLock(&trusted_file_lock);
 
     listp_for_each_entry(tmp, &trusted_file_list, list) {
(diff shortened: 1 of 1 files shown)


chiache:
Thanks for the patch. The patch makes sense to me and could have been a very useful feature. However, I am slightly worried about the security implication of bypassing security checks in CWD.

The reason that I have hesitated to allow "`file:`" as a target of `sgx.allowed_files` is that this rule can open the door to accessing literally every files under the CWD, even one that mismatches with one of the known checksums. Besides, a lazy user may just use "`file:`" to bypass all checks, even though they shouldn't. Of course, in that case, the user is responsible for his/her wrongdoing, but I wonder if we should open this Pandora's box at all.

Please feel free to dispute the points. I would prefer a full discussion before accepting this patch.

donporter:
Ok to test

donporter:
So is what you actually need the ability to create a file under a given directory from SGX?  And this is not currently allowed, right?

I'd rather apply least privilege principle here.  Can we just have an option like, allow creation of a new allowed file that does not exist?

chiache:
That's not the problem. Creating new files in a directory whitelisted as "allowed files" is already permitted. What's in question is whether we should allow assigning the whole CWD (a.k.a., "file:") to a whitelisting rule, similar to a DMZ.

donporter:
Right, but the motivating problem is creation of files for which it is hard to predict the name.

I think the solution proposed is overly permissive; I'm wondering if we can address the problem with a more precise rule.

dimakuv:
I agree with @chiache that allowing CWD is plain dangerous.

@rainfld, what is the particular application that creates files under the current directory? Are there options to dump new files under a subdir? Well-behaved applications usually allow to change the default directory for logs/dumps.

rainfld:
The motivation of this PR has already been address in the PR https://github.com/oscarlab/graphene/pull/182

I think we can close this unless there is a usage requiring this change.

rainfld:
The problem this PR tries to address has already been resolved by https://github.com/oscarlab/graphene/pull/182

Close this for now. If we see any potential usages for this feature, re-open in future.

PR #180, opened 2018-02-15
Fixes for the allocator/VMA code

### LINKED ISSUE #173: Graphene crashes when there are too many entries in a manifest
My manifest has about 2k lines in a form of `sgx.trusted_files.asdf = xyz`. Graphene crashes during startup when using it in the SGX mode.

This seems to be a few different bugs:
* `enclave_framework.c` uses `alloca` for storing manifest data during parsing (`alloca` should never be used for any not-very-small data, especially when the size is user-controlled).
* After fixing the code to use `malloc`, everything literally explodes. There are many crashes related to `slab_mgr::free_list` corruption. Although all of them are indeterministic, I've never managed to run my code without a crash related to such a corruption.

In the second problem, it seems that an uninitialized object was inserted into the list (the `next` and `prev` pointers are filled with `0xa5a5a5a5a5a5a5a5`, which is the value with which all heap buffers are memsetted in your debug mode).
The odd thing here is that there seems to be no list operation which did that (I added many asserts there to verify that), so this bug may be related to the VMA overlapping bug (#159): another buffer was allocated so that it overlapped with something already placed in the list, then the debug memset had overwritten the list node.

I'm debugging this issue right now and will post any new findings.

mkow:
Small fixups:
* Replaced `alloca` with `malloc` for potentially large buffers in `enclave_framework.c` (partially fixes #173).
* Many minor code readability improvements (e.g. removed a `while` loop implemented using `goto`...).
* Fixed int overflow in `calloc`.
* Fixed a few resource leaks (including two unreleased locks).
* Added `__attribute__((unused))` to a variable used in an ifdefed code to remove a compiler warning.
* Removed `print_vma_hash`, this function looks like it never worked (it doesn't even print anything?).
* Fixed many typos.

### DIFF
--- a/LibOS/shim/include/shim_checkpoint.h
+++ b/LibOS/shim/include/shim_checkpoint.h
@@ -50,7 +50,7 @@ extern char __migratable_end;
 /* TSAI 7/11/2012:
    The checkpoint scheme we are expecting is to support an easy syntax to
    implement migration procedure. A migration procedure can be written
-   in teh following syntax:
+   in the following syntax:
 
    BEGIN_CP_DEFINITION(exec)
    {
@@ -85,7 +85,7 @@ struct shim_cp_entry
     union
     {
         ptr_t cp_val;   /* interger value */
-        /* orignally there is a pointer, now we don't need them */
+        /* originally there is a pointer, now we don't need them */
     } cp_un;
 };
 
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -657,9 +657,11 @@ static inline int __ref_dec (REFTYPE * ref)
     register int _c;
     do {
         _c = atomic_read(ref);
-        assert(_c > 0);
-        if (!_c)
+        if (!_c) {
+            debug("Fail: Trying to drop reference count below 0\n");
+            bug();
             return 0;
+        }
     } while (atomic_cmpxchg(ref, _c, _c - 1) != _c);
     return _c - 1;
 }
@@ -708,8 +710,8 @@ extern void * migrated_memory_start;
 extern void * migrated_memory_end;
 
 #define MEMORY_MIGRATED(mem)                                    \
-        ((void *) mem >= migrated_memory_start &&               \
-         (void *) mem < migrated_memory_end)
+        ((void *) (mem) >= migrated_memory_start &&             \
+         (void *) (mem) < migrated_memory_end)
 
 extern void * __load_address, * __load_address_end;
 extern void * __code_address, * __code_address_end;
--- a/LibOS/shim/include/shim_vma.h
+++ b/LibOS/shim/include/shim_vma.h
@@ -118,9 +118,6 @@ void unmap_all_vmas (void);
 /* Debugging */
 void debug_print_vma_list (void);
 
-void print_vma_hash (struct shim_vma * vma, void * addr, uint64_t len,
-                     bool force_protect);
-
 /* Constants */
 extern unsigned long mem_max_npages;
 extern unsigned long brk_max_size;
--- a/LibOS/shim/include/shim_vma.h
+++ b/LibOS/shim/include/shim_vma.h
@@ -94,15 +94,18 @@ int bkeep_mmap (void * addr, uint64_t length, int prot, int flags,
                 struct shim_handle * file, uint64_t offset, const char * comment);
 
 /* Bookkeeping munmap() system call */
-int bkeep_munmap (void * addr, uint64_t length, const int * flags);
+int bkeep_munmap (void * addr, uint64_t length, int flags);
 
 /* Bookkeeping mprotect() system call */
-int bkeep_mprotect (void * addr, uint64_t length, int prot, const int * flags);
+int bkeep_mprotect (void * addr, uint64_t length, int prot, int flags);
 
 /* Get vma bookkeeping handle */
 void get_vma (struct shim_vma * vma);
 void put_vma (struct shim_vma * vma);
 
+/* Returns 0 on success, -E* on failure.
+   Calls `get_vma` on the result before returning it.
+*/
 int lookup_supervma (const void * addr, uint64_t len, struct shim_vma ** vma);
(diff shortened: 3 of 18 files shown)


mkow:
Also, your bot seems to be broken ;) (see the duplicated comment above)

mkow:
Two questions:
* Why `get_reserved_pages`: https://github.com/oscarlab/graphene/blob/2e9438f71c487c5a3ae2e7b3ca1c2d957dee6d3e/Pal/src/host/Linux-SGX/enclave_pages.c#L71 has such a strange semantics? It seems that it may allocate memory on a different address than the one given as an argument. This impacts `DkVirtualMemoryAlloc` behavior, which as a result may break the assert in the `malloc` implementation:
https://github.com/oscarlab/graphene/blob/181f9ea988698f9ffbb956728747b2ce2f554279/LibOS/shim/src/shim_malloc.c#L166-L169
* Why there is a loop in `cp_alloc`? Why would anyone want to try allocation again when previous allocation failed? See https://github.com/oscarlab/graphene/pull/180/files#diff-469674ea45a019c9057ec00c49fb6605.

chiache:
Answers to your questions:

(1) I think it is correct that `get_reserved_pages()` should return an address exactly as requested or simply fail.  When `DkVirtualMemoryAlloc `is given an address, it behaviors as `mmap()` with `MAP_FIXED`.

(2) This is due to the potential race condition between `get_unmapped_vma_for_cp()` and `DkVirtualMemoryAlloc()`. If `DkVirtualMemoryAlloc()` fails, it is reasonable to try again. I agree it is not ideal to loop indefinitely. Maybe we can set a bound to this loop.

mkow:
Should I squash the last fixup commit (67e4ca1) before merging, or you prefer to do this on your own?

1. I'll try to fix `get_reserved_pages` then (in another PR).
2. There was some retry limit before, but it was broken (the iteration count wasn't incremented at all). Should I implement this limit? How many iterations should use?
btw. I'm a bit skeptic about retrying allocations here and in general, I'm not sure if this makes any sense. The loop will probably execute fast enough to not see changes from other threads.

chiache:
Please squash all commits before merging.

I think retrying allocation (even indefinitely) is useful because the allocator will try to allocate at a different address. It doesn't matter if another thread has raced with the allocation. In most cases, the retry wouldn't take more than two or three times. If we don't allow retries, a fork() will inevitably have to return -ENOMEM even though that's not really the case.

mkow:
> Please squash all commits before merging.
You mean all of them into one commit, yes? (not only the last fixup)

Regarding retrying:
So, this is only because the address is randomized? Should I add a comment to the code explaining it?

donporter:
Ok to test

donporter:
On both questions, please add comments documenting the state of things.  I can squash when we do the merge.

In general, I would much rather have some sensible allocation protocol that won't have race conditions, than to have this retry loop.  But I'm ok with documenting what is going on here for now and perhaps dealing with this in a separate PR.

Are there any particular test cases or unit tests this fix addresses?

mkow:
Done.

> In general, I would much rather have some sensible allocation protocol that won't have race conditions, than to have this retry loop.

Totally agree. I added a TODO about this.

> Are there any particular test cases or unit tests this fix addresses?

I'm not aware of any. It fixes running Blender in an enclave though (only partially, more fixes coming soon). After fixing it entirely it may be a good idea to add headless Blender to the tests.

donporter:
Test this please

donporter:
So the CI failures seem semi-legit.  There is some variance, but in both the normal and debug (non-SGX) runs, the following LTP tests fail (some consistently, some only sometimes):

munmap01
munmap02
mmap001
mprotect04

I'm not sure whether these are real regressions, or perhaps just other latent bugs being exposed by these fixes.

But would you mind taking a look and seeing if you can repro/debug them?

Under LTP, you can run them by doing 'make' under LibOS/shim/test/apps/ltp and then './syscalls.sh'.  You can also run individual tests by hand, or under a debugger.

mkow:
Works now, this was a stupid bug I already fixed in the newer branch but forgot to backport it here (I mean the branch with more fixes which I'll PR after we merge this one).

donporter:
Test this please

donporter:
Test this please

donporter:
The last CI failure is from futex_wait03 LTP test.  I think we have seen this issue elsewhere.

Would you mind adding moving that one to the FLAKY list on your branch and see if CI passes?

mkow:
@donporter done.

donporter:
Test this please.

donporter:
Test this please

donporter:
Ok.  I'm going to go ahead and merge, but I would really like to get this apache issue debugged.

Nithinmelala:
I followed the wiki page's directions. When the server of the apache begins. I'm dealing with the bellow issue. 

graphene@graphene:~/Desktop/Progect/graphene/LibOS/shim/test/apps/apache$ make start-graphene-server
[core:warn] [pid 1] (92)Protocol not available: AH00076: Failed to enable APR_TCP_DEFER_ACCEPT
WARNING: no physical memory support, process creation may be slow.
shim_init() in restore_checkpoint (-13)
process creation failed
shim_init() in restore_checkpoint (-13)
process creation failed
shim_init() in restore_checkpoint (-13)
process creation failed
shim_init() in restore_checkpoint (-13)
process creation failed
shim_init() in restore_checkpoint (-13)
process creation failed
shim_init() in restore_checkpoint (-13)
process creation failed
shim_init() in restore_checkpoint (-13)
process creation failed

dimakuv:
In your case, Apache creates a new child process, and so Graphene wants to migrate state from the parent process to the child. But this migration fails at VMAs (internal memory regions). You should build Graphene with `make DEBUG=1` and specify `debug_type = inline` in the Apache manifest and re-run. You will see more detailed information about your bug.

mkow:
@Nithinmelala @dimakuv Please move this discussion to a separate issue. This is an old pull request which was closed more than a year ago and seems to be unrelated to your case.

PR #182, opened 2018-02-23
A manifest option allowing applications to create files which are not in trusted/allowed file lists

rainfld:
The PR includes a new option named "sgx.allow_file_creation", which is disabled by default. By turning on this option, users could allow the application to create the files which are not in trusted and allowed file lists. After the file is created, following open/read/write are allowed.

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_framework.c
+++ b/Pal/src/host/Linux-SGX/enclave_framework.c
@@ -156,6 +156,7 @@ int load_trusted_file (PAL_HANDLE file, sgx_stub_t ** stubptr,
        The created file is added to allowed_file list for later access */
     if (create && allow_file_creation) {
        register_trusted_file(uri, NULL);
+       *sizeptr = 0;
        return 0;
     }
 
--- a/Pal/src/host/Linux-SGX/enclave_framework.c
+++ b/Pal/src/host/Linux-SGX/enclave_framework.c
@@ -17,6 +17,8 @@ void * enclave_base, * enclave_top;
 
 struct pal_enclave_config pal_enclave_config;
 
+static int register_trusted_file (const char * uri, const char * checksum_str);
+
 bool sgx_is_within_enclave (const void * addr, uint64_t size)
 {
     return (addr >= enclave_base &&
@@ -120,9 +122,10 @@ DEFINE_LISTP(trusted_file);
 static LISTP_TYPE(trusted_file) trusted_file_list = LISTP_INIT;
 static struct spinlock trusted_file_lock = LOCK_INIT;
 static int trusted_file_indexes = 0;
+static int allow_file_creation = 0;
 
 int load_trusted_file (PAL_HANDLE file, sgx_stub_t ** stubptr,
-                       uint64_t * sizeptr)
+                       uint64_t * sizeptr, int create)
 {
     struct trusted_file * tf = NULL, * tmp;
     char uri[URI_MAX];
@@ -136,6 +139,10 @@ int load_trusted_file (PAL_HANDLE file, sgx_stub_t ** stubptr,
     if (uri_len < 0)
         return uri_len;
 
+    if (create && allow_file_creation) {
+       register_trusted_file(uri, NULL);
+       return 0;
+    }
     /* Normalize the uri */
     if (!strpartcmp_static(uri, "file:")) {
         SGX_DBG(DBG_E, "Invalid URI [%s]: Trusted files must start with 'file:'\n", uri);;
@@ -522,6 +529,12 @@ no_trusted:
 
 no_allowed:
     ret = 0;
+
+    if (get_config(store, "sgx.allow_file_creation", cfgbuf, CONFIG_MAX) <= 0) {
+        allow_file_creation = 0;
+    } else
+        allow_file_creation = 1;
+
 out:
     return ret;
 }
--- a/Pal/src/host/Linux-SGX/enclave_framework.c
+++ b/Pal/src/host/Linux-SGX/enclave_framework.c
@@ -124,6 +124,19 @@ static struct spinlock trusted_file_lock = LOCK_INIT;
 static int trusted_file_indexes = 0;
 static int allow_file_creation = 0;
 
+
+/* Function: load_trusted_file
+ * checks if the file to be opened is trusted or allowed,
+ * according to the setting in manifest
+ *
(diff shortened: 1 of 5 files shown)


donporter:
Ok to test

Pal/src/host/Linux-SGX/enclave_framework.c
```diff
@@ -17,6 +17,8 @@ void * enclave_base, * enclave_top;
 
 struct pal_enclave_config pal_enclave_config;
 
+static int register_trusted_file (const char * uri, const char * checksum_str);
```
donporter: @rainfld can you add a short comment for this function, explaining what it does, and what the arguments are.  It is somewhat obvious, but we are trying to (incrementally) beef up the code documentation.

Pal/src/host/Linux-SGX/enclave_framework.c
```diff
@@ -120,9 +122,10 @@ DEFINE_LISTP(trusted_file);
 static LISTP_TYPE(trusted_file) trusted_file_list = LISTP_INIT;
 static struct spinlock trusted_file_lock = LOCK_INIT;
 static int trusted_file_indexes = 0;
+static int allow_file_creation = 0;
 
 int load_trusted_file (PAL_HANDLE file, sgx_stub_t ** stubptr,
```
donporter: @rainfld Also, under the "campsite rule" (leave it better than you found it), would you mind adding a comment above this function as well: summarizing what it does, and its inputs and outputs?

Pal/src/host/Linux-SGX/pal_linux.h
```diff
@@ -105,7 +105,7 @@ typedef struct { char bytes[16]; } sgx_stub_t;
 
 int init_trusted_files (void);
 int load_trusted_file
```
donporter: @rainfld Please also copy the comment describing this function into the header.

rainfld:
Thanks for suggestions @donporter ! I have updated the PR based on your comments. 

I will add unite tests for this feature later. For your question, I would make the newly created file an allowed file. Because most of time this kind of file is a log or data file followed by many writes later.

donporter:
Although this could be a heisenbug, I think you may have introduced a regression in apache somehow.  It happens in both SGX and non-SGX runs.  Perhaps something in the manifest parsing logic?

Relevant output (until you get a jenkins account): 

Listen on 172.17.0.2:8001
./httpd.manifest -D FOREGROUND -C "ServerName 172.17.0.2" -C "Listen 172.17.0.2:8001" -C "PidFile logs/httpd-172.17.0.2-8001.pid"
[Tue Jan 06 08:46:13.833439 1970] [core:warn] [pid 1] (92)Protocol not available: AH00076: Failed to enable APR_TCP_DEFER_ACCEPT
WARNING: no physical memory support, process creation will be slow.
Internal memory fault at 0x665c8707da71 (IP = +0x9f7d6, VMID = 4012257512, TID = 1)
Trace/breakpoint trap (core dumped)
Makefile:111: recipe for target 'start-graphene-server' failed
make: *** [start-graphene-server] Error 133
Usage: httpd [-D name] [-d directory] [-f file]
             [-C "directive"] [-c "directive"]
             [-k start|restart|graceful|graceful-stop|stop]
             [-v] [-V] [-h] [-l] [-L] [-t] [-T] [-S] [-X]
Options:
  -D name            : define a name for use in <IfDefine name> directives
  -d directory       : specify an alternate initial ServerRoot
  -f file            : specify an alternate ServerConfigFile
  -C "directive"     : process directive before reading config files
  -c "directive"     : process directive after reading config files
  -e level           : show startup errors of level (see LogLevel)
  -E file            : log startup errors to file
  -v                 : show version number
  -V                 : show compile settings
  -h                 : list available command line options (this page)
  -l                 : list compiled in modules
  -L                 : list available configuration directives
  -t -D DUMP_VHOSTS  : show parsed vhost settings
  -t -D DUMP_RUN_CFG : show parsed run settings
  -S                 : a synonym for -t -D DUMP_VHOSTS -D DUMP_RUN_CFG
  -t -D DUMP_MODULES : show all loaded modules 
  -M                 : a synonym for -t -D DUMP_MODULES
  -t                 : run syntax check for config files
  -T                 : start without DocumentRoot(s) check
  -X                 : debug mode (only one worker, do not detach)
+ hostname -I

donporter:
Does this change to the unit test cause it to fail on the current master?  It may be as simple as adding a check that the file can be re-opened.

rainfld:
Looks like Jenkin building already fails before I check in the change to the unite test.  I added a Read permission when opening the file. Since the File.c will read the file after writting, without adding read permission, the read always fails

rainfld:
I'm looking at the reason why Jenkin fails

rainfld:
Looks like the error appeared in several previous builds as well. Maybe related to the bugs in memory management previously encountered?

donporter:
Ok on the unit test.  Have you seen this failure on master branch or other PRs?  I'm a little hesitant to merge this without a clean CI build.  But you don't necessarily need to debug this yourself - just be patient until I get a better understanding of the situation.

rainfld:
I reproduced this issue with Graphene setup (not Graphene-SGX) locally. Actually the exactly same issue occurs on both PR branch and latest master branch (https://github.com/oscarlab/graphene.git). 

I tried ./httpd.manifest -D FOREGROUND -C "ServerName 127.0.0.1" -C "Listen 127.0.0.1:8001" -C "PidFile logs/httpd-172.17.0.3-8004.pid"

[Mon Oct 12 00:42:37.178961 1970] [core:warn] [pid 1] (92)Protocol not available: AH00076: Failed to enable APR_TCP_DEFER_ACCEPT
WARNING: no physical memory support, process creation will be slow.
Internal memory fault at 0x3f0885519806 (IP = +0xa0627, VMID = 3008975461, TID = 1)
Trace/breakpoint trap (core dumped)

I compared the logs of different PR builds on Graphene setup. It seems like the reason why other PR builds did not hit that failure is because they failed earlier and apache test hasn't got chance to start yet.

Do you @donporter know the reasons about other failures on the Jenkins?

donporter:
There is a git timeout that is also causing some jobs to fail.  I'll try tuning some parameters and see if we can't get rid of this issue.

donporter:
Test this please

donporter:
Test this please

PR #194, opened 2018-03-31
Rewriting LGPL loader code in SHIM (1/3)

chiache:
This patch is still under testing. I pushed this just to see if it passes the Jenkins tests.

The purpose of this patch is to __largely reduce the trusted computing base__.


LibOS/shim/src/shim_rtld.c
```diff
@@ -371,3 +389,70 @@ int execute_elf_object (struct shim_handle * exec, int argc, const char ** argp,
     shim_do_exit(0);
     return 0;
 }
+
+BEGIN_CP_FUNC(link_map)
```
banne01: Why do we need to checkpoint any loaded binary ? the VMA checkpoint takes care of is for us.

chiache: To prevent reloading the binaries later when init_loader() is called. If there is no exec_map, init_loader() will think the process doesn't have an executable and try to load one.

banne01: That can be checked by weather we are parent ( first process ) or not.

chiache: That's not always true. A new process can be created with execve() as well. In that case, init_loaded() does have to load a new executable.

chiache: Besides, for the sake of code management, I will prefer not to rely on implicit assumptions and uses more straightforward logics such as "if (exec_map does not exist) loads a new executable."

chiache: I don't see checkpointing the loader and exec as such a big problem, especially if this step makes the implementation more straightforward. The invariant here is simple: if there is an executable, there has to be an exec_map. Same for the loader/interpreter.

chiache: Partially I agree that there is probably a design that we don't have to bookkeep any loaded binaries at all. The design might be slightly more obscure than the current design, but not completely impossible. Maybe it's better to take some time and think about the pros and cons.

banne01: I don't understand how this makes the implementation straight forward, What happens if we don't checkpoint/restore loader and exec, the answers I am coming up with is, nothing: the new fork or exec process works fine.

chiache: Not checkpointing/restoring exec_map and interp_map leads to a situation where these bookbeepings can be missing in the child processes, and that's how bugs and screwups happen under your nose. I will insist that either we always have the corresponding bookkeepings for consistency, or don't bookkeep at all. That's what I mean for "being straightforward". The fact that this even takes you or me quite a bit of time to figure out whether this is okay or not indicates that skipping this checkpointing is unstraightforward.

chiache: Then, how do you suggest we differentiate fork() and execve() in init_loader()?

banne01: How about returning from init_loader in case of  (PAL_CB(parent_process)) is present and don't do anything ?

chiache: That's true for both fork() and execve() in child processes. execve() also spawn a new process if there are multiple threads (example: vfork->execve).

banne01: right and exec will load a new binary and a loader in 'shim_do_execve_rtld' right ?

chiache: I am not talking about the case that init_loader() is called in shim_do_execve_rtld(), but rather the case that execve() creates a new process and the new process calls init_loader() as part of shim_init().

banne01: I am sure that init_loader can be refactored/parameterized for fork vs exec difference but the point being you don't need the old loader in any case, right ?

chiache: No, the problem is not about refactoring or parameterizing init_loader().

(1) The current code relies on the exec_map bookkeeping quite a lot. It determines whether there is a preloaded executable, where to start the execution, and how to fill in the auxiliary vector.  Keeping the pointer missing is not a wise thing to do. I have to emphasize: it is possible to completely eliminate this variable, but will take quite a bit of code change (nothing is impossible programmatically).

(2) You mentioned about large code base and TCB. That's actually a very nuanced argument: You only reduce the risk of bugs and vulnerabilities when you have *significantly* reduced the TCB (like orders of magnitude). Getting rid of less than 100 lines of code is not gonna improve your code quality; even worse, it can make the code more confusing when you try to cut corner and violate existing invariants.

(3) Moreover, checkpointing these libraries has very little effects on the migration time. Each of these ``link_map'' is smaller than 100 bytes and copying them is nothing compared to migrating the whole user memory.

banne01: I agree with 2 & 3 above and those comments were in response to the argument of code management simplicity. 
I still don't see a use of exec_map or interp_map in fork or exec child but I will take the argument that we should not loose exec_map just in case we need it at some point in our current design and checkpoint it.

chiache:
Closing the PR for now.

PR #208, opened 2018-05-10
Rewriting LGPL loader code in PAL (2/3)

chiache:
This patch is still under testing. I pushed this just to see if it passes the Jenkins tests.

The purpose of this patch is to __largely reduce the trusted computing base__.

PS. This should be merged after #194.


chiache:
Closing this PR for now.

PR #209, opened 2018-06-12
Implement sched_getaffinity system call

chiache:
This PR implements the missing sched_getaffinity system call at the library OS level. It also fixes some bugs in the Linux and SGX PAL for retrieving CPU information.


chiache:
Added PAL regression test for the CPU info in the control block. Also some bugfixes to make the CPU info reflect the actual values in /proc/cpuinfo in the host.

PR #210, opened 2018-06-14
Detecting invalid user memory given as system call inputs

chiache:
This PR introduces two APIs, `test_user_memory()` and `test_user_string()`, to check if user inputs point to readable / writable virtual memory regions. If the tests failed, the system calls should return -EFAULT or -EINVAL according to the Linux specification.

### DIFF
--- a/LibOS/shim/include/shim_tls.h
+++ b/LibOS/shim/include/shim_tls.h
@@ -73,6 +73,10 @@ typedef struct {
     void *                  debug_buf;
     int                     last_lock;
     struct lock_record      held_locks[NUM_LOCK_RECORD];
+    struct {
+        void * start, * end;
+        void * cont_addr;
+    } test_range;
 } shim_tcb_t;
 
 #ifdef IN_SHIM
--- a/LibOS/shim/src/bookkeep/shim_signal.c
+++ b/LibOS/shim/src/bookkeep/shim_signal.c
@@ -246,6 +246,15 @@ ret_exception:
 
 static void memfault_upcall (PAL_PTR event, PAL_NUM arg, PAL_CONTEXT * context)
 {
+    shim_tcb_t * tcb = SHIM_GET_TLS();
+    if (tcb->test_range.cont_addr && arg
+        && (void *) arg >= tcb->test_range.start
+        && (void *) arg <= tcb->test_range.end) {
+        assert(context);
+        context->rip = (PAL_NUM) tcb->test_range.cont_addr;
+        goto ret_exception;
+    }
+
     if (IS_INTERNAL_TID(get_cur_tid()) || is_internal(context)) {
 internal:
         internal_fault("Internal memory fault", arg, context);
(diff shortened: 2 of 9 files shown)


LibOS/shim/src/bookkeep/shim_signal.c
```diff
@@ -246,6 +246,15 @@ static void divzero_upcall (PAL_PTR event, PAL_NUM arg, PAL_CONTEXT * context)
 
 static void memfault_upcall (PAL_PTR event, PAL_NUM arg, PAL_CONTEXT * context)
```
donporter: Can you add some comments about the role of this function, when it is likely called, and how the arguments should be interpreted?

LibOS/shim/include/shim_tls.h
```diff
@@ -73,6 +73,10 @@ typedef struct {
     void *                  debug_buf;
     int                     last_lock;
     struct lock_record      held_locks[NUM_LOCK_RECORD];
+    struct {
```
donporter: Can I get a comment on the purpose of this field?

LibOS/shim/src/bookkeep/shim_signal.c
```diff
@@ -295,6 +304,71 @@ static void memfault_upcall (PAL_PTR event, PAL_NUM arg, PAL_CONTEXT * context)
     DkExceptionReturn(event);
 }
 
+bool test_user_memory (void * addr, size_t size, bool write)
```
donporter: Same request - add a comment documenting the function's job and expected behavior.

LibOS/shim/src/bookkeep/shim_signal.c
```diff
@@ -295,6 +304,71 @@ static void memfault_upcall (PAL_PTR event, PAL_NUM arg, PAL_CONTEXT * context)
...
+
+    has_fault = false;
+
+ret_fault:
+    tcb->test_range.cont_addr = NULL;
+    tcb->test_range.start = tcb->test_range.end = NULL;
+    return has_fault;
+}
+
+bool test_user_string (const char * addr)
```
donporter: Ditto on function-level comment.

LibOS/shim/src/ipc/shim_ipc.c
```diff
@@ -368,11 +368,20 @@ int send_ipc_message (struct shim_ipc_msg * msg, struct shim_ipc_port * port)
     assert(msg->size >= IPC_MSG_MINIMAL_SIZE);
     msg->src = cur_process.vmid;
 
-    int ret = DkStreamWrite(port->pal_handle, 0, msg->size, msg, NULL);
+    debug("send ipc message to port %p (handle %p)\n", port,
+          port->pal_handle);
+
+    PAL_HANDLE pal_handle = port->pal_handle;
+    rmb();
```
donporter: Why is a barrier needed here?  This also merits a comment.

donporter:
Also, I'm ok with removing lat_unix and the sendto LTP tests, and opening a separate issue to track that they need to eventually be put back.  I've found both to be a little flaky in CI.

chiache:
I have added some comments to explain the intuition of this PR.

Basically, the `test_user_memory()` and `test_user_string()` functions are purely for compatibility, not security. In fact, I am not sure whether we need to / is able to provide any race-free memory checking against a malicious user. From the perspective of the library OS, we can only assume the users are benign and won't actively exploit the concurrency to attack the library OS. If the users are malicious, then the whole library OS is compromised. Something like `copy_from_user() `may stop unaware attackers but cannot stop more advanced attacks.

In short, you can assume that the main purpose of this PR is nothing but to revive a big chunk of LTP tests that are previously failing because they didn't get EFAULT as errno.

donporter:
I am ok with the idea that the libOS cannot protect itself against an app that casts a pointer into libOS-internal memory and scribbles on it.  I suppose we might eventually hit an app that does a concurrent munmap or something.

I think this fix is an improvement, and we should take it.  It might be good to either open an issue documenting the concurrent munmap limitations, or maybe add a page on the wiki about this.

LibOS/shim/src/ipc/shim_ipc.c
chiache: I added some comment here, as well as in `shim_ipc_helper.c` where `port->deleted` is set to true. Basically we need the barrier here to ensure `port->deleted` is up-to-date in the memory.

LibOS/shim/src/bookkeep/shim_signal.c
chiache: Already added.

LibOS/shim/include/shim_tls.h
chiache: Added.

donporter:
Test this please

donporter:
The last CI job failed on dforkexec, which was disabled in master - not sure why Jenkins still ran this test.

chiache:
@donporter I think the lmbench-2.5 directory still points to the previous commit.

donporter:
I would have expected merging master to have picked up the latest LTP reference.

Issue #54, opened 2017-04-07
Can't execute shared objects anymore


adombeck:
Since commit 76967f39d3e197a037779f9adfccf7c6243bf0d4, the execution of shared objects in Graphene with SGX doesn't work anymore. I used to execute `python3.5`, which happens to be compiled as a shared object on Ubuntu 14.04, in Graphene. I will have to fix this. Sadly, this is a very big commit, which makes it hard to pin down the change that is responsible. Do you have an idea which change broke this?

donporter:
Hi @adombeck,

I think we have a pending fix; I'm trying to get to a bit more stable point before merging things to master going forward.  

I created the branch testing, which has some fixes for execution of objects on SGX.  I haven't tested python specifically yet, but would you mind taking a look and seeing if this addresses your issue.

In general, this branch can serve to let people try out pending fixes they are blocked on before the fix is fully vetted.

Thanks,
Don

adombeck:
Hi, thanks for the testing branch, I'm sure this will be useful. But sadly, it doesn't fix my issue, I still can't execute python3.5 with it.

adombeck:
My temporary workaround is to base my work on pre 76967f3, i.e. fffcb6a3663fa261600d19de54d78a7dc120c3f.

donporter:
Bummer.  Would it be too much of an imposition to create a simple unit test for the missing functionality?  Obviously running python would do it, but if there is a simpler example of "execute a shared library", that would help to make sure we don't break this in the future.

adombeck:
I created a simple shared object in https://github.com/oscarlab/graphene/pull/56 for testing this issue. I verified that the created shared object can be executed with the SGX Pal in fffcb6a, but in the current HEAD it fails with the same error as the `python3.5`:

```
enclave EINIT failed
initializing enclave failed: init_enclave: 1
```

Anyway, for other reasons I had to switch to Ubuntu 16.04, and it turns out that the `python3.5` executable in Ubuntu 16.04 is not a shared object and executes fine with Graphene. So this is no longer an urgent issue for me.

chiache:
Running shared objects is working now. Thanks for reporting the bug and contributing the test case.

PR #213, opened 2018-06-16
Add the shared object unit test to CI/standard unit test suites

donporter:
This puts the fixes for #54 and #56 into CI as part of the standard unit testing framework.


LibOS/shim/test/native/Makefile
```diff
@@ -27,12 +27,7 @@ static: %: %.c
 	@$(CC) $(CFLAGS) -o $@ -static $< \
 	$(shell echo $@ | sed 's/^[^\.]*//g' | sed 's/\./ -l/g')
 
-shared_object: %: %.c
-	@echo [ $@ ]
-	@$(CC) $(CFLAGS) -o $@ -fPIC -pie $< \
-	$(shell echo $@ | sed 's/^[^\.]*//g' | sed 's/\./ -l/g')
-
-pie: %: %.c
+xopie: %: %.c
```
chiache: Is there any reason of disabling the rule for pie?

donporter: No, this looks like a typo.  I only intended to take out the shared_object rule.

Issue #205, opened 2018-04-19
[SGX] PAL always returns 0 regardless of payload's exit code


omeg:
```
~/graphene/Runtime/pal-Linux-SGX test.manifest.sgx; echo "ret=$?"
...
[     1] ---- shim_exit_group (returning 1)
[     1] now kill other threads in the process
[     1] walk_thread_list(callback=0xb8de7a8)
[     1] now exit the process
[     1] ipc broadcast: IPC_CLD_EXIT(1, 1, -1)
[     1] found port 0xb8280c0 (handle 0xbd60f50) for process 0 (type 0002)
[     1] found port 0xb828048 (handle 0xbd606d0) for process 0 (type 0001)
[     1] parent not here, need to tell another process
[     1] ipc broadcast: IPC_CLD_EXIT(1, 1, -1)
[     1] found port 0xb8280c0 (handle 0xbd60f50) for process 0 (type 0002)
[     1] found port 0xb828048 (handle 0xbd606d0) for process 0 (type 0001)
[     1] this is the only thread
[     1] exiting ipc helper
[P22466] ipc helper thread terminated
[     1] deleting port 0xb8280c0 (handle 0xbd60f50) for process 0
[     1] deleting port 0xb828048 (handle 0xbd606d0) for process 0
[     1] process 22466 successfully terminated
ret=0
```

`test` is just doing `exit(1)`.

donporter:
Thanks for the report @omeg .  PR #216 should fix this issue.

mkow:
This was fixed by #216.

PR #216, opened 2018-06-22
Fix error code propagation (#205)

donporter:
This addresses issue #205.  I think propagating the exit code out of the PAL to $? is a reasonable default, although I do see an argument to be able to keep this inside the enclave (in SGX).  However, I still feel the default is to propagate out, and "future work" to add a manifest option to mask the exit code.

This fixes a bug in the SGX Pal's implementation of DkProcessExit.  PAL Unit test added.

This also fixes a bug in exit() and exit_group() that the exit code for the thread was set, but not the process (which is ultimately passed to DkProcessExit().  Shim unit test added.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/216)


donporter:
Test this please

donporter:
Test this please

PR #218, opened 2018-07-02
Adding a manifest option to allow graphene-sgx to run on non-avx platforms

rainfld:
By default, Graphene-SGX always enables AVX when generating XFRM attribute in SECS. There is no option to disable it. However, on some Intel platforms such as Intel NUC with SGX2 supported, AVX is not available, which causes the enclave creation failure raised up by the SGX driver. 

This PR adds an option named "sgx.disable_avx" to allow user to disable AVX in XFRM in manifest file.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/218)

### DIFF
--- a/Pal/regression/06_AvxDisable.py
+++ b/Pal/regression/06_AvxDisable.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python2
+
+import os, sys, mmap
+from regression import Regression
+
+loader = os.environ['PAL_LOADER']
+
+sgx = os.environ.get('SGX_RUN', False)
+
+def manifest_file(file):
+    if sgx:
+        return file + '.manifest.sgx'
+    else:
+        return file + '.manifest'
+
+if not sgx:
+  sys.exit(0)
+# Running AvxDisable 
+regression = Regression(loader, "AvxDisable")
+
+regression.add_check(name="Disable AVX bit in XFRM",
+    check=lambda res: "Illegal instruction executed in enclave" in res[0].log)
+
+rv = regression.run_checks()
+if rv: sys.exit(rv)
--- a/Pal/regression/AvxDisable.c
+++ b/Pal/regression/AvxDisable.c
@@ -1,0 +1,23 @@
+#include "pal.h"
+#include "pal_debug.h"
+#include <immintrin.h>
+#include <stdio.h>
+
+int main(){
+    /* Initialize the two argument vectors */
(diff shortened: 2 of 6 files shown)


donporter:
Ok to test

donporter:
Hi @rainfld , thanks for the fix!  Would you mind adding a PAL unit test (under Pal/regression) that exercises this option, and ensures that it works?  I think it would be sufficient to just copy one of the hello world/bootstrap tests and add this option, and then check that Graphene runs with this option selected.

Ideally, we would actually run under this condition, but I assume that would require hardware.

rainfld:
Sorry for my late repl @donporter. I'm happy to add this unit test. Could you be more specific about what Graphene need to check? Since this option is only consumed by the signing script (signer/pal-sgx-sign) for generating XFRM attributes of SECS, Graphene runtime will have no effect when the option is selected.  

One way we could test is to check the value of XFRM, but looks like we need expose this information in regression. I'm not sure if it's appropriate. Do you have any better idea how to achieve this?

And with hardware support, yes,  the enclave creation would succeed with this option selected on non-avx platform. Otherwise it will fail.

chiache:
So if this bit is set in the XFRM, using any AVX instruction will trigger #UD, right? Maybe add a test program that does this, with a manifest file setting the enclave to be non-AVX.

donporter:
Thanks @rainfld !  I like @chiache 's suggestion.  Just take one of the Bootstrap variants that does a hello world, but disable AVX at signing, and then test that we get a #UD.   Let us know if you have any other questions!

rainfld:
It's a good idea @chiache @donporter . I have written a simple program which is able to use AVX instruction as following:

```
#include"pal.h"
#include"pal_debug.h"
#include<stdio.h>

int main(){
  double num = 5.0;
  pal_printf("double num is %f\n", num);
  return 1;
}
```
When compiling with "-mavx" enabled, the compiler will generate vmovsd instruction (avx instruction) for manipulating the double type vairable. It's good enough for testing the feature disabling avx bit in XFRM I believe. 

However, this program hit segmentation fault when printing double type in pal_printf. Details are at https://github.com/oscarlab/graphene/issues/231

I'm now debugging this issue and will get back to this thread later.

donporter:
ignore: comment to trigger reviewable

donporter:
Retest this please

donporter:
Retest this please

donporter:
Retest this please

Issue #219, opened 2018-07-03
Bug: uninitialized `pipe.nonblocking` for the read file descriptor on `pipe()`


dimakuv:
(For simplicity, I only mention Graphene-SGX code but the bug holds also for other PALs)

When testing Redis 5 (https://github.com/antirez/redis/tree/5.0-rc1), I encountered the following bug.

Redis 5 introduces new functionality (modules) that uses non-blocking `read()` on a read end of the pipe:
```
src/module.c:4340   if (pipe(server.module_blocked_pipe) == -1) { ...  // create the pipe
...
src/module.c:4348   anetNonBlock(NULL,server.module_blocked_pipe[0]);  // set O_NONBLOCK via fcntl()
src/module.c:4349   anetNonBlock(NULL,server.module_blocked_pipe[1]);  // set O_NONBLOCK via fcntl() 
...
src/module.c:3588   while (read(server.module_blocked_pipe[0],buf,1) == 1);  // expects non-blocking read
```

In Graphene-SGX, the `pipe()` syscall ends up in `shim_pipe.c: create_pipes()`.

* The write file descriptor (`hdl2`) of the pipe is created via `DkStreamOpen() -> ... -> pipe_connect()`. In `pipe_connect()`, the file descriptor's `pipe.nonblocking` is correctly initialized to either TRUE or FALSE.

* The read file descriptor (`hdl1`) of the pipe is created via `DkStreamWaitForClient() -> ... -> pipe_waitforclient()`. The bug lurks here: `pipe_waitforclient()` *forgets* to initialize `pipe.nonblocking`, thus this variable contains garbage.

Now that `pipe.nonblocking` for the read FD contains garbage, the following happens:

1. Redis performs `fcntl(readfd, F_SETFL, O_NONBLOCK)` as part of `anetNonBlock()`

2. Graphene-SGX ends up in `LibOS/shim/src/fs/pipe/fs.c: pipe_setflags()` which verifies the correctness of requested syscall:
```
    if (attr.nonblocking) {
        if (flags & O_NONBLOCK)
            return 0;
        attr.nonblocking = PAL_FALSE;
    } else {
        if (!(flags & O_NONBLOCK))
            return 0;
        attr.nonblocking = PAL_TRUE;
    }
   ... actual fcntl syscall ...
```

3. Since `attr.nonblocking` for our read FD contains garbage, Graphene-SGX wrongly assumes that underlying FD is already non-blocking and simply does `return 0`.

[...]

PR #220, opened 2018-07-04
Apply the fix suggested for uninitialized pipe non-blocking (#219)

donporter:
In some code paths, the pipe.nonblocking field is not initialized, which can lead to obvious problems after memory has been recycled for a bit.

It is not clear to me if there is an easy unit test, other than perhaps in a debugging mode with memory poisoning enabled.  I think this is a bit beyond the scope of this simple PR.


chiache:
Test this please.

Issue #165, opened 2018-01-18
pal_loader doesn't handle incorrect file name properly


mkow:
If you run the following command:
```Bash
[...]/graphene/LibOS/shim/test/native/pal_loader SGX something_which_doesnt_exist
```
Then nothing happens (no errors or warnings on the console). This is very annoying in cases where you made a typo in the file name and are wondering what's wrong. Also, this seems to happen only in SGX mode.

Another problem occurs when you type `something.manifest` instead of `something`. AFAIR your Wiki said somewhere that it should also work, but I always get the following error when using the command with `.manifest`:
```
enclave EINIT failed - Invalid measurement
initializing enclave failed: init_enclave: 1
```

Tested on the commit c911f0213ab000d5393cd213d06947a544384072.

donporter:
These sound like legit issues to me.

mkow:
Couldn't reproduce on current master (https://github.com/oscarlab/graphene/tree/34b8eb1ec2b23d5538d8ede385eb1be50e29354d). Seems that this was fixed when refactoring pal-$HOST binaries interface.

PR #223, opened 2018-07-05
Fix various exec file specification issues

donporter:
This PR addresses the issues reported in  #165 .

It ensures that passing a binary or manifest that does not exist returns a sensible error message, and adds a PAL unit test for this case.

It also ensures that specifying a binary by manifest (at least with the .manifest.sgx extension on the SGX PAL) actually works.  This is somewhat complicated by the fact that the binary may be different than the manifest base name, and that it is ok to just preload libraries with no executable at all.  In the first case, we give loader.exec_uri precedence over inferring the binary from the manifest.  In the second case, we preserve the loading behavior as before with no executable, but add a warning that this is probably an error (unless you are @chiache :) ).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/223)

### DIFF
--- a/Pal/regression/00_Bootstrap.py
+++ b/Pal/regression/00_Bootstrap.py
@@ -100,10 +100,11 @@ if rv: sys.exit(rv)
     
 
 # Running Bootstrap2
-regression = Regression(loader, "Bootstrap2")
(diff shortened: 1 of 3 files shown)


donporter:
Test this please

PR #225, opened 2018-07-06
Simplifying the flaky memory availability test for PAL

chiache:
The "Get Memory Available Quota" test in the PAL regression tests is unstable because it is affected by the OS. This patch removes the test and adds a simpler test that only verifies the range of memory quota.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/225)


donporter:
I think the change is fine, although I am curious if you can explain a bit more what "is affected by the OS" means, and what the remaining test is actually testing for.

chiache:
DkMemoryAvailableQuota() returns the remaining free memory in the host. This tests assumes that after accessing 1000 pages, the free memory should reduce by 4MB, or at least 80% of it. This test is flaky because it is heavily affected by the paging of the host OS. In most of the case, this test should pass, but in very rare cases, the host OS may act differently, causing the test to be flaky.

donporter:
Would it make sense to just show that it reduces by at least a given amount, and accept that it can be more?

chiache:
Yes, that's exactly what the test is doing, and for most of the time, it works.

chiache:
Let's put a hold on this for a while. If the same test fails again, we can revive this PR.

chiache:
Test this please.

chiache:
test this please

PR #226, opened 2018-07-08
Suppressing GLIBC compilation warning

chiache:
This PR removed a few annoying warnings generated from building GLIBC 2.19. The suppressed warnings include:

(1) 
```
gawk: scripts/sysd-rules.awk:56: warning: gensub: third argument `' treated as 1
gawk: scripts/sysd-rules.awk:56: warning: gensub: third argument `' treated as 1
gawk: scripts/sysd-rules.awk:56: warning: gensub: third argument `' treated as 1
...
```

(2)
```
../include/signal.h:58:74: warning: right-hand operand of comma expression has no effect [-Wunused-value]
```
(Suppressed with -Wno-unused-value)

(3) 
```
ar: `u' modifier ignored since `D' is the default (see `U')
ar: `u' modifier ignored since `D' is the default (see `U')
ar: `u' modifier ignored since `D' is the default (see `U')
...
```


donporter:
I assume this does not suppress errors if compilation goes wrong?  If this is correct, I see no problem with this patch.

PR #227, opened 2018-07-08
Allowing concurrent PAL builds

chiache:
This PR separates the object files for different host targets (e.g., Linux and Linux-SGX), and doesn't require "make clean" before building another PAL. With this PR, no object files will be created in the shared source directory (i.e., Pal/src). Instead, host-independent object files will be moved to a ".obj" directory in each host-specific directory (e.g., Pal/src/host/Linux/.obj).


donporter:
Indeed, memory quota test failed again.  I think it is reasonable to reopen that PR.

chiache:
Test this please.

donporter:
This is blowing an assert in shim_ipc_nsimpl.h - I believe @chiache has a fix on a branch that could be landed.  I'll try to run one more time, as I think this is an intermittent error.

donporter:
Test this please

Issue #228, opened 2018-07-10
shim_do_epoll_wait() is terribly inefficient


dimakuv:
This are two performance bugs in [`shim_do_epoll_wait()`](https://github.com/oscarlab/graphene/blob/master/LibOS/shim/src/sys/shim_epoll.c). I found it while testing Redis 5.0. I concentrated only on Graphene-SGX, but should be applicable to other PALs.

1. Redis relies on `epoll_wait()` for bulk I/O on network and pipes. The expectation is that under heavy load (~100 clients) this syscall will return all events on all clients. After the syscall, Redis iterates through returned ~100 events and processes each of them.

In current Graphene, `shim_do_epoll_wait()` returns only *one* event. This leads to an overhead of 100x: Redis receives only one event, processes it, and has to issue `epoll_wait()` again (with other 99 events still pending).

In particular, the problem lies here:
```
    listp_for_each_entry(epoll_fd, &epoll->fds, list)
        if (polled == epoll_fd->pal_handle) {
            ... fill in revents of epoll_fd ...
            break;  // <-- forces subsequent logic to process only this epoll_fd !
        }
```

2. Another performance bottleneck is the invocation of `DkStreamAttributesQuerybyHandle()` during `shim_do_epoll_wait()`, for *each* returned event.

`DkStreamAttributesQuerybyHandle()` ends up in a PAL-specific [`socket_attrquerybyhdl()`](https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/db_sockets.c) for sockets, [`pipe_attrquerybyhdl()`](https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/db_pipes.c) for pipes, etc. Note that Redis uses `epoll_wait()` on both sockets and pipes.

These `xxx_attrquerybyhdl()` helpers return attributes for a particular handle. While most attributes are extracted directly from the handle metadata, two attributes (`pending_size` and `readable`) require OCALLs (`ocall_fionread` and `ocall_poll` respectively).

Now, in Redis case, for an epoll on its 100 clients and two opened pipes, there will be 204 OCALLs in the worst case during a single `shim_do_epoll_wait()` (given that issue 1 is fixed).

Ironically, `attr->pending_size` is not even checked during `shim_do_epoll_wait()`. The `attr->readable` can be actually extracted from the first `ocall_poll()`. I believe that the invocation of `DkStreamAttributesQuerybyHandle()` is redundant.

[...]

PR #229, opened 2018-07-11
Improving handle polling semantics

chiache:
This PR addresses the performance issues of #228. The existing PAL call DkObjectsWaitAny() is too restrictive for specifying the polling events and being unable to return multiple simultaneous events. The new DkObjectsWaitEvents() accepts additional arguments to specific and return various events (signal, read, write, and error).

For incremental development, the old DkObjectsWaitAny() is kept for compatibility against the library OS. Eventually we will replace the API with DkObjectsWaitEvents().

This PR is not complete yet; remaining to-dos:
- Reimplement epoll_wait() using DkObjectsWaitEvents()
- Evaluate performance improvements
- Writing a regression test for DkObjectsWaitEvents()

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/229)

### DIFF
--- a/LibOS/shim/src/sys/shim_poll.c
+++ b/LibOS/shim/src/sys/shim_poll.c
@@ -117,6 +117,7 @@ static int __do_poll (int npolls, struct poll_handle * polls,
     struct poll_handle * polling = NULL;
     struct poll_handle * p, ** n, * q;
     PAL_HANDLE * pals = NULL;
+    PAL_FLG * pal_events = NULL, * ret_events;
     int ret = 0;
 
 #ifdef PROFILE
@@ -289,8 +290,8 @@ done_finding:
     }
 
     pals = __try_alloca(cur, sizeof(PAL_HANDLE) * npals);
-    PAL_FLG * pal_events = __try_alloca(cur, sizeof(PAL_FLG) * npals);
-    PAL_FLG * ret_events = __try_alloca(cur, sizeof(PAL_FLG) * npals);
+    pal_events = __try_alloca(cur, sizeof(PAL_FLG) * npals * 2);
+    ret_events = pal_events + npals;
     npals = 0;
 
     n = &polling;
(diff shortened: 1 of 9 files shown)


donporter:
This seems like a nice improvement, to better capture the common polling requirements and make this more natural.

There are a couple of implicit design decisions I flagged that I think would benefit from a maintainer-level discussion.

thomasknauth:
I am unsure on how to proceed. I have re-implemented shim_do_epoll_wait() using the newly added PAL call DkObjectsWaitEvents(). I currently do not have a small test case though, since the code sort of works on some branch or even on master after applying a queued up PR.

In general, I feel the situation is becoming unsustainable with the sheer amount of pending PRs. We've repeatedly stumbled upon problems only to discover someone else already reported them or there might even already be a pending PR fixing it. Interaction of pending PRs is another worrisome problem.

I'm inclined to submit my code as is, just so it does not get lost. In my case, it fixes an issue I saw when running redis-benchmark on Graphene's DkObjectWaitEvents branch. Coming up with a smaller scale test case is possible, but would require more work on my end. Given the issues outlined above, I am reluctant to do so at this point.

donporter:
I'm a bit confused.  Was this submitted to the wrong PR?

Your concern is definitely appreciated about the backlog of PRs - we are working on it.  

I think we can be reasonable about test cases if they really are hard to come up with, but a lot of times they are not.  I think if no one, author or reviewer, can describe how to write a simple unit test, then we should accept the PR as is.

thomasknauth:
@donporter, possibly the wrong PR. I've just submitted #434. Still learning about the workflows. Wasn't sure if I could tack on my changes to this existing PR.

donporter:
I know you had a comment on #197 about this, and, ironically, I was holding up merge of that until we resolved that discussion.  Would you be ok with merging #197?

In general, your fixes would need to be in a separate PR.  A single PR should address one issue, and be from one author.

But I appreciate you reaching out and helping.

thomasknauth:
I did reply on #197 yesterday. Yes, please go ahead and close it.

PR #233, opened 2018-08-02
Deprecating RSA enclave keys

chiache:
Generation of a unique RSA key significantly slows down the startup time of Graphene-SGX. Hereby we deprecate this key and replace it with a simple random number. The performance improvement is as follows:

For creating a 256MB enclave:

- With RSA key generation: 1.01-1.05 seconds
- Without the key: 0.43 seconds

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/233)

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_framework.c
+++ b/Pal/src/host/Linux-SGX/enclave_framework.c
@@ -847,50 +847,38 @@ void test_dh (void)
 
 int init_enclave (void)
 {
+#if 0
+    /*
+     * This enclave-specific key is a building block for authenticating
+     * new pipe connections with other enclaves that are already
+     * authenticated. Since pipe protection is a future feature, this key
+     * is currently unused and hence deprecated.
+     */
     int ret;
     LIB_RSA_KEY *rsa = malloc(sizeof(LIB_RSA_KEY));
     lib_RSAInitKey(rsa);
 
     ret = lib_RSAGenerateKey(rsa, RSA_KEY_SIZE, RSA_E);
-    if (ret != 0) {
+    if (ret < 0) {
         SGX_DBG(DBG_S, "lib_RSAGenerateKey failed: %d\n", ret);
         return ret;
     }
 
-    PAL_NUM nsz = RSA_KEY_SIZE / 8, esz = 1;
-    uint8_t n[nsz], e[esz];
-
-    ret = lib_RSAExportPublicKey(rsa, e, &esz, n, &nsz);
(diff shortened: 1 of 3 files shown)


Pal/src/host/Linux-SGX/enclave_framework.c
```diff
@@ -847,56 +847,22 @@ void test_dh (void)
...
-    if (ret < 0)
-        goto out_free;
-
-    pal_enclave_config.enclave_key = rsa;
+    /*
+     * The enclave identifier is uniquely created for each enclave to
+     * prevent man-in-the-middle attack during local attestion.
+     * The identifier is injected into the local attestation report
+     * to be authenticated by another trustd process.
+     */
```
donporter: What happens if you get the same value for two enclaves?  Can we detect this case, or pass in a pid from the parent that we have reason to believe is unique?

chiache: The enclave id is a 256-bit random number, generated by the CPU's rng. The probability of collision on this id should be low enough given that enclave creation is quite slow. Relying on pids passed from the host wouldn't be secure in this scenario.

donporter: Can you write a bit longer description of the protocol in a comment (or point me to it if it is somewhere else)?  I am not confident I understand this yet, and I think this would be good documentation.  It would help to know what steps are done in which enclave, what messages are exchanged (and if they are in plaintext), and how the decision is made to accept the connection.

chiache:
I should clarify some misleading in the previously-written comments. The man-in-the-middle-attack is prevented by proving the the entities who perform Diffie-Hellman to be the same as the ones who generate the attestation with EREPORT. This part is done by injecting the hash of the session key into the  attestation. The enclave id is also hashed and sent through the secure channel, as a token for reconciling mutually-trusting enclaves when re-establishing connections.

rainfld:
Now it makes sense to me. Looks like Jenkins/Jenkins-Debug failed. Is it due to this change? They work under other PRs

donporter:
Jenkins is failing again with this heisenbug: assert failed ipc/shim_ipc_nsimpl.h:991 !qstrempty(&NS_LEADER->uri) (value:0)

Or this one: 

assert failed ipc/shim_ipc_helper.c:239 vmid != cur_process.vmid (value:0)

Were there some recent changes to the IPC code that merit a second look?

rainfld:
Test this please

PR #235, opened 2018-08-08
Asynchronous RPC syscalls for Graphene-SGX

dimakuv:
This PR adds the ability to invoke system calls asynchronously, similar to solutions from [Eleos by Orenbach](https://dl.acm.org/citation.cfm?id=3064219) and [SCONE by Arnautov et al](https://dl.acm.org/citation.cfm?id=3026930).

New directive `sgx.rpc_thread_num = X` in the manifest instructs Graphene-SGX to create X outside-enclave RPC threads that spin-wait for syscalls (and any OCALLs in general) on a shared queue. In-enclave threads do not exit the enclave now but instead enqueue the syscall in the queue and spin-wait for it to be processed by an RPC thread.

Omitting `sgx.rpc_thread_num` defaults to the old, synchronous way of executing syscalls.

This PR works on single-process applications and can provide 2-3X performance improvement.

The patch has the following issues:

1. Graphene-SGX uses `poll()` system call during initialization (and during IPC communication?). Currently, this syscall "becomes" asynchronous only after the enclave accepts some connection (which hints that initialization phase is over). The correct way would be to distinguish between Graphene's internal usage of `poll()` and original app's usage, and use async version only for app's usage. However, I didn't find an easy way to propagate this info to `ocall_poll`.

2. For correctness, whenever a signal is delivered to the enclave, we must abort blocking syscalls in RPC threads. Otherwise the RPC thread could become stuck in the blocking syscall while the corresponding enclave thread handles the signal and assumes that its syscall was aborted. Currently, we send dummy `SIGUSR1` to all RPC threads whenever termination signal is received (so they abort their blocking syscall). This could be implemented in a cleaner way.

3. If original app has a signal handler that uses syscalls (e.g., to dump logs), we switch to old synchronous way of invoking syscalls for safety. If issue 2 is solved in a clean way, this safety measure can be removed.

4. RPC threads are killed using `exit_group()` when all enclave threads are finished. This is probably not a clean way to close the app.

5. Currently I run into bugs in multi-process applications, most probably because of the Graphene-SGX IPC thread which is not aware of the changes. Not sure how to debug these issues.

I thought I will push it now so that people suggest how to solve above issues.

### DIFF
--- a/Pal/src/host/Linux-SGX/db_exception.c
+++ b/Pal/src/host/Linux-SGX/db_exception.c
@@ -33,6 +33,7 @@
 #include "pal_security.h"
 #include "api.h"
 #include "ecall_types.h"
+#include "rpcqueue.h"
 
 #include <atomic.h>
 #include <sigset.h>
@@ -52,6 +53,8 @@ typedef struct exception_event {
                   "movq %%rax, %0\r\n"              \
                   : "=b"(ptr) :: "memory", "rax")
 
+extern rpc_queue_t* rpc_queue;
+
 void _DkGenericEventTrigger (PAL_IDX event_num, PAL_EVENT_HANDLER upcall,
                              PAL_NUM arg, struct pal_frame * frame,
                              PAL_CONTEXT * context)
@@ -62,6 +65,8 @@ void _DkGenericEventTrigger (PAL_IDX event_num, PAL_EVENT_HANDLER upcall,
     event.context = context;
     event.frame = frame;
 
+    if (rpc_queue)
+        rpc_queue->in_signal_handler++;
     (*upcall) ((PAL_PTR) &event, arg, context);
 }
 
@@ -303,6 +308,12 @@ void _DkRaiseFailure (int error)
 
 void _DkExceptionReturn (void * event)
 {
+    if (rpc_queue) {
+        rpc_queue->in_signal_handler--;
+        if (rpc_queue->in_signal_handler < 0)
+            rpc_queue->in_signal_handler = 0;
+    }
+
     PAL_EVENT * e = event;
     sgx_context_t uc;
     PAL_CONTEXT * ctx = e->context;
--- a/Pal/src/host/Linux-SGX/Makefile
+++ b/Pal/src/host/Linux-SGX/Makefile
@@ -9,9 +9,9 @@ defs	= -DIN_PAL -DPAL_DIR=$(PAL_DIR) -DRUNTIME_DIR=$(RUNTIME_DIR)
 enclave-objs = $(addprefix db_,files devices pipes sockets streams memory \
 		 threading mutex events process object main rtld \
 		 exception misc ipc spinlock) \
-	       $(addprefix enclave_,ocalls ecalls framework pages untrusted) 
+	       $(addprefix enclave_,ocalls ecalls framework pages untrusted rpcqueue)
 enclave-asm-objs = enclave_entry 
-urts-objs = $(addprefix sgx_,enclave framework main rtld thread process exception graphene)
+urts-objs = $(addprefix sgx_,enclave framework main rtld thread process exception graphene rpcqueue)
 urts-asm-objs = sgx_entry
 graphene_lib = ../../.lib/graphene-lib.a
 headers	= $(wildcard *.h) $(wildcard ../../*.h) $(wildcard ../../../lib/*.h) \
--- a/Pal/src/host/Linux-SGX/ecall_types.h
+++ b/Pal/src/host/Linux-SGX/ecall_types.h
@@ -13,4 +13,5 @@ typedef struct {
     const char ** ms_arguments;
     const char ** ms_environments;
     struct pal_sec * ms_sec_info;
+    void * rpc_queue;
 } ms_ecall_enclave_start_t;
--- a/Pal/src/host/Linux-SGX/enclave_ecalls.c
+++ b/Pal/src/host/Linux-SGX/enclave_ecalls.c
@@ -7,8 +7,11 @@
 #include <api.h>
 
 #include "ecall_types.h"
+#include "rpcqueue.h"
 
 #define SGX_CAST(type, item) ((type) (item))
+
+extern rpc_queue_t * rpc_queue;  /* pointer to untrusted queue */
 
 void pal_linux_main (const char ** arguments, const char ** environments,
                      struct pal_sec * sec_info);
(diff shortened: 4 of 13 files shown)


donporter:
This is really cool!  I'm glad to see some work on this.  I think this is probably going to require some discussion to figure out the best way to deal with these issues.

In the interest of getting some discussion going:
1) I think the right way to propagate this might be as an argument to the PAL call that ultimately does the poll.

2, 3) I think there is a detail here that I don't yet understand.  Can you give an example of the case you are worried about?  In general, I like the invariant that these helper threads should not be blocking, but should do the minimal amount of work to hand off to the application threads to finish a system call.

4) I agree.  This feels like a separate issue, but probably deserves some attention.

5) Can we disable this feature as soon as you start up a second process, until we can debug the issue?  That might be one incremental path.

I suspect that solving (1) will probably address (5).

Thanks again!

Pal/src/host/Linux-SGX/db_exception.c
```diff
@@ -62,6 +65,8 @@ void _DkGenericEventTrigger (PAL_IDX event_num, PAL_EVENT_HANDLER upcall,
     event.context = context;
     event.frame = frame;
 
+    if (rpc_queue)
```
donporter: One thing we will need to be careful about is, anything imported from the URTS as an extern should probably be heavily commented on why this use of an untrusted input is safe (or, what is the worst thing that will happen if someone is adversarial).

donporter:
Ok to test

Pal/src/host/Linux-SGX/db_exception.c
dimakuv: Indeed. I recognized now that `rpc_queue` needs some refactoring for security. I will add security checks on `rpc_queue` during enclave start-up, as well as comments.

dimakuv:
> 1. I think the right way to propagate this might be as an argument to the PAL call that ultimately does the poll.

Yes, if @chiache approves, I can do it.

> 2, 3) I think there is a detail here that I don't yet understand. Can you give an example of the case you are worried about? In general, I like the invariant that these helper threads should not be blocking, but should do the minimal amount of work to hand off to the application threads to finish a system call.

**(2)** Consider the situation with networking. The enclave calls `read` which ultimately transforms into `ocall_read` which finally issues the `read` syscall. `read` blocks the issuing thread until socket data becomes ready.

In normal Graphene-SGX with *synchronous syscalls*, the enclave thread exits the enclave to issue `read` and blocks. Whenever a signal comes to this blocked thread, the syscall will be interrupted ([`signal` manpage](http://man7.org/linux/man-pages/man7/signal.7.html)), the enclave will be re-entered, and the signal will be handled inside the enclave. After signal handling, the app inside enclave will recognize that the syscall was interrupted and will react accordingly.

The problem is that with *asychronous syscalls*, there are suddenly two threads: the helper thread and the enclave thread. The enclave thread outsources the syscall to the helper thread and busy-waits; the helper thread blocks on `read`. Because the threads are decoupled, whenever a signal comes to the *enclave thread*, the enclave thread jumps into signal handler while the helper thread still blocks not knowing about the signal.

This breaks the semantics: after handling the signal, the enclave thread assumes that the syscall was interrupted. But the helper thread still waits on this syscall, thus this syscall's existence and result are "forgotten" by the enclave. This generally leads to hangs because the helper thread waits forever for the enclave thread to consume the result.

It is possible to make `read` synchronous, i.e., let the enclave thread exit, block on syscall, and re-enter. The problem with this approach, of course, is that *most* syscalls are blocking and thus must be marked as synchronous. Thus, there is no performance benefit (which is the sole purpose of this PR).

The approach I opted for is to "imitate" the original behavior on interrupted syscalls. Whenever a signal comes to the enclave thread, it will send a dummy signal to the corresponding helper (RPC) thread so that it interrupts the syscall. In this case, both threads become "synchronized". Note that for the first prototype I make two simplifications: (a) *all* helper threads are notified by the dummy signal instead of only one, and (b) enclave thread sends notifications only on termination (e.g. SIGINT) signals.

**(3)** Consider now if enclave's signal handler has a blocking `write` syscall. The helper thread may still be blocked on the previous `read`, thus the enclave thread finds no recipient for a new `write` and hangs. Ideally, this should not be a problem because the helper thread should interrupt the previous syscall and accept the new one, but for safety I chose to switch to synchronous behavior when inside a signal handler (assuming that signals are rare and thus not affect performance).

As I mentioned, logic for (3) can be removed if (2) is solved in a nice way.

P.S. What do you mean by "these helper threads should not be blocking, but should do the minimal amount of work to hand off to the application threads to finish a system call"? Some thread must block on a syscall, and it cannot be an application thread -- that would mean that the app thread must exit enclave and issue a syscall.

donporter:
I see.  This is a tricky problem.  Thanks for clarifying.

I like moving towards asynchronous behavior.  I think this is the key to performance, and worth pursuing.

Would it solve part of problem 2 to have a field (or bit) in the struct that is busy-waited upon that indicates the system call was interrupted/canceled?

dimakuv:
> I like moving towards asynchronous behavior. I think this is the key to performance, and worth pursuing.

Yes, in my experience it gives the biggest boost wrt performance.

> Would it solve part of problem 2 to have a field (or bit) in the struct that is busy-waited upon that indicates the system call was interrupted/canceled?

I am not sure what this field would accomplish. If the helper thread is interrupted, there is no need to set the field: the syscall of helper thread will restart/fail and the enclave thread will get the result at some point. But if the enclave thread is interrupted, setting the field doesn't help: the helper thread is blocked on syscall, and the only way to awake it is to send a signal.

donporter:
So perhaps I am still misunderstanding problem 2.  When we say "helper thread" do we both mean an in-enclave thread?

> If the helper thread is interrupted, there is no need to set the field: the syscall of helper thread will restart/fail and the enclave thread will get the result at some point. But if the enclave thread is interrupted, setting the field doesn't help: the helper thread is blocked on syscall, and the only way to awake it is to send a signal.

In the second case, is the issue that we don't control which thread in the enclave receives the signal?

dimakuv:
> So perhaps I am still misunderstanding problem 2. When we say "helper thread" do we both mean an in-enclave thread?

Oh sorry. I misunderstood you then. By "helper thread" I meant the "RPC" outside-of-enclave thread that performs syscalls on behalf of the "enclave" (in-enclave app) thread.

> > If the helper thread is interrupted, there is no need to set the field: the syscall of helper thread will restart/fail and the enclave thread will get the result at some point. But if the enclave thread is interrupted, setting the field doesn't help: the helper thread is blocked on syscall, and the only way to awake it is to send a signal.
> In the second case, is the issue that we don't control which thread in the enclave receives the signal?

Sorry again, we talked about different things. I was talking about the "helper" RPC outside-of-enclave thread which is blocked on a syscall.

donporter:
I think we are on the same page now, and no worries - these are tricky issues without well-established nomenclature.

Just to reiterate: the case you are worried about is when a signal ends up delivered to the enclave thread, which "abandons" a system call that the out-of-enclave helper thread is still waiting on?

dimakuv:
> Just to reiterate: the case you are worried about is when a signal ends up delivered to the enclave thread, which "abandons" a system call that the out-of-enclave helper thread is still waiting on?

Yes, this is exactly the issue.

PR #236, opened 2018-08-09
Update README by reorganize the section of building Graphene-SGX

rainfld:
Through a few users' feedback, current section of building Graphene-SGX is not quite intuitive and lack of  details instructions about how to run examples so the users can test Graphene-SGX is correctly installed. This PR reorganizing the flows and adding instructions of running built-in examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ To build Graphene library OS with debug symbols, run "make DEBUG=1" instead of
 
 __** Note: this step is optional. **__
 
-__** Note: for building with Intel:registered: SGX support, skip this step. **__
+__** Note: for building with Intel:registered: SGX support, skip this step, go to section 2.2 **__
 
 __** Disclaimer: this feature is experimental and may contain bugs. Please do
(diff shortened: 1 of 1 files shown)


dimakuv:
Couple nitpicks:

1. Subsubsections like "2.1.1" should be at least bold. Hard to discern them from the normal text.

2. `(The SDK and driver version must be 1.9 or LOWER)` -- this is not true anymore, can be removed. Tested with SDK 2.0 and 2.2.

3. `SGX=1 ./pal_loader HelloWorld` -- this shouldn't work. Since Linux is case-sensitive, the correct command must be `SGX=1 ./pal_loader helloworld`.

4. Please use `backticks` in 2.1.3 (2) when explaining commands to make and run Python.

chiache:
Test this please

chiache:
ok to test

chiache:
test this please

chiache:
Test this please

chiache:
Test this please

PR #237, opened 2018-08-12
Signal not handled properly when the current thread is dead

chiache:
I suspect this might be the cause of the heisenbugs that happen during Jenkins test. When a signal was caught after a thread or a process is in the middle of termination, the internal signal handler continues to process the signal, either triggering an internal segfault or calling the user handler. The right behavior should be to ignore the signal completely.

Testing if this PR passes the Jenkins test.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/237)

### DIFF
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -431,6 +431,8 @@ static inline PAL_HANDLE thread_create (void * func, void * arg, int option)
 static inline void __disable_preempt (shim_tcb_t * tcb)
 {
     //tcb->context.syscall_nr += SYSCALL_NR_PREEMPT_INC;
+    /* Assert if this counter overflows */
+    assert((tcb->context.preempt & ~SIGNAL_DELAYED) != ~SIGNAL_DELAYED);
     tcb->context.preempt++;
     //debug("disable preempt: %d\n", tcb->context.preempt & ~SIGNAL_DELAYED);
 }
@@ -446,6 +448,8 @@ static inline void disable_preempt (shim_tcb_t * tcb)
 static inline void __enable_preempt (shim_tcb_t * tcb)
 {
     //tcb->context.syscall_nr -= SYSCALL_NR_PREEMPT_INC;
+    /* Assert if this counter underflows */
+    assert(tcb->context.preempt > 0);
     tcb->context.preempt--;
     //debug("enable preempt: %d\n", tcb->context.preempt & ~SIGNAL_DELAYED);
 }
--- a/LibOS/shim/src/ipc/shim_ipc_pid.c
+++ b/LibOS/shim/src/ipc/shim_ipc_pid.c
@@ -158,7 +158,7 @@ int ipc_pid_kill_callback (IPC_CALLBACK_ARGS)
             break;
     }
 
-    assert(ret != -ESRCH);
+    //assert(ret != -ESRCH);
 
     SAVE_PROFILE_INTERVAL(ipc_pid_kill_callback);
     return ret < 0 ? ret : RESPONSE_CALLBACK;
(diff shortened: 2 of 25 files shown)


chiache:
Test this please

chiache:
Test this please

chiache:
Test this please

chiache:
Test this please

chiache:
Test this please

chiache:
So far the test has been going fine, except that the GNU website has been unresponsive since this morning. I change the Makefile to download glibc from multiple mirrors.

chiache:
Test this please

chiache:
Test this please

chiache:
Test this please

chiache:
I will add a few regression tests to this PR. It would be nice if we can also fix the notorious "signal queue is full" issue.

chiache:
Test this please

chiache:
Merged with #240.

donporter:
If the GNU mirror continues being a problem, I'd be happy to set up a local mirror at UNC, at least for the jenkins test.  @dcowhig - would you mind keeping an eye on this?

LibOS/shim/include/shim_internal.h
```diff
@@ -446,6 +448,8 @@ static inline void disable_preempt (shim_tcb_t * tcb)
 static inline void __enable_preempt (shim_tcb_t * tcb)
 {
     //tcb->context.syscall_nr -= SYSCALL_NR_PREEMPT_INC;
+    /* Assert if this counter underflows */
+    assert(tcb->context.preempt > 0);
```
donporter: Isn't preempt is an unsigned type?  If you only care about 0 this is fine, but if you want to check for a negative value, this needs a cast.

LibOS/shim/src/ipc/shim_ipc_pid.c
```diff
@@ -158,7 +158,7 @@ int ipc_pid_kill_callback (IPC_CALLBACK_ARGS)
             break;
     }
 
-    assert(ret != -ESRCH);
+    //assert(ret != -ESRCH);
```
donporter: What is up with this assertion?  Is this a bad assert (should it be deleted?), or should it be re-enabled?

LibOS/shim/src/sys/shim_clone.c
```diff
@@ -110,6 +110,7 @@ int clone_implementation_wrapper(struct clone_args * arg)
     }
     allocate_tls(my_thread->tcb, my_thread->user_tcb, my_thread);
     shim_tcb_t * tcb = &((__libc_tcb_t *) my_thread->tcb)->shim_tcb;
+    __disable_preempt(tcb);
```
donporter: A comment here would be good, similar to the ones above.

Pal/src/host/Linux/db_exception.c
```diff
@@ -207,23 +207,33 @@ static bool _DkGenericSignalHandle (int event_num, siginfo_t * info,
...
+struct pal_frame * __get_frame (uintptr_t rbp)
+{
+    if (!rbp)
+        asm volatile ("movq %%rbp, %0" : "=r"(rbp) :: "memory");
+
+    uintptr_t last_rbp = rbp - 64;
+    while (ADDR_IN_PAL(((uintptr_t *) rbp)[1])) {
         last_rbp = rbp;
-        rbp = *(unsigned long *) rbp;
+        rbp = *(uintptr_t *) rbp;
```
donporter: It might be worth a comment that we require the PAL to use frame pointers, and that we don't expect this to ever happen in the bottom-most PAL frame.

Pal/src/host/Linux/pal_host.h
```diff
@@ -230,12 +227,14 @@ struct arch_frame {
...
 
 #define PAL_FRAME_IDENTIFIER    (0xdeaddeadbeefbeef)
+#define PAL_LOCK_RECORDS        4
 
 struct pal_frame {
     volatile uint64_t           identifier;
     void *                      func;
     const char *                funcname;
     struct arch_frame           arch;
+    PAL_LOCK *                  lock_records[PAL_LOCK_RECORDS];
```
donporter: This could use a comment.  I assume the scheme is that you might acquire some locks in an exception handler and need a mechanism to free them?

chiache:
Closing this PR because it has been made obsolete due to #281 and #279.

Issue #117, opened 2017-11-25
LibOS and Pal don't build on Ubuntu 17.10


mkow:
I tried to build Graphene on Ubuntu 17.10, but it failed with the following errors:

### LibOS
```
graphene/LibOS$ make
Building glibc, may take a while to finish. Warning messages may show up. If this process terminates with failures, see "glibc-build/build.log" for more information.
cd glibc-build && ((make >> build.log) 2>&1 | tee -a build.log)
/usr/bin/ld: graphene/LibOS/glibc-build/csu/crt1.o: relocation R_X86_64_32S against symbol `__libc_csu_fini' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/7/crtbegin.o: relocation R_X86_64_32 against hidden symbol `__TMC_END__' can not be used when making a shared object
/usr/bin/ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
make[3]: *** [graphene/LibOS/glibc-build/iconv/iconvconfig] Error 1
make[2]: *** [iconv/others] Error 2
make[1]: *** [all] Error 2
make -C shim all
make[1]: Entering directory 'graphene/LibOS/shim'
make all -C src
make[2]: Entering directory 'graphene/LibOS/shim/src'
[ shim_parser.o ]
In file included from ../include/shim_internal.h:66:0,
                 from shim_parser.c:27:
shim_parser.c: In function ‘parse_syscall_args’:
shim_parser.c:451:40: error: first argument to ‘va_arg’ not of type ‘va_list’
     const char * arg_type = va_arg(ap, const char *);
                                        ^
shim_parser.c: In function ‘skip_syscall_args’:
shim_parser.c:463:41: error: first argument to ‘va_arg’ not of type ‘va_list’
     const char * arg_type = va_arg (ap, const char *);
                                         ^
shim_parser.c:466:20: error: first argument to ‘va_arg’ not of type ‘va_list’
         va_arg(ap, const char *);
                    ^
shim_parser.c:468:20: error: first argument to ‘va_arg’ not of type ‘va_list’
         va_arg(ap, void *);
                    ^
shim_parser.c:470:20: error: first argument to ‘va_arg’ not of type ‘va_list’
         va_arg(ap, int);
                    ^
Makefile:109: recipe for target 'shim_parser.o' failed
make[2]: *** [shim_parser.o] Error 1
[...]

PR #241, opened 2018-08-24
 Addressing GLIBC building issue with GCC 5/6/7/8

chiache:
This PR addresses the issue reported by #117.

GCC 5 makes -fPIC and -pie the default options. Building GLIBC 2.19 with GCC 5/6/7/8 triggers the following compilation failure:

```
/usr/bin/ld: graphene/LibOS/glibc-build/csu/crt1.o: relocation R_X86_64_32S against symbol `__libc_csu_fini' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/7/crtbegin.o: relocation R_X86_64_32 against hidden symbol `__TMC_END__' can not be used when making a shared object
/usr/bin/ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
make[3]: *** [graphene/LibOS/glibc-build/iconv/iconvconfig] Error 1
make[2]: *** [iconv/others] Error 2
make[1]: *** [all] Error 2
```

This PR adds a GLIBC patch that detects whether -pie is the default option and forcefully disable with -no-pie.

Tested with GCC 4.8/5/6/7/8.

### DIFF
--- a/LibOS/Makefile
+++ b/LibOS/Makefile
@@ -37,6 +37,7 @@ $(GLIBC_SRC)/configure: $(GLIBC_SRC).patch
 	tar -xzf $(GLIBC_SRC).tar.gz
 	cd $(GLIBC_SRC) && patch -p1 < ../$(GLIBC_SRC).patch
 	cd $(GLIBC_SRC) && patch -p1 < ../glibc-fix-warning.patch
+	cd $(GLIBC_SRC) && patch -p1 < ../glibc-no-pie.patch
 endif
 
 .PHONY: clean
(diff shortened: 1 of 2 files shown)


donporter:
Looks good to me.  Does this patch completely address issues on 17.10?  If so, perhaps we should add other supported ubuntu versions to the CI environment?

chiache:
I test this on 18.04, so I would assume it works on 17.10 as well.

Khallu:
https://github.com/oscarlab/graphene/issues/117#issuecomment-559931419

@chiache Pls check this out. Thanks in advance.

PR #242, opened 2018-09-13
gcc -dumpmachine returns x86_64-xxx-linux-gnu

jat001:
On Linux Desktop, ```gcc -dumpmachine``` will return ```x86_64-xxx-linux-gnu``` instead of ```x86_64-linux-gnu```.

OS: Ubuntu Desktop 16.04
gcc version: 5.3.0
```gcc -dumpmachine``` returns ```x86_64-unknown-linux-gnu```

OS: Ubuntu Server 16.04
gcc version: 5.4.0
```gcc -dumpmachine``` returns ```x86_64-linux-gnu```

OS: Manjaro Linux
gcc version: 5.5.0
```gcc -dumpmachine``` returns ```x86_64-pc-linux-gnu```

### DIFF
--- a/LibOS/Makefile
+++ b/LibOS/Makefile
@@ -12,7 +12,7 @@ GLIBC_TARGET = $(addprefix $(BUILD_DIR)/,libc.so.6 ld-linux-x86-64.so.2 libpthre
 all: $(GLIBC_TARGET)
 	$(MAKE) -C $(SHIM_DIR) all
 
-ifeq ($(SYS),x86_64-linux-gnu)
+ifeq ($(findstring linux,$(SYS)),linux)
 
 .INTERMEDIATE: $(BUILD_DIR)/Build.success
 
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by donporter** [comment]
Ok to test

**Comment by donporter** [comment]
I assume this is causing a problem for you; I am curious what sort of issue?

I hesitate to just pick the 'xxx' from the build system, as this will create a lot of non-uniformity, but it seems like it might be better to specify something (possibly 'graphene' if I'm being vain and the string just needs to be non-empty).  I'd rather pass it through from the host or specify it in the manifest if we need something dynamic that matches the local files in the runtime environment.

I think it would be clearer to me what the right thing to do is if I knew more about the problem that led to this patch.

File: LibOS/Makefile [hunk 1]
```diff
@@ -12,7 +12,7 @@ GLIBC_TARGET = $(addprefix $(BUILD_DIR)/,libc.so.6 ld-linux-x86-64.so.2 libpthre
 all: $(GLIBC_TARGET)
 	$(MAKE) -C $(SHIM_DIR) all
 
-ifeq ($(SYS),x86_64-linux-gnu)
+ifeq ($(findstring linux,$(SYS)),linux)
```
donporter: I think we really want to search for both linux and x86_64, as one might be building on a 32-bit linux system.

donporter:
BTW, the problem makes more sense now.

PR #245, opened 2018-09-27
Fixing memory corruption during exception handling 

rainfld:
This PR fixes the potential memory corruption in the in-enclave exception handler. The in-enclave exception handler uses the memory below RSP register to store the sgx register context: 
```
.Lhandle_exception:
	mov SGX_GPR_RSP(%rbx), %rsi
	sub $0x90, %rsi
```
It assumes the memory below the stack top won't be used. However, this is not always true. For example, The function native_cpuid:
```
static inline void native_cpuid(unsigned int *eax, unsigned int *ebx,
                                unsigned int *ecx, unsigned int *edx)
{
    /* ecx is often an input as well as an output. */
    asm volatile("cpuid"
        : "=a" (*eax),
        "=b" (*ebx),
        "=c" (*ecx),
        "=d" (*edx)
        : "0" (*eax), "2" (*ecx));
}
```
Will be compiled to:
```
   ...
   0x000000000040059b <+5>:     mov    %rdi,-0x10(%rbp)
   0x000000000040059f <+9>:     mov    %rsi,-0x18(%rbp)
   0x00000000004005a3 <+13>:    mov    %rdx,-0x20(%rbp)
   0x00000000004005a7 <+17>:    mov    %rcx,-0x28(%rbp)
   0x00000000004005ab <+21>:    mov    -0x10(%rbp),%rax
   0x00000000004005af <+25>:    mov    (%rax),%ecx
   0x00000000004005b1 <+27>:    mov    -0x20(%rbp),%rax
   0x00000000004005b5 <+31>:    mov    (%rax),%edx
   0x00000000004005b7 <+33>:    mov    %ecx,%eax
   0x00000000004005b9 <+35>:    mov    %edx,%ecx
   0x00000000004005bb <+37>:    cpuid
   0x00000000004005bd <+39>:    mov    %ebx,%esi
   0x00000000004005bf <+41>:    mov    -0x10(%rbp),%rdi
   0x00000000004005c3 <+45>:    mov    %eax,(%rdi)
   0x00000000004005c5 <+47>:    mov    -0x18(%rbp),%rax
   0x00000000004005c9 <+51>:    mov    %esi,(%rax)
   0x00000000004005cb <+53>:    mov    -0x20(%rbp),%rax
   0x00000000004005cf <+57>:    mov    %ecx,(%rax)
 =>  0x00000000004005d1 <+59>:    mov    -0x28(%rbp),%rax
   0x00000000004005d5 <+63>:    mov    %edx,(%rax)
   ...
```
The code the arrow points to will cause segmentation fault as it uses the memory below RSP/RBP to retrieve the values stored before exception (cpuid) but the memory has already been corrupted during the exception handling. 

This PR includes: 

(1) The implementation of using a per-thread memory (TLS field) to store sgx context and the app could retrieve after the exception;

(2) A test case for testing cpuid functionality and reproducing the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/245)

### DIFF
--- a/LibOS/shim/test/native/cpuid.c
+++ b/LibOS/shim/test/native/cpuid.c
@@ -1,0 +1,28 @@
+#include <stdio.h>
+
+static inline void native_cpuid(unsigned int *eax, unsigned int *ebx,
+                                unsigned int *ecx, unsigned int *edx)
+{
+        /* ecx is often an input as well as an output. */
+        asm volatile("cpuid"
+            : "=a" (*eax),
+              "=b" (*ebx),
+              "=c" (*ecx),
+              "=d" (*edx)
+            : "0" (*eax), "2" (*ecx));
+}
+
+int main(int argc, char **argv)
+{
+  unsigned eax, ebx, ecx, edx;
+
+  eax = 1; /* processor info and feature bits */
+  native_cpuid(&eax, &ebx, &ecx, &edx);
+
+  printf("stepping %d\n", eax & 0xF);
+  printf("model %d\n", (eax >> 4) & 0xF);
+  printf("family %d\n", (eax >> 8) & 0xF);
+  printf("processor type %d\n", (eax >> 12) & 0x3);
+  printf("extended model %d\n", (eax >> 16) & 0xF);
+  printf("extended family %d\n", (eax >> 20) & 0xFF);
+}
--- a/LibOS/shim/test/native/cpuid.c
+++ b/LibOS/shim/test/native/cpuid.c
@@ -3,26 +3,26 @@
 static inline void native_cpuid(unsigned int *eax, unsigned int *ebx,
                                 unsigned int *ecx, unsigned int *edx)
 {
-        /* ecx is often an input as well as an output. */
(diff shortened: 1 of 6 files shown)


donporter:
Ok to test

rainfld:
Thanks for suggestions @donporter . I will create a unit test at PAL layer with python scripts. Also, I think maybe it's worth putting cpuid test in native as well since it exercises a very important logic of graphene-sgx. But I don't know if any restrictions for the programs under native.   

BTW, the jenkin fails due to a bug in the PR. I have a fix already. will update the branch and re-run the test later soon.

dimakuv:
@rainfld Could you close this PR? This is superseded by #625.

HW42:
If you would like to keep the test case (mine is on Pal level, this is on shim level and more tests usually don't hurt) please make it a regression test (which then will be run automatically by Jenkins).

rainfld:
PR: https://github.com/oscarlab/graphene/pull/625 addressed the issue. Closed this one

Issue #246, opened 2018-10-08
Uninitialized hdl_hdr.fds


bigdata-memory:
In the Pal source files
    Pal/src/host/FreeBSD/db_streams.c:292
    Pal/src/host/Linux-SGX/db_streams.c:282
    Pal/src/host/Linux/db_streams.c:295
    1)      The struct hdl_hdr is defined as local variable
    2)      The ‘fds’ member of struct hdl_hdr is used without Initialization
    3)      The pointer to struct hdl_hdr is assigned to iov[0]
    4)      The iov is assigned to struct hdr as hdr.msg_iov
    5)      Do syscall with struct hdr as one of parameters
    The bug may cause the unexpected result from syscall.

rainfld:
This issue has been resolved and PR merged. Close

PR #247, opened 2018-10-08
Uninitialized hdl_hdr.fds bugfixes (#246)

bigdata-memory:
In the Pal source files
Pal/src/host/FreeBSD/db_streams.c:292
Pal/src/host/Linux-SGX/db_streams.c:282
Pal/src/host/Linux/db_streams.c:295
1)	The struct hdl_hdr is defined as local variable
2)	The ‘fds’ member of struct hdl_hdr is used without Initialization
3)	The pointer to struct hdl_hdr is assigned to iov[0]
4)	The iov is assigned to struct hdr as hdr.msg_iov
5)	Do syscall with struct hdr as one of parameters
The bug may cause the unexpected result from syscall.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/247)

### DIFF
--- a/Pal/src/host/FreeBSD/db_streams.c
+++ b/Pal/src/host/FreeBSD/db_streams.c
@@ -284,6 +284,7 @@ int _DkSendHandle (PAL_HANDLE hdl, PAL_HANDLE cargo)
     if (ret < 0)
         return ret;
 
+    hdl_hdr.fds = 0;
     hdl_hdr.data_size = ret;
     int fds[MAX_FDS];
     int nfds = 0;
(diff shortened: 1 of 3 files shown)


donporter:
Ok to test.

Pal/src/host/Linux-SGX/sgx_main.c
```diff
@@ -644,6 +646,7 @@ int load_manifest (int fd, struct config_store ** config_ptr)
 
     if (ret < 0) {
         SGX_DBG(DBG_E, "can't read manifest: %s\n", errstring);
+        free(config);
```
rainfld: It's also better to munmap "config_raw" as well.

bigdata-memory: Sure, I will do that, thanks,

PR #248, opened 2018-10-09
Not clean up "config" and "config_raw" in error handling code

bigdata-memory:
The config variable points to a new allocated memory block inside of load_manifest() but it isn't cleaned up on two error handling branches.
also the mapped config_raw is not unmapped on last error handling branch before return.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/248)

### DIFF
--- a/Pal/src/host/Linux-SGX/sgx_main.c
+++ b/Pal/src/host/Linux-SGX/sgx_main.c
@@ -617,6 +617,7 @@ static void create_instance (struct pal_sec * pal_sec)
 
 int load_manifest (int fd, struct config_store ** config_ptr)
 {
+    int retval = -EINVAL;
     int nbytes = INLINE_SYSCALL(lseek, 3, fd, 0, SEEK_END);
 
     if (IS_ERR(nbytes))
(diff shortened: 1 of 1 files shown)


donporter:
Ok to test

Pal/src/host/Linux-SGX/sgx_main.c
```diff
@@ -644,6 +646,8 @@ int load_manifest (int fd, struct config_store ** config_ptr)
 
     if (ret < 0) {
         SGX_DBG(DBG_E, "can't read manifest: %s\n", errstring);
+        INLINE_SYSCALL(munmap, 2, config_raw, nbytes);
+        free(config);
```
donporter: I think this function is right on the cusp of needing a single error case at the end with gotos.  I think the code is ok as-is, but I think one more system call/cleanup case would tip this into having one exception case at the end.

bigdata-memory:
Sure, I will do it. thanks.

bigdata-memory:
Hi @rainfld and @donporter, your suggestions have been merged to this PR, Please review it again, thanks.

donporter:
I think this can be merged on Sunday or when it gets a third approval.

donporter:
Test this please

Issue #239, opened 2018-08-20
shim_do_epoll_wait() ignores user-supplied timeout


dimakuv:
I test Redis 5.0 and noticed that [`shim_do_epoll_wait()`](https://github.com/oscarlab/graphene/blob/master/LibOS/shim/src/sys/shim_epoll.c) completely ignores user-supplied `timeout` argument.

In case of Redis in Cluster mode, this broke the failure-detection semantics because `poll()` waited indefinitely for client connections and thus did not send periodic heartbeat messages to other instances in the cluster. This led Redis instances to believe that all other instances are failed and panic.

I don't know why this logic is used in Graphene; I notice it was there since the very beginning. Is it just a bug or there is some rationale to ignore user-supplied timeout values?

I fixed this issue as follows:
```
diff --git a/LibOS/shim/src/sys/shim_epoll.c b/LibOS/shim/src/sys/shim_epoll.c
index b936005..e42de90 100644
--- a/LibOS/shim/src/sys/shim_epoll.c
+++ b/LibOS/shim/src/sys/shim_epoll.c
@@ -316,8 +316,15 @@ retry:

     unlock(epoll_hdl->lock);

-    PAL_HANDLE polled = DkObjectsWaitAny(nread ? npals + 1 : npals, pal_handles,
-                                         nread ? NO_TIMEOUT : 0);
+    /* BUGFIX: Before `timeout` was never considered */
+    PAL_HANDLE polled = NULL;
+    if (timeout >= 0) {
+        polled = DkObjectsWaitAny(nread ? npals + 1 : npals, pal_handles,
+                nread ? timeout : 0);
+    } else {
+        polled = DkObjectsWaitAny(nread ? npals + 1 : npals, pal_handles,
+                nread ? NO_TIMEOUT : 0);
+    }

     lock(epoll_hdl->lock);

```

(The `if (timeout >= 0)` is not strictly necessary, but I wanted to be on the safe side.)

donporter:
I am not aware of any rationale for this.  In general, the goal is compatibility with Linux, so this seems like a clear bug.

I think the fix is pretty straightforward.  Maybe just set timeout to NO_TIMEOUT and have one call to DkObjectsWaitAny?

It would be good to have a unit test for this.  I suspect one may already exist in LTP that just isn't passing.  

But otherwise, this fix looks like a no-brainer to me.

dimakuv:
PR is posted, with a unit test. See PR #249.

PR #249, opened 2018-10-09
Fixes oscarlab/graphene#239 (shim_do_epoll_wait ignores user-supplied timeout)

dimakuv:
A trivial fix and a (less trivial) unit test for regression.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/249)

### DIFF
--- a/LibOS/shim/test/regression/epoll_wait_timeout.c
+++ b/LibOS/shim/test/regression/epoll_wait_timeout.c
@@ -1,0 +1,126 @@
+/* -*- mode:c; c-file-style:"k&r"; c-basic-offset: 4; tab-width:4; indent-tabs-mode:nil; mode:auto-fill; fill-column:78; -*- */
+/* vim: set ts=4 sw=4 et tw=78 fo=cqt wm=0: */
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/epoll.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netdb.h>
+
+/* On success, a file descriptor for the new socket is returned.
+ * On error, -1 is returned. */
+static int create_and_bind (char *port) {
+    struct addrinfo hints;
+    struct addrinfo *result, *rp;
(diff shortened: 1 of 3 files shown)


LibOS/shim/test/regression/epoll_wait_timeout.c
```diff
@@ -0,0 +1,119 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/epoll.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netdb.h>
+
+static int create_and_bind (char *port) {
```
donporter: Consider adding a comment about how to interpret the return value.  This function also seems to have some failure cases that do not print a clear error.

LibOS/shim/test/regression/epoll_wait_timeout.c
```diff
@@ -0,0 +1,119 @@
...
+
+#define MAXEVENTS 64
+
+int main (int argc, char *argv[]) {
+    int sfd, s, n;
+    int efd;
+    struct epoll_event event;
+    struct epoll_event *events;
+
+    sfd = create_and_bind("8000");
```
donporter: You may need to make the port a command-line argument.  We have had a few issues with jenkins where, if the same ephemeral port is used in quick succession, the test will spuriously fail.

donporter:
Ok to test

dimakuv:
Updated the unit test as per @donporter comments.

donporter:
I think we can merge this Sunday or when we get a third approval.

donporter:
CI Failed on lat_proc exec, which is probably flaky.  Let's try again.

donporter:
Test this please

mkow:
I know I'm quite late, but I think it's worth fixing two things:
* Checking return value from `epoll_wait`
* I'm not sure if something lower than 1s wouldn't be better, is there any reason we wait here for so long? If not, I'd use something like 50ms.

What do you think?
We can do that in a separate PR if agree with these changes.

donporter:
I"m ok with these changes in a separate PR

PR #251, opened 2018-10-11
No suffix to integer literals cause overflow

bigdata-memory:
a left shift operator applys on a literal number 1 cause overflow
so unsigned long long type suffix should be appended to that literal
that makes it has same type as xfrm and xfrm for safety

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/251)


donporter:
Ok to test

PR #253, opened 2018-10-14
Fix kernel warning when VM debugging is enabled

woachk:
Just removing some spaces.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/253)


donporter:
This is a new one to me.  Can you point me to either documentation that the name to kmem_cache_create() should not have spaces, or give some specifics on the version you used to create this?  I am not able to reproduce this warning.

The change itself looks harmless, although the spaces make for easier reading.

woachk:
@donporter line 93 of https://github.com/torvalds/linux/blob/master/mm/slab_common.c is where the warning is triggered.

donporter:
Thanks!

PR #257, opened 2018-10-18
Make the Skeleton PAL buildable again

woachk:
This makes the Skeleton PAL buildable on macOS for the OS-dependent part.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/257)

### DIFF
--- a/Pal/src/host/Skeleton/db_events.c
+++ b/Pal/src/host/Skeleton/db_events.c
@@ -44,7 +44,7 @@ int _DkEventSet (PAL_HANDLE event, int wakeup)
     return -PAL_ERROR_NOTIMPLEMENTED;
 }
 
-int _DkEventWaitTimeout (PAL_HANDLE event, int timeout)
+int _DkEventWaitTimeout (PAL_HANDLE event, uint64_t timeout)
 {
     return -PAL_ERROR_NOTIMPLEMENTED;
 }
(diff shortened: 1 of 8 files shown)


donporter:
Very cool!  Thanks for the fixes.  

I assume you are testing this on OS X?

I'd recommend we add something to CI once this is merged to test the build, both on Linux and OS X.

donporter:
Ok to test.

woachk:
@donporter yeah, I'm personally testing on macOS.

PAL on Linux, Linux-SGX and FreeBSD uses ```int _DkCreatePhysicalMemoryChannel (PAL_HANDLE * handle, unsigned long * key)```, with unsigned long*, as does Pal/src/pal_internal.h, which is why the change was done, maybe transitioning those to PAL_NUM* would be a valid choice too. :)

donporter:
Got it.  I'll do a separate PR to normalize this.

donporter:
@woachk I'd like to get this PR landed.  Can I trouble you to rebase onto current master (and drop my merge of master into your branch)?  If CI doesn't pass, I'll take a more careful look at what is going wrong.

donporter:
@woachk Can you resolve the conflict with master that has come up?  I can also do this for you, but I'd rather not "steal" your credit.  Thanks!

woachk:
@donporter Yeah I will.

woachk:
@donporter done.

donporter:
lat_proc exec failure (again).  I'll go ahead and rebase this to master.

PR #258, opened 2018-10-18
Fixing warnings reported by -Wall

rainfld:
This PR fixes the code currently causing warning reported by -Wall.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/258)


donporter:
This is failing (again!) on futex_wait03.  This was on the flaky list, and should not have been placed back on the passed list.  As in #263 , this should only be reenabled if it runs in a loop and we have a credible bug fix.

Would you mind just deleting this test from the PASSED list in this PR?  I can also open a separate PR if you prefer.

PR #259, opened 2018-10-18
Fix null test and write test issues about the stat parameter

bigdata-memory:
This issue may cause SegFault if pass null to state,
and miss the validation by test_user_memory(...) if not null.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/259)


bigdata-memory:
@rainfld Yes, I will add it, thanks.

donporter:
Ok to test.

PR #260, opened 2018-10-18
Fix an error test on the result of mmap syscall

bigdata-memory:
The Macro IS_ERR should not be used to test the result value of address type.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/260)

### DIFF
--- a/Pal/src/security/Linux/main.c
+++ b/Pal/src/security/Linux/main.c
@@ -300,7 +300,7 @@ static int load_static (const char * filename, void ** load_addr,
     base = INLINE_SYSCALL(mmap, 6, NULL, maplength, c->prot,
                           MAP_PRIVATE | MAP_FILE, fd, c->mapoff);
 
-    if (IS_ERR(base)) {
+    if (IS_ERR_P(base)) {
         ret = -ERRNO_P(base);
         goto out;
     }


dimakuv:
I'm not sure I understand the difference between `IS_ERR()` and `IS_ERR_P()`. They are defined as aliases to `INTERNAL_SYSCALL_ERROR` and `INTERNAL_SYSCALL_ERROR_P` respectively, and they are defined as follows:

```c
#undef INTERNAL_SYSCALL_ERROR
#define INTERNAL_SYSCALL_ERROR(val) ((val) < 0)

#undef INTERNAL_SYSCALL_ERROR_P
#define INTERNAL_SYSCALL_ERROR_P(val) \
  ((unsigned long) (val) >= -4095L)
```

The `mmap()` syscalls returns `-1` on error. I struggle to understand then why `IS_ERR_P()` is better than `IS_ERR()` for checking `mmap` result.

donporter:
Ok to test.

donporter:
This build failed on the pipe04,1 LTP test, which seems possibly flaky (and definitely unrelated to this PR).

donporter:
@bigdata-memory Can I get you to rebase your change to the current master, and drop the merges of master from this branch (i.e., one commit with just your changes)?  If CI still doesn't pass, I'll take a deeper look at the CI issues.  I would like to land this PR.

Thanks!

PR #263, opened 2018-10-21
Take waitpid03 (back) out of CI. 

donporter:
It has been flaky, and should not have gone back into CI without pretty exhaustive testing.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/263)


donporter:
BTW, if you look at the history, I took this one out once before, and it was re-added erroneously because another fix was thought to address the problem (it didn't).

We should only put this test back if it runs in a loop many times without hanging (and with a credible bugfix for the root issue).

donporter:
This round of CI failed on the same test addressed in #225 .  Let's try again with CI, but also land 225.

PR #268, opened 2018-10-29
Normalize this function's definition.

donporter:
In reviewing #257, it became clear that this function's definition is not consistent between headers and implementation.  Changing to a uint64_t.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/268)


donporter:
Test this please

PR #269, opened 2018-10-30
Fix kernel warning with CONFIG_DEBUG_VM on

woachk:
Botched totally the other request so this is it made again...

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/269)

### DIFF
--- a/Pal/ipc/linux/graphene-ipc.c
+++ b/Pal/ipc/linux/graphene-ipc.c
@@ -982,7 +982,7 @@ static int __init gipc_init(void)
 #endif
 
 	/* Register the kmem cache */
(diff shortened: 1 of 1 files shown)


donporter:
Ok to test

chiache:
This patch looks harmless to me, but I am not sure what exactly is the rationale. Would you mind adding some more explanation to the PR description?

woachk:
@chiache line 93 of https://github.com/torvalds/linux/blob/master/mm/slab_common.c

donporter:
I know I had already approved, but is there a good reason not to use underscores "_" vs just smashing the words together?

woachk:
@donporter - and _ are safe to use

PR #272, opened 2018-11-01
Improving the warning messages in SGX PAL

chiache:
When the Intel SGX driver is not loaded, the SGX PAL shows the following warning message:

```
cannot open device /dev/gsgx, possibly the kernel module is not loaded.
```

This warning message is confusing because the Graphene-SGX driver is actually loaded. This PR fixes the code to show the proper warning messages. In addition, the PR also adds a warning message for the check for WRFSBASE instruction availability.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/272)


donporter:
@yamahata: Thanks for the review.  Would you mind resolving your open comments, or clarifying what else you think needs to be done here?

PR #278, opened 2018-11-02
Using leaf 0xb and subleaf 1 to get the correct cpu (logic core) number

rainfld:
Currently to get the number of cpus (logic cores), graphene uses cpuid with leaf 4 and extracts the values from EAX[26:31]. However, the number is not the correct value of cpu logic cores.  
According to SDM, "Bits 31 - 26: **Maximum number of addressable IDs for processor cores** in the physical package". The maximum number of addressable IDs for processor cores (not logic cores) is not necessarily equal to the number of logic cores the machine has. In practice, it often returns 8 on the platforms with 4 cores or 12 cores. So far at least causing the regression tests failed. 

To fix, we should use cpuid with leaf 0xb and subleaf 1, and extract the values from EBX[0:15] In SDM, this means "Bits 15 - 00: **Number of logical processors** at this level type." Subleaf 1 represents the number of logic cores across different physical cores.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/278)

### DIFF
--- a/Pal/src/host/FreeBSD/db_main.c
+++ b/Pal/src/host/FreeBSD/db_main.c
@@ -358,10 +358,18 @@ void _DkGetCPUInfo (PAL_CPU_INFO * ci)
     ci->cpu_brand = brand;
 
     cpuid(2, 1, words, 0);
-    ci->cpu_num      = BIT_EXTRACT_LE(words[WORD_EBX], 16, 24);
     ci->cpu_family   = BIT_EXTRACT_LE(words[WORD_EAX],  8, 12) + 1;
     ci->cpu_model    = BIT_EXTRACT_LE(words[WORD_EAX],  4,  8);
     ci->cpu_stepping = BIT_EXTRACT_LE(words[WORD_EAX],  0,  4);
+
+    /* According to SDM: EBX[15:0] is to enumerate processor topology 
+     * of the system. However this value is intended for display/diagnostic
+     * purposes. The actual number of logical processors available to
+     * BIOS/OS/App may be different. We use this leaf for now as it's the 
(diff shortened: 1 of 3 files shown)


chiache:
Is this something specific to Intel CPUs or it also applies to other vendors?

rainfld:
The two leaves only apply to Intel processors (AMD doesn't have these two leaves reserved). Looks like under other platforms except SGX, graphene doesn't try to get the cpu number. Under Linux-SGX, I think it's safe to assume running under Intel processor anyway.

chiache:
The same code is used for Intel platforms in Linux PAL and FreeBSD PAL, too:

https://github.com/oscarlab/graphene/blob/f07dfd052f476912617de74a0835f83d63f309eb/Pal/src/host/Linux/db_main.c#L391:
```    if (!memcmp(vendor_id, "GenuineIntel", 12)) {
        cpuid(4, 0, words);
        ci->cpu_num  = BIT_EXTRACT_LE(words[WORD_EAX], 26, 32) + 1;
    } else if (!memcmp(vendor_id, "AuthenticAMD", 12)) {
        cpuid(0x8000008, 0, words);
        ci->cpu_num  = BIT_EXTRACT_LE(words[WORD_EAX], 0, 8) + 1;
    } else {
        ci->cpu_num  = 1;
    }
```

donporter:
@rainfld I think we can go ahead and squash + rebase this PR.  I'll keep an eye on the CI issues, but I hope that I hope some pending PRs will address.

donporter:
@rainfld Please go ahead and squash this into one commit and rebase to master.

mkow:
Retest this please

PR #281, opened 2018-11-04
Fixing signal instability issues in Jenkins tests

chiache:
This PR fixes a majority of the instability issues I've seen in the Jenkins tests:

(1) Forbidding signal delivery during thread initialization and termination:

The previous implementation does not forbid a host-raised signal (e.g., SIGINT) or a thread-resuming signal (i.e., SIGCONT) to be delivered before a thread is fully initialized or after a thread is terminated. This bug causes additional upcalls to the user-defined signal handlers even AFTER a thread has died.

For stability, this PR also temporarily disable signal handling during fork(), clone(), and execve(). Signal handling is resumed after the user program resumes execution in the new thread or the new process.

(2) IPC port deletion should wake up blocking threads:

For some IPC calls, a thread needs to block on the IPC port to wait for the response. However, if the IPC helper detects failures on the IPC port, usually being closed by the other end of the port, the IPC port is deleted without notifying these blocking threads.

(3) Rewriting the signal handling convention in PALs:

The PAL design relies on a complex signal handling convention, which needs to rewind the stack to the latest LibOS-owned frame when receiving a signal within PAL. This convention causes lots of problems, including causing deadlocks when signals arrive in the middle of critical sections. The new convention simply postpones signal delivery until the ends of PAL calls. The only caveat is that any blocking operations (such as the futex() system calls) need to check if there is any pending signal when being interrupted by host signals.

(4) (Attempt) to fix an assertion: ipc/shim_ipc_nsimpl.h:991 !qstrempty(&NS_LEADER->uri)

This is supposedly a bug for the `__discover_ns()` routine, which discovers the leadership of a namespace. This assertion mainly occurs during the Bash or GCC regression tests. I am not completely sure about the root cause of this problem, but at least this PR has made the bug disappear OR will help us debug the problem in the future. The main changes in __discover_ns() are two:

- Fixing a race condition where `__discover_ns()` does not re-grab `cur_process.lock` after the IPC-based recovery has succeeded (we release the lock temporarily to prevent deadlocks).
- At the end of `__discover_ns()`, when the current process just assumes itself as the leader, set the VMID of the IPC port where the leader listens to as 0. First of all, in this case, we don't have to know who's the other end of the IPC port. Second, we normally will check if the IPC port is connecting to another process when adding the IPC port to the listening list.
- Adding assertions when `__discover_ns()` returns.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/281)


chiache:
Test this please.

chiache:
Test this please.

chiache:
Test this please

chiache:
Test this please.

chiache:
Test this please

chiache:
Test this please.

chiache:
Test this please.

chiache:
Test this please.

chiache:
Test this please.

chiache:
Test this please.

chiache:
Test this please.

chiache:
Test this please

chiache:
Test this please.

PR #287, opened 2018-11-06
Removing ^M in shim_brk.c

chiache:
I am not sure if this PR is necessary, but the issue has been annoying to me for years. `shim_brk.c` has a different file format (the "dos" format) from all of the other files. The most obvious difference is the return characters (^M) at the end of every line.

If possible, I prefer the file format to be consistent in the repository. Since we are in the middle of fixing code quality issues, I thought it will be good to address this issue. Of course, this means that we lose the ability to do `git blame` on this file, but we might as well just rip the band-aid off.

If the majority of maintainers consider this fix to be unnecessary, I am okay with keeping this issue as a small nuisance in my life.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/287)


dimakuv:
Is this the only file? I have no issues with this PR.

chiache:
@dimakuv Yes, AFAIK, this is the only file.

PR #290, opened 2018-11-07
Vdso support

yamahata:
LibOS: vDSO support

This patch series adds vDSO support to LibOS.
the first patch adds vDSO support to LibOS. It create dedicated page which contains necessary symbols in ELF format. plus fix of stack initialization to align 16bytes of stack
which is required by x86_64 ABI.

the second patch removes modification to glibc to avoid vDSO usage.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/290)


chiache:
Okay to test

chiache:
Ok to test

donporter:
I like splitting the patch up in to components.  This is nice.  It may be worth splitting some of this into smaller PRs (e.g., separating the reorg from the change, or separating out the stack alignment fix).

Can you also add some unit tests for this functionality, or point out LTP tests that would exercise this code?

yamahata:
two new PRs supercedes this PR. Let's continue there.

PR #291, opened 2018-11-07
fork optimization by cross memory attach

yamahata:
This pull request is to get comments/feedback before going further.
The current graphene ipc device driver isn't stable and unlikely for upstreaming.
so it's desirable to optimize fork with the existing linux kernel feature.

This is the first try to optimize fork. The patch series uses cross memory attach, process_vm_readv() system call so that one memory copy is done instead of two memory copy by unix pipe.
I implemented it only for Linux Pal for now and the code quality is not for merge yet.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/291)

### DIFF
--- a/LibOS/shim/include/shim_checkpoint.h
+++ b/LibOS/shim/include/shim_checkpoint.h
@@ -105,6 +105,10 @@ struct shim_gipc_entry {
 #endif
 };
 
+struct shim_cma_entry {
+    struct shim_mem_entry mem;
+};
+
 struct shim_palhdl_entry {
     struct shim_palhdl_entry * prev;
     PAL_HANDLE handle;
@@ -127,6 +131,11 @@ struct shim_cp_store {
     bool use_gipc;
     struct shim_gipc_entry * last_gipc_entry;
     int gipc_nentries;
+
+    /* entries of cma records */
+    bool use_cma;
+    struct shim_cma_entry * last_cma_entry;
+    int cma_nentries;
 
     /* entries of out-of-band data */
     struct shim_mem_entry * last_mem_entry;
@@ -424,6 +433,10 @@ struct newproc_cp_header {
         unsigned long entoffset;
         int nentries;
     } gipc;
+    struct cma_header {
+        unsigned long entoffset;
+        int nentries;
+    } cma;
 };
 
 struct newproc_header {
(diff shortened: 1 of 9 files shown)


donporter:
Thanks for your interest in the project and optimizing this code!

At a high level, the graphene-ipc driver is deprecated.  On newer kernels, we have found that it is not any faster than without.  @chiache could probably say more about his experiments here.  But if CMA can out-perform the current baseline, I don't see a lot of downside here.  Before we accept the change, I'd like to see some microbenchmark numbers.

The other high-level guidance I would give is that we want to avoid baking in Linux-host-specifics in the shim code.  Part of the overall goal is keeping the PAL interfaces expressive enough to communicate what the PAL needs to implement the functionality efficiently, but to avoid Linux idioms if we can.  It seems like you are on the right track, making the changes underneath DkPhysicalMemory* interfaces, but you might need to explain in more detail why cma-specific code needs to be added in the checkpoint/restore code.

dimakuv:
The PR proposed to use "cross memory attach" (https://man7.org/linux/man-pages/man2/process_vm_readv.2.html) to speed up forking: the child process performs `process_vm_readv()` to read forked-memory-regions from the parent process. This reduces the number of user-to-kernel data copies from two to only one, and delivers these forked-memory-regions directly to the child's LibOS.

This is a nice optimization for insecure PALs like `Linux` but it is unclear how this can improve fork performance of secure PALs like `Linux-SGX`. Recall that there is are no shared pages between two SGX enclaves, so the "cross memory attach" won't help with cross memory attach between two enclave memories.

This PR also contains an optimization to use `vmsplice()` on the parent-to-child pipe to directly copy pipe data (e.g. serialized PAL handles) instead of a classic `write()`/`read()`. Similarly to `process_vm_readv()`, this reduces the number of user-to-kernel data copies from two to one. This may be useful, but there seems to be little performance improvement (at least for Linux-SGX) -- in typical cases, fork happens at application startup (think Nginx, Apache with their pre-forked worker model) and there is very little metadata and process memory to copy over the pipe.

My last objection is that `splice()/vmsplice()` and `process_vm_readv()` are Linux-specific, whereas pipe `write()/read()` are universal across UNIXes. I see no point in sacrificing simplicity for features with unproven performance benefit.

I'm closing this PR for now. We could revive the `vmsplice()` part of this PR later, but only with benchmarking proving its usefulness for secure PALs (e.g. `Linux-SGX`).

PR #293, opened 2018-11-09
Pal: re-version _r_debug from GLIBC to PAL to reduce glibc patch

yamahata:
_r_debug symbol is interface for for debugger, like gdb, to understand
how dynamic loading is happening. So it's not used by dependent
library or executable. So there is no point to put _r_debug into GLIBC
version space for export. Put it into its own version (PAL_PRIVATE).
Then, the next patch will reduce glibc modification.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/293)

### DIFF
--- a/Pal/src/host/FreeBSD/db_rtld.c
+++ b/Pal/src/host/FreeBSD/db_rtld.c
@@ -50,9 +50,6 @@
 struct r_debug pal_r_debug =
         { 1, NULL, &pal_r_debug_state, RT_CONSISTENT, };
 
-extern __typeof(pal_r_debug) r_debug
-    __attribute ((alias ("pal_r_debug")));
-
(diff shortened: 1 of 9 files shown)


dimakuv:
Useful links to understand what's going on:
https://people.freebsd.org/~deischen/symver/freebsd_versioning.txt
https://github.com/lattera/glibc/blob/master/elf/dl-debug.c
https://github.com/lattera/glibc/blob/master/elf/rtld-debugger-interface.txt

donporter:
Please squash this into one commit and rebase to current master.  I think this one is ready.

mkow:
Retest this please

mkow:
Retest this please

PR #295, opened 2018-11-09
Pal: make _dl_debug_status versioned to PAL_PRIVATE not  GLIBC

yamahata:
dynamic symbol of _dl_debug_state is interface for debugger. Not
for depdent library or executable. So there is no point to put in
GLIBC version. this patch put _dl_debug_status symbol to PAL_PRIVATE.
So that the next patch eliminates rename in glibc. _dl_debug_status
-> __libc_dl_debug_status.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/295)

### DIFF
--- a/Pal/src/host/FreeBSD/pal.map
+++ b/Pal/src/host/FreeBSD/pal.map
@@ -1,7 +1,3 @@
-GLIBC {
-    global:
-        r_debug_state;
-};
 PAL {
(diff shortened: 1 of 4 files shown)


yamahata:
To clarify the PRs relations: the merge order is
294->295->296

PR #297, opened 2018-11-09
remove unnecessary file Pal/include/pal/.gdb_history.

yamahata:
The file, Pal/include/pal/.gdb_history, seems committed accidentally.
So remove it.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/297)


chiache:
Ok to test.

PR #301, opened 2018-11-15
Makefile: prevent gcc from fp code

yamahata:
Pal and LibOS shouldn't clobber FP registers in most cases.
i.e. system call emulation.
Update CFLAGS to prevent gcc from generating FP code
accidentally for safety for now.

It's exceptional FP is allowed. i.e. signal handlers for host,
internal thread(ipc and migration), code before entering user program.
Anyway those codes need to be checked carefully before allowing FP
code. For now, disable FP code at the moment.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/301)

### DIFF
--- a/LibOS/shim/src/Makefile
+++ b/LibOS/shim/src/Makefile
@@ -13,6 +13,7 @@ CFLAGS	= -Wall -fPIC -std=gnu99 -fgnu89-inline -Winline -Wwrite-strings \
 	  -fmerge-all-constants -Wstrict-prototypes \
 	  -Werror=implicit-function-declaration \
 	  -fno-stack-protector -fno-builtin -Wno-inline \
+	  -mno-sse -mno-mmx -mno-sse2 -mno-3dnow -mno-avx \
 	  -I../include -I../../../Pal/lib -I../../../Pal/include/pal
 ifeq ($(OMIT_FRAME_POINTER),yes)
 CFLAGS += -DOMIT_FRAME_POINTER=1
--- a/Pal/lib/Makefile
+++ b/Pal/lib/Makefile
@@ -38,7 +38,7 @@ ifeq ($(CRYPTO_PROVIDER),mbedtls)
 subdirs += crypto/mbedtls
 headers += $(wildcard crypto/mbedtls/mbedtls/*.h)
 # XXX: For now, allow aesni.c to compile.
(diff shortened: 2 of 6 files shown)


dimakuv:
Ignoring until we measure the slowdown of FP context save/restore on every syscall.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/301)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @donporter and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/301#-MP4N_39BHKNltOflTHR:-MP4N_39BHKNltOflTHS:bml42dt) (no related file):*
Blocking. I created a new PR #2036 that introduces the following logic:
- Graphene internally can freely use FP/SSE/AVX/... registers.
- The FP and SSE/AVX/... control words are saved (as part of the CPU context) before each syscall and restored after the syscall.
- The application must *not* assume that FP/SSE/AVX/... registers are preserved across syscalls. Since this is already mandated by the System V ABI, all sane applications will follow this rule.

This only adds ~20 cycles of latency on each system call. Sounds fare.

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/301)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv, @donporter, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/301#-MP4N_39BHKNltOflTHR:-MPBf8bd9cBwmuoyuZme:b4c9lyr) (no related file):*

Blocking. I created a new PR #2036 that introduces the following logic:
- Graphene internally can freely use FP/SSE/AVX/... registers.
- The FP and SSE/AVX/... control words are saved (as part of the CPU context) before each syscall and restored after the syscall.
- The application must *not* assume that FP/SSE/AVX/... registers are preserved across syscalls. Since this is already mandated by the System V ABI, all sane applications will follow this rule.

This only adds ~20 cycles of latency on each system call. Sounds fare.
</blockquote></details>

:thumbsup: 

---

PR #303, opened 2018-11-17
Addded a LOOPRUNS file to specify tests to be run multiple times.

bpjain:
The format is just like timeout file. It runs that test and its subtests the number of times specified.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/303)


dimakuv:
Closing this PR as inactive. Given that we discovered all (?) intermittent failures in LTP and submitted corresponding PRs, seems like this feature is not needed anymore.

PR #306, opened 2018-11-21
LibOS: sched_getaffinity return 0, not # of cpu

yamahata:
sched_getaffinity(2) return 0 on success, not # of cpu.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/306)


yamahata:
Oops, this behaviour is for glibc behavior, not raw system call.
this patch was wrong.

yamahata:
LibOS: sched_getaffinity return byte size, not bit size

    sched_getaffinity(2) returns the # of bytes, not bits.
    So calculate the # of bytes.
    Also it should honer input argument len.

PR #314, opened 2018-11-26
LibOS/shim/src/shim_checkpoint.c: retry DkStreamRead/DkStreamWrite

yamahata:
On EINTR/EAGAIN of DkStreaRead/DkStreamWrite, retry the PAL call
instead of error out.
Especially when it runs under debugger, EINTR often happens.
This patch makes debug easier.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/314)


donporter:
Ok to test

PR #322, opened 2018-11-26
LibOS: race condition on thread exiting

yamahata:
Once exiting thread notifies at shim LibOS level, other threads can
release stack memory and tls. But the existing thread can be still
executing resulting in SEGV. Switch stack and tls to private area to
avoid race condition.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/322)

### DIFF
--- a/LibOS/shim/include/shim_thread.h
+++ b/LibOS/shim/include/shim_thread.h
@@ -91,6 +91,10 @@ struct shim_thread {
 #ifdef PROFILE
     unsigned long exit_time;
 #endif
+
+    __libc_tcb_t exit_tcb;
+#define SHIM_THREAD_EXIT_STACK_SIZE (4096)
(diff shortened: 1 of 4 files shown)


alexandermerritt:
I submitted a PR which may be a valid regression test for this, #846

dimakuv:
This PR fails `alarm06,1` in LTP tests (both Jenkins and Jenkins-Debug). @yamahata, could you look into this?

dimakuv:
I assume that #641 fixed the underlying issues. I see no reason to keep this PR open. (But feel free to re-open it if you believe it is still useful.)

yamahata:
#641 is orthogonal to this issue.

This issue can happen with Pal/Linux (non-SGX).
The problem is  the emulation of clone(CLONE_CHILD_CLEARID) and guest pthread.

PR #324, opened 2018-11-27
Remove modelines from all source files

### LINKED ISSUE #304: Remove or update modelines in source files?
When the project started, we used a lot of Emacs and Vim modelines. Here is an example:
 
```
/* -*- mode:c; c-file-style:"k&r"; c-basic-offset: 4; tab-width:4; indent-tabs-mode:nil; mode:auto-fill; fill-column:78; -*- */
/* vim: set ts=4 sw=4 et tw=78 fo=cqt wm=0: */
```

I personally like them because they are convenient for per-project settings, but I don't know how other people feel about them. I assume with the new coding style guideline, we need to either update them or remove them once for all. Do we consider them too annoying? Maybe move them to something like a `.vimrc` file?

Suggestions are welcome.

chiache:
Fix #304.

In #304, I suggest we remove all the modelines. This PR will purge the modelines from all source and header files.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/324)


yamahata:
Although thia pathch itself looks good to me, the style gude sevtion in the contributing should be updated, at least mode line as example

PR #327, opened 2018-11-28
Linux-SGX/Makefile.Test: define SGX_SIGNER_KEY only when not defined

yamahata:
define SGX_SIGNER_KEY only when not defined so that user can define
SGX_SIGNER_KEY as error message.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/327)


donporter:
Ok to test

PR #332, opened 2018-11-29
Fix the local attestation protocol for process creation on SGX

chiache:
This PR fixes the protocol of exchanging the session key and local attestation during the creation of a new enclave. The fix is based on #233, which deprecated an useless RSA key generated during enclave initialization.  However, #233 does not correctly implement the protocol for creating a secure, mutually trusted RPC channel.

The main problem of the original code is that the local attestation from the Intel CPU is NEVER verified by the receiving side of the attestation. The SGX PAL needs to obtain the report key from the CPU and use the key to generate the AES-CMAC hash of the report body, to formally authenticate the communicating enclave.

This PR is also meant to document the requirements and specifications for process (enclave) creation on SGX.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/332)


dimakuv:
Jenkins, retest Jenkins please (`creat01` failed)

PR #337, opened 2018-12-04
glibc: optimize SYSCALLDB

yamahata:
Linux x86 syscall convention clobbers %rcx and %r11.
%rcx can be used instead of %rbx without save/restore.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/337)


donporter:
Ok to test

PR #338, opened 2018-12-04
glibc: optimize syscalldb for normal path

yamahata:
Now syscalldb in normal path on return include jmp.
jmp can be eliminated and penalize error path.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/338)


donporter:
Ok to test

PR #339, opened 2018-12-04
shim/Makefile: don't use -e for ld, but use ENTRY in linker script

yamahata:
For consistency use ENTRY(shim_start) in linker script, not command
line option of -e shim_start.
This also helps to understand how link is done. only shim.lds needs
to be read. don't have to dig into Makefile for command line.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/339)


donporter:
Looks like this PR is ready to go.  @yamahata Please go ahead and squash+rebase to current master.

PR #342, opened 2018-12-04
LibOS/glibc: syscall-template.S ignores redzone

yamahata:
For trivial system call, sysdeps/unix/syscall-template.S is used to generate
system call stub functions. which ignores redzone.
Teach it redzone.
Also use %rcx instead of %rbx without pushq/popq as %rcx is clobbered
register as system call abi.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/342)


donporter:
Ok to test

donporter:
Retest this please

PR #345, opened 2018-12-04
LibOS: syscallas.S should use GOTPCREL call instead of absolute call

yamahata:
libsysdb.so is complied as pic so GOT should be used.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/345)


donporter:
Ok to test

PR #346, opened 2018-12-05
LibOS: restore_context() clobbers redzone accidentally

yamahata:
restore_context() happens to use %rsp - 8 to return which is in redzone.
which clobbers redzone. It shouldn't be clobered.
So far it seems we're simply lucky.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/346)


donporter:
Ok to test

PR #347, opened 2018-12-05
RFC: Signal delivery in user program via signal stack frame

yamahata:
RFC: WIP: LibOS/shim_signal: deliver signal when user program running
    
    This patch is still Work-In-Progress.
    The purpose of this patch is to get feedback before going further
    and trigger discussion.
    I believe the patch is enough to show the idea and how to implement
    it.
    
    Deliver signal when executing in user program via signal stack
    frame, not direct calling handler within LibOS.
    
    With this patches, (emulated) signal is delivered only when
    user program is running by setting up signal stack frame.
    If LibOS or PAL is running with asynchronously signal, the (async)
    signal is queued and the signal stack frame is created right before
    system call is return.
    This also opens up the capability to support sigaltstack.
    
    The current implementation of invoking registered signal handler
    directly within LibOS is problematic.
    It causes several issues below and hard to solve them.
    
    - race condition in LibOS because invoking registered signal handler
      directly in host signal handler even when cpu is executing within
      LibOS. If the blocking emulated system call is executing, it's very
      likely. It incurs much complexity.
    
    - nested signal can't be handled well.
      If async signal(e.g. SIGALARM) is delivered and then synchronous
      signal(e.g. SIGSEGV) within signal handler is delieved,
      The current implementation tries to queue SIGSEGV, and loop in
      SIGSEGV.
    
    - sigaltstack can't be implemented well
      Typically language runtime uses sigaltstack. It's desirable to
      support those features well.
    
---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/347)

### DIFF
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -170,8 +170,12 @@ static inline void do_pause (void);
     do { debug("%s (" __FILE__ ":%d)\n", __func__, __LINE__); } while (0)
 
 /* definition for syscall table */
+extern unsigned long fpu_xstate_size;
+void handle_sysret_signal(void);
 void handle_signal (bool delayed_only);
+int handle_next_signal(ucontext_t * user_uc);
 long convert_pal_errno (long err);
+void __sigreturn(mcontext_t * user_uc);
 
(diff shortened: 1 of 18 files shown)


yamahata:
https://github.com/oscarlab/graphene/pull/1218 supersedes this.

PR #351, opened 2018-12-05
[LibOS] Improve the for condition expression to avoid dereference

bigdata-memory:
polls was constructed by nfds, with last index as nfds - 1. When calling __do_poll, nfds was passed as a parameter and used as index to refer polls as npolls.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/351)

### DIFF
--- a/LibOS/shim/src/sys/shim_poll.c
+++ b/LibOS/shim/src/sys/shim_poll.c
@@ -126,7 +126,7 @@ static int __do_poll (int npolls, struct poll_handle * polls,
 
     lock(map->lock);
 
-    for (p = polls ; p < &polls[npolls] ; p++) {
+    for (p = polls ; p < polls + npolls ; p++) {
         bool do_r = p->flags & DO_R;
         bool do_w = p->flags & DO_W;
 


bigdata-memory:
___
*[LibOS/shim/src/sys/shim_poll.c, line 129 at r1](https://reviewable.io/reviews/oscarlab/graphene/351#-LT-GAwaCUYIdn_48KQx:-LTONzGB8e7Do-d8KpjU:b-l07zlf) ([raw file](https://github.com/oscarlab/graphene/blob/50b35974f562bb7a2d05dbb1059d34b5bc06884e/LibOS/shim/src/sys/shim_poll.c#L129)):*
<details><summary><i>Previously, donporter (Don Porter) wrote…</i></summary><blockquote>

+1
</blockquote></details>

It is about the expression "pools[npolls]" not the "p < &polls[npolls]"

bigdata-memory:
The mark [482509]  is used for me to map  this PR to my local branch that helps me to prevent from lost in bunch of branches

bigdata-memory:
looks not an unit test that illustrates it.

bigdata-memory:
___
*[LibOS/shim/src/sys/shim_poll.c, line 129 at r1](https://reviewable.io/reviews/oscarlab/graphene/351#-LT-GAwaCUYIdn_48KQx:-LTP9P4I7wr3i2T7CK2I:bmx9myv) ([raw file](https://github.com/oscarlab/graphene/blob/50b35974f562bb7a2d05dbb1059d34b5bc06884e/LibOS/shim/src/sys/shim_poll.c#L129)):*

`pools[npolls]` is incorrect (i.e. it's undefined behavior), but `&pools[npolls]` is perfectly fine. There's an if in the standard for this case (taking address of an element after the last with `&arr[_countof(arr)]`).
I'll try to find the source for this and post it here (I don't remember where it was exactly and in which version of the standard).
</blockquote></details>

Thanks for the explanation. The iteration approach maybe a little bit confused, could be refactored as  
for (idx = 0; idx < npolls; p=&polls[++idx]) ?

bigdata-memory:
___
*[LibOS/shim/src/sys/shim_poll.c, line 129 at r1](https://reviewable.io/reviews/oscarlab/graphene/351#-LT-GAwaCUYIdn_48KQx:-LTPDGwd5qCa491-ttGI:bw38ymi) ([raw file](https://github.com/oscarlab/graphene/blob/50b35974f562bb7a2d05dbb1059d34b5bc06884e/LibOS/shim/src/sys/shim_poll.c#L129)):*

The iteration loop we have in the code is one of the standard ways of iterating an array and is pretty common in C sources.

In the version you suggested you need to declare two variables (`p` and `idx`) and initialize one of them outside the loop. IMO the loop we have right now is just simpler.
</blockquote></details>

Thanks for the information, Yes I got it, the functionality is correct. but how about using for (p = polls ; p < polls + npolls ; p++) ? that could be more efficient as you mentioned while not confusing it as the array declaration.

bigdata-memory:
Done, thanks.

bigdata-memory:
Hi, all change requests have been responded, please review, thanks.

donporter:
Ok to test

PR #353, opened 2018-12-05
[LibOS] Fix a memory leak in __map_elf_object() about variable pointer new_phdr

bigdata-memory:
new_phdr is allocated with malloc, if seek file or read file failed, the function will goto call_lose that will return NULL.
However, before returning NULL, there is no check inside call_lose to check if new_phdr was allocated successfully or not,
causing potencial resource leakage of new_phdr

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/353)

### DIFF
--- a/LibOS/shim/src/elf/shim_rtld.c
+++ b/LibOS/shim/src/elf/shim_rtld.c
@@ -404,10 +404,15 @@ call_lose:
     if (type == OBJECT_LOAD &&
         header->e_phoff + maplength <= (size_t) fbp_len) {
         ElfW(Phdr) * new_phdr = (ElfW(Phdr) *) malloc (maplength);
+        if (!new_phdr) {
+            errstring = "new_phdr malloc failure";
+            goto call_lose;
+        }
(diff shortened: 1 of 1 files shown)


donporter:
Ok to test

bigdata-memory:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/353#-LTO1fSZ8q1g2e73rJ37:-LTObL0vEF_qUv72NF0U:b-ot319r) (no related file):*

Could you please remove `[482645]` from the commit message?

For consistency: Could you please write in commit messages what a specific change does, instead of what's wrong with the code without it? This way it will be consistent with other commits in our repo. So, here I'd change `Variable new_phdr going out of scope leaks the storage it points to` to `[LibOS] Fix a memory leak in __map_elf_object()`.
</blockquote></details>

Revised according to your comment, thanks.

---

bigdata-memory:
@mkow Added the missing errorstring, thanks. 
@yamahata Yes, we will take care of that in another patch, thanks.

PR #354, opened 2018-12-05
[LibOS] Don't use uninitialized return value in walk_mounts()

bigdata-memory:
ret is declared without initialization. It depends on (ret = (*walk) (mount, arg) for the initial assignment;
however, the list might be NULL and the assignment can not be performed leaving ret in un-initialized state.
Thus ret should be properly initialized in this case

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/354)

### DIFF
--- a/LibOS/shim/src/fs/shim_fs.c
+++ b/LibOS/shim/src/fs/shim_fs.c
@@ -508,7 +508,7 @@ int walk_mounts (int (*walk) (struct shim_mount * mount, void * arg),
                  void * arg)
 {
     struct shim_mount * mount, * n;
-    int ret;
+    int ret = 0;
     int nsrched = 0;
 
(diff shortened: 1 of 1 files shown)


donporter:
Ok to test

bigdata-memory:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/354#-LTNxMO_0qbTv7VBVRQh:-LThnjUN8NqRz-kxdodb:b-r18wie) (no related file):*

Could you please remove `[482697]` from the commit message?

Also, I'd prefer to change the commit message from `Using uninitialized value ret in return statement` to `[LibOS] Don't use uninitialized return value in walk_mounts()`.

</blockquote></details>

done, thanks.

---

PR #358, opened 2018-12-05
[LibOS] Fix an uninitialized value n in wait_event

bigdata-memory:
Variable  n is used without initilization.
If if (!DkObjectsWaitAny(1, &e->event, NO_TIMEOUT)) condition is true,
then code will skip the assignment of n with n = DkStreamRead(e->event, 0, 1, &byte, NULL, 0);
and might finish the while loop. n should be properly initialized to prevent potenical undefined behavior
[482723]

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/358)


bigdata-memory:
done, thanks.

PR #359, opened 2018-12-05
[LibOS] Fix the storage leaks pointed by new_map in case of malloc fails to fd_new

bigdata-memory:
get_new_handle_map will return a pointer for new_map, which could be NULL after examing the implementation of get_new_handle_map function.
In this case, accessing new_map->fd_top will cause undefined behavior. We need to add a check for new_map before accessing its member.
if the code error out and function return the error, allocated new_map is not freed, causing potencial information leakage
[482567]

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/359)


bigdata-memory:
Done, thanks.

PR #367, opened 2018-12-06
Makefile: Use symlink instead of cp to populate Runtime dir

yamahata:
Currently cp to Runtime dir is used. Sometimes it is confusing when
binary re-build under Pal or LibOS. So symlink is better. Use symlink
instead of cp to populate Runtime dir.
This patch create ln_sf rule and use it instead of cp. and applies to
other recipe with cp.
In this context, creating symlink of glibc libraries is mess. this
patch also cleans it up to use Makefile instead of embedding the logic
into python script.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/367)


yamahata:
Now 344 has been merged. I rebased this patch for merge.

PR #369, opened 2018-12-07
[LibOS] Variable len is used without initilization

zhangckid:

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/369)


donporter:
Ok to test

PR #307, opened 2018-11-25
Clean up and optimize the integration of docker container and graphene-SGX

zhangckid:
The dockerfile instruction RUN will create new layers
    when it is executed, original codes will increase the size of
    the final image and causes potential cache issues.
    So we need to employ shell tricks to keep the layers as small as possible.
    
    The detail:
    https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/307)

PR #373, opened 2018-12-09
Jenkinsfiles: Optimize the docker file for ubuntu

zhangckid:
The dockerfile instruction RUN will create new layers
when it is executed, original codes will increase the size of
the final image and causes potential cache issues.
So we need to employ shell tricks to keep the layers as small as possible.
Like #307.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/373)


mkow:
Test this please

PR #375, opened 2018-12-10
Pal/{Linux, Linux-SGX, FreeBSD}: remove close() right before dup2()

yamahata:
As dup2() system call closes target file descriptor, close() right
before dup2() doesn't make sense. or worse, it opens a window for race
condition.
So remove close() right before dup2()

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/375)


mkow:
Test this please

PR #391, opened 2018-12-17
[Build] Remove residual references to nonexistent exec_victim

dimakuv:
PR #215 removed file `exec_victim.c` under `test/native`. However, there were still some residual references to this non-existent file in the `manifest`, which broke the build. This PR removes these residual references.

Note: this only affects the build of SGX PAL.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/391)


donporter:
Ok to test

PR #397, opened 2018-12-22
[Pal, Pal/Linux] Add fp registers definition to PAL_CONTEXT

yamahata:
Now fp registers aren't handled when unix signal is emulated.
As a first step to address it,  add fp registers to PAL_CONTEXT and
make Pal/Linux pass it. Pal/Linux-SGX part will follow as another patch.
LibOS part needs discussion on how unix(Linux) signal should be emulated.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/397)


donporter:
Ok to test

dimakuv:
@yamahata, please add another PR with support for SGX

yamahata:
#1238 supersedes this.

PR #401, opened 2018-12-26
[PAL] Fix the missing of closing the cached_file

bigdata-memory:
cached_file is a PAL_HANDLE, which in the case of out will be released by _DkObjectClose(cached_file).
In the normal case, before returning &obj->map, this resource allocated within the function should
also be released to prevent it leaking out side of the function scope.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/401)


donporter:
Ok to test

donporter:
Ok to test

PR #402, opened 2018-12-26
[PAL] Fix uninitialized field action.sa_flags

bigdata-memory:
action.sa_flags will only be initialized in the truth branch, while in the false branch,
only sa_handler is initialized. Before action.sa_flags |= SA_NOCLDWAIT,
action.sa_flags should be properly initialized (in this case 0x0u)

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/402)


donporter:
Ok to test

PR #406, opened 2018-12-28
[Pal/Linux] block signals instead of queuing signal

yamahata:
The malloc implementation of Pal/Linux isn't async signal safe. and
the queuing logic causes uninitialized PAL_CONTEXT when calling signal
handler of LibOS. On the other hand, LibOS signal handlers aren't
aware of uninitialized PAL_CONTEXT and PAL_CONTEXT is ignored when
returning from signal handler. which causes SEGV or other unexpected
issues.

Block all async signal while servicing signals so that the logic to
queue signal can be removed and the code is simplified.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/406)

### DIFF
--- a/Pal/src/host/Linux/db_exception.c
+++ b/Pal/src/host/Linux/db_exception.c
@@ -95,6 +95,10 @@ int set_sighandler (int * sigs, int nsig, void * handler)
         action.sa_flags |= SA_NOCLDWAIT;
 
     __sigemptyset((__sigset_t *) &action.sa_mask);
+    /* mask all the asynchronous signals whose signal hanlder is
+       _DkTerminateSighandler */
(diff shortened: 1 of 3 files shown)


dimakuv:
@yamahata , could you add one of those tests you mentioned (alarm, nanosleep) regression tests so we can move with this PR?

mkow:
@yamahata What's the status of this PR?

dimakuv:
This PR is outdated. The exception-handling logic in Linux PAL is re-worked. Now we don't have the issue of `malloc()`, and empty PAL context is handled correctly by the LibOS layer. The signals are still queued as pending, but there is no nesting of signals anymore. Closing this PR.

Issue #308, opened 2018-11-26
Internal memory fault at 0x0 when using SGX + xgboost library


Bizyroth:
Hello

I'm using graphene to run python program and especially python programs using the xgboost library.

When i'm trying to run a program containing `import xgboost` (even an Hello World), i've go an error:
`Internal memory fault at 0x0 (IP = +0xa1a9f, VMID = 1812475901, TID = 1)`

I try to increase the size of the enclave, the size of the stack, the maximum number of thread and the size of break limit but it doesnt change anything.

### Identification of the problem.
1) The first time i ran the program, i made a whole import of all the library of python (2.7 by the way).
I had the `Internal memory fault at 0x0 (IP = +0xa1a9f, VMID = 1812475901, TID = 1)` but i thought it was due to the size of all the python library.

2) Then i tried to import just the library needed. I had then one by one until i have to add the `libgfortran-ed201abd.so.3.0.0`. 

It must be a compatibility problem because the libgfortran is small (3KB).
Here is the manifest and 3 debug files:
* out_before_adding_libgfortran: logs before adding the libgfortran in the trusted files
* out_debug: logs with debug option
* out_with_bug: logs without debug option

[out_before_adding_libgfortran.txt](https://github.com/oscarlab/graphene/files/2615200/out_before_adding_libgfortran.txt)
[out_debug.txt](https://github.com/oscarlab/graphene/files/2615201/out_debug.txt)
[out_with_bug.txt](https://github.com/oscarlab/graphene/files/2615202/out_with_bug.txt)

[python.manifest.template.txt](https://github.com/oscarlab/graphene/files/2615206/python.manifest.template.txt)

The program is used is only:
`import xgboost`
`print("Hello world")`

If you want that i make more test ask me.

donporter:
Hmm...  This does seem like a legit bug.  

These lines in the debug output jump out at me: 

[     1] ---- shim_openat(AT_FDCWD,"/sys/devices/system/cpu",O_RDONLY|2204000,0000) = -2
[     1] handle signal (counter = 0)
[     1] ---- shim_time(0x0) = 1543232448
[     1] handle signal (counter = 0)
[     1] ---- shim_open("/sys/devices/system/cpu/online",O_RDONLY|2000000,200000) ...
[     1] ---- return from shim_open(...) = -2
[     1] handle signal (counter = 0)
[     1] ---- shim_open("/proc/stat",O_RDONLY|2000000,200000) ...
[...]

PR #408, opened 2019-01-01
Set the correct file type (regular file) for /proc/cpuinfo 

donporter:
The type for /proc/cpuinfo and /proc/meminfo should be a regular file, not a directory.  This addresses issue #308 (xgboost doesn't work in python).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/408)


donporter:
Ok to test

donporter:
Test this please

PR #409, opened 2019-01-02
Pal/Linux-SGX: map manifest file as read only

yamahata:
There is no point to map manifest file as writable.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/409)


donporter:
Ok to test

PR #410, opened 2019-01-04
[Pal/Linux-SGX]: clean up asm functions into macro and inline asm

yamahata:
This patch makes some of asm functions into static inline function
with inline assembly for clean-up. Which allows better register
scheduling.
This patch also introduces macro, WRFSBASE_RBX not to repeat same raw
byte sequence code.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/410)


donporter:
@yamahata I think this patch is ready.  Can you resolve the conflicts with master and squash into one commit?  Thanks!

dimakuv:
Interestingly, moving asm functions to normal C functions makes the GDB debugger use the RBX-based stack unwinding for `sgx_ocall()`, `sgx_entry()`, etc. This is because currently the `.cfi_startproc` and other debugging-helper macros force the compiler to generate a `.eh_frame` section which forces GDB to use the debugging-info-based stack unwinding (see https://stackoverflow.com/questions/42739893/force-gdb-to-use-frame-pointer-based-unwinding).

Because this PR moves asm functions which use these debugging-helper macros away from .S files, they are compiled without the `.eh_frame` section (no debugging info at all, actually).

The fact that GDB uses RBP-based stack unwinding allows for a better SGX-GDB experience where we can bridge the gap between enclave frames and untrusted-PAL frames.

PR #415, opened 2019-01-07
[Pal/Linux-SGX] use type for enclave_tls instead of void*

yamahata:
grp and thread member of struct enclave_tls is actually typed.
So use correct type for them instead of void *.
Also the use of void * to hide implementation is poor and not
type-safe. The right way is to hide the definition of struct.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/415)


mkow:
Test this please

PR #417, opened 2019-01-08
[Pal/Linux-SGX] Use proper sgx_ocall_* function signature for ocall_table

dimakuv:
`ocall_table[]` array contains function pointers to untrusted OCALL functions (which actually issue syscalls).

Previously, this array was `void* ocall_table`, i.e., function pointers were type-casted to `void*`.

This PR changes the type of entries to a proper function signature `int (*sgx_ocall_fn_t)(void*)`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/417)


donporter:
Ok to test

PR #424, opened 2019-01-14
[Pal/Linux-SGX] add rdtsc ocall and use it for rdtsc emulation

yamahata:
The current implementation always returns 0 for rdtsc.
It causes some application to result in error. (divide by zero or NaN)
So emulate rdtsc by ocall instead of returning 0.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/424)


dimakuv:
1. We decided to emulate `rdtsc` via gettimeofday() so we don't have two holes in SGX enclaves doing the same thing. Therefore, we close this PR and we will need another implementation.

2. We need to add a manifest bool option to explicitly enable unsafe `rdtsc`/`gettimeofday()` emulation.

PR #425, opened 2019-01-15
[LibOS] zero %rdx on entry of user program

yamahata:
As x86-64 unix ABI, %rdx on user program entry is a function pointer
which is registered as atexit. It should be zeroed unless LibOS needs
it. Otherwise random %rdx as function pointer is called by atexit and
can result in error. typically SEGV.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/425)


donporter:
Ok to test

donporter:
Shared object regression caused this failure

donporter:
Retest this please

PR #434, opened 2019-01-24
[LibOS and pal-Linux-SGX] Implement shim_do_epoll_wait() using DkObjectsWaitEvents()

thomasknauth:
[pal-Linux-SGX] Fix wrong subscript.
[LibOS] Implement shim_do_epoll_wait() using DkObjectsWaitEvents().

Running redis-benchmark (v5.0.0; redis/src/redis-benchmark -c 1 -n 100) against a redis-server will live-lock. Ultimately, LibOS polls for EPOLLOUT while the application is interested in EPOLLIN.

Please preserve the following checklist for reviewing:

- [ ] Pass all CI unit tests
- [ ] Resolve all discussions/requests from reviewers
- Reviewer approval (select one):
    - [ ] 3 approving reviews
    - [ ] 2 approving reviews and 5 days since the PR was created
    - [ ] The PR is from a maintainer; 1 approving review and 10 days since the PR was created

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/434)


thomasknauth:
#197 fixes an issue where the [e]poll() flags ultimately passed to the OS caused confusion in the application. While #197 attempts to fix this issue, it introduces another problem of ignoring the user-supplied timeout. If I recall correctly, DkObjectsWaitAny() decides on the events to poll for only on the internal LibOS state, completely ignoring what the application supplied.

While #197 fixes the redis-benchmark problem, it introduced another problem due to it ignoring timeouts in certain cases (e.g., if the fd is known to be writable, the call returns immediately, even if the application wasn't actually querying about the writable state). This caused DNS resolutions (e.g., getaddrinfo()) to fail, because it waited for an insufficient amount of time for the reply to arrive.

While #197 fixes some bug, it also introduces other unintended side-effects.

thomasknauth:
> Does this PR fix the test case you submitted in #197?

It should.

> > Reviewed 2 of 2 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/434#-:-LX0d7Wj9oP7qGCBu7gv:b8yxkwo)_ status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (1 more required, approved so far: ) (waiting on @thomasknauth)
> 
> _[LibOS/shim/src/sys/shim_epoll.c, line 348 at r1](https://reviewable.io/reviews/oscarlab/graphene/434#-LX0cexMDTh7K3qmPfDf:-LX0cexMDTh7K3qmPfDg:b68qaqu) ([raw file](https://github.com/oscarlab/graphene/blob/0488ac6961c56dd5f299f8952914c8b2bd2d1845/LibOS/shim/src/sys/shim_epoll.c#L348)):_
> 
> > ```c
> >     
> >     // TODO: Timeout must be updated in case of retries. Otherwise, we may
> >     // wait too long.
> > ```
> 
> Please open an issue for this, just so we don't lose track of it.

Done. #435 

> _[Pal/src/host/Linux-SGX/db_object.c, line 311 at r1](https://reviewable.io/reviews/oscarlab/graphene/434#-LX0dOHb2efwwIpZmDeB:-LX0dOHb2efwwIpZmDeC:b-5mftan) ([raw file](https://github.com/oscarlab/graphene/blob/0488ac6961c56dd5f299f8952914c8b2bd2d1845/Pal/src/host/Linux-SGX/db_object.c#L311)):_
> 
> > ```c
> >     uint64_t waittime = timeout;
> > ```
> 
> What is being changed here? Can we drop this line from the diff?

Nothing was changed. I introduced some extraneous blank line. Stupid question: I submit a new commit that removes this line?

PR #448, opened 2019-01-30
More compatible shebangs

### LINKED ISSUE #423: Fix shebang in all Python scripts
For better compatibility we should use `#!/usr/bin/env python2` (analogous with `python3`) in all our Python scripts.

mkow:
Fixes #423.

Please fill in the following form before submitting this PR:

## Affected components
- [ ] README and global configurations
- [X] Linux PAL
- [X] SGX PAL
- [X] FreeBSD PAL

## A brief description of this PR (describe the reasons and measures)

This PR makes shebangs in our scripts more compatible with non-standard Linux distros.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/448)

### DIFF
--- a/LibOS/buildglibc.py
+++ b/LibOS/buildglibc.py
@@ -1,5 +1,4 @@
-#!/usr/bin/python
-
+#!/usr/bin/env python2
 
 import sys, os, string, subprocess, shutil, fileinput, multiprocessing, re, resource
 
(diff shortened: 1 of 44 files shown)


donporter:
Should this be squashed to 1 commit, or was it intentional to have 2?

PR #464, opened 2019-02-11
Fix parse_stream_uri and parse_device_uri semantics

### LINKED ISSUE #463: parse_stream_uri doesn't match its documentation
https://github.com/oscarlab/graphene/blob/48f330c88b47d00340417e5eb531e5cfbdea917d/Pal/src/db_streams.c#L73-L136
`parse_stream_uri` doesn't match its documentation. The comment says that it separates the prefix, but in fact it always returns `*uri` in `*prefix`.
This makes a lot of callsites call `handle_ops::open` incorrectly (passing the whole URI in `type` argument, contrary to what `handle_ops::open` documentation says).

mkow:
Fixes  #463.

## Affected components

- [X] Linux PAL
- [X] SGX PAL
- [X] FreeBSD PAL
- [X] Common PAL code

See  #463.

Please verify that I patched all relevant places which depend on those functions.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/464)

### DIFF
--- a/Pal/src/db_streams.c
+++ b/Pal/src/db_streams.c
@@ -72,8 +72,8 @@ const struct handle_ops * pal_handle_ops [PAL_HANDLE_TYPE_BOUND] = {
 
 /* parse_stream_uri scan the uri, seperate prefix and search for
    stream handler which will open or access the stream */
-int parse_stream_uri (const char ** uri, const char ** prefix,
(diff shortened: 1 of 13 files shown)


mkow:
This PR is currently blocked by #465, which fixes a different bug exposed by this change.

PR #465, opened 2019-02-11
Null-terminate PAL_CPU_INFO strings

mkow:
## Affected components

- [X] Linux PAL
- [X] SGX PAL
- [X] FreeBSD PAL
- [ ] Common PAL code

Fixes a bug which was exposed by #464 (which changed the heap layout slightly). Strings in PAL_CPU_INFO were not null-terminated, but (un)luckily before #464 there were always zeros after the allocations, so the bug hasn't got exposed.

```Bash
cd Pal/regression
../../Runtime/pal_loader ./Bootstrap
```

This should output `CPU vendor` and `CPU brand` without any garbage appended, even after modifying the initial heap layout with some additional allocations (e.g. by merging this branch with #464).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/465)

### DIFF
--- a/Pal/src/host/FreeBSD/db_main.c
+++ b/Pal/src/host/FreeBSD/db_main.c
@@ -340,21 +340,25 @@ void _DkGetCPUInfo (PAL_CPU_INFO * ci)
 {
     unsigned int words[WORD_NUM];
 
-    char * vendor_id = malloc(12);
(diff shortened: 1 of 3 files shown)


mkow:
Retest this please

PR #492, opened 2019-02-26
[LibOS] Support of SGX PAL for test_user_memory/string

dimakuv:

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

Previously, `test_user_memory(addr, size)` and `test_user_string(addr)` failed on Linux-SGX PAL. These two functions are supposed to check if the user-supplied buffer/string are accessible inside LibOS.

The original implementation had only one option: these functions setup a segfault handler to catch memory errors (see `tcb.test_range` and `memfault_upcall()`) and access one byte of each page in the user-supplied buffer/string. If some byte is inaccessible, then an exception is raised which is caught by the segfault handler. The handler checks the faulting address (stored in `siginfo_t.si_addr`) and redirected back to the function (so it can return with failure).

This option doesn't work under SGX because there is no *trusted* field that contains the faulting address. SGX v1 doesn't have such a field at all, SGX v2 introduces a field in SSA.MISC region but only at the 4K-page granularity. Therefore, the above option doesn't work for Linux-SGX.

This PR adds a second option, specifically used for Linux-SGX PAL: `test_user_memory()` and `test_user_string()` consult the LibOS's internal VMA bookkeeping. If the buffer/string is not contained in any VMA, then these functions fail. This option is slightly slower than the first one since it requires traversing a list of VMAs. Thus, we only use this option for Linux-SGX and fallback to the previous option for all other PALs (by checking the PAL name string at initialization time).

New regression test `stat_invalid_args.c` and the corresponding Python script `30_stat.py` are added under `LibOS/shim/test/regression`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/492)

### DIFF
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -764,6 +764,19 @@ extern const char ** initial_envp;
 #define ALIGN_DOWN(addr)    \
     ((__typeof__(addr)) (((unsigned long) addr) & allocmask))
 
+void get_brk_region (void ** start, void ** end, void ** current);
+
+int reset_brk (void);
+int init_brk_region (void * brk_region);
+int init_heap (void);
+int init_internal_map (void);
+int init_loader (void);
+int init_manifest (PAL_HANDLE manifest_handle);
+
+bool test_user_memory (void * addr, size_t size, bool write);
+bool test_user_string (const char * addr);
+
+#ifdef __x86_64__
 #define switch_stack(stack_top)                                         \
     ({                                                                  \
         void * _rsp, * _rbp;                                            \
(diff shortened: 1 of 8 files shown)


dimakuv:
This PR depends on #259 .

dimakuv:
Retest Jenkins-Debug please

dimakuv:
Retest Jenkins-Debug please

dimakuv:
Retest Jenkins-Debug please

dimakuv:
Retest Jenkins please

PR #494, opened 2019-03-02
Get rid of insecure random

mkow:
## Affected components

- [X] Linux PAL
- [X] SGX PAL
- [X] FreeBSD PAL
- [X] Common PAL code

This PR:
* Removes insecure and not-too-useful function `getrand`.
* Ensures that both `/dev/random` and `/dev/urandom` outputs are cryptographically secure.
* Simplifies `DkRandomBitsRead()` interface.
* Adds missing return value checks when calling `DkRandomBitsRead()`.

Run standard tests.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/494)


mkow:
I had to remove my patch to sgx_main.c due to my misconception about the role of this file. Code from this file runs outside of SGX, so it's not security-relevant and can't use PAL functions.
Investigating the fields which were set there resulted in #647.

PR #503, opened 2019-03-08
[Pal/regression] Disable memory protection and deallocation tests on SGX

HW42:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

SGX1 does not support changing the protection of a page nor (un-)mapping
of a page.

Therefore we can't test memory protection.

Since the allocation/deallocation API is inspired by the mmap/munmap,
mapping over an already allocated area and freeing an unallocated area
are both fine so testing if a free was successful is also not possible
(without significant changes, like exposing the internal allocation
list).

By using SGX2 features it will be possible to implement this in the
future.

Run regression tests. With non-SGX PALs the memory protection/deallocation test should be runs, on SGX not.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/503)

### DIFF
--- a/Pal/regression/02_Memory.py
+++ b/Pal/regression/02_Memory.py
@@ -20,11 +20,11 @@ regression.add_check(name="Memory Allocation with Address",
 # Memory protection and deallocation can't be tested on SGX since we can't
 # unmap a page or change it's protection (SGX2 makes this possible).
 if not sgx:
-    regression.add_check(name="Memory Protection", flaky = sgx,
+    regression.add_check(name="Memory Protection",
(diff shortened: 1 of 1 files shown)


HW42:
With `flaky=True` the test script logs "This test is known not to work, but should be fixed". But this is working as expected (since SGX1 doesn't support to implement it differently).

Also other code there does the same, for example:
```
if sgx:
    print "Bulk IPC not supported on SGX"
    exit(0)
```

But if you want to cover cases like this with the flaky flag, then this is indeed not needed.

HW42:
Superseded by #520 and #521.

PR #507, opened 2019-03-11
[Pal/regression] Handle unopened broadcast stream

### LINKED ISSUE #501: Broadcast test fails without multicast route
The "Multi-Process Broadcast Channel Transmission" regression test fails if there's no multicast route configured.

I see 3 options:

1. Automatically add a route when running this test. This would require root.
2. Skip the test if there's no multicast route
3. Fail the test, but improve the error message.

Any preferences? (I think 2. or 3.).

HW42:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

If the broadcast stream is not opened this most likely means that
there's no configured multicast route so add a helpful warning. Also no
point in trying to read/write from/to a non-existent stream.

Resolves #501.

Run PAL regression tests with and without a configured multicast route.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/507)

### DIFF
--- a/Pal/regression/03_Process.py
+++ b/Pal/regression/03_Process.py
@@ -27,9 +27,17 @@ regression.add_check(name="Process Channel Transmission",
                       check_times("Process Write 2 OK",            res[0].log, 3) and
                       check_times("Process Read 2: Hello World 2", res[0].log, 3))
 
+def check_broadcast_result(res):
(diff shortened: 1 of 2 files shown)


donporter:
shared object regression/heisenbug

donporter:
Retest this please

Issue #509, opened 2019-03-11
handle_ecall and pal_linux_main trust arguments blindly


HW42:
`handle_ecall` and `pal_linux_main` currently don't verify the passed arguments. Since those are coming from the outside this is a security problem.

For example currently `pal_loader` can simply pass an pointer to a manifest outside the enclave which the enclave will accept.

There are many affected fields (`ms_ecall_enclave_start_t` and all it's recursive members). For those we need to check what needs to be done. Usually this will be one of those 3 options, I think:

 1. Don't read the value from the outside at all (use fixed address inside enclave, ask the CPU, etc.).
 2. Copy to enclave memory and then verify it.
 3. In cases where they should be allowed to be changed from the outside (like currently argv, but see #508) we probably still want to copy them inside the enclave to avoid TOCTOU problems later.

---

Implemented in the following PRs (checked means merged):
 - [x] #577: PAL addr/name
 - [x] #580: manifest size/addr
 - [x] #582: manifest/exec fd
 - [x] #588: exec addr/size
 - [x] #589: heap min/max
 - [x] #590: Other args

Some of these PRs will need to be updated depending on merge order of them (and of PR #573).

mkow:
Those fields and their subfields are also missing checks with `sgx_is_within_enclave` (the pointers *must not* point into enclave's address space, otherwise it can lead to leaking enclave memory, or other bad things).

HW42:
The `pal_sec` struct has some fields like `enclave_addr` or `mrenclave` which must be trustworthy and can be checked/calculated inside the enclave. But there are also fields like `ppid` or `uid` which the enclave can't check or calculate. So any use of those parameters needs be careful to not accidentally trust them.

What do you think how this should be handled? Split into two structs, prefix with `untrusted_`, ...

dimakuv:
I don't think we need to have any special namings for such fields as `pal_sec.ppid`. Is there an actual case which demonstrates where a special name would help?

I would suggest to have a minimal number of exceptions with the special naming convention (`untrusted_` is fine). These exceptions are basically pointers to those objects which live in untrusted memory -- e.g. a futex uaddr, a pointer to untrusted OCALL stack, etc. There must be a couple more such cases in the codebase.

[...]

PR #511, opened 2019-03-11
[Pal/Linux-SGX] Initialize enclave info in pal_sec inside enclave

HW42:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

The pal_sec struct has fields with enclave information (mrenclave,
mrsigner, enclave_attributes). Their initialization had two problems:

 - mrenclave and mrsigner were read from the enclave_secs struct, but
   those field were never populated so they were always zeroed bytes.
   This broke the fork detection in db_process.c.

 - The enclave trusted the data initialized by pal_loader (i.e. outside
   the enclave). This is a problem for example when used to check for
   forks in db_process.c (Part of #509).

Fix this by reading those values from a self report early inside the
enclave.

This fixes a bunch of regression tests which before failed since creating processes based on the same manifest were broken (since the MRENCLAVE was zeroed).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/511)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -152,7 +152,12 @@ void pal_linux_main(const char ** arguments, const char ** environments,
     setup_pal_map(&pal_map);
 
     /* initialize enclave properties */
-    init_enclave();
+    rv = init_enclave();
+    if (rv) {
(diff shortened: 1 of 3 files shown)


donporter:
This looks good to me.  Please go ahead and squash to one commit, or confirm if you intend this to remain as two commits.

donporter:
Ok to test

PR #438, opened 2019-01-25
[Pal/Linux-SGX] _DkEventWaitTimeout() doesn't return

yamahata:
_DkEventWaitTimeout() of Pal/Linux-SGX doesn't return. ETIMEDOUT and
ETRAYAGAIN are different meaing for futex(WAIT). but
unix_to_pal_error() combines into PAL_ERROR_TRYAGAIN. futex ocall
doesn't provide enough info.
So make ocall_futex() return unix error number and have
the callers of ocall_futex, e.g.  _DkEventWaitTimedout(), convert unix errno
into PAL ERROR NO.

_DkEventSet returned wrong value(positive value) on error. It should
return negated value.
_DkMutexLockTimeout has similar issue related to host futex call.
This patch fixes them too.

Signed-off-by: Isaku Yamahata <isaku.yamahata@gmail.com>

Please fill in the following form before submitting this PR:

- Affected components
    - [ ] README and global configurations
    - [x] Linux PAL
    - [x] SGX PAL
    - [x] FreeBSD PAL
    - [ ] Library OS (i.e., SHIM), including GLIBC

- A brief description of this PR (describe the reasons and measures)

- How to test this PR?
    - [ ] Documentation-only; no need to test
use futex-timeout.c which is included.

Please follow the [coding style guidelines](CODESTYLE.md). Do not submit PRs which violate these rules.
- [ ] Comments and commit messages: no spelling or grammatical errors
- [ ] 4 spaces per indentation level, at most 100 chars per line
- [ ] Asterisks (`*`) next to types: `int* pointer;` (one pointer each line)
- [ ] Braces (`{`) begin in the same line as function names, `if`, `else`, `while`, `do`, `for`, `switch`, `union`, and `struct` keywords.
- [ ] Naming: Macros, constants - `NAMED_THIS_WAY`; global variables - `g_named_this_way`; others - `named_this_way`.
- Other styling issues may be pointed out by reviewers.

Please preserve the following checklist for reviewing:

- [ ] Pass all CI unit tests
- [ ] Resolve all discussions/requests from reviewers
- Reviewer approval (select one):
    - [ ] 3 approving reviews
    - [ ] 2 approving reviews and 5 days since the PR was created
    - [ ] The PR is from a maintainer; 1 approving review and 10 days since the PR was created

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/438)

PR #514, opened 2019-03-12
Fix return code handling in _DkEventWaitTimeout

HW42:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

FUTEX_WAIT fails with EAGAIN if the supplied value doesn't match the
futex value and ETIMEDOUT in case of a timeout. PAL maps both return
errnos to PAL_ERROR_TRYAGAIN. So weed need to check which case happened
to avoid calling FUTEX_WAIT again in case of a timeout.

Also add regression test timeout handling when waiting for events.

See included regression test.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/514)

### DIFF
--- a/Pal/regression/02_Event.py
+++ b/Pal/regression/02_Event.py
@@ -1,0 +1,17 @@
+#!/usr/bin/env python2
+
+import os, sys
+from regression import Regression
+
+loader = os.environ['PAL_LOADER']
+
+regression = Regression(loader, "Event")
+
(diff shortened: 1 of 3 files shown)


dimakuv:
I would suggest to close this PR and open a new one with only the test. And mark the new PR that it tests changes in #438.

HW42:
Done (#519).

PR #522, opened 2019-03-14
[Pal/Linux-SGX] Ensure that ocall_exit never returns

HW42:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

OCALL_EXIT should never return. But the outside might try to trick the enclave, so add an infinite loop to ocall_exit to catch this.

This is probably exploitable but I didn't spend time to analyze which of the current usages was actually exploitable.

AFAICS it's hard to add a test which simulates a malicious host with the current code. So no test case provided. If you want to play with it you can change `sgx_ocall_exit` in `sgx_enclave.c`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/522)

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_ocalls.c
+++ b/Pal/src/host/Linux-SGX/enclave_ocalls.c
@@ -67,8 +67,13 @@ int printf(const char * fmt, ...);
 int ocall_exit(int exitcode)
 {
     int64_t code = exitcode;
-    SGX_OCALL(OCALL_EXIT, (void *) code);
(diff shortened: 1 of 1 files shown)


mkow:
Test this please

donporter:
Please squash this into one commit, and we should be set to merge.

mkow:
This was intended to be like this ;) Those are two unrelated changes.

PR #527, opened 2019-03-15
[Linux PAL] Ensure the usage of alternative stack for host exception handling

### LINKED ISSUE #486: [Pal/Linux] Signals are not delivered on altstack because SA_ONSTACK is not used
PR #281 introduces alternate signal stack context to handle signals (only for Linux PAL).

The alternate signal stack is malloced for each thread right after the TCB object and [sigaltstack is called](https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux/db_threading.c#L63) to install this alternate stack.

However, for a signal to be delivered on the alternate stack, its signal handler must be registered via `sigaction()` and contain `struct sigaction sa = {..., .sa_flags = SA_ONSTACK}`. Otherwise, the signal is delivered to the normal stack context. [See this](https://www.gnu.org/software/libc/manual/html_node/Flags-for-Sigaction.html#Flags-for-Sigaction).

**Fix**: all (?) signals in `Pal/src/host/Linux/db_exception.c` must be registered with the flag `SA_ONSTACK` set.

chiache:

## Affected components

- [x] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

Fixes #486.

The new host exception handling scheme in the Linux PAL is supposed to use an alternate stack to prevent mixing the PAL exception handler frame with LibOS/application frames. The previous implementation is not using the assigned alternate stack because `SA_ONSTACK` is not given as a flag to the `sigaction` system call. This PR adds the `SA_ONSTACK` flag.

This PR also fixes another issue with alternate stack assignment. The stack pointer given to the `sigaltstack` system call, i.e.,`ss.ss_sp`, should point to the bottom of the stack instead of the top.

A regression test is added for the Linux PAL only to check if the alternate stack is ever used.

Running the `Exception` test in the PAL regression test directory. A good output should look like this:
```
$ ../../Runtime/pal_loader Exception
stack in main: 0x7ffc29766618
Div-by-Zero Exception Handler 1: 0x40043d, rip = 0x40043d
stack in handler: 0x7ffc29765e40
Div-by-Zero Exception Handler 2: 0x40046a, rip = 0x40046a
stack in handler: 0x7ffc29765e40
Memory Fault Exception Handler: 0x1000, rip = 0x400484
stack in handler: 0x7ffc29765e40
```

The stack pointer in the main function should not be on the same page with the stack pointers inside the handlers.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/527)

### DIFF
--- a/Pal/regression/01_Exception.py
+++ b/Pal/regression/01_Exception.py
@@ -20,5 +20,24 @@ regression.add_check(name="Exception Handler Swap",
 regression.add_check(name="Exception Handling (Set Context)",
     check=lambda res: any([line.startswith("Div-by-Zero Exception Handler 1") for line in res[0].log]))
 
+if os.environ['PAL_HOST'] == 'Linux':
+    def check_altstack(res):
+        stacks = []
+        for line in res[0].log:
+            if line.startswith('stack in '):
+                try:
+                    stack = int(line.split(' ')[3][2:], 16)
+                except:
+                    return False
+                stacks.append(stack)
+        # The first stack will be the main stack
+        for stack in stacks[1:]:
+            # Handler stack cannot be on the same page with the main stack
+            if int(stack / 4096) == int(stacks[0] / 4096):
+                return False
+        return True
+
+    regression.add_check(name="Use host alternate stack (Linux only)", check=check_altstack)
+
 rv = regression.run_checks()
 if rv: sys.exit(rv)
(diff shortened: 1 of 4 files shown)


dcowhig:
Retest this please

dcowhig:
Retest this please

dcowhig:
Retest this please

dcowhig:
Retest this please

dimakuv:
@chiache should have another look at this.

mkow:
@yamahata @HW42 @dimakuv Is this PR still relevant? There was a bunch of changes in exception handling recently, so this may have already been covered by one of those PRs.

PR #569, opened 2019-03-29
[LibOS] Properly set shim_dentry.type

omeg:

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

- Properly set `shim_dentry.type` field in all cases (fixes incorrect `d_type` field in directory enumeration).
- Added standalone regression test for `getdents` syscall since the LTP one fails (`symlink` is unimplemented).

- Run the regression test.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/569)

### DIFF
--- a/LibOS/shim/src/fs/chroot/fs.c
+++ b/LibOS/shim/src/fs/chroot/fs.c
@@ -228,9 +228,9 @@ static int __query_attr (struct shim_dentry * dent,
     /* need to correct the data type */
     if (data->type == FILE_UNKNOWN)
         switch (pal_attr.handle_type) {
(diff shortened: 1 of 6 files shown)


donporter:
The last jenkins failure was on bw_tcp - which should be fixed in #618 , should we not be able to get a passing run.

dimakuv:
@donporter , just a slight correction: `bw_tcp` is a flaky test under Linux PAL, and doesn't work at all under Linux-SGX PAL (it is not included in SGX regression tests). I believe the failure that you see is an intermittent one, please rerun the pipeline.

PR #573, opened 2019-03-31
[Pal/Linux-SGX] Measure all memory

### LINKED ISSUE #505: Measure important enclave memory
As already [documented in the code](https://github.com/oscarlab/graphene/blob/d95c4f9da6d5a240497bb2e4b42a417cae3452d5/Pal/src/host/Linux-SGX/sgx_main.c#L340) there are important memory areas (like TCS) which are currently not measured.

We might even consider to measure all mapped pages by default.

HW42:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

Before this change some important memory areas, for example tcs and tls
were not measured. With this change all mapped enclave memory is
measured.

Closes #505.

In the current form this has the downside that it slows down enclave startup a lot. For very quick applications like the regression tests this is quite bad. I will take a look if measuring all pages with one syscall like in the old graphene kernel driver helps. Otherwise an option would be to not measure the heap but instead zero it directly after enclave startup (also not tested yet if or how much this helps).

Run SGX regression tests.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/573)

### DIFF
--- a/LibOS/shim/src/.gitignore
+++ b/LibOS/shim/src/.gitignore
@@ -1,3 +1,3 @@
 libsysdb.so.cached
-asm-offsets.h
-asm-offsets.s
+/asm-offsets.h
+/generated-offsets.s
--- a/LibOS/shim/src/generated-offsets.c
+++ b/LibOS/shim/src/generated-offsets.c
@@ -1,4 +1,4 @@
-#include <asm-offsets-build.h>
+#include <generated-offsets-build.h>
 
 #include <stddef.h>
 
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -22,12 +22,12 @@ echo-cmd = $(if $($(quiet)cmd_$(1)), echo '  $(call escsq,$($(quiet)cmd_$(1)))';
 cmd = @$(echo-cmd) $(cmd_$(1))
 
 
-quiet_cmd_asm_offsets_s = [ $@ ]
-      cmd_asm_offsets_s = $(CC) $(CFLAGS) $(defs) -S $< -o $@
+quiet_cmd_generated_offsets_s = [ $@ ]
+      cmd_generated_offsets_s = $(CC) $(CFLAGS) $(defs) -S $< -o $@
 
-asm-offsets.s: asm-offsets.c $(headers)
-	$(call cmd,asm_offsets_s)
-CLEAN_FILES += asm-offsets.s
+generated-offsets.s: generated-offsets.c $(filter-out asm-offsets.h,$(headers))
+	$(call cmd,generated_offsets_s)
+CLEAN_FILES += generated-offsets.s
 
 
 quiet_cmd_asm_offsets_h = [ $@ ]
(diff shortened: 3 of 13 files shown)


donporter:
Just FYI, 551 is merged now.

mkow:
@HW42 Could you squash this PR into intended number of commits and rebase on current master?

woju:
___
*[Pal/src/host/Linux-SGX/signer/pal-sgx-sign, line 412 at r1](https://reviewable.io/reviews/oscarlab/graphene/573#-LbelbXKD79OGl8NqdXt:-LddV51x-GYgvsE0QdXJ:b6afub1) ([raw file](https://github.com/oscarlab/graphene/blob/cb0065d7e30bc5db1a8c11d1d8c139cff466cdf0/Pal/src/host/Linux-SGX/signer/pal-sgx-sign#L412)):*
@hw42

Well, there are three modules that could do what you need: `struct`, `ctypes` and `cffi`. Both `struct` and `ctypes` are from standard library, but you have to define your structures separate, in python. IOW those cannot parse existing header files. CFFI is external library, but widely used by Python community. It has limited support for parsing C headers: 

```
[x86_64 pts/0] user@qubes-devel-tmp ~/aux/graphene/Pal/src/host/Linux-SGX [git||master] [16:25 1951]% python3
Python 3.6.8 (default, Mar 21 2019, 10:08:12) 
[GCC 8.3.1 20190223 (Red Hat 8.3.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import cffi
>>> ffi = cffi.FFI()
>>> ffi.cdef('''
... typedef struct {
...     uint64_t rax;
...     uint64_t rcx;
...     uint64_t rdx;
...     uint64_t rbx;
...     uint64_t rsp;
...     uint64_t rbp;
...     uint64_t rsi;
...     uint64_t rdi;
...     uint64_t r8; 
...     uint64_t r9; 
...     uint64_t r10;
...     uint64_t r11;
...     uint64_t r12;
...     uint64_t r13;
...     uint64_t r14;
...     uint64_t r15;
...     uint64_t rflags;
...     uint64_t rip;
...     uint64_t ursp;
...     uint64_t urbp;
...     uint32_t exitinfo;
...     uint32_t reserved;
...     uint64_t fsbase;
...     uint64_t gsbase;
... } sgx_arch_gpr_t;
... 
... struct enclave_tls {
...     uint64_t enclave_size;
...     uint64_t tcs_offset;
...     uint64_t initial_stack_offset;
...     void *   aep;
...     void *   ssa;
...     sgx_arch_gpr_t * gpr;
...     void *   exit_target;
...     void *   fsbase;
...     void *   stack;
...     void *   ustack_top;
...     void *   ustack;
... };
... 
... ''')
>>> ffi.offsetof('struct enclave_tls', 'stack')
64
>>> ffi.offsetof('struct enclave_tls', 'fsbase')
56
```

There are limitations, the biggest is that it does not support `#directives`, that is, no `#include`, no `#define`, not even the idiom`#ifndef _MY_HEADER_H #define _MY_HEADER_H`.

This would require separating typedefs and structs to separate files, similar to "bits" from linux-headers. I'd advise to leave `#ifndef _MY_HEADER_H` and just skip those lines while reading the file from python.

There is also Cython, which is basically python-to-C compiler with provision for #including real C headers. This emits .so python modules. But I think that would be overkill.

mkow:
Retest Jenkins-SGX please

mkow:
Retest Jenkins-Debug please

PR #578, opened 2019-04-01
[Pal/Linux] SEGV in udp_connect()

yamahata:
dest_addr can be NULL in udp_connect(). Add NULL check.
NOTE: Pal/Linux-SGX or Pal/FreeBSD doesn't have similar error because
they don't touch AF_INET6.

## Affected components

- [x] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

LibOS/shim/test/native/udp

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/578)


dimakuv:
Why does it conflict with #614? This one uses `udp.c` and #614 uses `tcp.c`. Or is there smth else?

dimakuv:
I am confused how `udp.c` is supposed to work under Graphene-SGX. In its current state, UDP support in Linux-SGX PAL looks incorrect (?).

Specifically, this PR fails on `SGX=1 <pal-loader> ./udp client` on `sgx_ocall_sock_connect()` because of this line:
```c
    /* ms->ms_addr is NULL because udp only specifies the ms_bind_addr address */
    ret = INLINE_SYSCALL(connect, 3, fd, ms->ms_addr, ms->ms_addrlen);
```

This call always fails with `-EINVAL` since there is no address to connect to.

dimakuv:
Ok, so after some debugging, it turned out that support for UDP was broken in SGX PAL.

- In LibOS's `do_sendmsg()` there are these lines:
```c
        if (sock->sock_state == SOCK_CREATED && !pal_hdl) {
            pal_hdl = DkStreamOpen("udp:", 0, 0, 0,
                                   hdl->flags & O_NONBLOCK);
            ...
            hdl->pal_handle = pal_hdl;
        }
```
So here we want to create an "addressless" UDP connection via `DkStreamOpen("udp:", ...)`.

- Linux-SGX's `udp_connect()` will parse this `udp:` URI into Linux-based `sockaddr` via:
```c
... socket_parse_uri(uri, &bind_addr, &bind_addrlen, &dest_addr, &dest_addrlen);
```

- Since URI is empty, both `bind_addr` and `dest_addr` are initialized to NULLs.

- We call `ocall_sock_connect()` which calls `sgx_ocall_sock_connect()`. Both these functions are generic and used for TCP and UDP.

- `sgx_ocall_sock_connect()` performs `INLINE_SYSCALL(connect, 3, fd, ms->ms_addr, ms->ms_addrlen);`. But connect() fails because there is `ms_addr = NULL`.

- Application's `sendto()` fails with `-EINVAL`.

dimakuv:
A simple fix is to add:
```c
    if (ms->ms_type & SOCK_STREAM) {
        ret = INLINE_SYSCALL(connect, 3, fd, ms->ms_addr, ms->ms_addrlen);
        ...
    }
```

I will do it as part of the final rebase and push the new version now.

dimakuv:
Retest Jenkins please

PR #580, opened 2019-04-02
[Pal/Linux-SGX] Don't get manifest addr/size from urts

HW42:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

Don't get manifest addr/size from urts.

Instead we use the fact that the manifest is always placed at the top of
the enclave address range. The manifest size is stored inside the TLS
like we have already done for the enclave size.

Part of issue #509.

Depends on PR #573 to make sense.

Run SGX regression test to see that it doesn't break things.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/580)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -195,6 +195,7 @@ free_and_err:
 }
 
 extern void * enclave_base;
+extern void * enclave_top;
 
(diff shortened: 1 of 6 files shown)


mkow:
Retest Jenkins-Debug please

PR #589, opened 2019-04-05
[Pal/Linux-SGX] Don't get heap min/max from urts

HW42:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

Don't get heap min/max from urts
Instead pass them through the measured TLS.

Part of issue #509.

This depends on #573 (but is based on master to make the reviewable.io diff more useful).

I will cleanup the argument handling of `pal_linux_main` in a separate PR, which will also remove the TOCTOU bug which this change has, because is uses `sec_info`.

Run SGX regression tests to see that it doesn't break things.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/589)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -215,16 +215,18 @@ void pal_linux_main(char * uptr_args, uint64_t args_size,
         return;
     }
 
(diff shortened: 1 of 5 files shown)


HW42:
I will update `pal-sgx-sign` depending on merge order of this and PR #507.

PR #591, opened 2019-04-07
[LibOS] Optimize do_exit_group for multi thread race condition

zhangckid:
If we try to running a C program in graphene, it may not be able to exit normally when running do_exit_group. Because have a status like that: another thread do clone syscall after do_exit_group,
the new thread can't get the KILL signal from do_kill_proc().
So, we loop in try_process_exit() to wait all thread exit.

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

Fix some APP have chance can't exit issue.

Try to run C hello-world 10 times, at least have 1 time can't exit normally.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/591)

### DIFF
--- a/LibOS/shim/src/sys/shim_exit.c
+++ b/LibOS/shim/src/sys/shim_exit.c
@@ -131,6 +131,7 @@ int thread_exit(struct shim_thread * self, bool send_ipc)
 int try_process_exit (int error_code, int term_signal)
 {
     struct shim_thread * cur_thread = get_cur_thread();
+    int ret = 0;
 
     cur_thread->exit_code = -error_code;
     cur_process.exit_code = error_code;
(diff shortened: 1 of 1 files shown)


zhangckid:
For the  ret = check_last_thread(cur_thread);

Yes, the original return value of try_process_exit still is 0, look meaningless.
And maybe I can change to ret = (int)check_last_thread(cur_thread) to keep the type.

For del_thread
In this case, the last thread will still alive. I tested use thread->is_alive to check that, but no one to make it dead. At the same time, I try to use thread_exit() here, it no effect to this case, do you have any other suggestion here?

dimakuv:
I am closing this PR since this behavior is not triggered on the master branch and there was no follow-up from the PR author.

PR #605, opened 2019-04-10
[*/regression] Migrate to Python 3

HW42:
## Affected components

- [ ] README and global configuration
- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code
- [x] Library OS (i.e., SHIM), including GLIBC

## Description of the changes (reasons and measures)

[*/regression] Migrate to Python 3

## How to test this PR? (if applicable)

Run regression tests.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/605)

PR #606, opened 2019-04-10
[Scripts/regression] Improved subprocess handling

HW42:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

[Scripts/regression] Improved subprocess handling

This fixes two problems with subprocess handling:

 1. Use subprocess.communicate to avoid that the pipe can fill up.

 2. Use a process group to ensure that we also kill the subprocess'
    children in case of a timeout.

Note: This does not cover the case that the child returns successfully
but leaves other child processes running. AFAICS this is not easy since
subprocess.communicate() wait(2)s for the process so the pid is no
longer guaranteed to be valid to kill. So we leave this to the outside
environment (Jenkins or whatever).

Run regression tests. To test the timeout case set it manually to something too small.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/606)

### DIFF
--- a/Scripts/regression.py
+++ b/Scripts/regression.py
@@ -1,4 +1,4 @@
-import sys, os, subprocess, re, time
+import sys, os, subprocess, re, time, signal
 
 class Result:
     def __init__(self, out, log, code):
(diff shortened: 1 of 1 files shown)


HW42:
Note this change is on top of #605, so until that gets merged the diff on GitHub/reviewable.io will show both changes. Look at the commit if you want to review it.

donporter:
Same issue as in #605 - no python3 installed in the Jenkins environment.

dimakuv:
Retest this please

Issue #152, opened 2017-12-19
size_t should be used for memory sizes (instead of int)


mkow:
A lot of current code uses `int` instead of `size_t` for storing in-memory object sizes, which is too small to hold anything bigger than 2GB (and will silently overflow when trying to do so). It will also trigger tons of warnings if you ever decide to use some static code checkers, obfuscating their output.

This issue already caused some problems in the past (e.g. https://github.com/oscarlab/graphene/pull/136), so it may be a good idea to fix this in the whole codebase someday. I'll check if there's any way to enable GCC warnings for this kind of issues (only for size_t, not for every data losing conversion) so you could easily prevent them in the future.

I saw that there were some recent fixes by @chiache to this kind of problems, thanks!

donporter:
I agree.

mkow:
Note: #741 partially fixed this issue.

PR #608, opened 2019-04-11
[Pal] Use size_t instead of int for size parameter

omeg:

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [X] Common PAL code

- Changed `size` parameter type of `_DkCheckMemoryMappable` from `int` to `uint64_t` to enable using large memory chunks (`int` overflowed).

Run regression tests.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/608)


mkow:
Side note: we should probably prioritize #152.

PR #618, opened 2019-04-13
[Pal/Linux-SGX] Set PAL_ERRNO to PAL_ERROR_INTERRUPTED

dimakuv:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

Set PAL_ERRNO to PAL_ERROR_INTERRUPTED when interrupted by host signal.

Current Linux-SGX PAL signal handling unwinds all inner PAL frames if host signal interrupts Graphene while it waits on a long syscall. Because PAL frames are unwound, all inner logic of PAL setting `PAL_ERRNO = PAL_ERROR_INTERRUPTED` (using the `_DkRaiseFailure()` flow) is skipped. This PR retrofits this logic.

This particular bug was found in `LibOS/shim/test/apps/lmbench/lmbench-2.5/bin/linux/bw_tcp` test under SGX.

Execute the test case `TCP socket bandwidth` of lmbench under Linux-SGX PAL. If manually:
```sh
lmbench-2.5/bin/linux$ SGX=1 ./pal_loader ./bw_tcp -s  # start server in one window
lmbench-2.5/bin/linux$ SGX=1 ./pal_loader ./bw_tcp 127.0.0.1  # start client in another window
# wait till client finishes talking to server
lmbench-2.5/bin/linux$ SGX=1 ./pal_loader ./bw_tcp -127.0.0.1  # start client again to terminate server
```

Without this PR, the server will crash after the second command because its own child process will send SIGCHLD while its parent process waits on `accept4()`. This will lead to incorrect PAL unwinding without setting a corresponding PAL_ERRNO.

With this PR, the test must finish correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/618)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_exception.c
+++ b/Pal/src/host/Linux-SGX/db_exception.c
@@ -410,6 +410,11 @@ void _DkHandleExternalEvent (PAL_NUM event, sgx_context_t * uc)
         arch_store_frame(&frame->arch);
     }
 
+    /* We only end up in _DkHandleExternalEvent() if interrupted during
+     * host syscall; Dk* function will be unwound, so we must inform LibOS
+     * layer that PAL was interrupted (by setting PAL_ERRNO). */
+    _DkRaiseFailure(PAL_ERROR_INTERRUPTED);
+
     if (!_DkGenericSignalHandle(event, 0, frame, NULL)
         && event != PAL_EVENT_RESUME)
         _DkThreadExit();


dimakuv:
@donporter, yes, this is how the failing output should look like. I can look into if and why this regression test doesn't fail loudly (locally on my machine).

donporter:
Thanks!

dimakuv:
So it's something in the way Jenkins scripts consume output of the `make SGX_RUN=1 regression` of lmbench. I will go deeper into Jenkins scripts.

mkow:
Retest this please

PR #639, opened 2019-04-24
[Pal/Linux-SGX] disallow nested async host exception in enclave_entry

yamahata:
The enclave is set as TCS.nssa = 2, one for page fault for host OS,
one for host signal to bounce into enclave. .Lhandle_exception assumes
that TCS.cssa == 1. uRTS can maliciously inject async signal
nestedly. In that case, it should fall into fail loop because the code
assumption was broken. Mask host async signal(SIGINT, SIGTERM and
SIGCONT) on signal.

TCS.nssa = 2 is enough because even if we set TCS.nssa = N > 2 and
allow nested async host signal. But if it nested up to
TCS.cssa = N - 1, the situation is same.
Please remember that .Lhandle_exception is called in signal context to
set up the actual exception handler with specified signal masked and
returns back from signal handler to normal execution. And then the
actual exception handler in enclave is triggered as normal execution
context, not as signal handler without any host signal masked.

## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/639)

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_entry.S
+++ b/Pal/src/host/Linux-SGX/enclave_entry.S
@@ -102,6 +102,20 @@ enclave_entry:
 	# PAL convention:
 	# RDI - external event
 
+	# Nested exceptions at the host-OS level are disallowed:
+	# - Synchronous exceptions are assumed to never happen during
+	#   handle_resume;
+	# - Asynchronous signals are not nested by benign host OS because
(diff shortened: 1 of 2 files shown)


HW42:
LGTM.

PR #676, opened 2019-05-09
[LibOS/PAL] Return proper exit code on abnormal process termination

### LINKED ISSUE #666: [LibOS] shim_terminate() causes the process to always exit with status 0
`shim_terminate()` is mostly used for early termination in case that shim initialization fails. It should accept an error code parameter (currently it's void).

omeg:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

Return proper process exit code when shim initialization fails (it's currently always 0).
Also fix PAL exit code on abnormal termination.
Fixes #666

The new `init_fail` regression test should pass.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/676)

### DIFF
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -145,7 +145,7 @@ static inline PAL_HANDLE __open_shim_stdio (void)
     return shim_stdio;
 }
 
-int shim_terminate (void);
+int shim_terminate (int err);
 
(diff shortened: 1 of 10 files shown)


omeg:
I also have a small fix for the same issue in PAL, should I add it to this PR or create another?

PR #686, opened 2019-05-15
[Makefile] drop WERROR=1 flag and enable -Werror

yamahata:

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/686)

### DIFF
--- a/Jenkinsfiles/Linux
+++ b/Jenkinsfiles/Linux
@@ -6,7 +6,7 @@ pipeline {
                 stage('Build') {
                     steps {
                         sh '''
-                            make WERROR=1
+                            make
                            '''
                     }
                 }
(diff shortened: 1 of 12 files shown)


yamahata:
Close this PR as the past discussion.
Personally I think the option should be opposite.
-Werror on by default. and it can be disabled by option.

Anyway I don't see much value to fight for it.

mkow:
Note for the future, in case this discussion would return someday: We discussed this on today's call and decided to leave current settings (i.e. `-Werror` not used by default).
Arguments against making it default:
* `-Werror` is annoying when debugging and prototyping code.
* Other versions of GCC may produce different warnings and refuse to compile Graphene on other people's machines because of some less-important warnings (e.g. signed vs unsigned comparison) which appear only on newer/older GCC.

Issue #539, opened 2019-03-19
Format existing code according to code style


HW42:
I'm sure this already planned but I didn't find a ticket for it.

Automatic reformatting is very useful to get consistent code style without much work and discussions. But due to the big amount of not yet reformatted code the `make format` target is not very helpful at the moment (too many unrelated changes to filter out).

mkow:
Yup, it's planned, but it's good to have a ticket for this. I'm waiting mostly for #385 to be merged to proceed with reformatting.

PR #697, opened 2019-05-20
Cleanup C macros

### LINKED ISSUE #694: Cleanup C macros
We need a major clean-up of C macros, because of the following issues:
* They are abused in a few places, e.g. modify control-flow (by containing `return` or `goto`).
* Their names should be uppercase where possible, to distinguish them from ordinary functions (this is very important, e.g. because of a risk of double-eval of their arguments).
* A lot of them should be normal functions, not macros.
* Macro arguments should (almost) always be parenthesized.
* Statement-like macros should be wrapped in `do { ... } while (0)`.

I'll prepare a PR fixing the above in the upcoming days.

mkow:
Fixes #694.

## Affected components

- [X] Linux PAL
- [X] SGX PAL
- [X] FreeBSD PAL
- [X] Common PAL code

## Description of the changes
This PR tries to fix issues listed in #694.

I'm still not sure whether some of the changes here make sense (e.g. capitalizing `barrier` macro, because the standard one is not), so don't hesitate to dispute them. I also didn't change some macros, as this would require huge rework of the code (e.g. `debug`, which uppercased collides with `DEBUG`, which marks our compilation mode).

I most likely also missed a bunch of things, but there are so many places using macros that it's hard to fix all of them correctly.

This is also a part of the efforts towards resolving #539.

Standard tests.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/697)


mkow:
FYI: I just pushed a fixup, because previously I forgot to grep for defines with spaces after `#`.

mkow:
Retest Jenkins-SGX please (timeout on bw_unix)

mkow:
Retest Jenkins-Debug please (`semctl01,2` and `semctl01,3` failed)

PR #726, opened 2019-05-31
[GLIBC] Fix the Makefile rules for building GLIBC

### LINKED ISSUE #725: Build requires manual intervention due to git-apply
After recent merges, when I do a non-clean build of Graphene (i.e., already have built files):
```bash
...
applying glibc-syscalldb-api.patch
The next patch would create the file elf/syscallas.S,
which already exists!  Assume -R? [n]  # REQUIRES MANUAL INPUT
Apply anyway? [n]  # REQUIRES MANUAL INPUT
Skipping patch.
1 out of 1 hunk ignored
...
```

We need to fix it (by adding some flags to git-apply I guess).

chiache:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

(1) Removing unnecessary dependencies
(2) Need to abort if one of the patches isn't applied properly
(3) Preventing repeated patching and building (fix #725)
(4) Clean up and repatch the GLIBC source when the dependencies are updated

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/726)

### DIFF
--- a/LibOS/Makefile
+++ b/LibOS/Makefile
@@ -48,7 +48,7 @@ $(BUILD_DIR)/Build.success: $(BUILD_DIR)/Makefile
 
 $(GLIBC_TARGET): $(BUILD_DIR)/Build.success
 
-$(BUILD_DIR)/Makefile: $(addprefix $(GLIBC_SRC)/,configure elf/Versions nptl/Versions dlfcn/Versions)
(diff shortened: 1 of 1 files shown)


chiache:
Retest Jenkins-SGX please

PR #736, opened 2019-06-05
[Pal, LibOS] Unify handling of ELF auxiliary vectors in LibOS and PAL

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

## Description of the changes 

PAL initializes a subset of ELF aux vectors on binary start. LibOS does
the same on execve(). Previously, the two flows were similar but not
unified. This led to subtle memory corruption bugs. This commit unifies
the requirements for the minimum supported number of ELF aux vectors as
well as the minimum additional space to hold their auxiliary data (such
as AT_RANDOM bytes). As a side-effect, it removes nauxv argument.

## How to test this PR? 

Run regression tests.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/736)

### DIFF
--- a/LibOS/shim/include/shim_thread.h
+++ b/LibOS/shim/include/shim_thread.h
@@ -349,6 +349,6 @@ bool check_on_stack (struct shim_thread * cur_thread, void * mem)
 
 int init_stack (const char ** argv, const char ** envp,
                 int ** argcpp, const char *** argpp,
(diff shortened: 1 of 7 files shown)


dimakuv:
Retest Jenkins-Debug please

PR #737, opened 2019-06-05
[LibOS] Fix int type cast bug in chroot FS

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Chroot FS implementation contained several bugs pertaining to
greater-int-type cast to smaller-int-type (uint64_t -> int). For files
more than 2GB, this resulted in negative length passed to other LibOS
functions, resulting in pointer arithmetic overflows and sefaults.

Note that this commit fixes one particular bug regarding type casts.
More changes are required to weed out such bugs from LibOS code.

## How to test this PR? 

This bug was triggered on a complicated Python3 setup with huge files. No simple test was constructed.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/737)

### DIFF
--- a/LibOS/shim/src/fs/chroot/fs.c
+++ b/LibOS/shim/src/fs/chroot/fs.c
@@ -464,7 +464,7 @@ static int chroot_open (struct shim_handle * hdl, struct shim_dentry * dent,
         return ret;
 
     struct shim_file_handle * file = &hdl->info.file;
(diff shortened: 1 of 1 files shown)


dimakuv:
Superseded by #741. Closing.

PR #748, opened 2019-06-12
[LibOS] Take into account adjacent VMAs on correct-memory check

### LINKED ISSUE #747: [LibOS][SGX]  test_user_memory issue
I have some code to test getcwd, but 'getcwd' will return NULL, and the errno is EFAULT
```
static char cwd_buf[PATH_MAX];

int test_getcwd()
{
    char *cwd = NULL;

    cwd = getcwd(cwd_buf, sizeof(cwd_buf));
    if (cwd) {
        std::cout << "getcwd: " << cwd << std::endl;
        return 0;
    }
    std::cout << "getcwd errno: " << errno << std::endl;

    return -1;
}
```
The buffer address passed to 'getcwd' is 0x6021c0, and the buffer size is 4096.
```
[     1] ---- shim_getcwd(0x6021c0,4096) = -14
```
It seems that the start and end of the buffer are in two vma:
```
[0x602000-0x603000] prot=00000003 flags=00000012 uri=file:hello offset=8192
[0x603000-0x604000] prot=00000003 flags=00000032 offset=0
```
so test_user_memory return false.

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Previously the check on correct memory region under Linux-SGX (called is_in_one_vma()) only checked that the whole memory region fits into one VMA. In some cases (e.g. DATA and BSS sections), memory regions can span several adjacent VMAs. This commit refines the check and renames it to is_in_adjacent_vmas(). It also adds a LibOS test.

Fixes #747.

## How to test this PR? 

The commit adds a LibOS regression test `getcwd.c`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/748)

### DIFF
--- a/LibOS/shim/include/shim_vma.h
+++ b/LibOS/shim/include/shim_vma.h
@@ -128,7 +128,7 @@ int lookup_vma (void * addr, struct shim_vma_val * vma);
 int lookup_overlap_vma (void * addr, size_t length, struct shim_vma_val * vma);
 
 /* True if [addr, addr+length) is found in one VMA (valid memory region) */
-bool is_in_one_vma (void * addr, size_t length);
+bool is_in_adjacent_vmas (void * addr, size_t length);
 
 /*
  * Looking for an unmapped space and then adding the corresponding bookkeeping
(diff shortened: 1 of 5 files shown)


dimakuv:
Retest Jenkins-SGX please (timeout on udp LibOS regression test)

dimakuv:
Retest Jenkins-Debug please

PR #759, opened 2019-06-18
Misc Makefile improvements

### LINKED ISSUE #189: Fix header dependencies tracking in the Makefiles
Currently modifying any header causes rebuilding of all objects, which is slow. Additionally, header dependencies between Pal and LibOS aren't tracked at all.

Proposed solution: use gcc/clang's `-MM` switch to generate accurate header dependencies.

### LINKED ISSUE #204: There should be a way to build Graphene without tests
It would be nice to have a way to build Graphene without tests. Two proposed solutions:
* Move all tests to a new target, "test".
* Add a switch to the main Makefile, e.g. `NO_TESTS=1`.

I prefer the first option more (AFAIR this is a widespread convention), but the second should work well also. @donporter @chiache Any preferences? I can implement this as a pull request once we decide which way to go.

HW42:
## Affected components

- [x] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

## Description of the changes 

See commit messages for description of the changes.

Fixes #189 and fixes #204.

## How to test this PR? 

Builds should work as before (note that building tests has been moved to a separate target).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/759)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 *.i
 *.s
 *.e
+*.d
 
 # object file/libraries
 *.a
--- a/Jenkinsfiles/Linux
+++ b/Jenkinsfiles/Linux
@@ -7,6 +7,7 @@ pipeline {
                     steps {
                         sh '''
                             make -j 8 WERROR=1
+                            make -j 8 WERROR=1 test
                             cd Pal/ipc/linux
                             make
                            '''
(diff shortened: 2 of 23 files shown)


dimakuv:
Retest Jenkins-SGX please

PR #768, opened 2019-06-20
[Pal/Linux-SGX] allocate dedicate stack for vfork() child

yamahata:
With vfork(), parent/child shares the stack. So the child has
to be careful to not modify parent's local variable.
Introduce dedicated function to allocate child local variables.
NOTE: compiler can use same stack location for different local variables.

Signed-off-by: Isaku Yamahata <isaku.yamahata@gmail.com>

## Affected components

- [ ] README and global configuration
- [x] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code
- [ ] Library OS (i.e., SHIM), including GLIBC

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/768)

PR #770, opened 2019-06-20
[Pal/{Linux, Linux-SGX}] block async signal during vfork

yamahata:
## Affected components

- [x] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 
To avoid vfork child from corrupting parent memory, block async signal
before vfork().

This PR depends on #768 

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/770)

### DIFF
--- a/Pal/src/host/Linux-SGX/sgx_exception.c
+++ b/Pal/src/host/Linux-SGX/sgx_exception.c
@@ -26,6 +26,7 @@
 #include "ecall_types.h"
 #include "ocall_types.h"
 #include "sgx_internal.h"
+#include "pal_linux.h"
 
 #include <atomic.h>
 #include <sigset.h>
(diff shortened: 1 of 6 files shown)


dimakuv:
There is some bug in this code which results in fails of these LTP regression tests:
```
Test 'kill01,1' did not pass in the current run!!
Test 'kill03,1' did not pass in the current run!!
Test 'kill09,1' did not pass in the current run!!
Test 'nanosleep03,1' did not pass in the current run!!
Test 'sendfile05,1' did not pass in the current run!!
Test 'sendfile05_64,1' did not pass in the current run!!
Test 'waitpid01,1' did not pass in the current run!!
Test 'waitpid01,2' did not pass in the current run!!
```

PR #774, opened 2019-06-20
enable stack protector

yamahata:
## Affected components

- [x] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

## Description of the changes 
Enable stack protector for Pal/Linux, Pal/Linux-SGX and LibOS
this PR depends on other PRs as preparation.

## How to test this PR? 
This requires gcc 8 or later. So on ubuntu I used the following patch locally.
(If you'd like test it with Jenkins, install gcc-8 and set CONFIG_PAL_STACK_PROTECTOR=y and CONFIG_SHIM_STACK_PROTECTOR=y)

diff --git a/Makefile.configs b/Makefile.configs
index 9659bb62..adbec4e7 100644
--- a/Makefile.configs
+++ b/Makefile.configs
@@ -1,8 +1,8 @@
 ifeq ($(origin CC),default)
-CC     = gcc
+CC     = gcc-8
 endif
 ifeq ($(origin AS),default)
-AS     = gcc
+AS     = gcc-8
 endif
 ifeq ($(origin AR),default)
 AR     = ar rcs
@@ -11,8 +11,8 @@ ifeq ($(origin LD),default)
 LD     = ld
 endif
 
-CONFIG_PAL_STACK_PROTECTOR = n
-# CONFIG_PAL_STACK_PROTECTOR = y
+# CONFIG_PAL_STACK_PROTECTOR = n
+CONFIG_PAL_STACK_PROTECTOR = y
 
-CONFIG_SHIM_STACK_PROTECTOR = n
-# CONFIG_SHIM_STACK_PROTECTOR = y
+# CONFIG_SHIM_STACK_PROTECTOR = n
+CONFIG_SHIM_STACK_PROTECTOR = y

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/774)


dimakuv:
Looks like we merged all PRs this PR depends on. @yamahata, could you rebase this PR to the new master so we can do a review of this PR-related changes only?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/774)* status: 0 of 32 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/774#-MP4TD_Y3z02Ar3bRywW:-MP4TD_Y3z02Ar3bRywX:b2x3gm2) (no related file):*
Summary of this PR:

GCC 8+ allows to use the "stack protector" feature (with advanced options `-mstack-protector-guard -mstack-protector-guard-reg -mstack-protector-guard-offset`). In a nutshell, the user program must specify (1) the register + offset at which a random stack canary resides and (2) the function `__stack_chk_fail` which is called when a stack corruption is found at runtime.

This PR adds stack protectors as follows:
- For LibOS, it is x86-64-specific `-mstack-protector`, with the stack canary put inside the thread TCB, and the `guard-reg` register is FS (same as was used for LibOS's TCB).
- For Linux PAL, it is x86-64-specific `-mstack-protector`, with the stack canary put inside the thread TCB, and the `guard-reg` register is GS (same as was used for PAL's TCB).
- For trusted part of Linux-SGX PAL, it is x86-64-specific `-mstack-protector`, with the stack canary put inside the thread TCB, and the `guard-reg` register is GS (same as was used for SGX PAL's TCB).
- For untrusted part of Linux PAL, it is simple `-fstack-protector` (since untrusted part of Linux PAL already uses Glibc and its provided TCB).
- For other parts of the code (in particular, under `Pal/lib`), stack protector is still disabled.

The PR uses a per-thread random stack canary via `DkRandomBitsRead()` on each thread creation, putting it inside a new field `canary` in the TCB. If `DkRandomBitsRead()` fails for some reason, we fall back to a value `0xbadbadbadbad`.

I'm not sure we want to introduce this feature. Stack protectors slightly increase the Graphene executable/libraries size and slightly degrade runtime performance.

On the other hand, it is another layer of protection (panic instead of silent corruption on enclave-stack overflow). Maybe we should revive this effort (in a new PR since so much code changed in the meantime).

---

PR #775, opened 2019-06-20
[LibOS] Fix memory leak of qstr->oflow string during checkpointing

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

During checkpointing, qstr->oflow was redirected to point to the memory
region in the checkpoint but the old memory region was not freed. This
commit fixes this memory leak. It also removes unnecessary check for the
case when qstr->oflow and qstr->len are set incorrectly.

## How to test this PR? 

Regression tests must not fail.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/775)


dimakuv:
Retest Jenkins please

PR #799, opened 2019-06-25
[LibOS] Fix bug of long relative paths (more than 256B)

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Previously, Graphene had a bug when dentry's rel_path was not initialized to actual relative path if parent + filename length was longer than 256B. The problem was that rel_path was a shim_qstr which could contain maximum 256 chars. This commit increases the limit to 4096B and adds checks to fail explicitly if rel_path is still too large. Also, it adds LibOS regression test.

Note that relative paths greater than 256 chars in length are not uncommon. Most OSes impose the limit of 4096 chars or no limit at all.

## How to test this PR? 

New test `fs_long_path` is added to LibOS regression suite.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/799)


dimakuv:
Retest Jenkins-Debug please (`kill01` failed)

PR #809, opened 2019-07-03
[GLIBC] Patches for 2.23 and 2.27

chiache:
## Affected components

- [ ] README and global configuration
- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code
- [x] Library OS (i.e., SHIM), including GLIBC

## Description of the changes 

This PR upstreams the GLIBC patches for 2.23 and 2.27. Users can select the GLIBC version with `make GLIBC_VERSION=xxx`. The PR works with both Ubuntu 16.04 and 18.04.

## How to test this PR? 

Default glibc version is updated to 2.27. Just run the regression tests as usual.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/809)

PR #810, opened 2019-07-03
[LibOS] shim handle corruption caused by ambiguous interpretation 

chiache:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

This PR is needed for #809.

In LibOS, a `shim_handle` contains a union structure `info` to store the FS-specific information. In [`fs/chroot/fs.c`](https://github.com/oscarlab/graphene/blob/7ac25ca5560523a48226abfb3b70f252387e8df7/LibOS/shim/src/fs/chroot/fs.c),  `handle->info` is exclusively intepreted as a `shim_file_handle`. However, in [`fs/namei.c:598-616`](https://github.com/oscarlab/graphene/blob/7ac25ca5560523a48226abfb3b70f252387e8df7/LibOS/shim/src/fs/shim_namei.c#L598-616), if the dentry being opened represents a directory, the code will treat `handle->info` as a `shim_dir_handle`. As a result, it corrupts the content of `shim_file_handle`, causing unexpected failure at run time.

One symptom of this bug is that if you open a directory (e.g., the current directory) and then use `fstat` to retrieve the metadata of the file directory, you will either get wrong values in the `stat` structure, or get an internal memory fault when closing the file descriptor. The exact behavior of this bug is up to the compiler.

To fix this bug, we need to separate the directory info (`shim_dir_handle`) from the FS-specific `handle->info` field since it is accessed in an FS-generic source file. `handle->info` should only be interpreted by FS implementation and should remain opaque for the virtual file system layer.

## How to test this PR? 

A test program `fstat_cwd` is added to the regression directory. Normally, this program should print out as follow:

    fstat returns the fd type as S_IFDIR

However, under the buggy implementation, the program will either get the fd type as something abnormal, or fault at closing the file descriptor:

    $ ./pal_loader fstat_cwd
    Internal memory fault at 0x7fa0000000fa (IP = +0x21cd5, VMID = 3290854439, TID=1)
    [1]    29735 trace trap (core dumped)  ./pal_loader fstat_cwd

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/810)

### DIFF
--- a/LibOS/shim/include/shim_handle.h
+++ b/LibOS/shim/include/shim_handle.h
@@ -349,7 +349,6 @@ struct shim_handle {
         struct shim_dev_handle    dev;
         struct shim_pipe_handle   pipe;
         struct shim_sock_handle   sock;
-        struct shim_dir_handle    dir;
         struct shim_shm_handle    shm;
         struct shim_msg_handle    msg;
         struct shim_sem_handle    sem;
@@ -357,6 +356,8 @@ struct shim_handle {
         struct shim_str_handle    str;
         struct shim_epoll_handle  epoll;
     } info;
+
+    struct shim_dir_handle    dir_info;
 
     int                 flags;
     int                 acc_mode;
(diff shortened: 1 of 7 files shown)


chiache:
Retest Jenkins-SGX please.

PR #832, opened 2019-07-16
[LibOS] Fixing LTP test clone03

### LINKED ISSUE #529: clone03 is flaky
`clone03,1` from LTP fails from time to time.

Bash script for repro:
```Bash
for i in `seq 1 3000`; do
    echo "[$i]"
    ../../../../pal_loader clone03 || break
done
```

Example output, after running in `LibOS/shim/test/apps/ltp/opt/ltp/testcases/bin`:
```
[...]
[1366]
WARNING: no physical memory support, process creation may be slow.
clone03     1  TBROK  :  clone03.c:93: read from pipe failed: errno=EINTR(4): Interrupted system call
clone03     2  TBROK  :  clone03.c:93: Remaining cases broken
```

chiache:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

This PR fixes #529. Just as @dimakuv investigated, the root cause of this bug is the interruption of `read()` on the pipe between the parent and child processes. The emulation of `read()` is interrupted by the host-level signal sent from the IPC helper in order to deliver the SIGCHLD signal from the terminated child process. The `clone03` test does not expect `read()` to be interrupted because SIGCHLD is set to be ignored by default.

The quick fix presented in this PR is to skip the delivery of a signal if the target thread is set to ignore the said signal. This fix should at least stabilize `clone03` and prevent similar corner cases to happen again. However, this fix does not prevent the target thread from being interrupted by **other** signals (e.g., a SIGINT from the host). We may need a more fundamental fix in the future.

## How to test this PR? 

The potential way to reproduce this issue is to run `./pal_loader clone03` in a loop while running the whole regression in parallel. `clone03` should no longer fail under Graphene or Graphene-SGX.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/832)

### DIFF
--- a/LibOS/shim/include/shim_signal.h
+++ b/LibOS/shim/include/shim_signal.h
@@ -136,9 +136,10 @@ int init_signal (void);
 void __store_context (shim_tcb_t * tcb, PAL_CONTEXT * pal_context,
                       struct shim_signal * signal);
 
-void append_signal (struct shim_thread * thread, int sig, siginfo_t * info,
-                    bool wakeup);
-void deliver_signal (siginfo_t * info, PAL_CONTEXT * context);
+// Need to hold thread->lock when calling this function
+void append_signal(struct shim_thread* thread, int sig, siginfo_t* info, bool need_interrupt);
+
+void deliver_signal(siginfo_t* info, PAL_CONTEXT* context);
 
 __sigset_t * get_sig_mask (struct shim_thread * thread);
 __sigset_t * set_sig_mask (struct shim_thread * thread,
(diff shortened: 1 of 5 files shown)


dimakuv:
Jenkins, retest Jenkins please (`kill12` failed, and this is after my IPC fixes... Interesting, need to investigate)

dimakuv:
Jenkins, retest Jenkins-Debug please (`kill12` failed, similar to Jenkins pipeline; let's see if this is something deterministic -- e.g., IPC fixes conflicting with this PR)

dimakuv:
Jenkins, retest Jenkins-Debug please (kill12 failed; clearly some non-deterministic bug is triggered)

PR #833, opened 2019-07-16
Pytest

woju:
## Affected components

- [ ] README and global configuration
- [x] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code
- [x] Library OS (i.e., SHIM), including GLIBC

## Description of the changes 
Test reorganisation. `pytest` is used in place of custom runner. Test results are aggregated in JUnit-XML file for consumption in Jenkins, which should generate nice report of which tests exactly failed and were fixed in which commits.

## How to test this PR? 
Run `Jenkinsfile`s.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/833)

PR #840, opened 2019-07-17
[LibOS,Pal/{Linux,Linux-SGX}] Fix issues with large directory listing

### LINKED ISSUE #471: Unable to find certain pip installations after changing the python version
So I am working on trying to run a project in Graphene SGX that requires a specific anaconda environment to build. In doing so I am running into an issue where some of the pip installed packages (but not all of them) are not able to be imported as modules. Curiously the issue appears to be that the site-packages directory doesn't contain the same contents inside and outside of Graphene.

I have gotten the anaconda version to work in general and with some pip installed packages, but not all of them.

When running a simple script to print the contents of site packages using os.listdir outside of Graphene I see that the direct contains 426 files or directories, but when run inside Graphene it only contains 204, excluding many packages I need. For example I cannot import past inside Graphene but I can do so outside and looking at the contents past is not a directory this is listed inside Graphene.

Its possible that I may be doing unnecessary imports because I was trying multiple path additions in order to try and locate these pip packages (and admittedly not doing so in the safest manner as I am still just trying to get it to run). Is it possible that memory sizes are restricting my imports? Based on some unscientific testing changes to the enclave size don't seem to effect the number of files imported. Any suggestions are appreciated.

boryspoplawski:
## Affected components

- [x] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 
Fixes bugs preventing large directories from being listed  + rewritten related functions for speed and clarity.
Probably fixes #471

## How to test this PR? 
There's a regression test added, I'll enable it after #833 gets merged.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/840)

### DIFF
--- a/LibOS/shim/include/shim_handle.h
+++ b/LibOS/shim/include/shim_handle.h
@@ -23,6 +23,8 @@
 #ifndef _SHIM_HANDLE_H_
 #define _SHIM_HANDLE_H_
 
+#include <stdalign.h>
+
 #include <shim_types.h>
 #include <shim_defs.h>
 #include <shim_sysv.h>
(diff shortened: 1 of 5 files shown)


dimakuv:
Jenkins, retest this please

dimakuv:
Jenkins, retest this please

dimakuv:
Jenkins, retest Jenkins-SGX please (`Base /proc path present` LibOS test failed)

mkow:
Jenkins, retest this please

Issue #824, opened 2019-07-15
[Pal/Linux-SGX] Program crashes on exit from multi-threaded execution


alexandermerritt:
I hope I am not conflating multiple problems together, but I am able to verify that two behaviors together cause undesired program exiting or crashing: exiting a program while threads created by main continue to execute.

**Symptom:**

In Linux-SGX, ~15% of the time (like 3 every 20 runs), the program experiences an AEX upon exiting the application and hangs.

This problem arose from testing Golang applications. The runtime spawns threads on creation of the process, even if you do not use any in the implementation (e.g., for garbage collection, etc.). When the program exits, these Golang threads are still around and result in multiple (nested) entries into `_DkExceptionHandler` which results in an AEX and program halt. Golang apps were used on an as-yet-unpublished branch of Graphene used internally at Intel. I've tried replicating it with a C program on the mainline master branch, documented below.

**Root cause:**

Have not yet identified. Not sure next steps to investigate.

**Examples**

On the master branch commit 124a418 compiled to use Linux-SGX you can try this C application, placed in `LibOS/shim/test/native`. You'll need to add `-pthread` to `LDFLAGS` or to the linker command line, and add `sgx.thread_num = 16` to `manifest.template` before building.

```C
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/mman.h>
#include <sys/types.h>
#include <unistd.h>

const int N = 8;   // total threads
_Atomic int S = 0; // threads which have started
_Atomic int E = 8; // threads which have ended

void* thread(void* arg) {
    const int n = 1<<14;
    for (int i = 0; i < n; i++) {
        void *m = mmap(NULL, 4096, PROT_READ, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
        if (m != MAP_FAILED) munmap(m,4096);
        if (i == n/8) S++;
    }
    E--;
    puts("thread done");
    return NULL;
}

int main() {
    pthread_t tids[N];
    for (int i = 0; i < N; i++) {
        if (0 != pthread_create(&tids[i], NULL, thread, NULL))
            return 1;
    }
    while (S < N) ; // wait for threads to be in-progress
    // while (E != 0) ; // wait for threads to complete
    puts("--------------- Done");
    return 0;
}
```

Shows the following errors in the output:

```
(initial output omitted)
[     9] bkeep_unmapped: 0x8097cf000-0x8097d0000
[...]

PR #641, opened 2019-04-24
[PAL/Linux-SGX] Reuse TLS/TCS pages for newly created threads after p…

rainfld:
Currently, Graphene keeps using new TLS/TCS pages for newly created threads. This PR reuses the TLS/TCS pages released after the thread exits. This supports applications which create a huge amount of short-live threads at runtime. 

Please fill in the following form before submitting this PR and ensure that your code follows our [coding style guideline](../blob/master/CODESTYLE.md).

## Affected components

- [ ] README and global configuration
- [ ] Linux PAL
- [X] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code
- [ ] Library OS (i.e., SHIM), including GLIBC

## Description of the changes (reasons and measures)
1. Clear Thread Context (TCS/TLS, etc.) after Thread exits
2. A unite test which creates 128 short-live threads, testing if 4 TCS/TLS pages can handle it

Fixes #517 

## How to test this PR? (if applicable)
A LibOS layer unite test "multi_pthread" provided as part of regression

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/641)

PR #846, opened 2019-07-22
[LibOS/shim/test/native] multithread on exit test

alexandermerritt:
This test attempts to trigger race conditions observed on program exit
from an SGX enclave by creating busy threads, and then having the main
thread exit without joining on the child threads.

Since the underlying behaviors are racy, this test may need to be
executed many times to trigger a crash or hang.

Testing on the master branch, I can cause a program hang.

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code
- [x] LibOS test

## Description of the changes 

Introduce new regression test. No changes made to core implementation.

## How to test this PR? 

Compile as normal and execute this test. Continue execution until program crash or hang.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/846)


alexandermerritt:
Related Issue #824

alexandermerritt:
Related PR #591

dimakuv:
I ran this test multiple times (~30 times) on both Linux PAL and Linux-SGX PAL and didn't encounter any bugs/hangs. I believe #641 fixed the underlying issue. Also, since #641 introduced a similar test, I see no reason to keep this PR open. (But feel free to re-open it if you believe it is still useful.)

PR #848, opened 2019-07-22
[Pal/lib/string] add GNU strcmp,strncmp funcs

alexandermerritt:
Files pulled from glibc-2.29 with minor edits to build in Pal.

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

## Description of the changes 

Add GNU C Library implementation of `strcmp` and `strncmp` from v2.29. This commit supports upcoming PRs I will be filing.

## How to test this PR? 

No test currently exists. I did not modify the implementation, but am happy to write a Pal regression test for this, if desired.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/848)

### DIFF
--- a/Pal/lib/api.h
+++ b/Pal/lib/api.h
@@ -86,6 +86,9 @@ long int atol (const char *nptr);
 
 char * strchr (const char *s, int c_in);
 
+int strncmp (const char *s1, const char *s2, size_t n);
+int strcmp (const char *p1, const char *p2);
+
 void * memcpy (void *dstpp, const void *srcpp, size_t len);
 void * memmove (void *dstpp, const void *srcpp, size_t len);
 void * memset (void *dstpp, int c, size_t len);
--- a/Pal/lib/string/strcmp.c
+++ b/Pal/lib/string/strcmp.c
@@ -1,0 +1,44 @@
+/* Copyright (C) 1991-2014 Free Software Foundation, Inc.
+   This file is part of the GNU C Library.
+
+   The GNU C Library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
+   License as published by the Free Software Foundation; either
+   version 2.1 of the License, or (at your option) any later version.
+
+   The GNU C Library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


alexandermerritt:
> Why do you split these pulls from glibc into multiple PRs?

No reason. It existed as two commits in my internal branch. I probably could have squashed them before submitting the PR.

> Also glibc code is by far not the prettiest one and I doubt we need e.g. uppercase `STRNCMP`
> Maybe copy it from musl? https://git.musl-libc.org/cgit/musl/tree/src/string

Agreed. Other utilities in Graphene originated from GNU libc, and I was debating to include libc implementation from a BSD, instead, but decided to keep the status-quo (given I am a new contributor).

I am not opposed to pushing needed functions from musl libc, instead, if that is more desirable.

alexandermerritt:
task for @alexandermerritt submit a new PR with musl libc implementation that obviates the need for this one

alexandermerritt:
Do folks want me to just push the musl implementations of **just** the string functions I need, or also replace other existing functions in Pal with the equivalent musl code? The former means we have a mix of GNU libc and musl libc implementations.

alexandermerritt:
Incomplete list of challenges I encountered replacing some string files with musl implementations

- Build environment: exposes system header files to Pal source files. We may need to expose system headers to some files in Pal, and not to others (musl implementation).

musl files contain `#include <stdio.h>` which pulls in `/usr/include/stdio.h` which is usually GNU. One conflict is `strtol` which is a wrapper over `strtox` and that uses functions and macros accepting `FILE*` defined differently by GNU (`/usr/include/x86_64-linux-gnu/bits/types/__FILE.h`) and musl (`src/internal/stdio_impl.h`).

- Build environment: cannot specify per-file `CFLAGS` (e.g., `CFLAGS-foo.o:`) due to use of wildcards in locating implementation files. Musl implementation cause build warnings and Graphene uses `-Werror` which would break the build.

In musl some code performs pointer math which results in a signed result, and this it compares with unsigned values, e.g., `string/strstr.c` has `if (z-h < l)` declared as `size_t l; unsigned char *z, *h;`

dimakuv:
Closing this PR as we decided to use Musl versions of string functions

PR #847, opened 2019-07-22
[Pal/lib/string] Add GNU strstr string function

alexandermerritt:
File pulled from glibc-2.29 with minor edits to build in Pal.

Signed-off-by: Alex Merritt <mail@alexmerritt.us>

## Affected components

- [ ] README and global configuration
- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code
- [ ] Library OS (i.e., SHIM), including GLIBC

## Description of the changes 

Add GNU C Library implementation of `strstr` from v2.29. This commit supports upcoming PRs I will be filing.

## How to test this PR? 

No test currently exists. I did not modify the implementation, but am happy to write a Pal regression test for this, if desired.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/847)

PR #849, opened 2019-07-22
[Pal] fix libos preload handling

alexandermerritt:
With this patch, Pal now invokes ELF initializers of preloaded libraries in the order
specified (those in the manifest file in `libos.preload`).

Also fix how CALL_ENTRY is invoked; it should only be invoked on the
LibOS library libsysdb.so. CALL_ENTRY executes the ELF entry point,
which does NOT return to the caller.

In the manifest, user is now required to add `loader.libos` to tell Pal
a substring name of the LibOS library file (retaining it in the
preload list). Adding this allows us to also decouple the actual name of
the LibOS library, and enable it to become position-independent in the
preload list.

For example:

loader.preload = file:libsysdb,file:libfoo.so,file:libbar.so
loader.libos = sysdb

ELF initializers will be invoked in a left-to-right order, with the ELF
entry point invoked on libsysdb.so.

Dependencies between dynamic libraries are NOT understood or caught by
this change set. If one library depends on the state initialized by
another, the user must currently ensure the correct order.

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

## Description of the changes 

See above.

A subsequent PR will allow an easy way to build libs to be used for preloading to LibOS.

Depends on PR #847 and #848  

## How to test this PR? 

This change set does not include a test.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/849)

### DIFF
--- a/LibOS/shim/test/benchmark/manifest.template
+++ b/LibOS/shim/test/benchmark/manifest.template
@@ -1,4 +1,5 @@
 loader.preload = file:../../src/libsysdb.so
+loader.libos = libsysdb
 loader.env.LD_LIBRARY_PATH = /lib
 loader.debug_type = none
 loader.syscall_symbol = syscalldb
--- a/LibOS/shim/test/inline/manifest.template
+++ b/LibOS/shim/test/inline/manifest.template
@@ -1,2 +1,3 @@
 loader.preload = file:../../src/libsysdb_debug.so
+loader.libos = libsysdb
 loader.debug_type = inline
--- a/LibOS/shim/test/native/exec_fork.manifest.template
+++ b/LibOS/shim/test/native/exec_fork.manifest.template
@@ -1,4 +1,5 @@
 loader.preload = file:$(SHIMPATH)
+loader.libos = libsysdb
 loader.env.LD_LIBRARY_PATH = /lib
 loader.debug_type = inline
 loader.syscall_symbol = syscalldb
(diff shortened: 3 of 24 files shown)


amerritt-intel:
___
*[Pal/src/db_rtld.c, line 1425 at r1](https://reviewable.io/reviews/oscarlab/graphene/849#-LkS5DgM-U0Rq1_KAVUr:-Ll49rIT1xRfq_ZnuM7n:b-jf4352) ([raw file](https://github.com/oscarlab/graphene/blob/14c41d38a81bc014beb3842ce1789d98a991b7a0/Pal/src/db_rtld.c#L1425)):*

Please separate this error into two (it's easy to check which case it is here)
</blockquote></details>

For some usages, as you stated in some comments on this PR, the library libsysdb.so is not used. If we add a warning message when we do not find `loader.libos` then it will incorrectly appear in usages which do not require the LibOS.

This was not an issue before, because `CALL_ENTRY` was invoked on some library in the preload list, so one could simply omit libsysdb.so and those usages would be fine.

Perhaps we can require `loader.libos` to always exist, but if it is empty (`loader.libos = `) then we assume it is intentionally not needed and the user is made aware of this; else if `loader.libos = file:<path>` is specified then we will load and enter it.

Thoughts?

dimakuv:

mkow:
FYI: I plan to get rid of this PAL-level library preloading altogether in my loader rework, stay tuned!

PR #850, opened 2019-07-22
[LibOS/libs] Add build area for LibOS libraries

alexandermerritt:
Creates LibOS/libs/ into which one can build new libraries that are
preloaded via `loader.preload` manifest option to add functionality to
LibOS.

Execution starts with the initializers of the library, demonstrated by
the library in example/

Users can build their library by invoking make inside LibOS/libs/foo if
their library is named "foo". This will build a dynamic library and
place it into Runtime/

To make a new library, copy the example directory including the
Makefile.

Libraries are intentionally NOT automatically built with Graphene; this
can be changed by adding the library subdirectory name to the makefile
LibOS/libs/Makefile

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

See above

## How to test this PR? 

Build the example library, add libexample.so to the preload list in the manifest, and execute the application within an enclave.

One can invoke `pal_printf` to the example library in an initializer function to see the message is printed prior to the application printing any output (e.g., from a helloworld).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/850)


alexandermerritt:
Depends on #849

dimakuv:

PR #851, opened 2019-07-22
[LibOS/libs/symtab] Symbol table preload library

alexandermerritt:
When included in the library preload list within an application's
manifest file, this library will selectively map in portions of the
application's section headers that enable discovery of symbol
information.

This change set also looks up the beginning of the text section of the application, as it exists in the binary file. The runtime location can be calculated by adding the base location it is placed in by Pal, e.g., adding `pal_control.executable_range.start`.

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

See above. We introduce one function to look up symbols, and a global symbol to determine the ELF address of the text section.

## How to test this PR? 

No test is added with this change set.

Add this library to the preload list for an application in the manifest file. You can lookup symbols via any other preload library as one function is exported.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/851)


alexandermerritt:
Depends on #849 and #850

dimakuv:

PR #859, opened 2019-07-24
Turn on CPSTORE_DERANDOMIZATION, fixes apache on SGX.

### LINKED ISSUE #826: Checkpoint on child's do_migration may overlap with parent's VMAs
Here is an intermittent bug (due to randomization of address spaces):

```
[P00907] checkpoint mapped at 0xb970000-0xba2b793
[P00907] 767891 bytes read on stream
[P00907] restore checkpoint at 0x0b970000 rebased from 0x53f0000
[P00907] memory entry [0xb973610]: 0x600000-0x601000
...
[P00907] memory entry [0xb971fb0]: 0xb9c5000-0xba05000
[P00907] bkeep_mmap: 0xb9c5000-0xba05000
... child process fails ...
```

What happens is that:
1. Parent checkpoints its VMAs to restore them on the client. Notice the last VMA `0xb9c5000-0xba05000`. The checkpoint of size 767891B is sent to child as one blob.
2. Child allocates 767891B somewhere in its own memory to hold the checkpoint blob. Notice that the child randomly chose the addresses `0xb970000-0xba2b793`.
3. Child starts to restore VMAs. But the very first VMA it tries to restore is `0xb9c5000-0xba05000` -- overlapping with the checkpoint blob itself! Child cannot overwrite the checkpoint memory region, so the child panics (in `__bkeep_unmapped()`) and dies with `-13 = -EACCES`.

*Solutions*:
1. There is a macro disabled by default `CPSTORE_DERANDOMIZATION`. If this macro is enabled, the child process will always map the checkpoint at the same address as parent, thus never experiencing an accidental overlap with parent's VMAs. But this is bad for fragmentation/security.
2. The child process must move the checkpoint to some address where there will be no overlap with restored parent's VMAs. This seems like a sensible solution to implement.

donporter:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

The Apache test on SGX deterministically hit an ongoing problem with fork - that the cpstore location was creating conflicts.  In the short-term, disable CPSTORE randomization, to reduce the risk of address space conflicts.  In the longer term, we will need to clean up cpstore.

This makes Apache work on Linux-SGX, and probably reduces the likelihood of other heisenbugs.  

This also includes a few whitespace fixes to the Jenkinsfile.

Fixes #826 

## How to test this PR? 

Run the SGX Jenkins pipeline.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/859)

### DIFF
--- a/Jenkinsfiles/Linux-SGX
+++ b/Jenkinsfiles/Linux-SGX
@@ -51,20 +51,20 @@ pipeline {
                         timeout(time: 5, unit: 'MINUTES') {
                             sh '''
                                 cd LibOS/shim/test/apps/python
-                                                  make SGX=1
-                                                  make SGX_RUN=1
+                                make SGX=1
+                                make SGX_RUN=1
                                 make SGX_RUN=1 regression
                             '''
-      }
-      timeout(time: 5, unit: 'MINUTES') {
+                        }
+                        timeout(time: 5, unit: 'MINUTES') {
                             sh '''
                                 cd LibOS/shim/test/apps/bash
                                 make SGX=1
                                 make SGX_RUN=1
                                 make SGX_RUN=1 regression
                            '''
-      }
-                        timeout(time: 5, unit: 'MINUTES') {
+                        }
+                        timeout(time: 10, unit: 'MINUTES') {
                             sh '''
                                 cd LibOS/shim/test/apps/gcc
                                 make SGX=1
--- a/Jenkinsfiles/Linux-SGX
+++ b/Jenkinsfiles/Linux-SGX
@@ -88,7 +88,6 @@ pipeline {
                             sleep 10
                             LOOP=1 CONCURRENCY_LIST="1 32" ./benchmark-http.sh `hostname -I|tr -d '[:space:]'`:8000
                             '''
-                            /*
                         sh '''
                             cd LibOS/shim/test/apps/apache
                             make SGX=1
@@ -97,7 +96,6 @@ pipeline {
                             sleep 15
                             LOOP=1 CONCURRENCY_LIST="1 32" ./benchmark-http.sh `hostname -I|tr -d '[:space:]'`:8001
                             '''
-                            */
                     }
                     post {
                         always {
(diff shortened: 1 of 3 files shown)


dimakuv:
GCC failed on Jenkins-SGX with a error:
```
./gcc.manifest test_files/helloworld.c -o test_files/hello
*** allocating out of heap: 0x2f936000 ***
shim_init() in do_migration (-13)
gcc: internal compiler error: Killed (program cc1)
```

It looks like main GCC executable issues `execve("cc1")`, which spawns a child process that receives the checkpoint but cannot use the required CPSTORE address (not sure why). Let's try again and see if it is deterministic. But this is bad news anyway.

dimakuv:

donporter:
Perhaps GCC needs a larger enclave?

dimakuv:
GCC already uses 2GB enclaves (all executable which are run by GCC). So it doesn't seem to be the issue.

dimakuv:
This PR was updated to use https://github.com/oscarlab/graphene-tests/pull/14. That PR describes the issue, but in a nutshell GCC regression test `execve`s several subprocesses which have different enclave sizes. So when the parent GCC process has bigger enclave size, it creates a checkpoint at a high address which will not be available to the child process, and child panics.

dimakuv:
Jenkins, retest Jenkins please (Jenkins slave machine was unplugged in the meantime)

dimakuv:
NOTE: Jenkins-SGX failed previously because GCC had a very strict 5-minute timeout. The new fix pushed the execution time of the GCC regression test slightly above this limit. So I updated the timeout to 10 minutes.

donporter:
Seems like apache on SGX failed with a socket reset by peer.  Wonder how consistent this is.

donporter:
Jenkins, retest Jenkins-SGX please.

donporter:
Jenkins, retest Jenkins-SGX please. (previous test passed)

dimakuv:
Jenkins, retest Jenkins-SGX please (`vfork_and_exec` failed)

PR #868, opened 2019-07-26
Fix cpuid parsing on a system with hyperthreads

donporter:
## Affected components

- [X] Linux PAL
- [X] SGX PAL
- [X] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Add code to figure out whether hyperthreads are enabled, so that one can detect the total number of cores on the system.

## How to test this PR? 

Run the Bootstrap PAL regression on a HT system.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/868)


dimakuv:
This PR still doesn't help. All pipelines failed on the cpuinfo Pal regression test.

PR #869, opened 2019-07-26
clone03 fix followup

yamahata:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 
    [LibOS] teach append_signal() how to handle SIG_IGN/SIGCHLD
    
    This patch is follow up for #832, "[LibOS] Fixing LTP test clone03".
    When signal is masked ans handler = SIG_IGN, it shouldn't queue
    signal to avoid memory leak.
    If signal is SIGCHLD, it can be discarded even if it's masked.
    
    Reference:
    - Linux implementation: do_notify_parent() @ linux/kernel/signal.c
    - Standards:
      https://pubs.opengroup.org/onlinepubs/9699919799/functions/_Exit.html
    
    With glibc-2.19, kill12 is broken due to the its emulation.
    It may or may not success depending on which, SIGCHLD from parent or
    child nanosleep, is earlier.
    If SIGCHLD is earlyer, it success.
    If child nanosleep is earlier, it hangs.
    With the following changeset, glibc is updated so that the issue
    doesn't exit.
    
    Related:
    glibc patch
    > commit 8c873bf0190740ac1102e13ff7aeb6c08048abfd
    >    Remove signal handling for nanosleep (bug 16364)
    >
    >    Linux 2.6.32 and forward do not show the issue regarding SysV SIGCHLD
    >    vs. SIG_IGN for nanosleep which make it feasible to use it for sleep
    >    implementation without requiring any hacking to handle the spurious
    >    wake up.  The issue is likely being fixed before 2.6 and git
    >    history [1] [2].
    >
    >    This patch simplifies the sleep code to call nanosleep directly by
    >    using the posix default version.  It also removes the early cancellation
    >    tests for zero argument, since nanosleep will handle cancellation
    >    in this case.
    >
    >    [1] https://lkml.org/lkml/2004/11/25/5
    >    [2] https://lkml.org/lkml/2003/11/8/50
    >
    >    Checked on x86_64, ppc64le, and aarch64.
    >
    >    [BZ #16364]
    >    * sysdeps/unix/sysv/linux/sleep.c: Remove file
    >    * sysdeps/posix/sleep.c (__sleep): Simplify cancellation handling.
    
## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/869)

### DIFF
--- a/LibOS/shim/src/bookkeep/shim_signal.c
+++ b/LibOS/shim/src/bookkeep/shim_signal.c
@@ -678,7 +678,8 @@ __handle_one_signal(shim_tcb_t* tcb, int sig, struct shim_signal* signal) {
 
 void __handle_signal (shim_tcb_t * tcb, int sig)
 {
-    struct shim_thread * thread = (struct shim_thread *) tcb->tp;
+    struct shim_thread * thread = tcb->tp;
+    assert(thread);
     int begin_sig = 1, end_sig = NUM_KNOWN_SIGS;
 
     if (sig)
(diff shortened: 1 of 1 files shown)


dimakuv:
Jenkins, retest Jenkins please (`kill12` failed)

dimakuv:
Jenkins, retest Jenkins-SGX please (`vfork_and_exec` failed, slave uncledeadly)

PR #950, opened 2019-08-23
Fix make clean

HW42:
## Affected components

- [x] README and global configuration
- [x] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code
- [x] Library OS (i.e., SHIM), including GLIBC

## Description of the changes 

Fix make clean. There were a bunch of files which weren't cleaned up. This also adds a test to CI to check for such error in the future.

Depends on
https://github.com/oscarlab/graphene-sgx-driver/pull/6
https://github.com/oscarlab/graphene-tests/pull/31

## How to test this PR? 

See included test.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/950)

PR #965, opened 2019-08-28
[Makefile] Central build config

HW42:
## Affected components

- [x] Linux PAL
- [x] SGX PAL
- [x] FreeBSD PAL
- [x] Common PAL code

## Description of the changes 

Before this change the build config (like SGX=1, DEBUG=1, ...) was
passed via make command lines. Subsequent make invocations needed to get
the same arguments (make SGX=1, make SGX=1 regression, make SGX=1 clean,
...).

With this change the build configuration is set once by calling the new
'build-setup' script. Example:

    $ ./build-setup SGX=1
    $ make
    $ make -C Pal/regression regression
    $ make clean

For now build-setup is a simple shell script which mostly passes the
variable definition 1-to-1 into build-config.mk. In the future this can
be expanded with better error detection, searching for libraries, etc.

Depends on #950 and oscarlab/graphene-tests#38.

## How to test this PR? 

See above or Jenkins files for usage examples.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/965)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 !/Runtime/pal_loader
 /Jenkinsfiles/Jenkinsfile-*
 /Jenkinsfiles/JenkinsfileSGX-*
+/build-config.mk
 
 .lib
 *.i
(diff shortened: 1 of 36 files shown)


dimakuv:
We are re-working the whole build system and moving from Makefiles to Meson. Thus, this PR is obsolete.

PR #974, opened 2019-09-03
[Pal] Delete '\' removal in read_config

boryspoplawski:
If a backslah ('\') was encountered in a quoted ('"') value in a manifest,
it was getting deleted, but since the manifest is mapped read-only it caused
a segfault.
Removed this faulty behavior, now values are passed exactly as seen in
a manifest file, e.g. `loader.env.XXX = "\xY"Z"` passes a environmental
variable `XXX` with a value `"\xY"Z"`

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/974)

### DIFF
--- a/Pal/lib/graphene/config.c
+++ b/Pal/lib/graphene/config.c
@@ -333,18 +333,12 @@ int read_config(struct config_store* store, int (*filter)(const char* key, int k
         char* val = NULL;
         int vlen;
         if (*ptr == '"') {
-            int shift = 0;
-            val       = (++ptr);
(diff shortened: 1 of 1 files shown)


dimakuv:
Jenkins, test this please

mkow:
Jenkins, retest this please

Issue #980, opened 2019-09-06
18.04 is flaky on Jenkins


HW42:
When the tests on Jenkins run on Ubuntu 18.04, graphene seems to sometimes hang.

Failed examples:
https://raw.githubusercontent.com/HW42/graphene-build-logs/jenkins-full-output/graphene-18.04/694/build-log
https://raw.githubusercontent.com/HW42/graphene-build-logs/jenkins-full-output/graphene-18.04/685/build-log
https://raw.githubusercontent.com/HW42/graphene-build-logs/jenkins-full-output/graphene-18.04/747/build-log

Successful examples:
https://raw.githubusercontent.com/HW42/graphene-build-logs/jenkins-full-output/graphene-18.04/695/build-log
https://raw.githubusercontent.com/HW42/graphene-build-logs/jenkins-full-output/graphene-18.04/749/build-log

On Jenkins it hits very often the `test_pal.py::TC_01_Bootstrap::test_104_manifest_as_executable_name` (aka Bootstrap2) PAL regression test (timeout). As can be seen above this is not always the case.

I haven't managed to reproduce this locally so far. Even running 4 parallel loops, with each executing Bootstrap2 10 000 times (on a 2 cores 4 threads machine) didn't help. SGX also works (note that I didn't find a recent Jenkins log where that test fails for SGX).

HW42:
If someone has some hints (for example if you are able to observer this locally on some other test), please leave a comment).

mkow:
Observation:
I checked `graphene-18.04` pipeline and it *always* succeeds on `slave-leeroy` and `slave-uncledeadly` and *always* fails on `slave-pepe`, `slave-statler`, `slave-beaker12`, `slave-waldorf` and `slave-bobothebear`.

I don't know what's the difference between them nor have ssh access to investigate further. Maybe different kernel on the host or thread/cores count?
CC: @donporter, @dcowhig

HW42:
Hahaha ...

That timeout in https://raw.githubusercontent.com/HW42/graphene-build-logs/jenkins-full-output/graphene-18.04/685/build-log got me on the wrong track.

Compare `Bootstrap.c`:
```
[...]
    /* check manifest name */
    char manifest[30];
    DkStreamGetName(pal_control.manifest_handle, manifest, 30);
    pal_printf("Loaded Manifest: %s\n", manifest);
[...]
```

With `Bootstrap2.c`:
```
[...]
    /* check manifest name */
    char manifest[30] = "";
[...]

PR #986, opened 2019-09-09
Fix test 104 on 18.04

donporter:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

## Description of the changes 

On my test 18.04 system, this initialization line was the culprit with this test hanging.  

It was actually being treated as a SIGILL, and, for some reason, being ignored and getting in an infinite loop.

I am not sure if this is the only issue, but starting here.  Other tests also fail locally.  Will continue looking at this and may update.

## How to test this PR? 

Run the 18.04 pipeline on a host as mentioned in #980

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/986)

### DIFF
--- a/Pal/regression/Bootstrap2.c
+++ b/Pal/regression/Bootstrap2.c
@@ -10,7 +10,7 @@ int main(int argc, char** argv, char** envp) {
     pal_printf("Loaded Executable: %s\n", pal_control.executable);
 
     /* check manifest name */
-    char manifest[30] = "";
+    char manifest[30];
     DkStreamGetName(pal_control.manifest_handle, manifest, 30);
     pal_printf("Loaded Manifest: %s\n", manifest);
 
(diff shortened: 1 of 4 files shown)


donporter:
This is a bad fix.  The actual issue is -mavx in the makefile.

HW42:
@donporter Where does the SIGILL happen. Do you have a backtrace?

Are you aware that those array initialization are fine AFAIK? (I initially confused those array initializes with string literals. That's why I deleted my comment in #980.) I.e., `char manifest[30] = "";` is equivalent to `char manifest[30] = {'\0'};`. So it should work fine.

HW42:
Ah, just saw your comment in #980, will move discussion there.

PR #987, opened 2019-09-10
[Pal] regression: only enable -mavx for AvxDisable

yamahata:
Other executable doesn't need avx, and it causes failure on platform
that doesn't support avx.

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/987)

### DIFF
--- a/Pal/regression/Makefile
+++ b/Pal/regression/Makefile
@@ -1,7 +1,7 @@
 include ../src/Makefile.Host
 
 CC	= gcc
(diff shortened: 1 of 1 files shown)


mkow:
Jenkins, retest Jenkins please (trying to avoid HDD nodes)

PR #999, opened 2019-09-16
[SGX] Switch from requiring AVX making it optional by default.

donporter:
## Affected components

- [ ] Linux PAL
- [X] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Our current signer always signs SGX manifests in such a way that it requires AVX, even if the application doesn't use AVX.  This creates problems on some SGX v2 systems (NUCs) that don't support AVX.

Change the default behavior for AVX to be off by default.

In a future PR, we will change this behavior to enabled, but not required, by default.

## How to test this PR? 

Run a Jenkins pipeline on such a node.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/999)

### DIFF
--- a/Pal/regression/AvxDisable.manifest.template
+++ b/Pal/regression/AvxDisable.manifest.template
@@ -1,3 +1,3 @@
-sgx.enable_avx = 0
-sgx.enable_avx512 = 0
+sgx.require_avx = 0
+sgx.require_avx512 = 0
 sgx.allowed_files.res = file:avxRes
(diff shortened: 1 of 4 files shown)


dimakuv:
Jenkins, retest Jenkins-SGX please. An interesting failure:
```
+ make SGX_RUN=1 start-graphene-server
./httpd.manifest -D FOREGROUND -C "ServerName 127.0.0.1" -C "Listen 127.0.0.1:8001" -C "PidFile logs/httpd-127.0.0.1-8001.pid"
...
shim_init() in restore_checkpoint (-13)
```

PR #1004, opened 2019-09-18
[Pal/Linux-SGX] gdb wrapper: use add-auto-load-safe-path

dimakuv:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Previously, the GDB wrapper of Graphene-SGX used "set auto-load safe-path GDB_SCRIPT". This contradicts newer GDB versions which have a libpthread-2.27.so-gdb.py script in a different system dir. Thus, our wrapper overwrote the native GDB safe-path and broke GDB. This commit ensures that our wrapper only adds paths to safe-path.

## How to test this PR? 

Try GDB on any executable under Ubuntu 18.04 (where PIE executables are default): `cd LibOS/shim/test/native && GDB=1 SGX=1 ./pal_loader helloworld`. GDB should not complain (without this PR, it says something like `warning: File "/usr/share/gdb/auto-load/lib/x86_64-linux-gnu/libpthread-2.27.so-gdb.py" auto-loading has been declined by your auto-load safe-path set to "graphene/Pal/src/host/Linux-SGX/debugger/pal-gdb.py".`).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1004)

### DIFF
--- a/Pal/src/host/Linux-SGX/debugger/gdb
+++ b/Pal/src/host/Linux-SGX/debugger/gdb
@@ -7,4 +7,4 @@ if [ -z "$INSIDE_EMACS" ]; then
 	set -x
 fi
 
-LD_PRELOAD=$GDB_SO gdb -iex "set auto-load safe-path $GDB_SCRIPT" "$@"
+LD_PRELOAD=$GDB_SO gdb -iex "add-auto-load-safe-path $GDB_SCRIPT" "$@"


dimakuv:

PR #1011, opened 2019-09-20
Makefile: don't include .manifest.sgx.d when cleaning as wildcard

yamahata:
Since Make tries to update those .manifest.sgx.d as dependency,
but it isn't wanted when cleaning.

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1011)


dimakuv:
We are moving the build system from Makefiles to Meson, so this PR becomes obsolete. Closing it.

PR #1032, opened 2019-09-26
add errno EAFNOSUPPORT

pangzi85480:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 
When redis runs, it tries to listen on the ipv6 port. If the listen()
returns -1 and the errno is EAFNOSUPPORT, it will try to listen to the
ipv4 port, otherwise it will exit the program. So EAFNOSUPPORT needs to
be handled by Pal.

## How to test this PR? 
Run redis on a host that does not support ipv6.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1032)

### DIFF
--- a/LibOS/shim/src/shim_init.c
+++ b/LibOS/shim/src/shim_init.c
@@ -104,6 +104,7 @@ static int pal_errno_to_unix_errno [PAL_ERROR_NATIVE_COUNT + 1] = {
         /* PAL_ERROR_ZEROSIZE       */  0,
         /* PAL_ERROR_CONNFAILED     */  ECONNRESET,
         /* PAL_ERROR_ADDRNOTEXIST   */  EADDRNOTAVAIL,
+        /* PAL_ERROR_AFNOSUPPORT    */  EAFNOSUPPORT,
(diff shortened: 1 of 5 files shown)


pangzi85480:
@dimakuv @yamahata 

This error can only occur if the following specific conditions are met:
1 host machine does not support ipv6.
2 no redis.conf is specified when the redis-server is started, or ipv4 is not set(eg. “bind 127.0.0.1”) in redis.conf.

In fact, if you set "bind 127.0.0.1" in redis.conf and specify redis.conf at startup, the error will not occur.
If we want to keep the min set of error codes in PAL,  EAFNOSUPPORT can be ignored.

dimakuv:
I would prefer to not add more ABIs unless really necessary. This issue looks like a mis-configuration of Redis.

dimakuv:
Jenkins, test this please

dimakuv:
Jenkins, retest this please

PR #1033, opened 2019-09-26
Fix and test case for resource leak on file close.

thomasknauth:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

A sequence of open(), read(), close() led to a situation where the underlying host FD would never be released. The fix no longer associates the PAL handle with the VMA structure, resulting in a different code path where the ref counting on the PAL handle actually drops to 0 and the underlying host FD is correctly released.

## How to test this PR? 

I've added a simple regression test that fails without this patch.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1033)


dimakuv:
Jenkins, test this please

dimakuv:
Jenkins, test Jenkins-18.04 please (`sigpending02` failed)

dimakuv:
Jenkins, retest this please (somehow this job was not picked up)

PR #1034, opened 2019-09-26
Fix return code on exit due to signal

### LINKED ISSUE #841: [PAL] Exit code is 0 on SIGSEGV
Both with and without SGX. It would be nice to fix for detecting crashing tests and similar things.

thomasknauth:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Exit due to signal produced a return code of 0. This correctly propagates the return code on exiting from a signal.

Closes #841.

## How to test this PR? 

Test case is provided.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1034)

### DIFF
--- a/LibOS/shim/src/sys/shim_exit.c
+++ b/LibOS/shim/src/sys/shim_exit.c
@@ -135,7 +135,7 @@ int try_process_exit (int error_code, int term_signal)
     struct shim_thread * cur_thread = get_cur_thread();
 
     cur_thread->exit_code = -error_code;
-    cur_process.exit_code = error_code;
+    cur_process.exit_code = term_signal ? term_signal : error_code;
     cur_thread->term_signal = term_signal;
 
     if (cur_thread->in_vm)
@@ -170,6 +170,15 @@ noreturn int shim_do_exit_group (int error_code)
     struct shim_thread * cur_thread = get_cur_thread();
     assert(!is_internal(cur_thread));
 
+    /* If exit_group() is invoked multiple times, only a single invocation proceeds past this
+     * point. Kill signals are delivered asynchronously, which will eventually kick the execution
+     * out of this loop.*/
+    static struct atomic_int first = ATOMIC_INIT(0);
+    if (atomic_cmpxchg(&first, 0, 1) == 1) {
+        while (1)
+            DkThreadYieldExecution();
+    }
+
     if (debug_handle)
         sysparser_printf("---- shim_exit_group (returning %d)\n", error_code);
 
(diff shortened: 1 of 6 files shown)


dimakuv:
Jenkins, test this please

dimakuv:
Jenkins, test this please

dimakuv:
Hm, things are not that simple. Linux PAL fails on LTP tests:
```
Test 'pipe04,1' did not pass in the current run!!
Test 'waitpid02,3' did not pass in the current run!!
Test 'waitpid05,2' did not pass in the current run!!
...
```

Thomas, could you try these and see why they fail?

dimakuv:
Linux-SGX PAL fails on the OpenMP LibOS regression test. Thomas, please debug.

dimakuv:
Jenkins, test Jenkins-SGX please (`openmp` passes on our local systems)

dimakuv:
Jenkins, test this please

dimakuv:
So there is a funny bug with `clone02` LTP test. This test does something **very exotic**:
```
$ strace -ff ./clone02
clone(..., flags=CLONE_VM|CLONE_FS|CLONE_FILES|CLONE_SIGHAND|SIGCHLD) = 13011
```

Notice how this test create the child *process* and not *thread*: it asks to share memory via `CLONE_VM` but doesn't ask to create the thread via `CLONE_THREAD`. This is exotic and makes little sense (you may read some hilarious comments here: https://sourceware.org/bugzilla/show_bug.cgi?id=10311).

Graphene LibOS code silently assumes the pairing `CLONE_VM | CLONE_THREAD` (create a normal thread) or `CLONE_VM | CLONE_VFORK` (create a process via vfork with subsequent execve). However, it never explicitly checked for this in `shim_do_clone()`.

This exotic clone invocation is only found in `clone02`, and this test worked previously *by accident*. Basically, Graphene thought of the parent and child processes as two threads of the same process. This PR revealed this bug: the child process does `exit_group()` -- in normal Linux it actually kills the process, but in Graphene it kills the "child thread". And this PR introduces the logic of "wait for all threads to die before finally killing myself" -- and because the parent is another "thread" in this scenario, the "child thread" waited forever for it to die. Hence, a hang under Graphene.

Since emulating the insanity of Linux is low in priority (I don't think any real-world application uses the model of shared-memory-but-separate-processes), we should simply disable `clone02` LTP test and mark it as insane. Then this PR will pass our Jenkins.

TLDR, actions to do:
1. Update `BLOCKED` file under `LibOS/shim/test/apps/ltp` to include `clone02` and add a comment like "tries to create shared-memory but separate processes (CLONE_VM without CLONE_THREAD/CLONE_VFORK); Graphene doesn't support this exotic model".
2. Create PR for (1) under https://github.com/oscarlab/graphene-tests.
3. Update this PR with the hash to the new PR in (2).
4. Add the check to disallow CLONE_VM without CLONE_THREAD/CLONE_VFORK in `shim_do_clone()` -- return error with debug output on this.

donporter:
Ok to test

dimakuv:
Jenkins fails because it cannot fetch submodule from Thomas's private graphene-tests fork. We'll have to copy that branch into our origin for Jenkins to succeed. This error is really annoying.

mkow:
> Jenkins fails because it cannot fetch submodule from Thomas's private graphene-tests fork. We'll have to copy that branch into our origin for Jenkins to succeed. This error is really annoying.

Fixed (I mirrored his branch in our repo).

mkow:
Jenkins, retest this please

mkow:
Jenkins, retest this please (I had to mirror https://github.com/oscarlab/graphene-tests/pull/46 in the non-remote repo)

PR #1048, opened 2019-10-01
Clean up asserts

### LINKED ISSUE #1018: Introduce macro static_assert and replace all _Static_assert
In Glibc, `static_assert` is an alias to the C11 keyword `_Static_assert`. Usually, `static_assert` is defined in the `assert.h` header. But our PAL has its own `Pal/lib/assert.h` which doesn't have the `static_assert` macro.

Our current Graphene code uses `_Static_assert`. In the future PR, we should:
- Introduce `static_assert` in `Pal/lib/assert.h`
- Replace all `_Static_assert` with `static_assert`.

mkow:
## Affected components

- [X] Linux PAL
- [X] SGX PAL
- [X] FreeBSD PAL
- [X] Common PAL code

## Description of the changes 

Various assert-related clean-ups:
* Make assert() a no-op in non-debug builds,
* Use static_assert for compile-time asserts,
* Fix assert() implementation (previous version didn't work for expressions with types larger than long, it also always printed `(value:0)`),
* Clean up calls to asserts.

When working on them, I also noticed that `PAL_HANDLE` type is used inconsistently, so I cleaned it up. And two typos fixed as a bonus (`extern struct hadnle_ops udppacket_ops;` and `mkdir & chdr`).

Fixes #1018.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1048)

### DIFF
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -34,6 +34,8 @@
 
 #define static_always_inline static inline __attribute__((always_inline))
 
+#include <api.h>
+#include <assert.h>
 #include <shim_types.h>
 #include <shim_defs.h>
 #include <atomic.h>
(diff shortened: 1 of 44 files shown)


mkow:

mkow:
Jenkins, retest Jenkins-Debug-18.04 please (`stat01_64,1` and `stat01_64,2` failed, weird)

PR #1051, opened 2019-10-01
No longer set SO_LINGER on socket.

thomasknauth:
## Affected components

- [x] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

For reasons that have long been forgotten, Linux PAL set SO_LINGER with a zero timeout before closing the socket. [Among other things](https://stackoverflow.com/questions/3757289/tcp-option-so-linger-zero-when-its-required), this leads to ab (Apachebench) reporting failed requests when running nginx/lighttpd on Graphene, because the TCP connection is not closed cleanly.

## How to test this PR? 

Run ab against nginx/lighttpd. Without this fix, you will likely see failed requests. With this patch, all requests should succeed.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1051)

### DIFF
--- a/Pal/src/host/Linux/db_sockets.c
+++ b/Pal/src/host/Linux/db_sockets.c
@@ -956,11 +956,6 @@ struct __kernel_linger {
 
 static int socket_close(PAL_HANDLE handle) {
     if (handle->sock.fd != PAL_IDX_POISON) {
-        struct __kernel_linger l;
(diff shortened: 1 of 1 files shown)


donporter:
Add to whitelist

Issue #627, opened 2019-04-18
Check for missing handling of extended state (FPU, SSE, AVX, etc.)


HW42:
PR #626 fixes leaking of information in "extended" registers.

We need to check where else we are missing handling of extended state:

 - Should we reset the extended state to a known state on enclave entry?
 - Before exiting for an ocall we only call `fxsave`. The Intel SDK uses `xsave` here. OTOH AFAIU only `MXCSR` and `FCW` are callee saved.
 - The exception handler needs to save extended state.

Anything else?

dimakuv:
> The exception handler needs to save extended state.
Is #397 a step in the right direction?

dimakuv:
> Should we reset the extended state to a known state on enclave entry?

Yes, we can re-use the clean XSAVE area state introduced in #626. This should go before `callq handle_ecall`.

yamahata:
To prevent information leak, those three above would cover it.

I dumped what I have as #632.
_DkHandleExternalEvent can call exception handler which requires the context of fp registers. 
Possibly we can simply zero the saving area except MXCSR and FCW instead of xsave/fxsave.
For correct ABI for user application, fp context on entering LibOS should be passed.
I don't know any application that uses fp context on entering kernel by signal handler, though.
In practice, it wouldn't matter.

HW42:
Ok, so AFAICS we have:

 - Clear before EEXIT is implemented in PR #626 and as part of PR #632.
 - PR #632 covers ocalls and exception handling, I expect that @yamahata will provide separated PRs soon-ish.
 - I can provide a simple follow up to PR #626 to clear before ecall handler.

@yamahata: What do you think? How show we resolve PR #626 vs your changes?

yamahata:
@HW42 sounds good.

dimakuv:
Btw, we are planning to merge 626 soon.

mkow:
FYI: #626 was just merged.

HW42:
> - I can provide a simple follow up to PR #626 to clear before ecall handler.

This is a bit more complex than I thought. The problem is that we didn't read the value of XFRM yet before calling the ecall handler. I see 3 options:

 1. Only do fxrstor in early assembly and delay clearing the rest (i.e. features not covered by fxrstor) to `enclave_init`. In this case I would wait for the XSAVE relared parts of PR 632 since I also need to call xrstor from C.
[...]

PR #1064, opened 2019-10-05
[Pal/Linux-SGX] Fixing XSAVE estended state on EENTER/EEXIT

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

This PR contains two fixes in `enclave_entry.S`:

1. Fix offset of MXCSR "reset" XSAVE area. Before EEXIT, Graphene-SGX resets the extended state (XSAVE) area to the default state (of mostly zero bytes). This prevents the leakage of  x87/SSE/AVX/MPX register values inside the enclave. However, the previous default state had an incorrectly calculated offset of MXCSR. This commit fixes the offset of MXCSR. (This particular error manifested in spurious "Numeric underflow (#U)" SSE hardware exceptions on OpenVINO.)

2. Reset FXSAVE extended state on EENTER. Previously, the FXSAVE extended state (ST, XMM, MXCSR registers) was not cleared on EENTER (i.e., enclave-thread enter). This could lead to maliciously crafted ST/XMM registers propagating into the enclave and subverting execution. This commit resets FXSAVE on every EENTER to a default mostly zero-byte state.

Note that this commit does not reset XSAVE state (YMM, ZMM registers). This will be fixed in a future commit.

## How to test this PR? 

This bug was found during testing OpenVINO. The minimal example to reproduce SIGFPE is:
```c
#include <xmmintrin.h>
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

void* child_print(void* arg) {
//    _mm_setcsr(0x1f80);  /* uncomment to get rid of SIGFPE in Graphene-SGX */
    float res = strtof("1.000000013351432e-10", NULL);
    printf("child: %f\n", res);
    return NULL;
}

int main(int argc, char** argv) {
    float res = strtof("1.000000013351432e-10", NULL);
    printf("parent: %f\n", res);

    pthread_t thread;
    pthread_create(&thread, NULL, child_print, NULL);
    pthread_join(thread, NULL);
    return 0;
}

// Graphene-SGX will fail with something like:
// [     2] arithmetic fault at 0x048a1cd8
// [     2] SIGFPE handled
```

More info on this bug can be found here: https://github.com/oscarlab/graphene/issues/627#issuecomment-538588324

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1064)

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_entry.S
+++ b/Pal/src/host/Linux-SGX/enclave_entry.S
@@ -91,6 +91,15 @@ enclave_entry:
 	xorq %r14, %r14
 	xorq %r15, %r15
 
+	# Clear "extended" state (FPU aka x87, SSE, AVX, ...).
+	# TODO: We currently clear only state covered by FXRSTOR but not by XRSTOR
+	#       (e.g., no clearing of YMM/ZMM regs). This is because we didn't read
+	#       the value of XFRM yet, so we don't know whether XRSTOR is safe at
+	#       this point.
+	leaq .Lxrstor_init_arg(%rip), %rax
+	fxrstor (%rax)
+	xorq %rax, %rax
+
 	# register states need to be carefully checked, so we move the handling
 	# to handle_ecall() in enclave_ecalls.c
 	callq handle_ecall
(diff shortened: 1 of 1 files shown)


dimakuv:
Jenkins, retest Jenkins-SGX-18.04 please (`setuid01` LTP test failed, unrelated)

dimakuv:
Jenkins, retest Jenkins-18.04 please (`setuid01` LTP test failed, unrelated)

PR #1066, opened 2019-10-06
[Pal/Linux-SGX] load_trusted_file(): fix uninitialized pointer

dimakuv:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Previously, `*stubptr` was not initialized to NULL in the corner case of a file created from inside of the SGX enclave. This led to a subsequent failure in logic which tried to mmap an empty file (because it observed that `*stubptr` contained some value).

## How to test this PR? 

This obvious bug doesn't seem to deserve a separate test. But this is the minimal test to reproduce the bug. Add this to the manifest:
```
sgx.allowed_files.testfile123 = file:tmp/testfile123
sgx.allow_file_creation = 1
```
Write this example:
```
int main(int argc, const char** argv) {
    int fd = openat(AT_FDCWD,"tmp/testfile123",O_WRONLY|O_CREAT|O_TRUNC,0666);
    printf("fd = %d\n", fd);
    ...
```
Graphene-SGX prints an error like this:
```
fd = -1
```

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1066)

### DIFF
--- a/Documentation/oldwiki/Graphene-SGX-Manifest-Syntax.md
+++ b/Documentation/oldwiki/Graphene-SGX-Manifest-Syntax.md
@@ -60,6 +60,15 @@ This syntax specifies the files that are allowed to be loaded into the enclave u
 These files are not cryptographically hashed and are thus not protected. It is insecure to allow
 files containing code or critical information; developers must not allow files blindly!
 
+### Allowing File Creation
+
+    sgx.allow_file_creation=[1|0]
+    (Default: 0)
+
+This syntax specifies whether file creation is allowed from within the enclave. Set it to 1 to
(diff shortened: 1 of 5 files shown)


mkow:
Jenkins, retest Jenkins-SGX please (`test_libos.TC_40_FileSystem.test_010_path` failed, this test got unstable recently for unknown reasons)

dimakuv:
Jenkins, retest Jenkins-SGX please (`proc-path` LibOS regression test failed, this should be unrelated but weird, let's see if it is deterministic)

PR #1071, opened 2019-10-08
Don't use __builtin_expect

mkow:
## Affected components

- [X] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes

We don't need this micro-optimization which just obfuscates the sources for negligible performance gains. (Un)likeliness of branches in almost all cases should be derived from profiling, not from hand-written hints, using profile-guided optimization.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1071)


mkow:
Jenkins, retest Jenkins-SGX please (`test_libos.TC_40_FileSystem.test_010_path` failed, seems unrelated)

PR #1079, opened 2019-10-12
[LibOS] Fix SIGABRT, SIGTERM, SIGINT to kill the whole process

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Previously, there was a bug that SIGTERM and SIGINT sent from the host OS killed only one thread but not the whole process (including child processes). Additionally, SIGABRT of the application also killed only one thread instead of the whole process. This commit fixes this.

## How to test this PR? 

Maybe I need to introduce a LibOS test to spawn several threads, and one of them calls `abort()`. Without this PR, it will hang. With this PR, it will kill all threads and finish.

This bug was found and fixed while debugging TensorFlow. So at least one program is affected by this fix.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1079)

### DIFF
--- a/LibOS/shim/src/bookkeep/shim_signal.c
+++ b/LibOS/shim/src/bookkeep/shim_signal.c
@@ -794,18 +794,35 @@ void append_signal(struct shim_thread* thread, int sig, siginfo_t* info, bool ne
 
 static void sighandler_kill (int sig, siginfo_t * info, void * ucontext)
 {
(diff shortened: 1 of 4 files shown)


dimakuv:
Jenkins, test this please

PR #1082, opened 2019-10-14
[LibOS] Reorder argv in user stack

blukat29:
Reorder argv strings in initial user stack in a way that the former string is located at lower address.

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes

Reorders the location of `argv` and `envp` strings in user stack when they are copied during `populate_user_stack()`. After this patch, following equations should hold:

```
argv[0] < argv[1] < .. < argv[argc - 1];  (increasing order)
argv[i + 1] - argv[i] == strlen(argv[i]) + 1;  (compact)
``` 

I found the need for change when I tried to run node.js in Graphene. As soon as the process starts, `libuv` complained at below check (https://github.com/libuv/libuv/blob/v1.32.0/src/unix/proctitle.c#L62):

```c
process_title.str = argv[0];
process_title.len = argv[argc - 1] + strlen(argv[argc - 1]) - argv[0];
assert(process_title.len + 1 == size); /* argv memory should be adjacent. */
```

According to the [SYS V ABI](https://www.uclibc.org/docs/psABI-x86_64.pdf) (p29), the order of the argv strings does not matter. However it would be useful to allow some programs that follow this argv ordering convention.

Regarding coding style, I have tried `clang-format` but it just changed everything. So here I followed existing code.

## How to test this PR?

I have modified the `bootstrap.c` LibOS regression test case to check the libuv condition. `make regression` in `LibOS/shim/test/regression` should succeed as before.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1082)

### DIFF
--- a/LibOS/shim/src/shim_init.c
+++ b/LibOS/shim/src/shim_init.c
@@ -326,10 +326,20 @@ static int populate_user_stack (void * stack, size_t stack_size,
 
     new_argv = stack_bottom;
     while (argv) {
+        /* Even though the SysV ABI does not specify the order of argv strings,
+           some applications (notably Node.js's libuv) assume the compact
+           encoding of argv where (1) all strings are located adjacently and
+           (2) in increasing order. */
+        int argv_size = 0;
+        for (const char ** a = argv ; *a ; a++)
+            argv_size += strlen(*a) + 1;
+        char * argv_bottom = ALLOCATE_TOP(argv_size);
+
         for (const char ** a = argv ; *a ; a++) {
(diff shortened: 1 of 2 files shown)


dimakuv:
Jenkins, test this please

dimakuv:
Jenkins, test this please

dimakuv:
Jenkins, test this please

PR #1084, opened 2019-10-15
Misc optimizations: use O2 for non-debug builds, allow up to 1024 enclave threads

### LINKED ISSUE #1040: LibOS and -O2 causes failure of libos regression test on Linux-SGX
Now LibOS is compiled with -O0 (default).
With -O2 enalbed, Linux-SGX libos regression test fails. 
LibOS regression test on Linux, Linux-18.04, Linux-SGX-18.04 succeed. Also with gcc-8, it succeeds. (with #1039 applied)

Ubuntu 16.04 has gcc-5.4. Since it's old, it may not be worthwhile to investigate it. 
We can switch to 18.04.
Thoughts?

options:
- disable -O2 if gcc 5
- drop 16.04 support
- won't try -O2

dimakuv:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Two fixes found during performance-testing OpenVINO on huge servers:

1. [LibOS] Use -O2 optimization level when building in non-debug mode.
2. [Pal/Linux-SGX] Increase MAX_DBG_THREADS constant from 64 to 1024.

Explanation for the second fix:  an internal Graphene structure for GDB metadata limits the number of enclave threads to MAX_DBG_THREADS. Previously, it was set to 64, which was enough for typical platforms. However, powerful servers have hundreds of logical cores. Graphene-SGX failed with error on such servers. This commit increases the limit to 1024 (should be more than enough for all modern platforms).

## How to test this PR? 

All tests should continue running.

Fixes #1040. 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1084)

### DIFF
--- a/LibOS/shim/src/Makefile
+++ b/LibOS/shim/src/Makefile
@@ -63,6 +63,8 @@ ifeq ($(DEBUG),1)
 CC += -gdwarf-2 -g3
 CFLAGS += -DDEBUG
 ASFLAGS += -DDEBUG
+else
+CFLAGS += -O2
 endif
 export DEBUG
 
--- a/LibOS/shim/src/sys/shim_clone.c
+++ b/LibOS/shim/src/sys/shim_clone.c
@@ -122,9 +122,6 @@ int clone_implementation_wrapper(struct clone_args * arg)
 
     int stack_allocated = 0;
 
-    object_wait_with_retry(arg->create_event);
-    DkObjectClose(arg->create_event);
-
     /* We acquired ownership of arg->thread from the caller, hence there is
      * no need to call get_thread. */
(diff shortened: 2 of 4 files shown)


dimakuv:
Some LibOS tests failed on Linux-SGX. Need to investigate.

dimakuv:
Jenkins, retest Jenkins-Debug-18.04 please (`dup201` failed)

dimakuv:
Jenkins, retest Jenkins-SGX please (my local tests do not exhibit any problems)

dimakuv:

PR #1085, opened 2019-10-16
Changes to support eventfd in graphene.

### LINKED ISSUE #678: [LibOS] eventfd2 enabled?
Symptom: I'm getting an error EINVAL (Invalid argument) when calling eventfd(count, flags). 

Context: eventfd.c (in ./LibOS/glibc-2.19/sysdeps/unix/sysv/linux) checks for the define of __NR_eventfd2. If defined, the call is actually routed to the eventfd2 syscall, which is intercepted via SHIM_SYSCALL_PASSTHROUGH (eventfd2, 2, int, int, count, int, flags) in file ./LibOS.shim/src/shim_syscall.c. Hence the call to eventfd(count, flags) is passed on to the kernel using eventfd2. This is what we'd expect to happen. 

Issue: The observed error can only happen, if neither __NR_eventfd2 nor __ASSUME_EVENTFD2 are defined - from which we can conclude that eventfd2 is not enabled even though readily available.

Question: Is __NR_eventfd2 defined during GLIBC build? Asked differently, shouldn't __NR_eventfd2 be defined and __ASSUME_EVENTFD2 remain undefined during build of GLIBC? Sorry, with the complex makefile setup, I wasn't yet able to pinpoint a location to do this.

sudharkrish:
## Affected components

- [X] Linux PAL
- [X] SGX PAL
- [ ] FreeBSD PAL
- [X] Common PAL code

### **Eventfd Usages:**

1. For Inter-process communication, between processes or threads within a process, similar to pipes. 
2. Eventfd can also be used by linux kernel to signal user-space applications. 
Refer to man-page, for more details->[http://man7.org/linux/man-pages/man2/eventfd.2.html]

### **Options to implement in graphene: (info below based on discussion with chia-che):**
**Option1->Eventfd’s counter, simulated using pipes:**
Design: There are many ways to simulate. One way described here. Write on eventfd, will write # of bytes equal to the value of counter in buffer. Read on eventfd, will read the pipe’s buffer, until pipe is empty, and return counter(equal to # of bytes read).
Cons->In general, simulating eventfd on top of pipes was in-efficient and limited, since eventfd counter, can be lot more than pipe's buffer. Not feasible to support second usage listed above

**Option2->Eventfd’s counter, in-memory, using notification event:.**
Design: Feasible to support within a single-process, using synchronization primitive such as notification event. Cons->Non-trivial amount of work needed for multi-process. Not feasible to support second usage listed above.

**Option3->Eventfd using ocall to host OS:.**
More efficient, and can support all usages. Cons->Need to determine strategy to secure the eventfd.

For this PR, based on analysis, discussions, and also because our customer use-cases, need support for second usage, we are adopting Option3.

**Details on Implementation using Option3:**
At LibOS level, new fs/eventfd added, to support fs calls on eventfd.
At PAL level, added db_eventfd.c, to route eventfd calls to host kernel
using ocalls. Polling on group of eventfds, works fine. 

## How to test this PR? 
Added test application->eventfd_test.c under LibOS/shim/test/native

In order to support the second usage, user needs to patch LibOS(using->git clone git@github.com:chiache/graphene.git -b ioctl-passthru), will update graphene wiki with more details.
But the idea is application, will get virtual file-descriptor(virtual FD), and in order for linux kernel to signal the real FD, application needs to send it to kernel via custom IOCTL. LibOS needs to be patched to trap this custom IOCTL, and then retrieve, LibOS shim_handle from VFD, and then retreive the PAL handle's eventfd, and send it down to kernel.

Closes #678.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1085)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -569,6 +569,7 @@ extern struct shim_mount chroot_builtin_fs;
 extern struct shim_mount pipe_builtin_fs;
 extern struct shim_mount socket_builtin_fs;
 extern struct shim_mount epoll_builtin_fs;
+extern struct shim_mount eventfd_builtin_fs;
 
 /* proc file system */
 struct proc_nm_ops {
--- a/LibOS/shim/include/shim_handle.h
+++ b/LibOS/shim/include/shim_handle.h
@@ -50,6 +50,7 @@ enum shim_handle_type {
     TYPE_FUTEX,
     TYPE_STR,
     TYPE_EPOLL,
+    TYPE_EVENTFD
 };
 
 struct shim_handle;
--- a/LibOS/shim/include/shim_table.h
+++ b/LibOS/shim/include/shim_table.h
@@ -480,6 +480,8 @@ ssize_t shim_do_recvmmsg(int sockfd, struct mmsghdr* msg, size_t vlen, int flags
 int shim_do_prlimit64(pid_t pid, int resource, const struct __kernel_rlimit64* new_rlim,
                       struct __kernel_rlimit64* old_rlim);
 ssize_t shim_do_sendmmsg(int sockfd, struct mmsghdr* msg, size_t vlen, int flags);
+int shim_do_eventfd2(unsigned int count, int flags);
+int shim_do_eventfd(unsigned int count);
 
 /* libos call implementation */
 int shim_do_msgpersist(int msqid, int cmd);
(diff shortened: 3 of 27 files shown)


dimakuv:
Jenkins, test this please

dimakuv:
@sudharkrish All Jenkins failed on build. Could you please test locally with Graphene rebuilt as `make WERROR=1`? And please test both Linux and Linux-SGX PALs (that they build and all regression tests under `Pal/regression` and `LibOS/shim/test/regression` pass, except for remote attestation).

dimakuv:
Jenkins, test this please

dimakuv:
Jenkins, test this please

dimakuv:
Jenkins, test this please

dimakuv:

dimakuv:
Jenkins, test this please

PR #1088, opened 2019-10-16
Linters

woju:
## Affected components

## Description of the changes 
Linters (currently `pylint` and `shellcheck`) are run as another step in Jenkins pipeline. This step is before build, so in case of problems the build is not attempted to save executor time, because the pipeline would fail anyway. It is assumed the linters are at least an order of magnitude faster than actual compilation.

Right now there are many problems, so linters are run twice, the first time for information only, the second one causes pipeline to fail only if the problematic lines were touched in the PR.

## How to test this PR? 
This PR is selftesting. It should run on itself, because part of is Python and shell code.
As part of first submission, it will have some problems intentionally left uncorrected. After that, I'll submit patches with correction.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1088)

### DIFF
--- a/.ci/prfilter
+++ b/.ci/prfilter
@@ -48,6 +48,7 @@ class Diff:
 
     The default diff is to the ``origin/master`` ref.
     '''
+    # pylint: disable=too-few-public-methods
 
     def __init__(self, ref=DEFAULT_REF):
         self._files = get_diff_ranges(ref)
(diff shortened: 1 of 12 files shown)


dimakuv:

mkow:
Jenkins, retest Jenkins-Debug-18.04 please (`apps.LTP.kill11` failed with `fork() failed: errno=EACCES(13): Permission denied`, weird, I've never seen such a failure before)

PR #1106, opened 2019-10-24
Allow zero-size read/write on pal stream

yamahata:
## Affected components

- [x] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

## Description of the changes 
Allow zero size read/write on PAL stream

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1106)


LibOS/shim/src/fs/pipe/fs.c
```diff
@@ -40,28 +40,22 @@
...
-    if (!count)
-        return 0;
-
     PAL_NUM bytes = DkStreamRead(hdl->pal_handle, 0, count, buf, NULL, 0);
 
     if (!bytes)
         return -PAL_ERRNO;
 
-    assert((ssize_t)bytes > 0);
+    assert(count == 0 || (ssize_t)bytes > 0);
```
sudharkrish: If we are removing !count check, before calling DkStreamRead, then we end up adding it part of assert. This makes an assumption that host kernel would have returned error(negative), for count == 0. I think we should either leave this, as-it-exists, (OR) think if we really need any assert here, why not let the user-application handle the error.

LibOS/shim/src/fs/pipe/fs.c
```diff
@@ -40,28 +40,22 @@
...
-    if (!count)
-        return 0;
-
     PAL_NUM bytes = DkStreamWrite(hdl->pal_handle, 0, count, (void*)buf, NULL);
 
     if (!bytes)
         return -PAL_ERRNO;
 
-    assert((ssize_t)bytes > 0);
+    assert(count == 0 || (ssize_t)bytes > 0);
```
sudharkrish: Ditto.

LibOS/shim/src/fs/socket/fs.c
```diff
@@ -81,7 +78,7 @@ static ssize_t socket_read(struct shim_handle* hdl, void* buf, size_t count) {
             }
         }
 
-    assert((ssize_t)bytes > 0);
+    assert(count == 0 || (ssize_t)bytes > 0);
```
sudharkrish: Ditto.

LibOS/shim/src/fs/socket/fs.c
```diff
@@ -127,7 +121,7 @@ static ssize_t socket_write(struct shim_handle* hdl, const void* buf, size_t cou
         return -err;
     }
 
-    assert((ssize_t)bytes > 0);
+    assert(count == 0 || (ssize_t)bytes > 0);
```
sudharkrish: Ditto.

yamahata:
Jenkins, retest this please.

dimakuv:
Jenkins, retest Jenkins-Debug please

dimakuv:

dimakuv:
I had to update this PR slightly since the already-merged PR on eventfd introduced a new kind of streams (eventfd streams), and that code had to be updated similarly to how pipe code was updated by this PR.

PR #1108, opened 2019-10-25
Add string length check in strcmp_static macro.

sudharkrish:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [X] Common PAL code

## Description of the changes 
Per description of macro->"check if the var is exactly the same as the static string".
Fix the check in macro->strcmp_static, to check if both str and var are of same length.
Without this strlen check, memcmp, can scan outside the length of var argument.

## How to test this PR? 
Regression should be enough.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1108)

### DIFF
--- a/Pal/lib/api.h
+++ b/Pal/lib/api.h
@@ -146,8 +146,10 @@ void *calloc(size_t nmemb, size_t size);
 #define force_literal_cstr(str)   ("" str "")
 
 /* check if the var is exactly the same as the static string */
-#define strcmp_static(var, str) \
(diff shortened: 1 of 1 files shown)


dimakuv:
Jenkins, test this please

bigdata-memory:
I think it might be better to implement an inline function "int strncmp_static(const char *s1, const char *s2, register size_t n)" first, because strcmp is not memory safe.

PR #1126, opened 2019-11-04
Clean up interaction between initialize_enclave() and create_enclave()

thomasknauth:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Clear up the responsibilities between create_enclave() and its caller.

create_enclave() now assumes its caller picks a suitable base address and size according to SGX's requirements. Use create_enclave()'s secs parameter for in and outputs to reduce total number of arguments (not sure if this violates some Graphene coding standard). Also, I believe the code tacitly assumes that create_enclave()'s mmap creates the mapping at whatever address we provide. Hence, I changed the secs->base = addr into an assert().

## How to test this PR? 

Regression tests still pass.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1126)

### DIFF
--- a/Pal/src/host/Linux-SGX/sgx_framework.c
+++ b/Pal/src/host/Linux-SGX/sgx_framework.c
@@ -136,10 +136,11 @@ bool is_wrfsbase_supported (void)
 }
 
 int create_enclave(sgx_arch_secs_t * secs,
-                   unsigned long baseaddr,
-                   unsigned long size,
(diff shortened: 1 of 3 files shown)


dimakuv:
Jenkins, retest Jenkins-Debug please (unrelated)

PR #1128, opened 2019-11-05
[LibOS,Pal] Rewrite poll and ObjectWait logic from scratch

### LINKED ISSUE #228: shim_do_epoll_wait() is terribly inefficient

### LINKED ISSUE #1117: [LibOS,Pal] Rewrite poll and ObjectWait logic from scratch
Our polling code is plain terrible. It's time to rewrite it from scratch.

- [x] delete `_DkObjectWaitOne()`
- [x] delete `__try_alloca()` and everything it entails
- [x] clean up `struct shim_epoll_fd`:
    - [x] rename to `struct shim_epoll_item` (it is an internal item in the array maintained for the particular `epfd / epoll handle`)
    - [x] struct item `fd` is kinda redundant but nice to have as a shortcut (update: keeping it)
    - [x] struct item `pal_handle` is useless since we can always get it via `handle->pal_handle`
    - [x] struct item `back` seems to be for bookkeeping of in which epfds this fd/handle is registered; rename it (update: keeping the name, couldn't find a better one)
- [x] `__do_poll`: remove the logic of cherry-picking handles-to-poll (especially remove `hdl->fs->fs_ops->poll()`)
- [x] `struct poll_handle` doesn't need anything except `fd` and `flags`. It may be more readable to remove this struct and just pass two corresponding arrays to `__do_poll()`.
- [x] `_DkObjectsWaitAny()` clean up:
     - [x] leave "remove duplicated handles" logic in there and delete from LibOS layer
     - [x] remove special-case of one handle
- [x] delete all logic for `TRACE_HEAP_LEAK`
- [x] delete `DEFAULT_QUANTUM`
- [ ] check if we can replace `_DkObjectsWaitAny()` with `_DkObjectsWaitEvents()` everywhere.

Also see https://github.com/oscarlab/graphene/pull/960.

dimakuv:
## Affected components

- [x] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [x] Common PAL code

## Description of the changes 

- [Pal/{Linux,Linux-SGX}] Cleanup of db_object.c:
    - Remove _DkObjectWaitOne() and incorporate its special-case of
      waiting on single mutex/event in _DkObjectsWaitAny().
    - Refactor _DkObjectsWaitAny() for readability.
    - Remove unused DEFAULT_QUANTUM and TRACE_HEAP_LEAK.

- [LibOS] Comprehensive cleanup of poll()/ppoll()/select()/pselect()

- [LibOS] Comprehensive cleanup of epoll_create()/epoll_ctl()/epoll_wait()

- [Pal/Linux-SGX] db_pipes.c: fix segfault due to redundant deletion of pipe

- [LibOS, Pal/{Linux,Linux-SGX}] Better emulation of polling; improves the emulation of polling mechanisms (select, pselect, poll, ppoll, epoll_wait) and cleans up the corresponding code:
    - New DkObjectsWaitEvents() PAL interface, replaces the inefficient DkObjectsWaitAny() interface. This interface closely resembles Linux/POSIX poll() in semantics.
    - Improved shim_do_epoll_wait() implementation, now using the new DkObjectsWaitEvents() interface.
    - Improved shim_do_poll() implementation, now using the new DkObjectsWaitEvents() interface.

This PR is based on #960 but additionally cleans up all the old (and badly written) code.

See issue #1117 for more details and performance numbers.

Closes #960. Closes #1117. Closes #229. Closes #228.  Closes #434.

## How to test this PR? 

All tests must pass. Plus I tested with Redis (in two versions: select and epoll).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1128)

### DIFF
--- a/Documentation/oldwiki/PAL-Host-ABI.md
+++ b/Documentation/oldwiki/PAL-Host-ABI.md
@@ -523,11 +523,25 @@ This API clears a notification event or a synchronization event.
 #### DkObjectsWaitAny
 
     #define NO_TIMEOUT ((PAL_NUM)-1)
-    PAL_HANDLE DkObjectsWaitAny(PAL_NUM count, PAL_HANDLE* handleArray, PAL_NUM timeout_us);
+    PAL_HANDLE DkObjectsWaitAny(PAL_NUM count, PAL_HANDLE* handle_array, PAL_NUM timeout_us);
 
-This API polls an array of handles and returns one handle with recent activity. `timeout` is the
-maximum time that the API should wait (in microseconds), or `NO_TIMEOUT` to indicate it is to be
-blocked until at least one handle is ready.
+This API polls an array of handles and returns one handle with recent activity. `timeout_us` is
+the maximum time that the API should wait (in microseconds), or `NO_TIMEOUT` to indicate it is to
+be blocked until at least one handle is ready.
+
+#### DkObjectsWaitEvents
+
+    #define NO_TIMEOUT ((PAL_NUM)-1)
+    PAL_BOL DkObjectsWaitEvents(PAL_NUM count, PAL_HANDLE* handle_array, PAL_FLG* events,
+                                PAL_FLG* ret_events, PAL_NUM timeout_us);
+
+This API polls an array of handles with user-defined events `events` and returns polled-handles'
+events in `ret_events`. `timeout_us` is the maximum time that the API should wait (in
+microseconds), or `NO_TIMEOUT` to indicate it is to be blocked until at least one handle is ready.
+It returns true if there was an event on at least one handle and false otherwise.
+
+This API is a more efficient version of `DkObjectsWaitAny()` and closely resembles Linux poll
+semantics. Therefore, `DkObjectsWaitAny()` should be considered deprecated.
 
 #### DkObjectClose
 
--- a/LibOS/shim/include/shim_handle.h
+++ b/LibOS/shim/include/shim_handle.h
@@ -304,16 +304,13 @@ DEFINE_LIST(shim_epoll_item);
 DEFINE_LISTP(shim_epoll_item);
 struct shim_epoll_handle {
     int maxfds;
-    int nread;
-    int nwaiters;
+    int waiter_cnt;
 
-    int npals;
+    int pal_cnt;
     PAL_HANDLE* pal_handles;
 
     AEVENTTYPE event;
-    LISTP_TYPE(shim_epoll_item) fds; /* this list contains all the
-                                    * shim_epoll_item objects in correspondence
-                                    * with the registered handles. */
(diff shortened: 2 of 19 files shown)


dimakuv:
Pipelines fail on the following LTP tests:
- [ ] `msgget02`
- [ ] `pipe04`
- [ ] `clock_nanosleep01`
- [ ] `waitpid03`

Also, `bash` failed on SGX.

PR #1162, opened 2019-11-24
[LibOS] Thread exit race fix via emulating CLONE_CHILD_CLEARTID

### LINKED ISSUE #230: Bug in returning from signal handler in Linux-SGX
I encountered an elusive bug while testing Redis 5.0. It has to do with exception (signal) handling in the Linux-SGX PAL.

Redis installs a SIGINT handler to gracefully shutdown. (The details are not important though.) In my situation, the main Redis thread ran in an infinite loop inside the enclave, received SIGINT, correctly handled the signal, but after returning to normal stack, its GPRs were corrupted with garbage values -> Redis got SIGBUS and crashed.

Here is the root cause of this bug. Graphene with Linux-SGX has two logical paths: (1) starting from `_DkHandleExternelEvent()` if thread received a signal while *not* in enclave mode, and (2) starting from `_DkExceptionHandler()` if thread was in enclave mode. Both are defined in [`db_exception.c`](https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/db_exception.c).

The bug lurks in the `pal_frame* frame` processing:
```
/* _DkHandleExternelEvent() -- no bug */
    if (!frame) {
        frame = __alloca(sizeof(struct pal_frame));
        frame->identifier = PAL_FRAME_IDENTIFIER;
        frame->func = &_DkHandleExternelEvent;
        frame->funcname = "_DkHandleExternelEvent";
        arch_store_frame(&frame->arch);
    }

/* _DkExceptionRealHandler -- bug! compare with previous function */
    if (frame) {
        frame = __alloca(sizeof(struct pal_frame));
        frame->identifier = PAL_FRAME_IDENTIFIER;
        frame->func     = &_DkExceptionRealHandler;
        frame->funcname = "_DkExceptionRealHandler";
        store_register(rsp, frame->arch.rsp);
        store_register(rbp, frame->arch.rbp);
        unsigned long * last_frame = ((unsigned long *) frame->arch.rbp) + 1;
        last_frame[0]  = (unsigned long) arch_exception_return;
        last_frame[1]  = context->rax;
        last_frame[2]  = context->rbx;
        last_frame[3]  = context->rcx;
        last_frame[4]  = context->rdx;
        last_frame[5]  = context->rsi;
        last_frame[6]  = context->rdi;
        last_frame[7]  = context->r8;
        last_frame[8]  = context->r9;
        last_frame[9]  = context->r10;
        last_frame[10] = context->r11;
        last_frame[11] = context->r12;
        last_frame[12] = context->r13;
        last_frame[13] = context->r14;
        last_frame[14] = context->r15;
        last_frame[15] = context->rip;
    }
```

So, if enclave got the interrupt in enclave mode, it eventually calls `_DkExceptionRealHandler()`. This function actually has two bugs:

1. First line must say `if (!frame)`. Current logic does "if I have a frame, otherwrite it". (This is actually not the bug that I hit.)

2. `last_frame[xxx]` logic overwrites the previous stack frame. Note two things: (a) Graphene-SGX *already* installed a correct stack frame to jump back to normal execution in [`.Lhandle_exception`](https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/enclave_entry.S), and (b) SGX architecture defines a *different* order of storing GPRs in [`struct sgx_arch_gpr_t`](https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/sgx_arch.h).

So what happened in my case is that `last_frame[xxx]` incorrectly overwrote my frames, Redis returned from exception handling, and my normal execution tried to `mov (%rdx)` but RDX contained some garbage.

In general, I found that the exception handling logic Linux-SGX requires no `frame` at all (it never uses `frame` but instead uses `context`).

My quick fix was to:
```
/* _DkExceptionRealHandler -- do not instantiate frame at all */
#if 0
    struct pal_frame * frame = get_frame(uc);
#else
    struct pal_frame * frame = NULL;  /* no bug now */
#endif
```

I guess the correct fix would be to copy logic from well-behaved `_DkHandleExternelEvent()`.

### LINKED ISSUE #1070: [LibOS] Thread exit race
When a thread exits it  wakes up its creator if it CLONE_CHILD_CLEARTID was set during `clone` call (using `release_clear_child_id`), but at this point it is still alive (not for long but still) and using some resources, most notably stack that might have been provided by `clone` caller. Caller might decide to free or reuse that stack (as pthreads do) causing some nasty to debug issues.

(This is one of the problems of #969)

dimakuv:
**UPDATE:** I am deprecating this PR. Instead, I will submit three separate PRs for easier reviewing (and the CLONE_CHILD_CLEARTID mechanism will be significantly different).

## Affected components

- [x] Linux PAL
- [x] SGX PAL
- [x] FreeBSD PAL
- [x] Common PAL code

## Description of the changes 

This is a large fix to correctly emulate `CLONE_CHILD_CLEARTID` behavior and signal handling during thread exits.

Fixes #230. Fixes #1070. Closes #322.

Summary of commits:

1. [Pal/Linux-SGX] Correctly propagate arguments to _DkHandleExternalEvent()

    Previously, Case A in the in-enclave exception handler in enclave_entry.S
    incorrectly set RDI register to -EINTR (to indicate interrupted syscall)
    of the current CPU context, instead of setting RDI register of the interrupted
    CPU context (which is accessed via SGX_GPR_RDI). This led to very rare data
    races that hanged the interrupted thread/corrupted its memory, because
    -EINTR incorrectly propated to SGX_GPR_RSI that holds external event (signal
    number) and this led to buffer overflows.

2. [Pal/Linux-SGX] Remove calculation of frame from _DkExceptionHandler()

    Graphene-SGX has two logical paths: (1) starting from _DkHandleExternalEvent()
    if thread received a signal while not in enclave mode, and (2) starting from
    _DkExceptionHandler() if thread was in enclave mode (both in db_exception.c).

    _DkExceptionHandler() contained two racy bugs pertaining to the frame variable.
    Bugs are described in detail in issue #230. The frame tries to find the outer-
    most enclosing Dk* PAL function, but since _DkExceptionHandler() happens
    during enclave-mode execution, there may be no such frame at all. That is why
    _DkExceptionHandler() always calculates and passes further the enclave
    context. Thus, the frame is not needed at all and can be removed, thus
    eliminating the corresponding bugs. As a result, an intermediate function
    _DkExceptionRealHandler() was also removed.

3. [LibOS,Pal] Correctly emulate CLONE_CHILD_CLEARTID

    When child thread exits, it wakes up its parent if CLONE_CHILD_CLEARTID was set
    during clone() call. Previously, this was done by the child thread itself as
    part of its own clean-up in release_clear_child_id(). But this child thread is
    still alive at this point and uses some resources, most notably the stack (that
    might have been provided by the parent) and the SGX TCS slot. Upon waking up,
    the parent might decide to free that stack (as Pthreads do) or re-use the TCS
    slot, causing data races.

    This commit introduces a correct emulation of CLONE_CHILD_CLEARTID:
    - A new argument `PAL_PTR clear_child_tid` is added to DkThreadCreate();
      it points to memory that is erased on child exit to notify parent.
    - At PAL layer, when thread finally exits, it sets *clear_child_tid = -1;
      this signals to LibOS layer that the thread stopped using resources.
    - At LibOS layer, Async Helper thread is set up to wait for the signal from
      PAL; it is now the responsibility of Async Helper thread to call
      release_clear_child_id() to wake up the parent thread.
    - Async Helper thread waits for clear_child_tid == -1 and then sets it to 0
      and wakes up all waiting parents via del_futex_waiter_wakeup().

    Note that for Linux-SGX PAL, clear_child_tid is set to -1 not immediately
    but as part of handle_thread_reset, otherwise the TCS slot could be still
    occupied when LibOS wakes up parents.

    This commit also fixes all regression tests to use the new signature of
    DkThreadCreate() and increases the number of SGX threads slightly (to
    accommodate the newly used Async Helper thread).

## How to test this PR? 

All tests must pass.

Since these are data races, the usually manifest in multi-threaded tests with Pthreads and signals. In particular, LibOS tests `abort_multithread` and `spinlock` are good candidates for stress testing. I ran tests in an endless loop something like this: `bash -c "exit 134"; while [ $? -eq 134 ]; do SGX=1 ~/graphene/Runtime/pal_loader abort_multithread; done`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1162)

### DIFF
--- a/Documentation/oldwiki/PAL-Host-ABI.md
+++ b/Documentation/oldwiki/PAL-Host-ABI.md
@@ -407,10 +407,11 @@ This API changes the name of an opened stream.
 
 #### DkThreadCreate
 
-    PAL_HANDLE DkThreadCreate(PAL_PTR addr, PAL_PTR param);
+    PAL_HANDLE DkThreadCreate(PAL_PTR addr, PAL_PTR param, PAL_PTR clear_child_tid);
 
 This API creates a thread in the current process. `addr` is the address of an entry point of
 execution for the new thread. `param` is the pointer argument that is passed to the new thread.
+`clear_child_tid` is the pointer to memory that is erased on child exit to notify parent.
 
 #### DkThreadDelayExecution
 
--- a/Documentation/oldwiki/Signal-Handling-in-Graphene.md
+++ b/Documentation/oldwiki/Signal-Handling-in-Graphene.md
@@ -152,8 +152,6 @@ On the example of SIGINT, until we arrive into `_DkGenericSignalHandle()`.
 | | | +                                                                                           |a
 | | | | PAL_CONTEXT ctx = copy(uc)  <ctx contains interrupted-context frame>                      |v
 | | | |                                                                                           |e
-| | | | _DkExceptionRealHandler(PAL_EVENT_SUSPEND, ctx)                                           |
-| | | |                                                                                           |
 + + + + _DkGenericSignalHandle(PAL_EVENT_SUSPEND, ctx)                                            v
           < ... >
 ```
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -456,10 +456,9 @@ static inline void enable_locking (void)
         lock_enabled = true;
 }
 
-static inline PAL_HANDLE thread_create (void * func, void * arg)
-{
+static inline PAL_HANDLE thread_create(void* func, void* arg, void* clear_child_tid) {
     assert(lock_enabled);
-    return DkThreadCreate(func, arg);
+    return DkThreadCreate(func, arg, clear_child_tid);
 }
 
 static inline int64_t __disable_preempt (shim_tcb_t * tcb)
@@ -796,6 +795,8 @@ uint64_t get_rlimit_cur(int resource);
 void set_rlimit_cur(int resource, uint64_t rlim);
 
 int object_wait_with_retry(PAL_HANDLE handle);
+
+void release_clear_child_id(IDTYPE caller, void* clear_child_tid_ptr);
 
 #ifdef __x86_64__
 #define __SWITCH_STACK(stack_top, func, arg)                    \
--- a/LibOS/shim/src/bookkeep/shim_thread.c
+++ b/LibOS/shim/src/bookkeep/shim_thread.c
@@ -759,7 +759,7 @@ BEGIN_RS_FUNC(running_thread)
                                  NUM_SIGS);
 
     if (cur_thread) {
-        PAL_HANDLE handle = DkThreadCreate(resume_wrapper, thread);
+        PAL_HANDLE handle = DkThreadCreate(resume_wrapper, thread, thread->clear_child_tid);
         if (!thread)
             return -PAL_ERRNO;
 
--- a/LibOS/shim/src/ipc/shim_ipc_helper.c
+++ b/LibOS/shim/src/ipc/shim_ipc_helper.c
@@ -780,6 +780,7 @@ noreturn static void shim_ipc_helper(void* dummy) {
     free(object_list);
     free(palhandle_list);
 
+    __disable_preempt(self->shim_tcb);
     put_thread(self);
     debug("IPC helper thread terminated\n");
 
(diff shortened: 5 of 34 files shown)


dimakuv:
This PR was superseded by:

- #1171  (main PR)
- #1169 (data race fix)
- #1170 (another data race fix)

Closing this PR.

PR #1174, opened 2019-11-27
[LibOS] async helper: race on async_list

yamahata:
- don't trigger callback during list traverse to avoid race
- fix memory leak in the case of install_async_event(object != NULL)
  This case is used for AIO signaling. which is one shot.

async_list is protected by async_list_lock.
async_helper unoock async_list_lock before calling callback.
Once the lock is released, the list is unstable. install_async_event
may delete/append the entry. If next of FOR_EACH_LIST_SAFE
points can be freed while trigering callback. then next iteration may
reference freed pointer.
create a temporal list of triggered entry, and then trigger
callback after list traverse.

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1174)

### DIFF
--- a/LibOS/shim/src/shim_async.c
+++ b/LibOS/shim/src/shim_async.c
@@ -193,6 +193,10 @@ static void shim_async_helper(void * arg) {
             break;
         }
 
+        LISTP_TYPE(async_event) triggered;
+        INIT_LISTP(&triggered);
+
+    again:;
         /* Iterate through all async IO events and alarm/timer events to:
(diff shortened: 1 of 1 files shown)


dimakuv:
Jenkins, retest Jenkins-SGX please (`cannot connect to AESMD service`, some unrelated issue with SGX service)

Issue #969, opened 2019-09-03
"Out-of-memory in library OS" when creating and destroying a lot of threads


lejunzhu:
I'm using Graphene SGX latest code on main branch. I ran a C program that simply creates a thread, destroy it, then going on to create another one. After such loop is repeated 6445 times, the process dies with "Out-of-memory in library OS"

The C program I use is:

```C
#include <pthread.h>
#include <stdio.h>
int i = 0;
void *threadproc(void *p)
{
        fprintf(stdout, "%d\n", i);
        fflush(stdout);
        i++;
        return NULL;
}
int main()
{
        pthread_t t;
        int j;
        for (j = 0; j < 10000; ++j) {
                while (i != j) {}
                pthread_create(&t, NULL, threadproc, NULL);
                pthread_join(t, NULL);
        }
        printf("Done\n");
        return 0;
}
```

The inline debug log is attached.
[log.txt](https://github.com/oscarlab/graphene/files/3568174/log.txt)

mkow:
Thanks for including the repro code! Unfortunately I'm not surprised by the outcome, our codebase is plagued by memory leaks.
Could you check whether it also fails without SGX? (i.e. on Linux PAL) Does it also fail if you remove `fprintf` and `fflush` from the thread function?

lejunzhu:
Same thing happens when I remove fprintf and fflush.
Non SGX version can complete 10,000 loops, but the memory is much larger also, so I'm not sure if I hit a memory leak or not.

yamahata:
#1030 in Pal/Linux case:
Now PAL/Linux stack and tcb is allocated/freed by malloc/free.
On the other hand, free(in PAL stack case) in pal is actually nop!
See __free() in Pal/src/slab.c.

In Pal/Linux-SGX case, I'm not sure.
Can you please enable SLAB_DEBUG (see slabmgr.h) and track down the source of memory use?
(or other thing?)

dimakuv:
With https://github.com/oscarlab/graphene/pull/1199, I was able to surpass >30,000 threads under Graphene-SGX and then I got bored and killed the process. This PR removes the `shim_thread` leak.

But I still observe leaks of some other objects. Maybe it's PAL_HANDLEs? I didn't look into PAL logic.

boryspoplawski:
[...]

PR #1199, opened 2019-12-05
[LibOS] Remove the data race on thread::is_alive

dimakuv:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Previously, there was a data race on `thread::is_alive` between one thread checking whether it is the last thread alive via `check_last_thread()` and another thread exiting via `thread_exit()`. The former checks if `is_alive` is true, the latter sets it to false. However, the exiting thread will truly exit only after it called `DkThreadExit()`, thus the race on `is_alive` led to scenarios where two threads believe to be the last threads alive and compete on terminating Async Helper/IPC threads and exiting the whole process.

This PR introduces `cleanup_thread()` called by Async Helper to set `is_alive` to false and delete the thread, freeing its resources. The data race is thus removed, and shim_thread object leak is prevented.

Additionally, I found a bug in thread-exiting in Linux-SGX PAL:

In Linux-SGX PAL, the main thread's stack is provided by host Linux, unlike child threads which stacks are mmapped by the PAL. Therefore, the main thread's stack must not be munmapped, unlike child threads' stacks. Previously, PAL tried to munmap stack of even the main thread, leading to spurious segfaults in e.g. `abort_multithread` test.

Finally, I re-enabled the LibOS regression test `exit_group` which was disabled before (due to the bugs during thread exiting on SGX which were fixed in this and previous PRs).

This PR removes a `shim_thread` data leak, as part of this issue: https://github.com/oscarlab/graphene/issues/969 (but I observe some residual leakage still, so this PR doesn't fix all problems in that issue!).

Ironically, this PR reverts some changes done in https://github.com/oscarlab/graphene/pull/1171. But this is good, I significantly simplified the logic of "ask Async Helper thread to cleanup after exiting thread".

## How to test this PR? 

All tests must pass. A couple good ways to check stability:

- `bash -c "exit 134"; while [ $? -eq 134 ]; do SGX=1 ~/graphene/Runtime/pal_loader abort_multithread; done`
- `bash -c "exit 0"; while [ $? -le 4 ]; do SGX=1 ~/graphene/Runtime/pal_loader exit_group; done`
- `bash -c "exit 0"; while [ $? -eq 0 ]; do SGX=1 ~/graphene/Runtime/pal_loader multi_pthread 2>/dev/null; done`

My tests show both non-SGX and SGX versions never fail on our regression tests.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1199)

### DIFF
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -41,6 +41,8 @@
 #include <atomic.h>
 #include <shim_tcb.h>
 
+noreturn void shim_clean_and_exit(int exit_code);
+
 /* important macros and static inline functions */
 static inline unsigned int get_cur_tid(void)
 {
@@ -144,8 +146,6 @@ static inline PAL_HANDLE __open_shim_stdio (void)
     return shim_stdio;
 }
 
-noreturn void shim_terminate (int err);
-
 #define USE_PAUSE       0
 
 static inline void do_pause (void);
@@ -160,7 +160,7 @@ static inline void do_pause (void);
     do {                                                                    \
         __SYS_PRINTF("BUG() " __FILE__ ":%d\n", __LINE__);                  \
         PAUSE();                                                            \
-        shim_terminate(-ENOTRECOVERABLE);                                   \
+        shim_clean_and_exit(-ENOTRECOVERABLE);                              \
     } while (0)
 
 #define DEBUG_HERE() \
@@ -769,9 +769,6 @@ static inline bool memory_migrated(void * mem)
 extern void * __load_address, * __load_address_end;
 extern void * __code_address, * __code_address_end;
 
-/* cleanup and terminate process, preserve exit code if err == 0 */
-noreturn void shim_clean_and_exit(int err);
-
 unsigned long parse_int (const char * str);
 
 extern void * initial_stack;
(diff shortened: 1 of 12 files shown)


dimakuv:
Jenkins, retest Jenkins-Debug-18.04 please (`socketpair01` LTP test failed)

dimakuv:
Jenkins, retest Jenkins-Debug-18.04 please (`creat03` LTP test failed)

dimakuv:
Something funny started to happen with LTP on Ubuntu 18.04 in Debug configuration. Need to debug.

dimakuv:
Jenkins, retest Jenkins-Debug-18.04 please. I cannot reproduce these errors on my local Ubuntu 18.04.

dimakuv:
Ok, it looks like the problem was in locking `shim_thread` in `cleanup_thread()`. Since there is no-one accessing this thread at this point except `check_last_thread()`, it is enough to lock on `thread_list_lock()`. Let's see if Jenkins passes now.

dimakuv:
Jenkins, retest this (run 2/3, previous run was successful).

I just want to check that there are no spurious failures introduced in this PR.

dimakuv:
Jenkins, retest this please (run 2/3, previous run was successful).

I just want to check that there are no spurious failures introduced in this PR.

dimakuv:
Jenkins, retest this please (run 3/3, previous run was successful).

mkow:
Jenkins, retest Jenkins-SGX-18.04-apps please

mkow:
Jenkins, retest Jenkins-18.04 please (`apps.LTP.write05` timed out)

PR #1209, opened 2019-12-11
[LibOS/regression] Add test for host root FS

### LINKED ISSUE #1179: Graphene-SGX manifests should support chroot-mounts on root directory
The feature should support the following configuration in GSGX manifest.
> fs.mount.root.type = chroot
> fs.mount.root.path = /
> fs.mount.root.uri = file:///
This way is able to allow GSGX to mount the specified root directory by chroot as default fs used by application running inside enclave. Thanks.

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

This PR does two things:

1. [LibOS] Allow inaccessible files during `getdents()`. Previously, if user performed `getdents()` on a directory containing inaccessible files (because user doesn't have permission), whole getdents failed with `-EACCES`. This is incorrect behavior: files must still be listed. This commit fixes the root cause of this bug by marking inaccessible files as `DENTRY_NEGATIVE`.

2. [LibOS/regression] Add test for host root FS. Graphene support a special "root FS" parameter in the manifest: `fs.root`. This commit adds a test that uses this parameter.

This PR is in response to https://github.com/oscarlab/graphene/issues/1179. It looks like Graphene supports host-root out of the box (bar a tiny fix in this PR). To close #1179, we'll need more tests on real applications though.

I am also not sure about the correctness of the proposed fix (see first commit). This may be too naive.

## How to test this PR? 

All tests must pass.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1209)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -104,7 +104,7 @@ struct shim_fs_ops {
 };
 
 #define DENTRY_VALID        0x0001  /* this dentry is verified to be valid */
-#define DENTRY_NEGATIVE     0x0002  /* negative, recently deleted */
+#define DENTRY_NEGATIVE     0x0002  /* recently deleted or inaccessible */
 #define DENTRY_RECENTLY     0x0004  /* recently used */
 #define DENTRY_PERSIST      0x0008  /* added as a persistent dentry */
 #define DENTRY_HASHED       0x0010  /* added in the dcache */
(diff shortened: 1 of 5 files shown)


dimakuv:
Jenkins, test this please (need fresh logs to debug issues)

PR #1219, opened 2019-12-19
Correctly release resources during str_close().

thomasknauth:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Fixes a resources leak where repeatedly opening and closing files in, for example, proc, led to memory exhaustion.

## How to test this PR? 

This PR comes with a regression test. Regression should still pass.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1219)

### DIFF
--- a/LibOS/shim/src/fs/str/fs.c
+++ b/LibOS/shim/src/fs/str/fs.c
@@ -76,6 +76,13 @@ int str_close(struct shim_handle* hdl) {
     }
 
     str_dput(hdl->dentry);
+
+    if (hdl->info.str.data) {
+        free(hdl->info.str.data->str);
+        free(hdl->info.str.data);
+        hdl->info.str.data = NULL;
+    }
+
     return 0;
 }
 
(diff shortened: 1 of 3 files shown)


dimakuv:
I looked into `exit_group` failing on this PR.

Turns out it has nothing to do with this PR. This is an unfortunate timing of this PR: it builds on a master branch that already incorporated commit https://github.com/oscarlab/graphene/commit/860e1083a2d60a0e6e72c1389a0a6278d5b7d186 (`[LibOS] test/regression: Enable exit_group test on SGX`). But this PR still doesn't have a commit https://github.com/oscarlab/graphene/commit/6cfaf420b3640134b67c613cff5b035b97443231 (`[LibOS] Rework futexes implementation`).

So what happens in `exit_group` is something like this: 5 threads are created and they compete on `futex(wait/wake)`. Whoever dies first, sends `kill` signal to the others that sleep on their futexes.

The old implementation of futexes had a couple data races, in particular there was a race on `put_thread(waiter->thread)`. So `exit_group` probabilistically failed (about 1/20 runs on my local machine) on trying to access `thread->ref_count` on a freed thread. This was fixed in the new implementation of futexes by Borys. When I rebased this to the newest master (with Borys's futexes re-implementation), this test doesn't fail anymore. So we just need to rebase it.

PR #1232, opened 2020-01-02
[Pal/Linux-SGX] Add sgx.ra_accept_configuration_needed manifest option

### LINKED ISSUE #1231: [Linux-SGX] remote attestation: need a way to bypass CONFIGURATION_NEEDED
There is already an option to bypass the GROUP_OUT_OF_DATE status returned from IAS, but not the CONFIGURATION_NEEDED status. According to the document:

> CONFIGURATION_NEEDED - The EPID signature of the ISV enclave QUOTE has been verified correctly, but additional configuration of SGX platform may be needed (for further details see Advisory IDs). The platform has not been identified as compromised and thus it is not revoked. It is up to the Service Provider to decide whether or not to trust the content of the QUOTE, and whether or not to trust the platform performing the attestation to protect specific sensitive information.

When such status is returned, PAL always ends the process. I suggest that when ra_accept_group_out_of_date is set, CONFIGURATION_NEEDED is bypassed as well.

dimakuv:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

This PR adds `sgx.ra_accept_configuration_needed` manifest option to handle the case when SGX remote attestation from IAS returns status `CONFIGURATION_NEEDED` (which means that this SGX platform is not known to be compromised but has e.g. hyper-threading enabled).

See #1231 and https://software.intel.com/en-us/forums/intel-software-guard-extensions-intel-sgx/topic/798777.

Fixes #1231.

## How to test this PR? 

All tests must pass.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1232)

### DIFF
--- a/Pal/regression/test_pal.py
+++ b/Pal/regression/test_pal.py
@@ -585,7 +585,7 @@ class TC_50_Attestation(RegressionTestCase):
             # Check the attestation status
             if line.startswith("Attestation status:"):
                 status = line[19:].strip()
-                self.assertIn(status, ["OK", "GROUP_OUT_OF_DATE"])
+                self.assertIn(status, ["OK", "GROUP_OUT_OF_DATE", "CONFIGURATION_NEEDED"])
 
             # Check the timestamp
             if line.startswith("Attestation timestamp:"):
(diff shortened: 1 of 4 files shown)


mkow:
Jenkins, retest Jenkins-SGX please (`test_libos.TC_01_Bootstrap.test_202_fork_and_exec` and `test_libos.TC_01_Bootstrap.test_200_exec` failed, weird)

PR #1233, opened 2020-01-02
[LibOS] Disallow eventfd emulation by default

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 

Eventfd emulation currently relies on the host, thus eventfd syscalls are disallowed by default due to security concerns. To use them, they must be explicitly allowed through `sys.allow_insecure_eventfd` in manifest.

## How to test this PR? 

LibOS regression test `eventfd` now has its own manifest file with `sys.allow_insecure_eventfd = 1`. One can remove this line and see how this test fails with `eventfd failed: Function not implemented`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1233)


mkow:
For bookkeeping: This is a fixup for #1085 which shouldn't be enabled by default (because it's not secure).

PR #1218, opened 2019-12-19
[LibOS] dedicated syscall stack and signal emulation through app stack

yamahata:
For syscall emulation, use dedicated stack for it. and emulates signal delivery more correctly.

If application stack is small(e.g. go goroutines), stack overflow happens
and crash. Especially when pushing all the GP registers at the beginning
of syscalldb.
In order to avoid it, allocate dedicated stack area for syscall emulation
and switch them back and forth.

When LibOS emulates unix signal,
- If app is interrupted, setup signal frame directly and return.
- If Pal or LibOS is interrupted, it means app invoked system call and
   interrupt. Queue it and check if the signal is queue when LibOS returns to
   app and setup signal frame.
- implemented sigreturn and sigaltstack
 So LibOS isn't re-entered by signal emulation because app signal handler
 isn't triggered while Pal or LibOS is interrupted.

supersedes https://github.com/oscarlab/graphene/pull/347
closes https://github.com/oscarlab/graphene/issues/348

## Affected components

- [ ] README and global configuration
- [ ] Linux PAL
- [ ] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code
- [x] Library OS (i.e., SHIM), including GLIBC

## How to test this PR? 
normal CI should pass

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1218)

PR #1238, opened 2020-01-08
[Pal/Linux-SGX] use dedicated stack for signal handler

yamahata:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] FreeBSD PAL
- [ ] Common PAL code

## Description of the changes 
use dedicated stack for signal handler:

In order to handle host signal reliably, allocate stack dedicated for
signal instead of stack application provides.
This patch makes host exception handler/ocall exception handler
properly save/restore fp registers. including initializing them
right before EEXIT.
Because size of fp register is large, ocall may cause stack overflow.
So for ocall case, use dedicated signal too and returning path are
unified among host exception and ocall exception case.
Now signal stack is allocated, eframe rewinding is broken and useless.
So remove it.

This is a preparation for https://github.com/oscarlab/graphene/pull/1218
This PR superces #632 
Closes https://github.com/oscarlab/graphene/pull/632
Supersedes https://github.com/oscarlab/graphene/pull/397

The final goal is

- to make LibOS's signal delivery more correct to avoid re-enter LibOS.
- to support sigaltstack systemcall
- to correctly handle FP registers
- to make interrupt handler more reliable even with small application stack (for golang)

If LibOS interrupt handler is executed with the application stack, it becomes difficult for LibOS to setup a signal frame for application. Actually #1218 hits the issue and I started this PR first. After this PR is done, I'll resume #1218 (and fixing signal delivery of LibOS)
sigaltstack is typically used by language runtime to do something reliably, e.g. report its diagnose in danger situation(e.g. stack overflow). So it's important as we'd like to support python, java, golang etc at high quality.

## How to test this PR? 
All CI pass.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1238)


yamahata:
Jenkins, retest Jenkins-SGX-18.04 please. (apps.LTP.socketpair01 timeout)

yamahata:
Jenkins, retest Jenkins-18.04 please.  (apps.LTP.socketpair01 timeout)

mkow:
Jenkins, retest Jenkins-Debug-18.04 please (`test_libos.TC_01_Bootstrap.test_401_exit_group` timed out)

PR #1256, opened 2020-01-15
Remove forcing CPUID subleaf to zero.

### LINKED ISSUE #1252: xsavec
While experimenting with PyTorch on Graphene-SGX, I ran into an issue with the xsavec instruction. glibc's rtld has several variants of a function to resolve dynamic functions. One of them is _dl_runtime_resolve_xsavec(). As the name suggests, it uses xsavec to stow away the processor's execution context. For some reason, the space calculated by glibc to hold the xsavec state seems to be insufficient, leading to a stack corruption.

A work around is to disable glibc's use of the xsavec variant by setting the environment variable GLIBC_TUNABLES to `glibc.tune.hwcaps=-XSAVEC_Usable`.

thomasknauth:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] Common PAL code

## Description of the changes 

Among other things, CPUID is used by glibc to calculate the area to set aside for xsave[c]. Silently forcing the subleaf to zero leads to a stack corruptions within glibc.

The original reason to force the subleaf to zero was forgotten (likely to reduce the attack surface through untrusted inputs).

Fixes #1252.

## How to test this PR? 

Regressions still pass. My PyTorch example suddenly works (to be submitted in a future PR).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1256)


dimakuv:
Jenkins, retest this please (internal Jenkins DNS issues)

woju:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 256 at r6](https://reviewable.io/reviews/oscarlab/graphene/1256#-LzxNAaZ1xm14dCOHgsU:-Lzxq5EfC9mOLSdg5OXg:b-6et3uv) ([raw file](https://github.com/oscarlab/graphene/blob/09891f99fe07152a143f0dbe90179efc84ea2a4b/Pal/src/host/Linux-SGX/db_misc.c#L256)):*

`#save_size_bytes` -> `save_size_bytes` (or the same but wrapped in backticks, as you prefer)
</blockquote></details>

If this was Doxygen comment, this would actually link to the variable, so this is kinda OK.

PR #1257, opened 2020-01-15
[LibOS] Allow all getsockopt() before bind()

### LINKED ISSUE #644: Issue with calling getsockopt before bind
I am debugging in an issue with gRPC which calls `getsockopt` before `bind` (see https://github.com/grpc/grpc/blob/master/src/core/lib/iomgr/tcp_server_utils_posix_common.cc#L165). When I run this program in Graphene, `getsockopt` fails because a `pal_handle` isn't assigned until `bind` is called. It seems like a `pal_handle` should be assigned earlier when `socket` is called. Below is a minimal test case.

```
#include <sys/socket.h>
#include <netinet/tcp.h>
#include <netinet/ip.h>
#include <unistd.h>
#include <assert.h>

int main(int argc, char **argv)
{
    int fd, ret, flags = 1;
    socklen_t len = sizeof(flags);

    fd = socket(AF_INET, SOCK_STREAM, IPPROTO_IP);
    ret = getsockopt(fd, SOL_TCP, TCP_NODELAY, (void *)&flags, &len);
    close(fd);

    assert(ret == 0);

    return ret;
}
```

### LINKED ISSUE #754: [LibOS] setsockopt does not handle optlen properly
Our implementation of setsockopt does not check optlen.

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] Common PAL code

## Description of the changes 

Previously in Graphene, some `getsockopt()` syscalls, e.g. TCP_NODELAY, failed because an underlying PAL handle was't created for the LibOS handle until `bind()` was called. Thus, a sequence of `accept()` and `getsockopt()` failed. This commit fixes this by returning default socket options (possibly augmented with setsockopt values). Test case is also provided.

Fixes #644. Fixes #754.

## How to test this PR? 

LibOS test `getsockopt` is augmented with an additional case of TCP_NODELAY that fails without this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1257)

### DIFF
--- a/LibOS/shim/src/sys/shim_socket.c
+++ b/LibOS/shim/src/sys/shim_socket.c
@@ -1550,8 +1550,7 @@ static void __populate_addr_with_defaults(PAL_STREAM_ATTR* attr) {
     attr->socket.tcp_nodelay    = PAL_FALSE;
 }
 
-static bool __update_attr(PAL_STREAM_ATTR* attr, int level, int optname, char* optval, int optlen) {
-    __UNUSED(optlen);
+static bool __update_attr(PAL_STREAM_ATTR* attr, int level, int optname, char* optval) {
     assert(attr);
 
     bool need_set_attr = false;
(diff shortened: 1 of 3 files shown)


dimakuv:
Jenkins, retest Jenkins-SGX please (`vfork_and_exec` LibOS test failed)

dimakuv:
Jenkins, retest Jenkins-Debug please (`exit_group` LibOS test timeout)

PR #1264, opened 2020-01-17
[LibOS] Fix type of checkpoint's total memory size variable

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] Common PAL code

## Description of the changes 

Checkpoint's total memory size is stored in `shim_cp_store::mem_size` field. Previously, this field was of type `int`. When a process allocates more than 2GB of memory and then tries to spawn a child,
the checkpoint send/receive fails due to int overflow of `mem_size`. This PR simply changes mem_size type to `size_t`. This is enough to make the bug go away on e.g. a huge Python app with TensorFlow.

## How to test this PR? 

This bug was triggered on a huge Python app with TensorFlow. Generally, this bug is triggered on huge apps. I don't think it deserve a separate test.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1264)

### DIFF
--- a/LibOS/shim/include/shim_checkpoint.h
+++ b/LibOS/shim/include/shim_checkpoint.h
@@ -112,7 +112,7 @@ struct shim_cp_store {
     /* entries of out-of-band data */
     struct shim_mem_entry* last_mem_entry;
     int mem_nentries;
-    int mem_size;
+    size_t mem_size;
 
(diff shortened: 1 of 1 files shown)


mkow:
Related issue: https://github.com/oscarlab/graphene/issues/152

PR #1266, opened 2020-01-20
[Pal/Linux-SGX] Unify SGX_OCALL_PREPARED and SGX_STACK

HW42:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] Common PAL code

## Description of the changes 

When reviewing PR #1238 I notice that with in the current version of `enclave_entry.S` we can get rid of the code emulation in `.Lsetup_exception_handler` when we avoid using two variables to keep track of the "we prepared for an ocall" state.

This PR is on top of #1238 to make reviewing and merging easier (So remember to look at the right diff).

## How to test this PR? 

Should not break any regression tests.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1266)

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_entry.S
+++ b/Pal/src/host/Linux-SGX/enclave_entry.S
@@ -55,7 +55,7 @@ enclave_entry:
 
 	# Only jump to .Lreturn_from_ocall if we have prepared the stack for
 	# it.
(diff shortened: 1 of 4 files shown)


HW42:
@yamahata: Please take a careful look if I have missed some corner case.

PR #1284, opened 2020-01-31
Attestation based on pseudo-file interface.

thomasknauth:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] Common PAL code

## Description of the changes 

This PR introduces a new pseudo-file based attestation interface to Graphene.

The following pseudo-files exist under /proc/sgx_attestation/
- report: Reading returns the sgx_report_t data structure.
- ias_header: Reading returns the HTTP headers returned on interaction with Intel Attestation Service.
- ias_report: Reading returns the HTTP body returned by the Intel Attestation Service.
- my_target_info: Reading returns the current enclave’s sgx_target_info_t.
- target_info: Read/write the sgx_target_info_t of peer enclave in local attestation flow.
- report_data: Read/write the sgx_report_data_t to include in local attestation report.
- quote: Read the sgx_quote_t obtained from quoting enclave.

A separate PR will provide a library to provide higher-level abstractions based on this new interface, e.g., to produce RA-TLS certificates and verify them.

## How to test this PR? 

Comes with a regression test case.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1284)


Pal/src/host/Linux-SGX/sgx_platform.c
```diff
@@ -265,109 +270,14 @@ int contact_intel_attest_service(const char* subkey, const sgx_quote_nonce_t* no
...
-
-    // Now return the attestation data, including the IAS response, signature, and the
-    // certificate chain back to the caller.
-    attestation->ias_report     = https_output;
-    attestation->ias_report_len = https_output_len;
-    attestation->ias_sig        = ias_sig;
-    attestation->ias_sig_len    = ias_sig_len;
-    attestation->ias_certs      = ias_certs;
-    attestation->ias_certs_len  = ias_certs_len;
+    https_header = NULL; // Don't free the HTTPS output
```
sudharkrish: Fix the comment here to specify HTTPS header.

LibOS/shim/src/fs/proc/attestation.c
```diff
@@ -0,0 +1,399 @@
...
+    return 0;
+}
+
+static int proc_sgx_report_mode(const char* name, mode_t* mode) {
+    __UNUSED(name);
+    *mode = 0444;
+    return 0;
+}
+
+static int proc_sgx_ias_header_open(struct shim_handle* hdl, const char* name, int flags) {
```
sudharkrish: Dont like the idea of having 2 separate pseudo interfaces, one for ias_header and one for ias_report, given that we get both in one call to DkIASReport.  There could be discrepancy, if caller has a wrong sequence, for example, calls ias_header, updates report_data, and then fetches ias_report.

LibOS/shim/src/fs/proc/attestation.c
```diff
@@ -0,0 +1,399 @@
...
+}
+
+static __sgx_mem_aligned sgx_report_t report;
+static __sgx_mem_aligned sgx_target_info_t target_info;
+static __sgx_mem_aligned sgx_target_info_t my_target_info;
+static __sgx_mem_aligned sgx_report_data_t report_data;
+
+/* IAS interactions are cached and only repeated if this variable is false. */
+static int    ias_valid = 0;
+static char   ias_report[10 * 1024];
```
sudharkrish: We dont need 10K buffer for ias_report.

Pal/src/host/Linux-SGX/db_main.c
```diff
@@ -393,11 +393,6 @@ void pal_linux_main(char * uptr_args, uint64_t args_size,
     __pal_control.manifest_preload.start = (PAL_PTR) manifest_addr;
     __pal_control.manifest_preload.end = (PAL_PTR) manifest_addr + manifest_size;
 
-    if ((rv = init_trusted_platform()) < 0) {
```
sudharkrish: I am fine(and agree) with this change. But still curious, if the removal of init_trusted_platform, which was previously used to initiate RA with IAS, upon startup, has been discussed with @chiache and others in graphene core team.

PR #1285, opened 2020-01-31
[Pal/Linux-SGX] Do not show SGX context debug info in non-debug build

dimakuv:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] Common PAL code

## Description of the changes 

PR https://github.com/oscarlab/graphene/pull/1238 was merged. But it accidentally prints debug information even in non-debug mode. This is annoying and confuses people. This PR simply changes debug level to not show this info in non-debug mode.

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1285)


dimakuv:
@yamahata Actually, maybe we should just remove this debug message completely? What do you think?

PR #1286, opened 2020-01-31
[Pal/Linux-SGX] Update internal-PAL pal_vmas when freeing memory

dimakuv:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] Common PAL code

## Description of the changes 

Linux-SGX PAL uses `pal_vmas[PAL_VMA_MAX]` array to keep track of allocated internal-PAL memory regions. Previously, Graphene only added items to this array (during DkVirtualMemoryAlloc) but never removed them from the array (during DkVirtualMemoryFree). Also, the code used an assert() instead of proper error handling, which led to Heisenbugs in non-debug builds.

This PR fixes this but in a rather inefficient way. Memory allocation in SGX is sub-par and will need a complete rewrite.

## How to test this PR? 

All tests must pass.

Redis under SGX with 100 clients must work now:
```
# fails without this PR!
~/graphene/LibOS/shim/test/apps/redis$ ./src/src/redis-benchmark -h 127.0.0.1 -p 6379 -c 100 -n 1000000 -t get,set
```

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1286)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_memory.c
+++ b/Pal/src/host/Linux-SGX/db_memory.c
@@ -35,8 +35,8 @@
 
 #include "enclave_pages.h"
 
-#define PAL_VMA_MAX     64
-
+/* TODO: Having VMAs in an array is extremely inefficient */
+#define PAL_VMA_MAX 64
 static struct pal_vma {
     void * top, * bottom;
 } pal_vmas[PAL_VMA_MAX];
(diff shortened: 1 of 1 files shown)


dimakuv:
Jenkins, retest Jenkins-SGX-18.04 please (`signalexit_multithread` LibOS test failed)

PR #1290, opened 2020-01-31
[Jenkins] Test 'make clean' also on 18.04

HW42:
## Affected components

- [x] README and global configuration
- [ ] Linux PAL
- [ ] SGX PAL
- [ ] Common PAL code
- [ ] Library OS (i.e., SHIM), including GLIBC

## Description of the changes 

See commit messages.

## How to test this PR? 

See Jenkins.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1290)

PR #1291, opened 2020-01-31
Improved download handling

HW42:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [x] Common PAL code

## Description of the changes 

Instead of implementing downloading of external resources in every
Makefile again, use a one script. This script adds the following
features:

 - Always check the download against a known SHA-256 hash.

 - Support caching of downloaded resource (set DL_CACHE=/some/dir).

 - Allow offline builds if all files are cached. If DL_OFFLINE=true the
   build will never attempt to download something.

Matching PR: https://github.com/oscarlab/graphene-tests/pull/70.

## How to test this PR? 

Building should work as before.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1291)

### DIFF
--- a/.ci/run-shellcheck
+++ b/.ci/run-shellcheck
@@ -15,6 +15,7 @@ shellcheck "$@" \
     Scripts/clean-check \
     Scripts/clean-check-prepare \
     Scripts/clean-check-test-copy \
+    Scripts/download \
     Scripts/list-all-graphene.sh \
     Scripts/memusg \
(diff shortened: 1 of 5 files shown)


HW42:
This has a hidden dependency on "[Scripts/clean-check] Ignore git repos" from PR #1290 (I think it's redis' version check, which calls git).

PR #1293, opened 2020-01-31
[LibOS] Update TYPE_DEV handles with correct function pointers in dev_ops

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] Common PAL code

## Description of the changes 

For device handles, `info.dev.dev_ops` contains function pointers into LibOS. They may become invalid due to relocation of LibOS text section in the child process on fork. This commit forces an update of these function pointers.

In particular, this bug was found on Ubuntu 18.04 while working on Apache with SSL/TLS. For some reason, LibOS text segment (`libsysdb.so`) is relocated in another place in memory in the child process, so Apache tried to access `/dev/urandom` that ended up in a function pointer that pointed to the old (parent's) value in memory, and we had a segfault. See https://github.com/oscarlab/graphene-tests/pull/68.

Closes #1277.

## How to test this PR? 

Apache with SSL/TLS must pass now (included as a separate commit in this PR).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1293)

### DIFF
--- a/Jenkinsfiles/Linux
+++ b/Jenkinsfiles/Linux
@@ -56,10 +56,12 @@ pipeline {
                                 make check
                             '''
                         }
-                        sh '''
-                            cd LibOS/shim/test/apps/bash
-                            make regression
+                        timeout(time: 5, unit: 'MINUTES') {
+                            sh '''
+                                cd LibOS/shim/test/apps/bash
+                                make regression
                             '''
+                        }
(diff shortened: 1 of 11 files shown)


dimakuv:
Jenkins, retest Jenkins-SGX please (`exec` LibOS test failed)

PR #1296, opened 2020-02-03
[Pal/Linux-SGX] Make the pal-sgx-sign code importable

HW42:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] Common PAL code

## Description of the changes 

This moves the Python implementation of pal-sgx-sign into pal_sgx_sign.py
and replaces the cmdline tool with a tiny wrapper. This allows other
code to import it as a module and thereby re-use it's code.

Designing an API and then putting it into a proper Python package is left
for the future.

## How to test this PR? 

Signing manifests for SGX should work as before. Additionally you can now `import pal_sgx_sign` in Python assuming its dir is in `sys.path`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1296)


HW42:
Jenkins, retest Jenkins-Debug-18.04

PR #1268, opened 2020-01-21
Update list.h to support additional list macros

sudharkrish:
## Affected components

- [ ] README and global configuration
- [ ] Linux PAL
- [X] SGX PAL
- [X] Common PAL code
- [ ] Library OS (i.e., SHIM), including GLIBC

## Description of the changes 
These are changes to update list.h, to support additional features like size, and ability to sort list items. Changes are WIP and needs cleanup, submitting this PR, to get initial feedback.
Note: These updates were needed for protected-fs changes in PAL.

## How to test this PR? 
PAL regression testing

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1268)

PR #1325, opened 2020-02-12
[Linux-SGX] Add protected files implementation (SGX SDK file format)

omeg:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [x] Common PAL code

## Description of the changes 
Protected files (PF) are a new type of file that can be specified in
the manifest (SGX only). They are encrypted on disk and transparently
decrypted when accessed by the Graphene payload.

Other features:
- data is integrity protected (tamper resistance)
- file swap protection (a PF can only be accessed when in a specific path)
- transparency (Graphene payload sees PFs as regular files, no need to modify
  the payload)

The following new manifest elements are added:

```
sgx.protected_files_key = <16-byte hex value>
sgx.protected_files.<name> = file:<host path>
```

`sgx.protected_files_key` specifies the encryption key and **is only a temporary
implementation. This key should be provisioned with local/remote attestation
in the future**.

Paths specifying PF entries can be files or directories. If a directory is
specified, all files/directories within are registered as protected
recursively (and are expected to be encrypted in the PF format).

`Linux-SGX/tools` directory contains the `pf_crypt` utility that converts files
to/from the protected format.

Internal protected file format in this version was ported from the SGX SDK:
https://github.com/intel/linux-sgx/tree/master/sdk/protected_fs

TODO:
* Truncating files is not yet implemented.
* The recovery file feature is disabled, this needs to be discussed if it's needed in Graphene.
* Tests for invalid/malformed/corrupted files need to be ported to the new format.

Fixes https://github.com/oscarlab/graphene/issues/1610 (as a side effect).

## How to test this PR? 
Tests in `LibOS/shim/test/fs` now contain PF tests too (target is still `test`). Target to run just non-PF tests is `fs-test`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1325)

### DIFF
--- a/LibOS/shim/test/fs/manifest.template
+++ b/LibOS/shim/test/fs/manifest.template
@@ -32,3 +32,7 @@ sgx.trusted_files.libpthread = file:../../../../Runtime/libpthread.so.0
 sgx.trusted_files.libgcc_s = file:$(ARCH_LIBDIR)/libgcc_s.so.1
 
 sgx.allowed_files.tmp_dir = file:tmp/
+
+sgx.protected_files_key = ffeeddccbbaa99887766554433221100
+sgx.protected_files.input = file:tmp/pf_input
+sgx.protected_files.output = file:tmp/pf_output
--- a/Pal/src/host/Linux-SGX/enclave_pf.c
+++ b/Pal/src/host/Linux-SGX/enclave_pf.c
@@ -1,0 +1,613 @@
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
+/* Copyright (C) 2018-2020 Invisible Things Lab
+ *                         Rafal Wojdyla <omeg@invisiblethingslab.com>
+ */
+
+#include <linux/fs.h>
+
+#include "pal_crypto.h"
+#include "pal_internal.h"
+#include "pal_linux.h"
+#include "pal_linux_error.h"
+#include "spinlock.h"
+
+/*
+ * At startup, protected file paths are read from the manifest and the specified files
+ * or directories registered. For supported I/O operations, handlers (in db_files.c)
+ * check if the file is a PF to perform the required operation transparently.
+ *
+ * Since PF's "logical" size is different than the real FS size (and to avoid potential
+ * infinite recursion in FS handlers) we don't use PAL file APIs here, but raw OCALLs.
+ */
+
+/* List of map buffers */
+LISTP_TYPE(pf_map) g_pf_map_list = LISTP_INIT;
+
+/* Callbacks for protected files handling */
+static pf_status_t cb_read(pf_handle_t handle, void* buffer, uint64_t offset, size_t size) {
+    int fd = *(int*)handle;
+    size_t buffer_offset = 0;
+    size_t to_read = size;
+    while (to_read > 0) {
+        ssize_t read = ocall_pread(fd, buffer + buffer_offset, to_read, offset + buffer_offset);
+        if (read == -EINTR)
+            continue;
+
+        if (read < 0) {
+            SGX_DBG(DBG_E, "cb_read(%d, %p, %lu, %lu): read failed: %ld\n",
+                    fd, buffer, offset, size, read);
+            return PF_STATUS_CALLBACK_FAILED;
+        }
+
+        /* EOF is an error condition, we want to read exactly `size` bytes */
+        if (read == 0) {
+            SGX_DBG(DBG_E, "cb_read(%d, %p, %lu, %lu): EOF\n", fd, buffer, offset, size);
+            return PF_STATUS_CALLBACK_FAILED;
+        }
+
+        to_read -= read;
+        buffer_offset += read;
+    }
+    return PF_STATUS_SUCCESS;
+}
+
+static pf_status_t cb_write(pf_handle_t handle, const void* buffer, uint64_t offset, size_t size) {
+    int fd = *(int*)handle;
+    size_t buffer_offset = 0;
+    size_t to_write = size;
+    while (to_write > 0) {
+        ssize_t written = ocall_pwrite(fd, buffer + buffer_offset, to_write,
+                                       offset + buffer_offset);
+        if (written == -EINTR)
+            continue;
+
+        if (written < 0) {
+            SGX_DBG(DBG_E, "cb_write(%d, %p, %lu, %lu): write failed: %ld\n",
+                    fd, buffer, offset, size, written);
+            return PF_STATUS_CALLBACK_FAILED;
+        }
+
+        /* EOF is an error condition, we want to write exactly `size` bytes */
+        if (written == 0) {
+            SGX_DBG(DBG_E, "cb_write(%d, %p, %lu, %lu): EOF\n", fd, buffer, offset, size);
+            return PF_STATUS_CALLBACK_FAILED;
+        }
+
+        to_write -= written;
+        buffer_offset += written;
+    }
+    return PF_STATUS_SUCCESS;
+}
+
+static pf_status_t cb_truncate(pf_handle_t handle, uint64_t size) {
+    int fd = *(int*)handle;
+    int ret = ocall_ftruncate(fd, size);
+    if (IS_ERR(ret)) {
+        SGX_DBG(DBG_E, "cb_truncate(%d, %lu): ocall failed: %d\n", fd, size, ret);
+        return PF_STATUS_CALLBACK_FAILED;
+    }
+    return PF_STATUS_SUCCESS;
+}
+
+#ifdef DEBUG
+static void cb_debug(const char* msg) {
+    SGX_DBG(DBG_D, "%s", msg);
+}
+#endif
+
+static pf_status_t cb_aes_gcm_encrypt(const pf_key_t* key, const pf_iv_t* iv,
+                                      const void* aad, size_t aad_size,
+                                      const void* input, size_t input_size, void* output,
+                                      pf_mac_t* mac) {
+    int ret = lib_AESGCMEncrypt((const uint8_t*)key, sizeof(*key), (const uint8_t*)iv, input,
+                                input_size, aad, aad_size, output, (uint8_t*)mac, sizeof(*mac));
+    if (ret != 0) {
+        SGX_DBG(DBG_E, "lib_AESGCMEncrypt failed: %d\n", ret);
+        return PF_STATUS_CALLBACK_FAILED;
+    }
+    return PF_STATUS_SUCCESS;
+}
+
+static pf_status_t cb_aes_gcm_decrypt(const pf_key_t* key, const pf_iv_t* iv,
+                                      const void* aad, size_t aad_size,
+                                      const void* input, size_t input_size, void* output,
+                                      const pf_mac_t* mac) {
+    int ret = lib_AESGCMDecrypt((const uint8_t*)key, sizeof(*key), (const uint8_t*)iv, input,
+                                input_size, aad, aad_size, output, (const uint8_t*)mac,
+                                sizeof(*mac));
+    if (ret != 0) {
+        SGX_DBG(DBG_E, "lib_AESGCMDecrypt failed: %d\n", ret);
+        return PF_STATUS_CALLBACK_FAILED;
+    }
+    return PF_STATUS_SUCCESS;
+}
+
+static pf_status_t cb_random(uint8_t* buffer, size_t size) {
+    int ret = _DkRandomBitsRead(buffer, size);
+    if (IS_ERR(ret)) {
+        SGX_DBG(DBG_E, "_DkRandomBitsRead failed: %d\n", ret);
+        return PF_STATUS_CALLBACK_FAILED;
+    }
+    return PF_STATUS_SUCCESS;
+}
+
+/* Wrap key for protected files.
+   TODO: In the future, this key should be provisioned after local/remote attestation. */
+static pf_key_t g_pf_wrap_key = {0};
+
+/* Collection of registered protected files */
+static struct protected_file* g_protected_files = NULL;
+
+/* Collection of registered protected directories */
+static struct protected_file* g_protected_dirs = NULL;
+
+/* Lock for operations on global PF structures */
+static spinlock_t g_protected_file_lock = INIT_SPINLOCK_UNLOCKED;
+
+/* Take ownership of the global PF lock */
+void pf_lock(void) {
+    spinlock_lock(&g_protected_file_lock);
+}
+
+/* Release ownership of the global PF lock */
+void pf_unlock(void) {
+    spinlock_unlock(&g_protected_file_lock);
+}
+
+/* Exact match of path in g_protected_files */
+struct protected_file* find_protected_file(const char* path) {
+    struct protected_file* pf = NULL;
+
+    pf_lock();
+    HASH_FIND_STR(g_protected_files, path, pf);
+    pf_unlock();
+    return pf;
+}
+
+/* Find registered pf directory starting with the given path */
+static struct protected_file* find_protected_dir(const char* path) {
+    struct protected_file* pf  = NULL;
+    struct protected_file* tmp = NULL;
+    size_t len = strlen(path);
+
+    pf_lock();
+    // TODO: avoid linear lookup
+    for (tmp = g_protected_dirs; tmp != NULL; tmp = tmp->hh.next) {
+        if (tmp->path_len < len &&
+                !memcmp(tmp->path, path, tmp->path_len) &&
+                (!path[tmp->path_len] || path[tmp->path_len] == '/')) {
+            pf = tmp;
+            break;
+        }
+    }
+
+    pf_unlock();
+    return pf;
+}
+
+/* Find PF by handle */
+struct protected_file* find_protected_file_handle(PAL_HANDLE handle) {
+    char uri[URI_MAX];
+    int uri_len;
+
+    /* TODO: this logic is inefficient, add a PF reference to PAL_HANDLE instead */
+    uri_len = _DkStreamGetName(handle, uri, URI_MAX);
+    if (uri_len < 0)
+        return NULL;
+
+    /* uri is prefixed by "file:", we need path */
+    assert(strstartswith_static(uri, URI_PREFIX_FILE));
+    return find_protected_file(uri + URI_PREFIX_FILE_LEN);
+}
+
+static int register_protected_path(const char* path, struct protected_file** new_pf);
+
+/* Return a registered PF that matches specified path
+   (or the path that is contained in a registered PF directory) */
+struct protected_file* get_protected_file(const char* path) {
+    struct protected_file* pf = find_protected_file(path);
+    if (pf)
+        goto out;
+
+    pf = find_protected_dir(path);
+    if (pf) {
+        /* path not registered but matches registered dir */
+        SGX_DBG(DBG_D, "get_pf: registering new PF '%s' in dir '%s'\n", path, pf->path);
+        int ret = register_protected_path(path, &pf);
+        __UNUSED(ret);
+        assert(ret == 0);
+        /* return newly registered PF */
+    }
+
+out:
+    SGX_DBG(DBG_D, "get_pf(%s) = %p\n", path, pf);
+    return pf;
+}
+
+#define S_ISDIR(m) ((m & 0170000) == 0040000)
+
+static int is_directory(const char* path, bool* is_dir) {
+    int fd = -1;
+    struct stat st;
+
+    *is_dir = false;
+    int ret = ocall_open(path, O_RDONLY, 0);
+    if (IS_ERR(ret)) {
+        /* this can be called on a path without the file existing, assume non-dir for now */
+        ret = 0;
+        goto out;
+    }
+
+    fd = ret;
+    ret = ocall_fstat(fd, &st);
+    if (IS_ERR(ret)) {
+        SGX_DBG(DBG_E, "is_directory(%s): fstat failed: %d\n", path, ret);
+        goto out;
+    }
+
+    if (S_ISDIR(st.st_mode))
+        *is_dir = true;
+
+out:
+    if (fd >= 0) {
+        int rv = ocall_close(fd);
+        if (IS_ERR(rv)) {
+            SGX_DBG(DBG_E, "is_directory(%s): close failed: %d\n", path, rv);
+        }
+    }
+
+    return IS_ERR(ret) ? unix_to_pal_error(ERRNO(ret)) : ret;
+}
+
+/* Register all files from the given directory recursively */
+static int register_protected_dir(const char* path) {
+    int fd = -1;
+    int ret = -PAL_ERROR_NOMEM;
+    size_t bufsize = 1024;
+    void* buf = malloc(bufsize);
+
+    if (!buf)
+        return -PAL_ERROR_NOMEM;
+
+    ret = ocall_open(path, O_RDONLY | O_DIRECTORY, 0);
+    if (IS_ERR(ret)) {
+        SGX_DBG(DBG_E, "register_protected_dir: opening %s failed: %d\n", path, ret);
+        ret = unix_to_pal_error(ERRNO(ret));
+        goto out;
+    }
+    fd = ret;
+
+    size_t path_size = strlen(path) + 1;
+    int returned;
+    do {
+        returned = ocall_getdents(fd, buf, bufsize);
+        if (IS_ERR(returned)) {
+            ret = unix_to_pal_error(ERRNO(returned));
+            SGX_DBG(DBG_E, "register_protected_dir: reading %s failed: %d\n", path, ret);
+            goto out;
+        }
+
+        int pos = 0;
+        struct linux_dirent64* dir;
+
+        while (pos < returned) {
+            dir = (struct linux_dirent64*)((char*)buf + pos);
+
+            if (!strcmp_static(dir->d_name, ".") || !strcmp_static(dir->d_name, ".."))
+                goto next;
+
+            /* register file */
+            size_t sub_path_size = URI_PREFIX_FILE_LEN + path_size + 1 + strlen(dir->d_name);
+            char* sub_path = malloc(sub_path_size);
+            ret = -PAL_ERROR_NOMEM;
+            if (!sub_path)
+                goto out;
+
+            snprintf(sub_path, sub_path_size, URI_PREFIX_FILE "%s/%s", path, dir->d_name);
+            ret = register_protected_path(sub_path, NULL);
+            if (ret != 0) {
+                free(sub_path);
+                goto out;
+            }
+            free(sub_path);
+next:
+            pos += dir->d_reclen;
+        }
+    } while (returned != 0);
+
+    ret = 0;
+out:
+    if (fd >= 0)
+        ocall_close(fd);
+    free(buf);
+    return ret;
+}
+
+/* Register a single PF (if it's a directory, recursively) */
+static int register_protected_path(const char* path, struct protected_file** new_pf) {
+    int ret = -PAL_ERROR_NOMEM;
+    struct protected_file* new = NULL;
+
+    char* normpath = malloc(URI_MAX);
+    if (!normpath)
+        goto out;
+
+    size_t len = URI_MAX;
+    ret = get_norm_path(path, normpath, &len);
+    if (ret < 0) {
+        SGX_DBG(DBG_E, "Couldn't normalize path (%s): %s\n", path, pal_strerror(ret));
+        goto out;
+    }
+
+    /* discard the "file:" prefix */
+    if (strstartswith_static(normpath, URI_PREFIX_FILE))
+        path = normpath + URI_PREFIX_FILE_LEN;
+    else
+        path = normpath;
+
+    if (find_protected_file(path)) {
+        ret = 0;
+        SGX_DBG(DBG_D, "register_protected_path: file %s already registered\n", path);
+        goto out;
+    }
+
+    new = calloc(1, sizeof(*new));
+    if (!new) {
+        ret = -PAL_ERROR_NOMEM;
+        goto out;
+    }
+
+    new->path_len = strlen(path);
+    /* This is never freed but so isn't the whole struct, PFs persist for the whole lifetime
+       of the process. */
+    new->path = malloc(new->path_len + 1);
+    if (!new->path) {
+        ret = -PAL_ERROR_NOMEM;
+        goto out;
+    }
+
+    memcpy(new->path, path, new->path_len + 1);
+    new->refcount = 0;
+    new->writable_fd = -1;
+
+    bool is_dir;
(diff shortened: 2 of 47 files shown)


sudharkrish:
@omeg file's content has confidentiality, and integrity protected.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files.c, line 50 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M04wKO09gBamzXF1tQN:-M04wKO09gBamzXF1tQO:b-gvdxyc) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files.c#L50)):*
> ```C
>  */
> 
> #include <linux/fs.h>
> ```

This file protected_files.c is 2000+ lines, has all the code from SGX SDK library, that was spread across several files, based on functionality, like file_read_write.c, file_flush.c, file_crypto.c and so on. Can we have a sub-directory under host/Linux-SGX, like pf or protected-fs and split this feature into a bunch of files, its hard to review, such a huge file.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 163 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M04jh6w82nEnERFdjQ-:-M04jh6w82nEnERFdjQ0:bs1ofv5) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/pal_linux.h#L163)):*
> ```objc
> 
> /* Used to track map buffers for protected files */
> DEFINE_LIST(pf_map);
> ```

Can you clarify or add some comments on how these pf_map buffers are used while flushing to the PF and whether it has any dependency with flushing from lru_cache to PF on disk.

sudharkrish:
___
*[LibOS/shim/test/fs/manifest.template, line 37 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M04yeUoEd83c4vwkNAh:-M04yeUoEd83c4vwkNAi:b-v2oqj1) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/LibOS/shim/test/fs/manifest.template#L37)):*
> ```
> sgx.protected_files_key = ffeeddccbbaa99887766554433221100
> sgx.protected_files.input = file:tmp/pf_input
> sgx.protected_files.output = file:tmp/pf_output
> ```

This part is not clear. In this example manifest, tmp/pf_input, tmp/pf_output, are these input and output files, (OR) are they directory-paths to input files and output files.
For example, if pf_input is a file(unencrypted) under tmp directory, and new protected file will be generated under tmp as pf_output(encrypted).

sudharkrish:
___
*[LibOS/shim/test/fs/test_pf.py, line 20 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0Q2tM_AOU3eZ2-rfXS:-M0Q2tM_AOU3eZ2-rfXT:b-ta68sp) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/LibOS/shim/test/fs/test_pf.py#L20)):*
> ```Python
> 
> @unittest.skipUnless(HAS_SGX, 'Protected files require SGX support')
> class TC_50_ProtectedFiles(TC_00_FileSystem):
> ```

Just wondering, if it would be possible to add some tests that shows performance of protected file access, like file read, file write and other protected file operations, when compared to regular non-protected files, when running within graphene. It will be helpful to have this peformance data.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/db_files.c, line 131 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0QAI1IEWR1VgJgd1-0:-M0QAI1J1C2YJTbHA0Sc:b-qey8ao) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/db_files.c#L131)):*
> ```C
>             }
> 
>             pf->refcount++;
> ```

Just wondering, if we would need a spinlock to protect increment or decrement of this reference count.

sudharkrish:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0V4pBK2kuURDfK0QbW:-M0V4pBLCosocW3bhPv3:bg95o1j) (no related file):*
@omeg,  sgx-protect fs has the option where user can provide custom-key(like how you are providing through manifest file) or it can be automatically generated using platform's sealing key. Using automatically generated keys(using platform key) will be a good feature, for applications that do not have usages to move protected files from one platform to another. Wondering what was the reason, for removing that functionality.

---

sudharkrish:
___
*[LibOS/shim/test/fs/manifest.template, line 35 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0V5TL508JsmguKw0HI:-M0V5TL6EFpmw9wnbBl8:b-hagiwj) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/LibOS/shim/test/fs/manifest.template#L35)):*
> ```
> sgx.allowed_files.tmp_dir = file:tmp/
> 
> sgx.protected_files_key = ffeeddccbbaa99887766554433221100
> ```

Trying to think option, to avoid having key in the manifest. There is plan to have a generic mechanism for secret provisioning in graphene after remote attestation. For example, after remote attestation, graphene library can output a sealed(using SGX sealing) secret to filesystem. And then the application that uses, protected-files feature, can unseal the blob, to retrieve the user-provided custom key. For the short-term, if we can add support for provisioning sealed custom key, that will be better than having the key in the manifest file. This would also mean that we add support for SGX seal and unseal in graphene.

LibOS/shim/test/fs/manifest.template
```diff
@@ -31,3 +31,7 @@ sgx.trusted_files.libpthread = file:../../../../Runtime/libpthread.so.0
 sgx.trusted_files.libgcc_s = file:/lib/x86_64-linux-gnu/libgcc_s.so.1
 
 sgx.allowed_files.tmp_dir = file:tmp/
+
+sgx.protected_files_key = ffeeddccbbaa99887766554433221100
+sgx.protected_files.input = file:tmp/pf_input
```
sudharkrish: It would be nice to support multiple protected-directory paths, probably separated by a delimiter(like colon or semi-colon).

sudharkrish:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0VPDP6BocfAzrhvcc-:-M0VPDP7-rqLTab5E3bQ:b-2r4bzj) (no related file):*
@omeg , from looking at the code, your providing ->file swap protection (a PF can only be accessed when in a specific path), by the check in this api->/* Exact match of path in protected_file_list */
struct protected_file* find_protected_file(const char* path). Is this understanding correct.

---

sudharkrish:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0V4pBK2kuURDfK0QbW:-M0Zqp0L0UcGfe4BBQcX:bnspjlh) (no related file):*
<details><summary><i>Previously, omeg (Rafał Wojdyła) wrote…</i></summary><blockquote>

I think that the application only needs to protect its files if it had some secrets provisioned to it. If there's no secret data, there's no point of using protected files IMO. And if secrets are provisioned (after remote attestation for example) then the PF key can also be provided.
</blockquote></details>

@omeg, I am not sure, if we can assume, that an application needs to protect its files, only if it had some secrets provisioned to it. These 2 aspects can be independent. Maybe something we can discuss in the graphene meeting. During run-time application can generate content that needs protection. And even if secrets are provisioned, they can be something else, not necessarily a key to protect files. I think it will be good to give the option to the end-user, for generating files based on automatically generated platform key. The code to use the sealing key is already there in the updated version we provided, just needs to be enabled.

---

sudharkrish:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0VPDP6BocfAzrhvcc-:-M0ZvFSn7Cm0zthRlTAd:b-2olash) (no related file):*
<details><summary><i>Previously, omeg (Rafał Wojdyła) wrote…</i></summary><blockquote>

This check is to determine if a file should be treated like a protected file. The actual path check against PF's metadata is in `ipf_init_existing_file()`.
</blockquote></details>

@omeg , looks like we have an issue here. In ipf_open, you have the check->  if (filename && strnlen(filename, FULLNAME_MAX_LEN) >= FULLNAME_MAX_LEN - 1), but then inside ipf_init_new_file, ipf_init_existing_file, you are using FILENAME_MAX_LEN, which is much less than FULLNAME_MAX_LEN.  the current sgx-sdk implementation removes the path-prefix, and only maintains the actual filename in the meta-data, with 260 characters as max length for filename->  char     clean_filename[FILENAME_MAX_LEN]; In linux systems, filename(including path-prefix) can be upto 4KB. Given that you are NOT removing the path-prefix in ipf_open, and storing the full-path and filename in clean_filename the length of this array needs to be bumped to be atleast equal to  FULLNAME_MAX_LEN  or as per lengths in linux systems.

---

sudharkrish:
___
*[Pal/src/host/Linux-SGX/lru_cache.c, line 128 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_3zN181cfgz8P9I6T:-M0_3zN181cfgz8P9I6U:b-p7l5cu) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/lru_cache.c#L128)):*
> ```C
> }
> 
> uint32_t lruc_size(lruc_context_t lruc) {
> ```

I have updated the list.h macros PR, to get the size in one call using LISTP_SIZE, instead of having to loop through. Can you try to use the changes in this PR>https://github.com/oscarlab/graphene/pull/1268

sudharkrish:
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 199 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_6v086wx_JbRchjpR:-M0_6v086wx_JbRchjpS:b-gljur8) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/enclave_pf.c#L199)):*
> ```C
> 
> /* Exact match of path in protected_file_list */
> struct protected_file* find_protected_file(const char* path) {
> ```

Given that this would get called first, for any PF related api, it can introduce latency, can we use a hashtable here to speed up protected file lookup.

sudharkrish:
___
*[Pal/src/host/Linux/pal_linux_error.h, line 12 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M04eLVe3truyEsUrzOS:-M0_8G-QDqrUmFi171Kn:b-ql5ury) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux/pal_linux_error.h#L12)):*
<details><summary><i>Previously, omeg (Rafał Wojdyła) wrote…</i></summary><blockquote>

Without this change a `0` errno was converted to `-PAL_ERROR_DENIED` which complicated some code paths in the mbedtls crypto adapter.
</blockquote></details>

In general, this unix_to_pal_error, should only get called if errno is non-zero. I noticed in enclave_pf.c::is_directory, you are calling without an extra check. Can you try modifying that. return IS_ERR(ret) ? unix_to_pal_error(ERRNO(ret)) : ret;

Pal/src/host/Linux-SGX/enclave_pf.c
```diff
@@ -0,0 +1,623 @@
...
+
+out:
+    if (fd >= 0) {
+        int rv;
+        if ((rv = ocall_close(fd)) < 0) {
+            SGX_DBG(DBG_E, "is_directory(%s): close failed: %d\n", path, rv);
+        }
+    }
+
+    return unix_to_pal_error(ERRNO(ret));
```
sudharkrish: Can you change to return IS_ERR(ret) ? unix_to_pal_error(ERRNO(ret)) : ret;

sudharkrish:
___
*[LibOS/shim/test/fs/manifest.template, line 36 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M-ufaDNkH0bv-OlF8Lz-r1-36:-M04ybMIA-gbyvawqtoS:b-ckrk9m) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/LibOS/shim/test/fs/manifest.template#L36)):*
<details><summary><i>Previously, omeg (Rafał Wojdyła) wrote…</i></summary><blockquote>

You can have multiple entries in the manifest, like this:
```
sgx.protected_files.pf_1 = file:tmp/some_file
sgx.protected_files.pf_2 = file:tmp/some_dir
sgx.protected_files.pf_3 = file:tmp/another_dir/some_file
```
</blockquote></details>

That sounds great. Maybe the manifest file of the test application can illustrate that or in the README.

sudharkrish:
___
*[LibOS/shim/test/fs/test_pf.py, line 20 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0Q2tM_AOU3eZ2-rfXS:-M0_B2xg7Ufj4wnQSKfd:b-w7erwk) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/LibOS/shim/test/fs/test_pf.py#L20)):*
<details><summary><i>Previously, omeg (Rafał Wojdyła) wrote…</i></summary><blockquote>

I'm working on that actually (and collecting FS usage patterns from some example real-world payloads for possible optimization).
</blockquote></details>

Okay sounds good.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/db_files.c, line 131 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0QAI1IEWR1VgJgd1-0:-M0_BCrkCZk_qb0McrwA:bh68lzn) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/db_files.c#L131)):*
<details><summary><i>Previously, omeg (Rafał Wojdyła) wrote…</i></summary><blockquote>

I need to properly review how concurrency is handled in Graphene FS handlers. I'm preparing some multithreaded/multiprocess FS tests for Graphene that should help find related bugs as well.
</blockquote></details>

Okay sure, you can also ask people in the slack group regarding this.  Maybe some folks already know the answer.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files.c, line 1937 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_QnuM4DNbe9rZhxuE:-M0_QnuN35y45pmK6661:b-87ieqz) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files.c#L1937)):*
> ```C
>                     pf_file_mode_t mode, bool create, bool enable_recovery, const pf_key_t* key,
>                     pf_context_t* context) {
>     if (!check_callbacks())
> ```

How about having a global variable like g_pf_lib_init_done, in init_protected_files api. And check if (g_pf_lib_init_done) instead of using check_callbacks in these pf_* apis.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 39 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_W4MX9ljotjKQu63K:-M0_W4MX9ljotjKQu63L:ble272f) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/enclave_pf.c#L39)):*
> ```C
> 
> /* Callbacks for protected files handling */
> static void* cb_malloc(size_t size) {
> ```

Just thinking, if we really need this callback->cb_malloc, why cant you directly use the malloc or calloc api(to avoid memset after malloc) in protected_files.c source code.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files_internal.h, line 196 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_bByz1sjwV5BcS4E4:-M0_bBz--iKzWTTX4uy6:bgix9o) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files_internal.h#L196)):*
> ```objc
>     meta_data_encrypted_t encrypted_part_plain; // encrypted part of meta data node, decrypted
>     file_node_t root_mht; // the root of the mht is always needed (for files bigger than 3KB)
>     pf_handle_t file;
> ```

Just wondering, is there a need for this pf_handle_t and whether it needs to be void * pointer. Since you are getting the OS file descriptor after call to open in db_files.c::db_open, we can just store this as ->int fd in struct pf_context. Not sure, if I am missing something else here.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files.c, line 91 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_dehvBgPoEOkYdWLs:-M0_dehwBSEBa0VQs1VY:bhp7ohx) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files.c#L91)):*
> ```C
> #define PF_DEBUG_PRINT_SIZE_MAX 4096
> 
> /* Debug print without function name prefix. Implicit param: pf (context pointer). */
> ```

These macros should be NOOPs, unless #ifdef DEBUG is enabled. But I know, #ifdefs are frowned upon.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files_internal.h, line 209 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_e63w4fcRZd_AQjIC:-M0_e63w4fcRZd_AQjID:bkdtmy2) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files_internal.h#L209)):*
> ```objc
>     char recovery_filename[RECOVERY_FILE_MAX_LEN]; // might include full path to the file
>     lruc_context_t cache;
>     char* debug_buffer; // buffer for debug output
> ```

This debug_buffer needs to be there, only if #ifdef DEBUG is enabled. But I know, #ifdefs are frowned upon.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/uthash.h, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_fc5j9eVhb-IFeC7q:-M0_fc5j9eVhb-IFeC7r:b-styll5) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/uthash.h#L1)):*
> ```objc
> /*
> ```

How about moving this file to Pal/include/lib, where we already have other related files such as list.h

sudharkrish:
___
*[Pal/src/host/Linux-SGX/tools/pf_util.c, line 551 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_gVBu7-nuOCNVqBEk:-M0_gVBu7-nuOCNVqBEl:b-lynd75) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/tools/pf_util.c#L551)):*
> ```C
> };
> 
> static int process_files(const char* input_dir, const char* output_dir, const char* prefix,
> ```

Nice work, with these utility apis in pf_util.c, will be helpful for people using this feature.

sudharkrish:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_hn9I7SSQ3Do5IYw6:-M0_hn9I7SSQ3Do5IYw7:b-3pa07h) (no related file):*
Just curious, if an application tries to rename a protected file, how will it get handled. I am sure, it will be dis-allowed, but it will be good to have a test case for that. Not sure, if you already have one for protected file rename.

---

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files.c, line 66 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_jnkx8eZNDFkV5h-B:-M0_jnkx8eZNDFkV5h-C:b74ocfn) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files.c#L66)):*
> ```C
> #endif
> 
> static void erase_memory(void *buffer, size_t size) {
> ```

Just curious, why you are using this custom api, instead of memset.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 118 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_kovWCGf_8pRFX7Gt:-M0_kovXC-9Yutj9r3On:btzl4fd) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/enclave_pf.c#L118)):*
> ```C
> static pf_status_t cb_open(const char* path, pf_file_mode_t mode, pf_handle_t* handle,
>                            size_t* size) {
>     // TODO (only used for recovery files)
> ```

I understand that you are disabling file-recovery feature during file-open. But since you are anyway having the recovery related code, we can just add the code here for OS call for file-open in cb_open, just for the sake of completeness. Later on, if we decide to enable file-recovery, it will be handy.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files.c, line 1982 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_nK3-1SU1JoLUBwER:-M0_nK3-1SU1JoLUBwES:b4bkbco) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files.c#L1982)):*
> ```C
>         pf->offset = pf->encrypted_part_plain.size;
>         DEBUG_PF("extending the file from %lu to %lu\n", pf->offset, size);
>         if (ipf_write(pf, NULL, size - pf->offset) != size - pf->offset)
> ```

I suppose in order to support this file extension you have this-> memcpy_or_zero api inside ipf_write, to append NULL bytes at the end.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/db_files.c, line 565 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_rLhBAmBLYxH88pxY:-M0_rLhBAmBLYxH88pxZ:b-maxkc) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/db_files.c#L565)):*
> ```C
>     __attribute__((unused)) pf_status_t pfs = pf_get_size(pf->context, &size);
>     assert(PF_SUCCESS(pfs));
>     attr->pending_size = size;
> ```

I see you are closing the PF if that is the only open instance of that PF. In the case, where another instance of PF is already open, would it help to set some of the other attributes of the open PF, using  ocall_fstat(fd, &stat_buf), where fd is the OS's fd of that PF stored in pf_context.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files.c, line 714 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_xMKz078KSeMChPsh:-M0_xMKz078KSeMChPsi:b-1a8ys6) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files.c#L714)):*
> ```C
> // file_flush.cpp
> 
> #if 0 // unused
> ```

Can be removed.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files.c, line 1154 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_xd1z2e1oDxGeqW2U:-M0_xd1z2e1oDxGeqW2V:b-bo077f) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files.c#L1154)):*
> ```C
> }
> 
> #if 0 // unused
> ```

All these unused apis under #if 0 can be removed.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files.c, line 1216 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_xlnm7EAIGMn5iiKG:-M0_xlnm7EAIGMn5iiKH:bs4q9op) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files.c#L1216)):*
> ```C
> }
> 
> #if 0 // unused
> ```

Ditto.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files.c, line 1229 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_xoTz6CnrGpzb_YoZ:-M0_xoTz6CnrGpzb_Yo_:bs4q9op) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files.c#L1229)):*
> ```C
> #endif
> 
> #if 0 // unused
> ```

Ditto.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files.c, line 1262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_xsV8Cl765FPiB30E:-M0_xsV8Cl765FPiB30F:bs4q9op) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files.c#L1262)):*
> ```C
> }
> 
> #if 0 // unused
> ```

Ditto.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files.c, line 1300 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_y0Cg3XvUH9Xwda-q:-M0_y0Cg3XvUH9Xwda-r:b3oqlez) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files.c#L1300)):*
> ```C
> 
> // memcpy src->dest if src is not NULL, zero dest otherwise
> static void memcpy_or_zero(void* dest, const void* src, size_t size) {
> ```

how about memcpy_or_zero_initialize

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files_internal.h, line 70 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_yLUjEw0F59ocbyTB:-M0_yLUjEw0F59ocbyTC:b-vxz811) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files_internal.h#L70)):*
> ```objc
> 
>     pf_keyid_t meta_data_key_id;
>     //sgx_cpu_svn_t    cpu_svn;
> ```

I think it will be good to bring back the code that uses automatic keys generated based on platform sealing key.

omeg:
Jenkins, retest Jenkins-Debug please (apps.LTP.setsockopt03 failed)

omeg:
Jenkins, retest Jenkins-SGX please (apps.LTP.setsockopt03 failed)

sudharkrish:
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 199 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0_6v086wx_JbRchjpR:-M0tVVGH16kgpFSr35F9:b-v6ne4t) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/enclave_pf.c#L199)):*
<details><summary><i>Previously, omeg (Rafał Wojdyła) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

That is great. Yes, I was thinking you should be able to leverage the apis in uthash.h.

sudharkrish:
___
*[Pal/src/host/Linux-SGX/protected_files.c, line 50 at r1](https://reviewable.io/reviews/oscarlab/graphene/1325#-M04wKO09gBamzXF1tQN:-M0tVtYwCkII1w1DbOp4:b-4nog2f) ([raw file](https://github.com/oscarlab/graphene/blob/c2f91a02918b142a9bf5e945d16695f755301ddc/Pal/src/host/Linux-SGX/protected_files.c#L50)):*
<details><summary><i>Previously, omeg (Rafał Wojdyła) wrote…</i></summary><blockquote>

Agreed, I'll split the implementation into smaller files.
</blockquote></details>

I noticed that you added a directory protected_files, which is good. Just curious, I thought you were saying, you will also break that 1 huge file->protected_files.c, that will be nice to have, also for future maintenance.

sudharkrish:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1325#-M0VPDP6BocfAzrhvcc-:-M2_q5VI5ceFZfbfP5v4:b-nthi23) (no related file):*
<details><summary><i>Previously, omeg (Rafał Wojdyła) wrote…</i></summary><blockquote>

I changed size of the path buffer to `FULLNAME_MAX_LEN`.
</blockquote></details>

#define PATHNAME_MAX_LEN      (512)
#define FILENAME_MAX_LEN      (260)
#define FULLNAME_MAX_LEN      (PATHNAME_MAX_LEN+FILENAME_MAX_LEN)

We should probably document in the README, regarding this limitation for file-path length. 

---

omeg:

omeg:
Jenkins, test Jenkins-18.04 please

omeg:
Jenkins, retest Jenkins-SGX-18.04 (weird build failure)

omeg:
Jenkins, retest Jenkins-18.04 please (lmbench d/l failed)

ravikumarbhattiprolu:
With the latest mainline code this pull request is causing merge conflicts. Can this request be rehashed over current mainline so that these conflicts are removed? This will help till code reviews are complete and PR is merged with mainline.

omeg:
Jenkins, retest Jenkins-Debug please (apps.LTP.kill01 failed)

PR #1326, opened 2020-02-12
[Pal/Linux-SGX] resolve race condition to use ustack for ocall and AEX

### LINKED ISSUE #486: [Pal/Linux] Signals are not delivered on altstack because SA_ONSTACK is not used

yamahata:
Untrusted stack is used for ocall. The argument is allocated by subtracting
from untrusted stack. which is managed by enclave_tls::ustack.
On the other hand, for AEX, GPR::ursp is used. As result the area below
the untrusted stack can be clobbered by signal handler in untrusted code.

Use sgx_pal_gpr_t::ursp instead of enclave_tls::ustack to manage the argument
for ocall so that signal handler for AEX doesn't clobber them.
I guess that the reason why this doesn't matter much is, usually the size
of argument is less than red zone (128 bytes). So the signal frame doesn't
overlap with it.

(Other option is, allocate dedicated stack for ocall.)

Closes #527. Closes #1025. Fixes #486.

## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] Common PAL code

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1326)

### DIFF
--- a/LibOS/shim/test/ltp/ltp-sgx.cfg
+++ b/LibOS/shim/test/ltp/ltp-sgx.cfg
@@ -1853,7 +1853,6 @@ skip = yes
 skip = yes
 
 [mprotect04]
-timeout = 60
 skip = yes
 
 [mq_notify01]
--- a/LibOS/shim/test/ltp/ltp.cfg
+++ b/LibOS/shim/test/ltp/ltp.cfg
@@ -1439,6 +1439,10 @@ must-pass =
 [mmap01]
 skip = yes
 
+# uses siglongjmp() from exception handler back into normal execution; not supported by Graphene
+[mmap05]
+skip = yes
+
 [mmap06]
 skip = yes
 
@@ -1446,6 +1450,10 @@ skip = yes
 skip = yes
 
 [mmap12]
+skip = yes
+
+# uses siglongjmp() from exception handler back into normal execution; not supported by Graphene
+[mmap13]
 skip = yes
 
 [mmap14]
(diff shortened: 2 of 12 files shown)


dimakuv:
@yamahata Some LTP tests failed: `mmap05`, `mmap13`, `mprotect04`, `munmap01`, `munmap02` on Linux PAL. This is related to the newly added signal-stack for Linux PAL. Could you look into this?

dimakuv:
Jenkins, retest Jenkins please (internet connectivity)

dimakuv:
Jenkins, retest Jenkins-SGX-18.04 please (`fork_and_exec` LibOS test failed, seems unrelated)

mkow:
Jenkins, retest Jenkins-SGX-18.04-apps please (lmbench download failed)

PR #1320, opened 2020-02-11
[Pal/Linux-SGX] cache mmap of untrusted area for read/write/recv/send

yamahata:
So that the number of ocall can be reduced.

## Affected components

- [ ] README and global configuration
- [ ] Linux PAL
- [x] SGX PAL
- [ ] Common PAL code
- [ ] Library OS (i.e., SHIM), including GLIBC

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1320)

PR #1327, opened 2020-02-12
[LibOS] chrootfs: remove mmap logic to read/write

yamahata:
This patch makes chroot_read/write() simply call DkStreamRead/Write()
for regular file and eliminate mmap trick.

The mmap trick actually causes overhead(one extra memory copy) on Pal/Linux-SGX.
It can be done in Pal/Linux-SGX cleanly.
(which is the future TODO. comment is added there)

The current execution flow is
- if it's not mapped yet,
  - find the unmapped region and
  - map the file to the region if necessary
  Please notice that LibOS passes address for mapping region to Pal.
  Pal/Linux-SGX has to copy memory because the region is in enclave.
- memcpy

Its goal seems trying to avoid memory copy and mmap/munmap unnecessary.
In Pal/Linux-SGX, actually DkStreamMap() does one more memory copy in read case
because LibOS passes mapping address for it.

In Pal/Linux case, Either way do only one copy. so mmap()/munmap() is deleted.

In Pal/Linux-SGX case, this patch will simply invoke ocall_read/write() to
eliminate one memory copy.
With this patch, the flow is
- mmap untrusted region for buffer of read/write
- ocall_read/write + memcpy (two memory copy)
- munmap untrusted region

NOTE:
https://github.com/oscarlab/graphene/pull/1320 will optimize out mmap/munmap().

## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] Common PAL code

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1327)

### DIFF
--- a/LibOS/shim/include/shim_handle.h
+++ b/LibOS/shim/include/shim_handle.h
@@ -86,11 +86,6 @@ struct shim_file_handle {
     enum shim_file_type type;
     off_t size;
     off_t marker;
-
-    enum { FILEBUF_MAP, FILEBUF_NONE } buf_type;
-    size_t mapsize;
-    off_t mapoffset;
-    void* mapbuf;
 };
 
 #define FILE_HANDLE_DATA(hdl)  ((hdl)->info.file.data)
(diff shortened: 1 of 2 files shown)


yamahata:
Jenkins, retest Jenkins-SGX please. TC_01_Bootstrap.test_202_fork_and_exec  AssertionError: 'child exited with status: 0' not found in 'test completed successfully\n'

yamahata:
Jenkins, retest Jenkins-Debug please. apps.LTP.lseek07 timeout

dimakuv:
Jenkins, retest Jenkins-SGX please (`vork_and_exec` LibOS test failed)

dimakuv:
Jenkins, retest Jenkins-18.04 please (downloading LTP suite failed)

Issue #1337, opened 2020-02-19
UTF-8 characters in trusted/allowed file names


vahldiek:
Hi,

Just ran into an issue with UTF-8 encoded files names (i.e., file names with ü) when used in the manifest to create trusted/allowed files. To recreate you'd have to create a file with such a character, include it as a trusted file in your favorite manifest, then run the pal-sgx-signer (which will break), token generation and then the loader (which will break again).

I've attached an updated version of the [pal_sgx_sign.py.txt](https://github.com/oscarlab/graphene/files/4222673/pal_sgx_sign.py.txt) that includes the required *.encode calls and encoding="utf-8" additions within python. I've not dug further into the manifest initialization within the libOS when loading an application, but I expect the code to be equally unprepared.

This issue is mainly to document my findings.

mkow:
Thanks! Could you create a PR with the mentioned fixes?

woju:
I posted this comment in #1355 (here with minor fixes):

Filenames are not required to be valid UTF-8 either, so hardcoding `.encode('utf-8')` etc is wrong. In principle, filenames are sequences of bytes (octets) and the only requirement is that they do not contain U+0 (`\0`) and U+2F (`/`) characters, because those are separators for C string and path, respectively. (Filesystems may impose limits on total path length, which they do at about 256 octets or slightly less, but that's implementation, not API).

Python has two mechanisms to cope with that:
- https://docs.python.org/library/os#os.PathLike and https://docs.python.org/library/os#os.fspath
- `errors='surrogateescape'` on paths (https://docs.python.org/library/sys#sys.getfilesystemencodeerrors).

Both of those mechanisms are since Python 3.6, but we support xenial (16.04 LTS), which has Python 3.5, which is a problem. On Python<=3.5 argument to `open()` has to be `str`, which implies that it is correctly encoded.

If we'd like to fix this correctly, we should allow for arbitrary byte sequences in paths. We could:
- specify that manifest must be encoded in UTF-8 (and open it with `encoding='utf-8');
- allow `\`-escapes or some other way to specify arbitrary bytes;
- tell people that incorrectly encoded file paths (wrt to current locale) are unsupported on python3.5/xenial.

To fix the path encoding problems on <=3.5, I usually do something like this:

```python
try:
    from os import fspath
except ImportError:
    # python<3.6
    fspath = str
```

Then one can use also `bytes` and `pathlib.Path` as file paths:
```python
with open(fspath(filepath)) as file:
    ...
```

[...]

PR #1355, opened 2020-03-02
[SGX PAL] pal_sgx_sign.py: UTF-8 characters in trusted file URIs

vahldiek:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] Common PAL code

## Description of the changes 

Updated pal_sgx_sign.py to allow trusted files to include UTF-8 characters. As suggested in https://github.com/oscarlab/graphene/issues/1337, the pal_sgx_sign.py fails to generate a signature for an application with a trusted filename including UTF-8 characters.

These changes are only concerned about the pal_sgx_sign.py and not about the actual initialization of Graphene. As a result, Graphene fails during the initialization of a manifest file with UTF-8 characters in trusted file URIs. 

## How to test this PR? 

1) Create a file with a UTF-8 character in its name (i.e. ♥♥♥.txt)
2) Create a manifest file including
```
sgx.trusted_file.love = file:♥♥♥.txt
```
3) Run pal_sgx_sign.py 
```
graphene/Pal/src/Host/Linux-SGX/signer/pal-sgx-sign 
    -libpal graphene/Runtime/libpal-Linux-SGX.so 
    -key graphene/Pal/src/Host/Linux-SGX/signer/enclave-key.pem 
    -output app.manifest.sgx
    -manifest app.manifest 
```

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1355)

### DIFF
--- a/Pal/src/host/Linux-SGX/signer/pal_sgx_sign.py
+++ b/Pal/src/host/Linux-SGX/signer/pal_sgx_sign.py
@@ -59,7 +59,7 @@ def parse_size(value):
 def read_manifest(filename):
     manifest = dict()
     manifest_layout = []
-    with open(filename, "r") as file:
+    with open(filename, "r", encoding="utf-8") as file:
         for line in file:
             if line == "":
                 manifest_layout.append((None, None))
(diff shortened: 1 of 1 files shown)


dimakuv:
Feels like this PR doesn't really fix the problem. We should investigate this issue more and do what @woju suggests. What about copying Wojtek's comment into the issue #1337 and closing this PR?

woju:
Copied the comment an closing.

Issue #1359, opened 2020-03-03
NodeJS 8.10.0 periodically failing on recent master


dimakuv:
Currently debugging NodeJS 8.10.0 on Ubuntu 18.04 (installed the default nodejs version via apt).

It very infrequently fails the simple helloworld script with:
```
$ SGX=1 ./pal_loader nodejs.manifest.sgx helloworld.js
#
# Fatal process OOM in heap setup
#

Unknown or illegal instruction at RIP 0x000000083b408a89
```

The last error happens because of `ud2` in the `/usr/bin/nodejs` assembly:
```
$ objdump -D /usr/bin/nodejs | less
0000000000c93a80 <_ZN2v84base2OS5AbortEv@@Base>:
  c93a80:       80 3d d1 05 70 00 00    cmpb   $0x0,0x7005d1(%rip)        # 1394058 <_ZN2v88platform7tracing24g_category_group_enabledE@@Base+0x178>
  c93a87:       74 02                   je     c93a8b <_ZN2v84base2OS5AbortEv@@Base+0xb>
  c93a89:       0f 0b                   ud2
  c93a8b:       ...
```

So NodeJS forces the "illegal instruction" failure explicitly. Now trying to understand what `Fatal process OOM` in NodeJS means.

# Root Cause

NodeJS requires ~1.7GB of heap space to function correctly. There is some randomization involved when initializing its heap allocator, so NodeJS fails only periodically on this.

Our current manifest specifies only 1GB as the enclave size, so this is not enough for NodeJS's heap allocator, and thus it periodically fails as above. I'm preparing the PR to fix it.

References:
* https://github.com/nodejs/node/issues/13018
* https://stackoverflow.com/questions/55933041/fatal-process-oom-in-heap-setup-when-running-node-with-ulimit
* https://github.com/minosworld/minos/issues/57

PR #1360, opened 2020-03-03
[LibOS] test/apps: Increase enclave size to 2GB so that NodeJS doesn't fail

dimakuv:
## Affected components

- [ ] Linux PAL
- [ ] SGX PAL
- [ ] Common PAL code

## Description of the changes 

NodeJS expects around 1.7GB of heap on startup. Previously, enclave size was set to only 1GB, which led to NodeJS probabilistically complaining about "Fatal process OOM in heap setup".

Corresponding PR is https://github.com/oscarlab/graphene-tests/pull/74.

Fixes https://github.com/oscarlab/graphene/issues/1359.

## How to test this PR? 

Nothing was changed in the test infrastructure.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1360)


dimakuv:
Jenkins, retest Jenkins-SGX please (`exec` LibOS test failed)

PR #1363, opened 2020-03-05
[Pal/Linux-SGX] Consolidate memory allocation logic and work around bugs

dimakuv:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] Common PAL code

## Description of the changes 

This PR contains three commits:

1. [Pal/Linux-SGX] Remove pal_vmas array to consolidate VMA logic in enclave_pages.c. Previously, Linux-SGX PAL contained a hack to bookkeep PAL-internal VMAs in the pal_vmas array with 64 VMAs. This limit led to out-of-PAL-memory errors for bigger applications, e.g., with huge number of trusted files. This commit removes the hack with pal_vmas and merges all its logic into the existing VMA bookkeeping logic in `enclave_pages.c`. This eliminates the PAL-internal limit and also improves performance, since there is no need to traverse an additional array (pal_vmas).

2. [Pal/Linux-SGX] Get rid of recursive deadlock of malloc(heap_vma_object). In Linux-SGX PAL, bookkeeping of enclave pages relies on a list of heap-VMA objects. Previously, these objects themselves were allocated via the same malloc mechanism, and if the pre-allocated memory pool got exhausted, heap-VMA allocation would end up in the bookeeping logic again, leading to a deadlock. To break this recursion, this commit introduces a pre-allocated BSS memory region to manage heap-VMA objects separately. This memory region is big enough to manage many heap-VMAs, and the existing logic of merging heap-VMA objects keeps the total number of heap-VMAs manageable, at least for typical workloads.

3. [Pal/Linux-SGX] Keep 8MB of enclave heap for internal PAL objects. This commit adds a hack to keep some heap (currently 8MB) for internal PAL objects allocated at runtime (recall that LibOS does not keep track of PAL memory, so without this hack it could overwrite internal PAL memory). This hack is probabilistic and brittle, but sufficient for now.

Closes #1340.

## How to test this PR? 

All tests must pass. Additionally, I tested locally on a memcached with 50,000 trusted files and new GSC with many files (not sure if we want to somehow add this to Jenkins).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1363)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_files.c
+++ b/Pal/src/host/Linux-SGX/db_files.c
@@ -199,7 +199,7 @@ static int file_map(PAL_HANDLE handle, void** addr, int prot, uint64_t offset, u
         return -PAL_ERROR_DENIED;
     }
 
-    mem = get_enclave_pages(mem, size);
+    mem = get_enclave_pages(mem, size, /*is_pal_internal=*/false);
     if (!mem)
         return -PAL_ERROR_NOMEM;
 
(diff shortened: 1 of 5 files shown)


dimakuv:

PR #1366, opened 2020-03-11
[Pal/Linux-SGX] Assorted SGX-related fixups

dimakuv:
## Affected components

- [ ] README and global configuration
- [ ] Linux PAL
- [x] SGX PAL
- [ ] Common PAL code
- [ ] Library OS (i.e., SHIM), including GLIBC

## Description of the changes 

Summary of commits in this PR:

- [Pal/Linux-SGX] Rename aesm_targetinfo to qe_targetinfo.
- [Pal/Linux-SGX] sgx_arch.h: Update sgx_report_body_t struct
- [Pal/Linux-SGX] sgx_api.h: Improve comments and fix formatting
- [Pal/Linux-SGX] Introduce generic sgx_get_report() and fix formatting

This PR is the first in the series of Remote Attestation PRs (re-worked #1284). Also see issue #1365.

## How to test this PR? 

All tests must pass. This PR is only formatting/renaming.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1366)

Issue #1365, opened 2020-03-10
[LibOS, Pal/Linux-SGX] Remote Attestation: Discussion of PRs


dimakuv:
Previously, we had Thomas working on this PR that exposes Remote Attestation to the LibOS and application layers (via `/dev/attestation/` pseudo-FS): https://github.com/oscarlab/graphene/pull/1284.

I am taking over this PR. Currently I'm leaning towards submitting a series of PRs (rebased on the current master) that will look like this:

- [x] PR 1 (#1366): Tiny bug fixes of SGX-related functions and structs. In particular:
    - `enclave_framework.c`: adding `sgx_get_report()`
    - `Linux-SGX/pal_linux.h`: assert on `pal_enclave_state` being 64B
    - `sgx_api.h`: better comments on `sgx_report()` and `sgx_getkey()`
    - `sgx_arch.h`: more fields in `sgx_report_body_t`
    - renaming `aesm_targetinfo` -> `qe_targetinfo`, `init_aesm_targetinfo` -> `init_quoting_enclave_targetinfo`

- [x] PR 2 (https://github.com/oscarlab/graphene/pull/1373): Removing most of Chia-Che's "Simple RA" code (won't be needed since useless self-verification will be replaced by pseudo-FS RA):
    - Removing `init_trusted_platform()`
    - Changing `ocall_get_attestation()` to return IAS header and report
    - Removing `sgx_attestation_t` and changing `ms_ocall_get_attestation_t`
    - Removing `parse_x509{_pem}()`
    - Significantly simplifying `sgx_verify_platform()` and renaming to `sgx_get_attestation()`
    - Significantly simplifying `contact_intel_attest_service()` to return IAS header and report instead of parsing them
    - `pal.h`: removing `attestation_status` and `attestation_timestamp`

- [x] PR 3 (#1367): Introduce `get_quote()` Linux-SGX PAL functions (as preparation for pseudo-FS RA):
    - `ocall_get_quote()`
    - `sgx_get_quote()`
    - `sgx_ocall_get_quote()`
    - `ms_ocall_get_quote_t` and `OCALL_GET_QUOTE`
    - `retrieve_quote()`

- [x] PR 4 (#1383): Introduce `DkAttestationReport()` and `DkAttestationQuote()` PAL APIs:
    - stubs for Skeleton and Linux PALs
    - `Linux-SGX/db_misc.c`: `_DkAttestationReport()` and `_DkAttestationQuote()`

[...]

PR #1367, opened 2020-03-12
[Pal/Linux-SGX] Introduce sgx_get_quote() OCALL

dimakuv:
## Affected components

- [ ] Linux PAL
- [x] SGX PAL
- [ ] Common PAL code

## Description of the changes 

New `sgx_get_quote()` OCALL allows to retrieve the SGX Quote from the Quoting Enclave. The retrieved quote is based on this enclave's SGX report and corresponds to this enclave. The caller of this OCALL may forward the quote to the remote user for remote attestation. A follow-up commit will introduce the actual callers.

This PR is the second in the series of Remote Attestation PRs (re-worked #1284). Also see issue #1365.

It is better to merge it after #1366 (but not necessary).

## How to test this PR? 

All tests must pass. Technically, this PR has no tests (the changes introduced here are dead code by themselves) since it is only a prerequisite for the follow-up PR.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1367)

### DIFF
--- a/Pal/src/host/Linux-SGX/ecall_types.h
+++ b/Pal/src/host/Linux-SGX/ecall_types.h
@@ -1,3 +1,5 @@
+#include <stddef.h>
+
 enum {
     ECALL_ENCLAVE_START = 0,
     ECALL_THREAD_START,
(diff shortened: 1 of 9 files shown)


dimakuv:
Jenkins, retest Jenkins-SGX please (`thread` LibOS test failed)

Issue #1380, opened 2020-03-22
[Graphene-SGX] Why have two independent base addresses?


vahldiek:
Hi,

I came across an application which starts PIE executables via fork/exec, but itself is non PIE. As a result, you have to specify sgx.static_address=1 in the PIE executable's manifest, otherwise the memory layouts are not compatible and it fails during fork. After chatting with @dimakuv we may want to discuss removing the differentiation and instead only use one base address for the enclave. The respective code can be found in [sgx_main.c](https://github.com/oscarlab/graphene/blob/8d76f9d3fb221a71f972461319e0179a823a5876/Pal/src/host/Linux-SGX/sgx_main.c#L300).

Is there a good reason to keep this differention? Otherwise, I'd suggest to remove the option of static_address from the manifest, the if/else branch in sgx_main.c, and change the pal_sgx_sign to not consider this option as well.

Happy to hear any concerns.

mkow:
So, you mean making everything effectively non-PIE? Doesn't sounds good, security-wise.

How hard would it be to fix the underlying problem with memory layouts? I don't remember what was the problem here.

dimakuv:
The annoying issue is this these lines of code basically: https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/sgx_main.c#L300

```c
    if (get_config(enclave->config, "sgx.static_address", cfgbuf, sizeof(cfgbuf)) > 0 && cfgbuf[0] == '1') {
        /* !!! this sets enclave base address to 0x0 */
        enclave->baseaddr = ALIGN_DOWN_POW2(heap_min, enclave->size);
    } else {
        /* !!! this sets enclave base address to 0x800000000 */
        enclave->baseaddr = ENCLAVE_HIGH_ADDRESS;
        heap_min = 0;
    }
```

And this `sgx.static_address` manifest option is derived from whether the executable is PIE or non-PIE. So different executables will have `sgx.static_address = 0` or `sgx.static_address = 1`.

But such executables become incompatible because of our (not amazing) checkpointing implementation. The checkpoint *must be* restored at the exact same addresses in the child enclave, but if the child enclave starts at a completely different address, the checkpoint restore fails!

vahldiek:
In addition, there is actually no ASLR going on here. In both cases the addresses are hard-coded, either 0 (first case) or ENCLAVE_HIGH_ADDRESS. (1 bit randomness).

[...]

PR #1384, opened 2020-03-24
[LibOS] Introduce test for system() and fix bug with Glibc 2.31

### LINKED ISSUE #1381: [Graphene] Since Glibc 2.31, system() libc call has different behavior and fails
The system() libc call has a different behavior between version 2.27 and 2.31. In 2.27, libc called fork and exec to start the shell. In 2.31, it leads to a particular clone call with arguments that Graphene currently does not support. Since master moved to 2.31 last week, this issue is triggered.

```
[     1] ---- shim_clone(CLONE_VM|CLONE_VFORK|SIGCHLD,0x4349eff0,0x439c8740,0x3000000030,0x0) ... clone
[     1] clone called with unsupported flags argument.
[     1] ---- return from shim_clone(...) = -22
```

To trigger this issue call `python` with `-c 'import os;os.system("ls")'`. The error occurs in [shim_clone.c](https://github.com/oscarlab/graphene/blob/master/LibOS/shim/src/sys/shim_clone.c#L194). It is supposed to be handled by the first condition check (VFORK), but due to its arguments being non-NULL, it faills. A simple removal of the non-NULL checks does not resolve the issue. 

I will look into this issue closer next week. Just a heads up, if someone currently uses glibc 2.31 and system().

dimakuv:
## Description of the changes: 

Glibc 2.31 implements system() POSIX function call via a combination of `clone(CLONE_VM|CLONE_VFORK|SIGCHLD, <child_stack_addr>, ...)` and `execve()`. This special case was not supported by Graphene. Moreover, Glibc 2.31 allocates a child stack and puts some variables on this child stack. After clone, the child process tries to access these variables on the child stack. However, previously Graphene always assumed that the child process returns on the same (parent) stack. Therefore, restored RSP in the child did not point to the child stack but to the parent stack, which triggered segfaults in Glibc.

This PR fixes this by temporarily rewiring the current thread (the one issuing `clone(VFORK, ...)`) so that it points to the provided child stack, and then rewiring it back to the original parent stack after the clone (in the parent process). The child process will resume execution with the correct child stack.
To test this bugfix, a LibOS test for `system()` is introduced. Also, all LibOS tests are forced to start enclaves at address 0x0. This removes checkpointing issues on mixes of PIE- and non-PIE executables (in particular, this happens in the `system()` test).

Fixes #1381. See detailed description of the bug there.

See #1380 regarding the need for `sgx.static_address = 1`.

## How to test this PR? 

New `system` LibOS regression test is added. It fails on previous Graphene with Glibc 2.31 and succeeds with this PR with Glibc 2.31 (i.e., on Ubuntu 18.04+).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1384)

### DIFF
--- a/LibOS/shim/src/sys/shim_clone.c
+++ b/LibOS/shim/src/sys/shim_clone.c
@@ -190,11 +190,48 @@ int shim_do_clone (int flags, void * user_stack_addr, int * parent_tidptr,
     int * set_parent_tid = NULL;
     int ret = 0;
 
-    /* special case for vfork. some runtime uses clone() for vfork */
-    if (flags == (CLONE_VFORK | CLONE_VM | SIGCHLD) &&
-        user_stack_addr == NULL && parent_tidptr == NULL &&
-        child_tidptr == NULL && tls == NULL) {
-        return shim_do_vfork();
+    /* special case of vfork: call shim_do_vfork() */
+    if (flags == (CLONE_VFORK | CLONE_VM | SIGCHLD)) {
+        /* some runtimes (e.g. Glibc 2.31+) specify user_stack_addr so that the child process
+         * must resume on this supplied stack; we mimic it by temporarily rewiring the current
+         * thread's stack values to the supplied user_stack_addr */
+        void* old_stack_top = self->stack_top;
+        void* old_stack_red = self->stack_red;
+        void* old_stack     = self->stack;
+        unsigned long old_stack_rsp = self->shim_tcb->context.regs->rsp;
+
+        if (user_stack_addr) {
+            struct shim_vma_val vma;
+            lookup_vma(ALLOC_ALIGN_DOWN_PTR(user_stack_addr), &vma);
+            self->stack_top = vma.addr + vma.length;
+            self->stack_red = vma.addr;
+            self->stack     = vma.addr;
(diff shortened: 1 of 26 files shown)


dimakuv:
Jenkins, retest Jenkins-SGX-18.04-apps please (internet connectivity)

dimakuv:
Jenkins, retest Jenkins-SGX-18.04 please (`exec` LibOS test failed, cannot reproduce locally)

mkow:
Jenkins, retest Jenkins-Debug please (`apps.LTP.lseek01` timed out)

mkow:
Jenkins, retest Jenkins-SGX-apps please

This failure looks weird. [The log](https://leeroy.cs.unc.edu/job/graphene-sgx-apps/779/consoleFull) says that Blender finished rendering, but also timed out:
```
Saved: '/data/images/simple_scene.blend0001.png'
 Time: 00:12.84 (Saving: 00:00.47)


Blender quit
Cancelling nested steps due to timeout
Sending interrupt signal to process
Makefile:72: recipe for target 'check' failed
make: *** [check] Terminated
Terminated
script returned exit code 143
```
CC: @boryspoplawski

PR #1399, opened 2020-04-02
[Pal/lib] Build in-Graphene mbedTLS crypto lib with minimal config

dimakuv:
## Description of the changes 

Graphene now moved to mbedTLS version 2.21.0. This new version decouples the crypto functionality in a separate lib (libmbedcrypto). This separate lib has its C code and headers under mbedtls/crypto subdirectory. Previously, we forgot to copy our in-Graphene minimal config `config.h` into this new subdirectory. This meant that the crypto lib was built with a default and bloated configuration, not needed for our Graphene purposes. This commit fixes this.

## How to test this PR? 

Check out `graphene/Pal/lib/crypto/mbedtls/crypto/include/mbedtls/config.h` after Graphene build. It must contain our minimal configuration.

PR #1400, opened 2020-04-02
[Pal/Linux-SGX] Encrypt all pipes/socketpairs with TLS-PSK

### LINKED ISSUE #1235: [Linux-SGX] Encrypted IPC
This is a design document for encrypted IPC in Graphene-SGX.

# General

Previously, Graphene-SGX performed all Inter-Process Communication (between SGX enclaves) in plaintext. In particular, these IPC mechanisms are vulnerable to attacks due to lack of integrity and encryption:

1. Process checkpoint send/receive.
2. Process inheritance of host resources (file descriptors in Linux-SGX), implemented via [sendmsg/recvmsg](https://sumitomohiko.wordpress.com/2015/09/24/file-descriptor-passing-with-sendmsg2-and-recvmsg2-over-unix-domain-socket/).
3. Communication over pipes and socketpairs.
4. Communication over multicast (*Update: this is removed from Graphene completely*).

We want to have robust, minimal, encrypted and integrity-verified IPC communication. Design principles are:
1. Encapsulate all modifications in Linux-SGX PAL. There is no reason to introduce it at other levels (LibOS, generic PAL, other non-secure PALs).
2. Use a minimalistic crypto mechanism. We must use established crypto instead of implementing our own ad-hoc mechanism (read: TLS).
3. Re-use already-existing mutual local SGX attestation between enclaves (see #332). In other words, use TLS-PSK (Pre-Shared Keys).
4. Start small.

# Implementation

We currently build on mbedTLS-provided TLS-PSK (without DH) chipersuite: `MBEDTLS_TLS_PSK_WITH_AES_128_GCM_SHA256`. Since every pair of SGX enclaves creates a new shared key via SGX local attestation, there is no need for additional forward secrecy.

- [x] **Process checkpoint send/receive.** Implemented in PR #1229.
- [ ] **Process inheritance of host resources.** This is probably not needed because the host OS already knows basically all metadata on host resources. But we must carefully sanitize PAL handle metadata sent from parent to child (see #1229 for an example of shared-key sanitization).
- [ ] **Communication over pipes and socketpairs.** In Graphene, pipes are emulated via socketpairs, so implementation boils down to hardening socketpairs. Two issues: (1) cannot share socketpairs among 3+ enclaves; (2) need to establish new TLS-PSK connection when host socketpair is inherited via fork/clone. *Need to discuss!*
- [x] **Communication over multicast.** Could this be simply removed from Graphene?

# Crypto Details

- We use the current implementation of Pal-lib's crypto adapters. At this point, it starts dangerously growing in complexity. Not sure if we need to have a clearer way of introducing TLS-PSK.

- We cannot use host-based sources of entropy (e.g. `dev/random`) because the host is malicious. The only source of entropy then is whatever is provided by the CPU. We use the following in mbedTLS:
```
#define MBEDTLS_ENTROPY_HARDWARE_ALT  /* use mbedtls_hardware_poll() */
#define MBEDTLS_NO_PLATFORM_ENTROPY  /* do not use host-based sources */

/* mbedTLS entropy logic expects this function where we use rdrand x86 instruction */
int mbedtls_hardware_poll(void* data, unsigned char* output, size_t len, size_t* olen) {
    for (size_t i = 0; i < len; i += 8) {
        unsigned long long rand64;
        while (__builtin_ia32_rdrand64_step(&rand64) == 0) /*nop*/;
        memcpy(output + i, &rand64, sizeof(rand64));
    }
    *olen = len;
}
```

- We do not use TLS reset or renegotiation. Renegotiation is known to be vulnerable and is excluded from TLS 1.3. Reset is not needed since we expect a benign TLS session between two SGX enclaves on the same machine to never fail.

- Currently, after the checkpoint is sent/received, the parent/child processes downgrade their sockets to plain non-secure ones (by disabling TLS). This is done because child may spawn a grandchild that also wants to communicate with grandparent, but it is impossible for multiple processes to share same TLS context. *TODO: how to fix this and always have encrypted IPC?*

- PSK identity is currently hard-coded to `"dummy"`. Since the pre-shared key is always unique to a pair of SGX enclaves, I don't see a need for a dynamic PSK identity. Maybe we should reconsider this decision when we figure out how to allow 3+ enclaves to share a socketpair (then a server enclave will have two PSK identities, one for each of client enclaves).

dimakuv:
## Description of the changes 

Previously, Linux-SGX PAL did not encrypt pipe/socketpair communication (only process checkpoint send/receive was encrypted). This PR encrypts all pipe/socketpair IPC between threads of
the same enclave and between enclave processes. In particular, all offsprings of the "first" enclave inherit the same master key and derive IPC session keys from this master key based on pipe ID. When two pipe/socketpair endpoints are first created, they establish a TLS-PSK session via intra-enclave handshake (requires a spawn of an intermediate enclave thread). During clone/fork/exec, endpoints' TLS contexts are serialized and sent to the child that deserializes them (using `mbedtls_ssl_context_{save,load}` functions).

Note that multicast pipes (with more than two communicating entities) are not supported since TLS protocol doesn't support it.

This commit modifies the PAL `SendHandle` test to correctly test pipe communication, as well as adds the LibOS `pipe` test.

Fixes #1235. See it for details.

It is better to merge #1399 before this one.

## How to test this PR? 

PAL `SendHandle` test and LibOS `pipe` test are modified/added. All other tests also must work correctly. You can test that pipe communication is indeed encrypted like this: `SGX=1 strace -ff -s 500 ./pal_loader ./pipe`. Without this PR, you will send send/recv syscalls with plaintext "Hello from write end". With this PR, these syscalls will send encrypted TLS messages.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1400)

### DIFF
--- a/LibOS/shim/test/regression/.gitignore
+++ b/LibOS/shim/test/regression/.gitignore
@@ -44,6 +44,7 @@
 /mprotect_file_fork
 /multi_pthread
 /openmp
+/pipe
 /poll
 /poll_many_types
 /ppoll
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -35,6 +35,7 @@ c_executables = \
 	mprotect_file_fork \
 	multi_pthread \
 	openmp \
+	pipe \
 	poll \
 	poll_many_types \
 	ppoll \
--- a/LibOS/shim/test/regression/pipe.c
+++ b/LibOS/shim/test/regression/pipe.c
@@ -1,0 +1,47 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+int main(int argc, char** argv) {
+    int pipefds[2];
+    char buffer[1024];
+    size_t bufsize = sizeof(buffer);
+
+    if (pipe(pipefds) < 0) {
+        perror("pipe error");
+        return 1;
+    }
+
+    int pid = fork();
+
+    if (pid < 0) {
+        perror("fork error");
+        return 1;
+    } else if (pid == 0) {
+        /* client */
+        close(pipefds[1]);
+
+        if (read(pipefds[0], &buffer, bufsize) < 0) {
+            perror("read error");
+            return 1;
+        }
+        buffer[bufsize - 1] = '\0';
+
+        printf("read on pipe: %s\n", buffer);
+    } else {
+        /* server */
+        close(pipefds[0]);
+
+        snprintf(buffer, bufsize, "Hello from write end of pipe!");
+        if (write(pipefds[1], &buffer, strlen(buffer) + 1) < 0) {
+            perror("write error");
+            return 1;
+        }
+
+        wait(NULL); /* wait for child termination, just for sanity */
+    }
+
+    return 0;
+}
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -480,6 +480,10 @@ class TC_80_Socket(RegressionTestCase):
         self.assertIn('pselect() on write event returned 1 file descriptors', stdout)
         self.assertIn('pselect() on read event returned 1 file descriptors', stdout)
 
+    def test_090_pipe(self):
+        stdout, _ = self.run_binary(['pipe'], timeout=60)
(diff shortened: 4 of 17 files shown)


dimakuv:
Jenkins, retest Jenkins-SGX-apps please (internet connectivity on LMBench)

PR #1401, opened 2020-04-03
LibOS: Prevent signal handler from disappearing due to SA_RESETHAND

stefanberger:
This patch prevents a signal handler that has the SA_RESETHAND flag set
from disappearing when a signal is appended. We do this by passing an
additional boolean to the __get_sighandler() function that allows us to
control whether the signal handler may be reset or not. It is only
desirable to reset the signal handler if an actual invocation of the
handler function occurs, not when only asking 'where is the handler?'

I noticed signals not being handled anymore while porting Graphene to PowerPC64.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1401)

### DIFF
--- a/LibOS/shim/src/bookkeep/shim_signal.c
+++ b/LibOS/shim/src/bookkeep/shim_signal.c
@@ -668,7 +668,7 @@ __sigset_t * set_sig_mask (struct shim_thread * thread,
     return &thread->signal_mask;
 }
 
-static __rt_sighandler_t __get_sighandler(struct shim_thread* thread, int sig) {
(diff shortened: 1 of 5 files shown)


dimakuv:
Jenkins, test this please

mkow:
Jenkins, retest this please

mkow:
Jenkins, retest Jenkins-SGX please (`test_libos.TC_01_Bootstrap.test_600_multi_pthread` timed out)

dimakuv:

PR #1407, opened 2020-04-08
[LibOS,Pal] Replace 32-bit pipeid with 256-bit pipe name

dimakuv:
## Description of the changes 

Previously, Graphene used the notation "pipe:<uint32_t>" to emulate pipes, socketpairs, and UNIX domain sockets. In particular, pipes and socketpairs received random integer IDs, and sockets received deterministic integer IDs. However, 32-bit IDs are bad for crypto. This commit changes pipe IDs (pipeid) from `uint32_t` to `char[96]` that can hold a 256-bit random sequence sufficient for crypto.

This is a prerequisite for #1400.

## How to test this PR? 

All tests must pass. If you do smth like `strace -ff Runtime/pal_loader Pal/regression/Pipe`, you'll see much longer names now.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1407)

### DIFF
--- a/LibOS/shim/include/shim_handle.h
+++ b/LibOS/shim/include/shim_handle.h
@@ -127,11 +127,7 @@ struct shim_dev_handle {
 };
 
 struct shim_pipe_handle {
-#if USE_SIMPLE_PIPE == 1
-    struct shim_handle* pair;
-#else
-    IDTYPE pipeid;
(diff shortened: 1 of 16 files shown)


dimakuv:
Jenkins, test this please

PR #1423, opened 2020-04-14
[LibOS,Pal] Support Glibc logic of line-buffered/fully-buffered STDOUT

dimakuv:
## Description of the changes 

Glibc has special handling of STDOUT: it checks whether STDOUT is TTY (app writes to a terminal) or non-TTY (app writes to a pipe/file) and makes STDOUT line-buffered or fully-buffered respectively. Glibc uses a trick with IOCTL(TCGETS) to find out whether it is TTY or non-TTY. Previously, Graphene always returned `-EINVAL` on this IOCTL, so STDOUT in Graphene was always fully-buffered. This inconsistency led to e.g. `close(STDOUT)` without printing out buffered strings.

This PR fixes this by propagating info on TTY/non-TTY from the host.

## How to test this PR? 

Unfortunately, I didn't find a way to test it in our LibOS tests. Our PyTest framework redirects STDOUT to a pipe (no surprises here), so STDOUT is always detected to be non-TTY and thus fully-buffered. I didn't come up with a way to force PyTest to make STDOUT line-buffered.

One can manually test it though. Just add `close(STDOUT_FILENO)` to `shared_object.c` LibOS test and run. The original app will output `Hello World`; old Graphene will not output anything; and Graphene with this patch will correctly output `Hello World`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1423)

### DIFF
--- a/LibOS/shim/src/sys/shim_ioctl.c
+++ b/LibOS/shim/src/sys/shim_ioctl.c
@@ -41,7 +41,7 @@
 #define TERM_DEFAULT_CFLAG (B38400 | CS8 | CREAD)
 #define TERM_DEFAULT_LFLAG (ICANON | ECHO | ECHOE | ECHOK | ECHOCTL | ECHOKE | IEXTEN)
 
-static int ioctl_termios(struct shim_handle* hdl, unsigned int cmd, unsigned long arg) {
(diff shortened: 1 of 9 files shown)


dimakuv:
This is too much code and hacks for a non-essential feature. Closing it until some app will really need this.

PR #1440, opened 2020-04-21
[DO NOT MERGE] Make sched_yield() no-op for performance testing

dimakuv:
## Description of the changes 

**DO NOT MERGE**.

This PR is to test the performance of Graphene-SGX on multi-threaded and synchronization-heavy workloads like OpenVINO. We'll also report the results here.

This PR simply stubs `sched_yield()` to be a no-op. (Previously, Graphene-SGX would perform an OCALL to do a host-level `sched_yield()`.)


dimakuv:
Closing this PR but keeping the branch for further perf testing.

PR #1443, opened 2020-04-22
[Pal] Force PIE executable to be located at address 0x555555554000

dimakuv:
## Description of the changes 

Previously, we forced PIE executables to be located at address `0x00400000` (4MB). However, there is a bug in the ELF relocation code in LibOS that leads to double-relocation. To circumvent this bug,
relocation code checks if the offset was already relocated. But if the offset itself exceeds the base address of PIE executable (i.e., exceeds 4MB), then the offset is not relocated and segfaults follow.

This PR changes base address from `0x00400000` to `0x555555554000`, similar to what Linux does.

## How to test this PR? 

Run NodeJS example manually. It will work now.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1443)

### DIFF
--- a/Pal/src/pal_internal.h
+++ b/Pal/src/pal_internal.h
@@ -316,8 +316,8 @@ int _DkAttestationQuote(PAL_PTR user_report_data, PAL_NUM user_report_data_size,
         _DkProcessExit(exitcode);                                                       \
     } while (0)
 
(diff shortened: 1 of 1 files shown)


dimakuv:
Jenkins, retest Jenkins please (Glibc build failed with weird error)

PR #1454, opened 2020-04-29
[LibOS] Add support for FIFOs (named pipes)

dimakuv:
## Description of the changes 

This PR adds support for FIFOs and the corresponding syscalls `mknod()` and `mknodat()`. Internally, FIFOs are emulated as pseudo-files in chroot mount points (not visible in host FS). FIFOs' read/write operations are emulated via pipes at PAL level (this means that they are transparently encrypted under SGX PAL). Generally, emulation of FIFOs is similar to emulation of named UNIX domain sockets, i.e., they are "ephemeral" and only allow communication between two related processes. FIFOs cannot be O_RDWR (opening fails with EINVAL).

FIFOs can be blocking and non-blocking. The only caveat with non-blocking FIFOs is that `open(O_WRONLY)` doesn't fail with ENXIO if the other end is not opened yet (but this is a rare corner-case probably not used by real-world apps).

## How to test this PR? 

New LibOS test is added: multi-process `mkfifo`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1454)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -525,6 +525,7 @@ extern struct shim_d_ops str_d_ops;
 
 extern struct shim_mount chroot_builtin_fs;
 extern struct shim_mount pipe_builtin_fs;
+extern struct shim_mount fifo_builtin_fs;
 extern struct shim_mount socket_builtin_fs;
 extern struct shim_mount epoll_builtin_fs;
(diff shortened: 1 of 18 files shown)


dimakuv:
Jenkins, retest Jenkins-Debug please (LMBench internet connectivity)

dimakuv:
Jenkins, retest Jenkins please (`setrlimit02` LTP test failed)

PR #1475, opened 2020-05-07
[LibOS] Disable reinitialization of brk after fork

### LINKED ISSUE #1472: Crashes due to ASLR on the brk (noticed on ppc64)
I noticed that nginx wasn't working reliably anymore on ppc64 port. The reason was that there is randomization occurring on the brk that causes the program to fail. I do not know whether the same problem occurs on x86_64 since I am only running a limited amount of programs there. Though here's the patch that I am using for solving the issue on powerpc64:

```
iff --git a/LibOS/shim/src/sys/shim_brk.c b/LibOS/shim/src/sys/shim_brk.c
index 719e77c0..89c76171 100644
--- a/LibOS/shim/src/sys/shim_brk.c
+++ b/LibOS/shim/src/sys/shim_brk.c
@@ -62,7 +62,7 @@ int init_brk_region(void* brk_start, size_t data_segment_size) {
             && brk_max_size <= (uintptr_t)PAL_CB(user_address.end)
             && (uintptr_t)brk_start < (uintptr_t)PAL_CB(user_address.end) - brk_max_size) {
         size_t offset = 0;
-#if ENABLE_ASLR == 1
+#if ENABLE_ASLR == 1 && !defined(__powerpc64__)
         int ret = DkRandomBitsRead(&offset, sizeof(offset));
         if (ret < 0) {
             return -convert_pal_errno(-ret);
@@ -72,6 +72,8 @@ int init_brk_region(void* brk_start, size_t data_segment_size) {
         offset %= MIN((size_t)0x2000000,
                       (size_t)((char*)PAL_CB(user_address.end) - brk_max_size - (char*)brk_start));
         offset = ALLOC_ALIGN_DOWN(offset);
+#else
+        int ret;
 #endif
         brk_start = (char*)brk_start + offset;
 
@@ -86,8 +88,10 @@ int init_brk_region(void* brk_start, size_t data_segment_size) {
             return ret;
         }
     } else {
+#if !defined(__powerpc64__)
         /* Let's try mapping brk anywhere. */
         brk_start = NULL;
+#endif
     }
 
     if (!brk_start) {
```

I don't think the code is right. With ASLR enabled we basically will always get a random brk but we cannot have this after a fork()!

Either we get randomization in the if branch: https://github.com/oscarlab/graphene/blob/master/LibOS/shim/src/sys/shim_brk.c#L61

Or then via brk_start in the else branch: https://github.com/oscarlab/graphene/blob/master/LibOS/shim/src/sys/shim_brk.c#L88

You may be lucky if your architecture in general or a particular program doesn't use brk, but if it does it may not work (correctly).

boryspoplawski:
## Description of the changes 
Fixes #1472

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1475)

### DIFF
--- a/LibOS/shim/src/sys/shim_brk.c
+++ b/LibOS/shim/src/sys/shim_brk.c
@@ -37,9 +37,20 @@ static struct {
     char* brk_end;
 } brk_region;
 
-static struct shim_lock brk_lock;
+static struct shim_lock brk_lock = { .lock = NULL };
 
 int init_brk_region(void* brk_start, size_t data_segment_size) {
+    if (!create_lock(&brk_lock)) {
+        debug("Creating brk_lock failed!\n");
+        return -ENOMEM;
+    }
+
+    /* TODO: this needs a better fix. Currently after fork, in the new child process, `shim_init`
+     * is run, hence this function too - but forked process will get its brk from checkpoints. */
+    if (brk_region.brk_start) {
+        return 0;
+    }
+
     size_t brk_max_size = DEFAULT_BRK_MAX_SIZE;
     data_segment_size = ALLOC_ALIGN_UP(data_segment_size);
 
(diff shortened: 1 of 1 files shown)


dimakuv:
Jenkins, retest Jenkins-SGX please (`sighandler_reset` LibOS test failed)

dimakuv:
Jenkins, retest Jenkins-Debug please (LMBench internet connectivity)

PR #1484, opened 2020-05-11
[Pal] Properly calculate address of futex variable

stefanberger:
Rather than passing the mutex_handle's address, which happens to have the
futex variable as its first variable, properly calculate the address of the
futex.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1484)


mkow:
Jenkins, test this please

dimakuv:
Jenkins, test this please

PR #1487, opened 2020-05-11
[Pal] Fix more IPv6 related stack buffer overflows

stefanberger:
This series of patches fixes more IPv6 related stack buffer overflows . We first add checks to the size of the passed buffers, which requires each caller to pass in the size of the buffer. If the buffers are too small for what they are used for, we fail the function before touching the buffer. Adapt all callers and extend their buffers by using sockaddr_storage, which also fits a sockaddr_in6.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1487)


mkow:
Jenkins, retest this please

mkow:
Jenkins, retest this please

PR #1491, opened 2020-05-11
[Pal] Move page size definitions to new pal_user.h

stefanberger:
Move page size definitions to a new pal_user.h , which would be similar to sys/user.h where this definition would typically be found.

Define PAGE_SIZE and use it in a regression test that maps memory of a page.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1491)


mkow:
Jenkins, retest this please

PR #1501, opened 2020-05-14
Use proper integer types in socket-related syscalls and ocalls

mkow:
## Description of the changes 

This contains two socket-related changes, which would be hard to factor out into two commits: (because of compilation warnings)

- Making ocalls use `size_t` in their interface. It's not the same as Linux syscall interface anyway, so why not make it sane?
- Fixing sycalls arguments types to match the ones use by kernel. Note: there's no such type as `socklen_t` in the kernel, so it got removed.

This is an SGX follow-up to #1487. It's also a small step towards resolving #152.

## How to test this PR? 

Just Jenkins, I guess.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1501)


mkow:
Jenkins, retest Jenkins please (LTP timed out, known issue)

PR #1503, opened 2020-05-14
[Pal] Relax addrlen check to require buffer to be 'large enough'

stefanberger:
Relax the addrlen check in inet_create_uri to only ensure that the passed buffer is large enough but doesn't need to be precise the needed size. This addresses the caller from udp_receivebyaddr that calls us like this:

```C
struct sockaddr_storage conn_addr;
socklen_t conn_addrlen = sizeof(conn_addr);

struct msghdr hdr;
struct iovec iov;
iov.iov_base       = buf;
iov.iov_len        = len;
hdr.msg_name       = &conn_addr;
hdr.msg_namelen    = conn_addrlen;
[...]

int ret = inet_create_uri(addr_uri, addr + addrlen - addr_uri,
                          (struct sockaddr*)&conn_addr, hdr.msg_namelen);
```

hdr.msg_namelen indicates a buffer that is bigger than needed, so we should be able to accept it.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1503)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_sockets.c
+++ b/Pal/src/host/Linux-SGX/db_sockets.c
@@ -153,7 +153,7 @@ static int inet_create_uri(char* uri, int count, struct sockaddr* addr, size_t a
     int len = 0;
 
     if (addr->sa_family == AF_INET) {
-        if (addrlen != sizeof(struct sockaddr_in))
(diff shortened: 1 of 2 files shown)


mkow:
Jenkins, retest this please

PR #1557, opened 2020-06-05
[PAL] Refactor code getting bogomips from host's proc filesystem

stefanberger:
Refactor the code for getting bogomips from the host's proc filesystem so that some functions can be reused for other architectures that will parse the same type of file.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1557)


dimakuv:
Jenkins, test this please

Issue #1561, opened 2020-06-07
[Linux-SGX] Children argv goes through untrusted host


mkow:
Currently Linux-SGX PAL passes arguments to newly spawned children through host's execve, which is not secure. Linux-SGX should be changed to work the same as Linux PAL, which passes them inside the checkpoint (which is encrypted on Linux-SGX).

dimakuv:
> Linux-SGX should be changed to work the same as Linux PAL, which passes them inside the checkpoint

The part about "Linux PAL passes argv inside the checkpoint" is incorrect. Linux PAL works the same as Linux-SGX PAL where it simply passes argv during `execve()`:
```c
// db_process.c
static int __attribute_noinline child_process(struct proc_param* proc_param) {
    int ret = ARCH_VFORK();
    if (ret)
        return ret;

    /* child */
    ...
    int res = INLINE_SYSCALL(execve, 3, PAL_LOADER, proc_param->argv, linux_state.environ);
    ...
}
```

mkow:
> The part about "Linux PAL passes argv inside the checkpoint" is incorrect.

I think I actually got this info from you ;) Anyways, should we fix both PALs? Fixing it in Linux PAL doesn't make that much sense, but OTOH doing the change in both would make the PALs more consistent.

dimakuv:
Yeah, I'm fixing it in both PALs. But first I need to refactor the checkpoint/migrate code...

Issue #1053, opened 2019-10-01
RFC: graphene invocation


woju:
# Introduction

I'd like to propose a unified way of running files in graphene. This will certainly be a very visible, very incompatible change, so I'd like to write some explanation and seek review, so we can get this right. Also, I'm not that knowledgeable about the architecture, so I apologize in advance if i had some misconceptions.

### Requirements
- command line invocations (hard)
- invocation from shebang's manifest (nice to have)
- easy way of running under GDB (nice to have, developers love it, so probably a must)

### Status quo ante
Today it is mainly done with `Runtime/pal_loader` script. Script 

### Related work
- #895 
- that CMake/meson/whatever effort to rewrite buildsystem, which I can't find the issue number for
- #994 

# the proposal (DRAFT)

After proper installation (`apt-get install` or `make DESTDIR=/ install`), `/usr/bin/graphene` and `/usr/bin/graphene-sgx` are binaries or symlinks. No `Linux` or `FreeBSD`, since user knows what the `uname` is or should be. In both cases the interface is the same:

1. `graphene[-sgx] [--gdb] {--manifest|-f} MANIFEST`, and manifest contains path to executable and argv
2. `graphene[-sgx] [--gdb] {--manifest|-f} MANIFEST [--] EXECUTABLE [ARGS ...]`, and the manifest contains neither executable nor argv
3. `graphene[-sgx] [--gdb] [--] EXECUTABLE [ARGS]` for the cases the manifest is not needed

This makes it possible to use it from manifest's shebang like awk: `#!/usr/bin/graphene -f`. If the manifest is invoked directly (`./manifest`), it is not possible to automatically run GDB, but as always, it is possible to use `graphene --gdb -f ./manifest`.

It is an error to specify arguments twice, both in command line and in manifest, even if they match.

Environment variables are not accepted.

# potential problems

### shebangs under FreeBSD
*BSD people don't like putting alien binaries in `/usr/bin`, and would likely prefer to have those binaries installed in `/usr/local/bin`. This will break shebangs as proposed. The Python way of writing `#!/usr/bin/env python3` (`#!/usr/bin/env graphene -f`) would not work, since `graphene -f` would be a single argument to `env` (not two arguments: `graphene` and `-f`). One solution would be to follow the Perl way of installing a symlink `/usr/bin/graphene` → `/usr/local/bin/graphene`.

# to consider
[...]

Issue #870, opened 2019-07-26
Use YAML or TOML syntax for manifests


bigdata-memory:
YAML is easier for humans to read and write than other common data formats, it is also flexible enough to accommodate much more complex structure of configuration. Further, there are libraries available in most programming languages for working with YAML.

dimakuv:
Just an additional note: the manifest syntax should also be able to unset envvars. Here is an example for Redis:
```
# Explicitly unset all environment variables used by Redis. This is just a
# precaution so that attacker cannot influence the default behavior of Redis
# by setting unexpected environment variables.
#
# TODO(dimakuv): Graphene's PAL incorrectly handles this currently. See
# `read_environments()`, the end of function. Below strings create envvars
# with empty strings, which forces Redis to do incorrect things like thinking
# it was started by systemd/upstart (due to UPSTART_JOB=="").
#loader.env.UPSTART_JOB =
#loader.env.NOTIFY_SOCKET =
```

woju:
TOML would be better:
- It is closer to current syntax (AFAIU, current files would only need quotes around strings to be compatible).
- It is simpler to parse. YAML has many features, most of them rarely used (~unneeded) and sometimes even underspecified, which leads to differences between dialects.

https://github.com/toml-lang/toml
https://en.wikipedia.org/wiki/TOML

dimakuv:
+1 for TOML. Looks very simple and clear.

dimakuv:
It was agreed to use TOML and migrate to it completely (removing the old ad-hoc parser). I am modifying the existing https://github.com/cktan/tomlc99 parser for our specific Graphene needs (no deps on Glibc, no setlongjumps, C-style error handling) and will submit a PR soon.

Issue #508, opened 2019-03-11
Provide a way to pass verified argv and environments


HW42:
Currently the process arguments and environment variables are passed from `pal_loader` without any verification.

There should be a way to specify those in the manifest such that those are measured and can be trusted. This is particular useful for running (mostly) unmodified applications inside the enclave.

mkow:
We have a fix for this in Graphene-ng, but it requires protected files to make sense (it adds an option to manifest to load argv from a file, which can be mounted as protected). We could also add third option: specify argv in the manifest itself, but this would require a lot of changes in manifest parsing to allow providing arbitrary argv and doesn't allow to provide the arguments in runtime (instead of build-time).

We could merge what we have as is, because it doesn't depend on protected files per se, but it would only start to be useful after also merging filesystem protection. Any thoughts?

dimakuv:
I suggest you submit a PR with your solution, and protected-files feature in another PR in the future. At least we'll have a better understanding how your solution works.

We are talking only about argv or envvars are also handled by this mechanism?

Is it possible to pass argv in a special envvar which will be protected somehow? (To avoid the dependency on protected-files.)

mkow:
> I suggest you submit a PR with your solution, and protected-files feature in another PR in the future. At least we'll have a better understanding how your solution works.

Ok, I'll wait a bit more to give everyone a chance to reply on this and prepare the PR afterwards.

> We are talking only about argv or envvars are also handled by this mechanism?

We used this only for argv, but adding the same thing for env sounds trivial, the code should be almost identical (and I plan to include both in the PR).

> Is it possible to pass argv in a special envvar which will be protected somehow? (To avoid the dependency on protected-files.)

Isn't protecting envs exactly as hard as protecting argv?

dimakuv:
> Isn't protecting envs exactly as hard as protecting argv?

Then ignore my question :)

PR #1562, opened 2020-06-07
Protected argv

mkow:
## Description of the changes 

_This is an RFC PR. After we converge on the final version I'll implement a similar PR for envs._

This PR implements a way to pass runtime arguments from a file, with intention to use it in conjunction with either trusted or protected files.

Unfortunately, argv passing is quite ugly because of `loader.exename` and very limited manifest syntax. After fixing #870 and #1053 the implementation and usage could be simplified.

One known issue with the current code is that argv of children goes through the untrusted host under Linux-SGX PAL (through `execve` arguments, see #1561).

This PR is a breaking change and there's (AFAIK) no better way to implement it - the old manifest syntax was insecure-by-default (allowing unstrusted args). Users will need to use `loader.insecure__use_cmdline_argv` if they don't want to use the file arguments.
For apps which are intended to run without any special arguments, the recommended way is to set `loader.execname` and don't pass any additional arguments. This will get a bit more intuitive after allowing inline argv in the manifest (see a comment in the code).

This resolves a half of #508.

## How to test this PR? 

New LibOS test included. Also, verify if this doesn't break any examples you use but aren't tested in Jenkins (or aren't committed to the repo).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1562)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -60,6 +60,28 @@ is run via the manifest (e.g. ``./app.manifest arg1 arg2 ...``). If the string
 is not specified in the manifest, the PAL will use the path to the manifest
 itself (standard UNIX convention).
 
+Command-Line Arguments
+^^^^^^^^^^^^^^^^^^^^^^
+
+::
+
+   loader.insecure__use_cmdline_argv = 1
+
+or
+
+::
+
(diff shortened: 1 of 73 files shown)


mkow:
Jenkins, retest Jenkins-Debug-18.04 please (fork11 failed, known issue)

PR #1542, opened 2020-05-29
[LibOS,Pal] Emulate SIGPIPE via creating and sending this signal on EPIPE

dimakuv:
## Description of the changes 

Previously, Graphene simply forwarded SIGPIPE generated by the host to LibOS/app. Unfortunately, SIGPIPE generation is a process-wide feature and there is no portable way to restrict it only to a subset of pipes, UNIX domain sockets, etc. This led to sporadic Graphene failures because Graphene's internal use of pipes and sockets may result in an unexpected (to application) SIGPIPE.

This commit removes the forwarding of SIGPIPE. Instead, PALs explicitly ignore SIGPIPE. This forces the host to return EPIPE error code, which is checked only on a subset of LibOS handles (the ones created by the app), and if required, LibOS generates a SIGPIPE for the application.

While adding this logic, the whole PAL exception code was refactored, both in Linux and Linux-SGX. 

Fixes https://github.com/oscarlab/graphene/issues/1534.

## How to test this PR? 

Tests for SIGPIPE are now enabled for both Linux and Linux-SGX PALs. In particular, `sighandler_sigpipe` LibOS regression test.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1542)

PR #1576, opened 2020-06-11
[PAL] Handle SIGTERM in application

stefanberger:
The dummy server in python-simple does not terminate upon SIGTERM
anymore. This patch fixes this by having the application handle the
signal.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1576)

### DIFF
--- a/Pal/src/host/Linux/db_exception.c
+++ b/Pal/src/host/Linux/db_exception.c
@@ -176,7 +176,7 @@ static void handle_async_signal(int signum, siginfo_t* info, struct ucontext* uc
 
     uintptr_t rip = uc->uc_mcontext.gregs[REG_RIP];
 
-    if (!ADDR_IN_PAL(rip)) {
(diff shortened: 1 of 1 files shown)


stefanberger:
@dimakuv You may know more about this. Previously we had the following at the end of the `_DkTerminateSighandler`:
```
-    // Call the event handler. If there is no handler, terminate the thread
-    // unless it is a resuming event (then ignore the event).
-    if (!_DkGenericSignalHandle(event_num, NULL, uc) && event_num != PAL_EVENT_RESUME)
-        _DkThreadExit(/*clear_child_tid=*/NULL);
```

dimakuv:
I debugged the issue (thanks to @stefanberger for noticing this!). Here is the correct fix: https://github.com/oscarlab/graphene/pull/1579. It was my stupid mistake in a previous PR on signal handling in PAL (https://github.com/oscarlab/graphene/pull/1542) which led to infinite loop.

PR #1555, opened 2020-06-04
[Testing - DO NOT MERGE] Test fsgsbase patch in Jenkins

donporter:
## Description of the changes 

This is just for testing.  This change disables the checks for the gsgx module, and runs SGX jobs on a host with the patched 5.7 +FSGSBASE patches.

## How to test this PR? 

In Jenkins.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1555)

PR #1583, opened 2020-06-15
[SGX] Dynamically determine whether FSGSBASE is supported

donporter:
## Description of the changes 

Add support to handle either the gsgx module or the proposed Linux kernel support for fsgsbase.  

This is a more polished version of the code originally proposed by @dimakuv . 

## How to test this PR? 

I am also testing these changes on a patched kernel in #1555, but wanted to separate out this change from other, incidental changes to make Jenkins work.  This should be backward compatible with code that uses gsgx as well.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1583)


dimakuv:
I'll take over this PR.

PR #1594, opened 2020-06-18
[Pal/Linux-SGX] Support DCAP 1.35 driver

jrbergsten:
## Description of the changes 

This PR adds support for the DCAP 1.35 driver (https://github.com/intel/SGXDataCenterAttestationPrimitives/tree/e3998c5fbd5d3fd065436b934b53822120e67be6/driver/linux). The PR makes changes to mmap and mprotect to support newer protection setting requirements. Supposed to be backward compatible with prior releases of other Intel SDK drivers.

More context: the DCAP driver (https://github.com/intel/SGXDataCenterAttestationPrimitives/tree/master/driver/linux) follows the upstreaming Intel SGX driver for Linux (https://lkml.org/lkml/2020/6/1/165). It was recently rebased to v32 of the upstreaming driver (see links), which has changes to mmap and mprotect. Thus, Graphene failed to correctly create, add-pages, and init the SGX enclave. This PR fixes these problems.

## How to test this PR?

Run any Graphene app using SGX and DCAP 1.35.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1594)

PR #1595, opened 2020-06-18
Jrbergsten/sgx signer app versus python

jrbergsten:
## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1595)

### DIFF
--- a/Pal/src/host/Linux-SGX/sgx_framework.c
+++ b/Pal/src/host/Linux-SGX/sgx_framework.c
@@ -159,7 +159,12 @@ int create_enclave(sgx_arch_secs_t * secs,
 
     uint64_t addr = INLINE_SYSCALL(mmap, 6, secs->base, secs->size,
                                    PROT_NONE, /* newer DCAP driver requires such initial mmap */
+#ifdef SGX_DCAP_16_OR_LATER
+                                   (MAP_FIXED + MAP_PRIVATE + MAP_ANONYMOUS), g_isgx_device, 0);
+                                   /* DCAP 1.5 requires private and anonymous */                                   
+#else
                                    flags|MAP_FIXED, g_isgx_device, 0);
+#endif
 
     if (IS_ERR_P(addr)) {
         if (ERRNO_P(addr) == 1 && (flags | MAP_FIXED))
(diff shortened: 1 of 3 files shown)


dimakuv:
Leaving only a similar PR #1594. Closing this one.

dimakuv:
___
*[Pal/src/host/Linux-SGX/signer/pal-sgx-get-token, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/1595#-MA8CAuvEtUDAqvGSXX_:-MA8aw4jEHx9WkaJT_Ou:brrdnvr) ([raw file](https://github.com/oscarlab/graphene/blob/231a43c3c6031f8b2a0a06ff8f09476bb439592b/Pal/src/host/Linux-SGX/signer/pal-sgx-get-token#L1)):*

Isn't the root of the problem in the fact that someone copy-pasted protobuf output into our repo instead of generating it build-time? I'm talking about signer/aesm_pb2.py.
</blockquote></details>

No. I mean, that's also a "bad smell" and we should get rid of the auto-generated `aesm_pb2.py` and generate it on the fly.

But this particular issue with Python 3.8 being literally unable to import Protobuf. I think there is incompatibility in default versions (see e.g. https://github.com/protocolbuffers/protobuf/issues/2350). But we didn't care to look deeper into this.

jrbergsten:
___
*[Pal/src/host/Linux-SGX/signer/pal-sgx-get-token, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/1595#-MA8CAuvEtUDAqvGSXX_:-MA8jxly1NRAdjwgVQ7_:bkm98sl) ([raw file](https://github.com/oscarlab/graphene/blob/231a43c3c6031f8b2a0a06ff8f09476bb439592b/Pal/src/host/Linux-SGX/signer/pal-sgx-get-token#L1)):*

Reading this issue it sounds like this is just someone's local misconfiguration (read: old versions), not that protobufs don't work on default installation in some distros with Python 3.8.
</blockquote></details>

I just poked around and it appears that there IS a compatible Python 3.8 version of protoc with Ubuntu 20.04.  So, in theory the aesm.proto (or whatever) could be compiled as part of make as long as the compiler is installed (sudo apt-get install protobuf-compiler).  I think.

Issue #1598, opened 2020-06-19
Optimize the system time function via TSC for GSGX


bigdata-memory:
## Description of the problem
In GSGX, the operation of fetching system time is cost due to triggering gettimeofday syscall each time when GSGX workloads trying to get time info. 
There are some significant performance overhead that has already been observed. 
@sudharkrish may help to provide some evidences.

## Steps to reproduce
Run function gettimeofday() repeatedly.
the syscall can be observed via debug mode or strace.

## Expected results
Few number of gettimeofday syscalls

## Actual results
A lot of gettimeofday syscalls observed when workloads relies on it to measure traffics.

PR #1599, opened 2020-06-19
[PAL/Linux-SGX] Optimize the system time function via TSC

bigdata-memory:
## Description of the changes 

This patch optimizes the function _DkSystemTimeQuery()
by using RDTSC instead of gettimeofday ocall/syscall.
This optimization won't take effect if there is no reliable
TSC source available to use i.e. nonstop TSC.
The TSC drift has been controlled by adaptive algorithm and
precisely aligned with system clock accuracy.

Fixes https://github.com/oscarlab/graphene/issues/1598.

## How to test this PR? 
The original time and gettime testcases covered it.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1599)

### DIFF
--- a/Pal/include/arch/x86_64/cpu.h
+++ b/Pal/include/arch/x86_64/cpu.h
@@ -28,6 +28,13 @@ static inline void cpuid(unsigned int leaf, unsigned int subleaf, unsigned int w
 }
 
 
+/* this function is used to get TSC cycle */
+static inline unsigned long get_tsc(void) {
+    unsigned long lo, hi;
+    __asm__ volatile("rdtsc" : "=a" (lo), "=d" (hi));
+    return lo | (hi << 32);
+}
+
 #define CPU_RELAX() __asm__ __volatile__("rep; nop" ::: "memory")
 
 /* some non-Intel clones support out of order store; WMB() ceases to be a nop for these */
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -294,6 +294,7 @@ void pal_linux_main(char * uptr_args, uint64_t args_size,
     init_enclave_key();
 
     init_cpuid();
+    init_tsc();
 
     /* now we can add a link map for PAL itself */
     setup_pal_map(&pal_map);
--- a/Pal/src/host/Linux-SGX/db_misc.c
+++ b/Pal/src/host/Linux-SGX/db_misc.c
@@ -23,12 +23,125 @@
 #include "sgx_api.h"
 #include "sgx_attest.h"
 
+#define CPUID_LEAF_INVARIANT_TSC 0x80000007
+#define CPUID_LEAF_TSC_FREQ 0x15
+#define TSC_REFINE_INIT_TIMEOUT_USECS 10000000L
+#define TSC_REFINE_THRESHOLD_PERCENT 10
+#define TSC_REFINE_SHIFT_THRESHOLD 25
+
+static int64_t g_tsc_hz = 0;
+static int64_t g_recal_usec = 0;
+static int64_t g_recal_shift = 0;
+static int64_t g_start_tsc = 0;
+static int64_t g_start_usec = 0;
+static PAL_LOCK g_tsc_lock = LOCK_INIT;
+
+/* this function determines whether TSC is reliable to use */
+static bool is_tsc_usable(void) {
+    bool retval = false;
+    unsigned int words[PAL_CPUID_WORD_NUM];
+
+    _DkCpuIdRetrieve(CPUID_LEAF_INVARIANT_TSC, 0, words);
+    retval = (words[PAL_CPUID_WORD_EDX] & 1 << 8) && true;
+
+    return retval;
+}
+
+/* this function is used to fetch the baseline freq of TSC */
+static int64_t get_tsc_hz(void) {
+    int64_t retval = 0;
+    unsigned int words[PAL_CPUID_WORD_NUM];
+    int64_t crys_hz;
+
+    _DkCpuIdRetrieve(CPUID_LEAF_TSC_FREQ, 0, words);
+    if (words[PAL_CPUID_WORD_EBX] > 0 && words[PAL_CPUID_WORD_EAX] > 0) {
+        /* nominal frequency of the core crystal clock in kHz */
+        crys_hz = words[PAL_CPUID_WORD_ECX];
+        if (crys_hz > 0) {
+            retval = (double)crys_hz * words[PAL_CPUID_WORD_EBX] /
+                words[PAL_CPUID_WORD_EAX];
+        }
+    }
+    return retval;
+}
+
+/**
+ * Initialize the data structures used for time/date emulation.
+ */
+void init_tsc(void) {
+    if (is_tsc_usable()) {
+        g_tsc_hz = get_tsc_hz();
+    }
+}
+
 unsigned long _DkSystemTimeQuery(void) {
-    unsigned long microsec;
-    int ret = ocall_gettime(&microsec);
-    if (ret)
-        return -PAL_ERROR_DENIED;
-    return microsec;
+    unsigned long usec = 0;
+    int64_t tsc_usec = 0, tsc_cyc1, tsc_cyc2, tsc_cyc;
+    int64_t tsc_hz = g_tsc_hz, tsc_hz_recal;
+    long tsc_hz_recal_diff;
+    int ret;
+
+    if (tsc_hz > 0) {
+        _DkInternalLock(&g_tsc_lock);
+        if (g_start_tsc > 0 && g_start_usec > 0) {
+            /* calculate the TSC based time */
+            tsc_usec = g_start_usec +
+                ((double)(get_tsc() - g_start_tsc) * 1000000 / tsc_hz);
+            /* determine whether it needs to be refined */
+            if (tsc_usec < g_recal_usec +
+                (TSC_REFINE_INIT_TIMEOUT_USECS << g_recal_shift)) {
+                usec = tsc_usec;
+            }
+        }
+
+        if (!usec) {
+            /* refining the TSC freq */
+            tsc_cyc1 = get_tsc();
+            ret = ocall_gettime(&usec);
+            tsc_cyc2 = get_tsc();
+            if (!ret) {
+                /* the ocall_gettime() is a time consuming operation.   *
+                 * it includes EENTER and EEXIT instructions, our best  *
+                 * estimation is the timestamp obtained in the middle   *
+                 * time point, therefore, the tsc_cyc as baseline will  *
+                 * be calibrated precisely in this way.                 */
+                tsc_cyc = ((tsc_cyc2 - tsc_cyc1) >> 1) + tsc_cyc1;
+                if (g_start_tsc == 0 && g_start_usec == 0) {
+                    /* setup baseline data */
+                    g_start_usec = usec;
+                    g_start_tsc = tsc_cyc;
+                } else {
+                    g_recal_usec = usec;
+                    /* avoid #DIV/0! error */
+                    if ((int64_t)usec != g_start_usec) {
+                        tsc_hz_recal = (double)(tsc_cyc - g_start_tsc) *
+                            1000000 / (usec - g_start_usec);
+                        tsc_hz_recal_diff = tsc_hz_recal - tsc_hz;
+                        if (tsc_hz_recal_diff < 0)
+                            tsc_hz_recal_diff *= -1;
+                        /* the change cannot exceed 10% derail */
+                        if ((double)tsc_hz_recal_diff * 100 / tsc_hz <
+                            TSC_REFINE_THRESHOLD_PERCENT) {
+                            g_tsc_hz = tsc_hz_recal;
+                            /* reduce the chance to refine it again */
+                            if (tsc_hz_recal_diff < TSC_REFINE_SHIFT_THRESHOLD)
+                                ++g_recal_shift;
+                        }
+                    }
+                }
+            } else {
+                _DkInternalUnlock(&g_tsc_lock);
+                _DkRaiseFailure(PAL_ERROR_DENIED);
+            }
+        }
+        _DkInternalUnlock(&g_tsc_lock);
+    } else {
+        /* fallback to gettimeofday syscall */
+        ret = ocall_gettime(&usec);
+        if (ret)
+            _DkRaiseFailure(PAL_ERROR_DENIED);
+    }
+    return usec;
 }
 
 int _DkInstructionCacheFlush(const void* addr, int size) {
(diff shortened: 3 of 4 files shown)


bigdata-memory:
@yamahata Thanks for the comments, I have updated the PR according to your feedback, and the freq of TSC should be re-calibrated otherwise it will not align with wall clock precisely, thanks.

bigdata-memory:
@yamahata Let me answer your other questions here:
Why "/2"?
"microsec - _Start_usec" includes whole round trip of ecall(. i.e. both of EEXIT and EENTER.
So tsc cycle should includes both, not half.
Am I missing something?
@bigdata-memory it try to get the expected timepoint when the gettimeofday syscall get called.

How do you guaranee that time is wall clock?
@bigdata-memory I cannot say guarantee but the algorithm re-calibrates the tsc freq very well, and the draft is managed under about 50 micro-seconds from long run.

Why do you need to re-calibrate tsc frequency?
What problem you trying to solve? I don't see any necessity of it.
@bigdata-memory The tsc freq should be re-calibrated otherwise the draft will quickly go >10,000 micro-second, and it is not acceptable after long run.

bigdata-memory:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1599#-MAA5YCo-8T-i7bDQDAG:-MADZxrKB46kgB9C9v20:bjm6t9j) (no related file):*
>  I cannot say guarantee but the algorithm re-calibrates the tsc freq very well, and the draft is managed under about 50 micro-seconds from long run.

---

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 36 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADP_ddAhzkdjSaJkLF:-MADaS2nEAvFNQ3vR0NC:b-m2g57d) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L36)):*

This is wrong, you have a typo: `&&` instead of `&`.

Simplify it as this: `return words[PAL_CPUID_WORD_EDX] & (1 << 8);`
</blockquote></details>

Yes, my bad

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 83 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADTNa66u_1xWlO7PgB:-MADj2h-AH5kNeE8Xvpt:bg0ftrv) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L83)):*

I would prefer to flip this check to decrease the indentations:
```
if (!tsc_khz) {
    /* no reliable TSC detected: fallback to gettimeofday() syscall */
    ret = ocall_gettime(&usec);
    if (ret < 0)
        return -PAL_ERROR_DENIED;
    return usec;
}

/* we have reliable TSC: calculate time using it */
_DkInternalLock(&g_tsc_lock);
...
```
</blockquote></details>

the return val will be explained as timestamp in usec.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 99 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADWBr92uSNS8CB_Wag:-MADjH9k-i6PBA7ncWMC:b72d0fn) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L99)):*

Here you silently ignore an error. Please do something like:
```
if (ret < 0) {
    _DkInternalUnlock(&g_tsc_lock);
    return -PAL_ERROR_DENIED;
}

... continue with tsc_cyc ...
```
</blockquote></details>

Yes, I see, maybe I need to take another pal api to handle it.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 107 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADXA444vMrt02IKnbD:-MADjYqwAlw0Uj-5Rmgf:bxcijg5) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L107)):*

Why `double` if all variables here are integers?
</blockquote></details>

I tried, if not it will trigger 100% cpu utilization in SGX enclave, it caused by ERESUME instruction

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 125 at r1](https://reviewable.io/reviews/oscarlab/graphene/1599#-MA9mT8V1nGr2-fdZWkE:-MADjlSs-v65QnRKDB3s:b6jcl7l) ([raw file](https://github.com/oscarlab/graphene/blob/9ac304596b3ef94071acdf482590b5cf7413270e/Pal/src/host/Linux-SGX/db_misc.c#L125)):*

> ```
> >> 1
> ```

Why "/2"?
"microsec - _Start_usec" includes whole round trip of ecall(. i.e. both of EEXIT and EENTER.
So tsc cycle should includes both, not half.
Am I missing something?
</blockquote></details>

it try to get the expected timepoint when the gettimeofday syscall get called

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 90 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADUtDA7aIkVd3hjcEr:-MADfAKn7ld7ci3fz8V2:b-pwj3w7) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L90)):*

What is this magic number? Replace with a macro with good name.

Also, what is the deal with `g_recal_shift`? You only increment it and never reset it. So at some point it may become a big number and overflow the 64-bit register...
</blockquote></details>

it will never reach overflow in lifetime.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 89 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADVHJL41GIDrw-eSHe:-MADoXhyC7mlz1RtEOKL:b-qvb647) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L89)):*

Please explain more. It looks like you check whether `10,000,000 ticks * recal_shift` elapsed, and in this case you will do recalibration. If my understanding is correct, please add it as a comment.
</blockquote></details>

yes, it is

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 30 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADOD825mPDxfWz3p-5:-MATWbzi08ti5FDgb7vl:bmeubst) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L30)):*

What's the point in a separate function? Just merge it into `init_tsc()`.
</blockquote></details>

well organized code according to its functionalities.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 75 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADSNhW2npxlPTZLzvK:-MATXin-3hU_XWvRCl-I:b3kre) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L75)):*

Please move this while function in `Pal/include/arch/x86_64/cpu.h`.
</blockquote></details>

?

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 83 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADT9FSAZ5TTjy2gERH:-MATXtos7s5KV49_Rey1:bu2w5v2) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L83)):*

Why do you need this helper variable `tsc_khz`? Just use `g_tsc_khz` everywhere.
</blockquote></details>

Minimize the impact on global variable by local variable.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 100 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADWXbsAV6k6ubKxEJL:-MATbHNa9dwd4pdc-eDl:bhrrlb4) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L100)):*

I need a comment why you're doing this
</blockquote></details>

it is for establish the baseline timestamp.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 117 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADWyOsB0l_zMQdzxpx:-MATbfn7BAVpRqmraiAX:bd49877) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L117)):*

I do not understand this whole logic of recalibration...
</blockquote></details>

explained in MS Team.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 100 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADWXbsAV6k6ubKxEJL:-MATbNST0G1NOO4ckvq0:bief1lf) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L100)):*

Yes, I guess your assumption is: `we performed an OCALL above, which includes one EENTER, one syscall execution, and one EEXIT; we assume that EENTER and EEXIT have approximately the same latency so we project that the TSC value at the time of actual gettimeofday() syscall is the middle point`. If I understand correctly, please add a comment in the code.
</blockquote></details>

Done. I have added the detail comments to code, thanks.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 117 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADWyOsB0l_zMQdzxpx:-MATbkyoB2G_7K3VrCIc:b60rgnz) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L117)):*

Unfortunately, I and Isaku still don't understand the need for it. You could post the explanations here (in Reviewable) so that other reviewers may weigh in.

My current understanding is that the TSC frequency obtained from CPUID is slightly off, and this recalibration is to bring the TSC frequency to a "better" value. However, this was only tested on one machine (if I understand correctly) and is not guaranteed to work properly in general. Also, it seems like an overkill because the drift is already contained by your `TSC_REFINE_INIT_TIMEOUT_USECS` check.
</blockquote></details>

Done.
It is not overkill, the reason are.
1) the algorithms avoided the periodically reset the timestamp.
2) the periodic timestamp reset using expensive ocall and syscall may cause observable performance downgrade for some sensitive workloads.
e.g. some Java application cannot even bear the parallel GC scanning
3) if the drift don't refined, the time interval sampling always contains the drift, it will cause the statistical difference for benchmarking.
4) the periodic timestamp reset introduces the security vulnerability since it always need to read untrusted external clock source at fixed timeout.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 107 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADXA444vMrt02IKnbD:-MATbmoJ2UVOUWFSu4oo:brxcodd) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L107)):*

I debugged on my local machine and I don't see 100% CPU utilization with and without this `double`. I don't think this is a real issue.

For context, I'll copy my comment here:

I tried this [test] code [from Gary]. I replaced `sleep(15)`  with `sleep(0)` . Everything works.

I think what happened in your case is this:
1. You sleep 1000 times for 15 seconds. This results in waiting for 250 minutes for this test to finish.
2. You didn't see any `Current timestamp...`  output. This is a known "feature" of Graphene: it buffers all output until the very end of program execution (unless explicitly fflush()).
3. You looked with perf at what's going on and saw a line with ERESUME -- but this simply means that perf cannot peek inside the SGX enclave, and EENTER/ERESUME is the last instruction "visible" to perf.

So I'm pretty sure you saw not a bug in Graphene but normal behavior. It has nothing to do with double .
</blockquote></details>

Done,

I have manually observed all values in this expression, 
it is not cause by overflow at all.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 45 at r3](https://reviewable.io/reviews/oscarlab/graphene/1599#-MAXjc003UZ-_9FdGRvG:-MAZXlXY3koyGHci3ezh:bxa71zs) ([raw file](https://github.com/oscarlab/graphene/blob/b107fa9eb13302faacb81fe87cc63f9f60f5d9f5/Pal/src/host/Linux-SGX/db_misc.c#L45)):*

I don't think you need `&& true` here.
</blockquote></details>

the type of left variable is bool, it much clear do so for readability.
the compiler may opt. it out

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 83 at r2](https://reviewable.io/reviews/oscarlab/graphene/1599#-MADT9FSAZ5TTjy2gERH:-MATbQQe10kXLEMJY3w3:bvyaogh) ([raw file](https://github.com/oscarlab/graphene/blob/6a9c2343e58263b95c9fe339f70aabaa693dd178/Pal/src/host/Linux-SGX/db_misc.c#L83)):*

I'm not sure I understand. The global variable is initialized only once (in `init_...`) and then used as read-only. There is not performance penalty of reading this global variable.
</blockquote></details>

The g_tsc_hz got updated in the refining process, and the code needs to keep original value of it, so has a local variable to hold its value for referring in refining process.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 13 at r4](https://reviewable.io/reviews/oscarlab/graphene/1599#-MAbvXzX6HcYz9yq-kl1:-MAcbaBaDlGvLAAOhAcT:b-hqqqxm) ([raw file](https://github.com/oscarlab/graphene/blob/baad81d319d0c5f5ad13b87b55d229a209bbc0c1/Pal/src/host/Linux-SGX/db_misc.c#L13)):*

You should add `#include "cpu.h"` (don't forget to sort the include lines)
</blockquote></details>

I can compile it successfully w/ no warnings. looks it has been included somewhere else.

dimakuv:
This PR significantly complicates the TSC-reading logic, however the performance benefit is unclear. We didn't observe any issues with the basic implementation that Graphene currently has. So let's close this PR until we really observe any performance/jitter issues in real-world apps.

PR #1608, opened 2020-06-22
[Testing] Only build the part of LTP we care about

donporter:
## Description of the changes 

Currently, we build all of LTP, even though we only use the syscall tests.  On a 5.7 kernel, some the test cases we don't use also do not build.

The simple answer is to be more targeted in our makefile, and not build code that is neither ours nor code we particularly care about.

## How to test this PR? 

Just run the CI pipeline.  Or build and run LTP manually

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1608)

### DIFF
--- a/LibOS/shim/test/ltp/Makefile
+++ b/LibOS/shim/test/ltp/Makefile
@@ -16,8 +16,9 @@ $(SRCDIR)/configure: $(SRCDIR)/Makefile
 
 $(BUILDDIR)/runltp: $(SRCDIR)/configure
 	cd $(SRCDIR) && ./configure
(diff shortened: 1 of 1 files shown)


dimakuv:
Jenkins, retest Jenkins-SGX-18.04 please (`openmp` LibOS test failed with `Unknown or illegal instruction at RIP 0x000000000af32ba0\n*** Unexpected AEX vector occurred inside PAL!` -- very unusual)

mkow:
FYI: I reproduced this openmp test failure on current master (without this PR). I just opened #1623 for this.

PR #1616, opened 2020-06-24
portable bash shebangs

Mic92:
Unlike /usr/bin/env which is a Posix standard, there is no guarentee
that /bin/bash exists. This is the case for operating systems such
as FreeBSD, NixOS and Guix. By using /usr/bin/env we also give
the user the option to provide their own bash in a different path
by setting the $PATH environemnt variable.
Distributions usually provide their own packaging wrappers
to fixup shebangs upon installation, however those are not convienent
to use when developing in the source tree.

See also other upstream discussions about the topic:

- https://github.com/systemd/systemd/pull/5816

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1616)

### DIFF
--- a/Scripts/clean-check
+++ b/Scripts/clean-check
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -eu -o pipefail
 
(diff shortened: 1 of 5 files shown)


dimakuv:
Jenkins, test this please

mkow:
Jenkins, retest this please

PR #1621, opened 2020-06-24
[Pal/Linux-SGX] Optimize the system time function using TSC cycles

### LINKED ISSUE #1598: Optimize the system time function via TSC for GSGX

bigdata-memory:
## Description of the changes 
This patch optimizes the function _DkSystemTimeQuery()
by using RDTSC instead of gettimeofday ocall/syscall.
This optimization won't take effect if there is no reliable
TSC source available to use i.e. nonstop TSC.
The TSC drift has been bound by resetting baseline timestamp
and tsc cycle aligned with system clock periodically.

Fixes #1598 

## How to test this PR? 
the existing time related test cases

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1621)

### DIFF
--- a/Pal/include/arch/x86_64/cpu.h
+++ b/Pal/include/arch/x86_64/cpu.h
@@ -3,6 +3,8 @@
 #ifndef CPU_H
 #define CPU_H
 
+#include <stdint.h>
+
 static inline void cpu_pause(void) {
     __asm__ volatile("pause");
 }
@@ -27,6 +29,12 @@ static inline void cpuid(unsigned int leaf, unsigned int subleaf, unsigned int w
                "c" (subleaf));
 }
 
+
+static inline uint64_t get_tsc(void) {
+    unsigned long lo, hi;
+    __asm__ volatile("rdtsc" : "=a" (lo), "=d" (hi));
+    return lo | ((uint64_t)hi << 32);
+}
 
 #define CPU_RELAX() __asm__ __volatile__("rep; nop" ::: "memory")
 
--- a/Pal/src/db_misc.c
+++ b/Pal/src/db_misc.c
@@ -15,8 +15,14 @@
 
 PAL_NUM DkSystemTimeQuery(void) {
     ENTER_PAL_CALL(DkSystemTimeQuery);
-    unsigned long time = _DkSystemTimeQuery();
-    return time;
+    uint64_t time;
+    int ret = _DkSystemTimeQuery(&time);
+    if (ret < 0) {
+        _DkRaiseFailure(-ret);
+        // TODO: Fix this interface to allow returning errors.
+        time = 0;
+    }
+    LEAVE_PAL_CALL_RETURN(time);
 }
 
 PAL_NUM DkRandomBitsRead(PAL_PTR buffer, PAL_NUM size) {
--- a/Pal/src/host/Linux-SGX/db_exception.c
+++ b/Pal/src/host/Linux-SGX/db_exception.c
@@ -116,8 +116,13 @@ static void emulate_rdtsc_and_print_warning(sgx_cpu_context_t* uc) {
                        "gettime() syscall.\n");
     }
 
+    uint64_t usec;
+    int res = _DkSystemTimeQuery(&usec);
+    if (res < 0) {
+        SGX_DBG(DBG_E, "_DkSystemTimeQuery() failed in unrecoverable context, exiting.\n");
+        _DkProcessExit(1);
+    }
     /* FIXME: Ideally, we would like to scale microseconds back to RDTSC clock cycles */
-    uint64_t usec = _DkSystemTimeQuery();
     uc->rdx = (uint32_t)(usec >> 32);
     uc->rax = (uint32_t)usec;
 }
--- a/Pal/src/host/Linux-SGX/db_main-x86_64.c
+++ b/Pal/src/host/Linux-SGX/db_main-x86_64.c
@@ -30,6 +30,9 @@
 #define BIT_EXTRACT_LE(value, start, after) \
     (((unsigned long)(value) & RIGHTMASK(after)) >> start)
 
+#define CPUID_LEAF_INVARIANT_TSC 0x80000007
+#define CPUID_LEAF_TSC_FREQ 0x15
+
 static char* cpu_flags[] = {
     "fpu",    // "x87 FPU on chip"
     "vme",    // "virtual-8086 mode enhancement"
@@ -89,6 +92,28 @@ static double get_bogomips(void) {
     buf[len] = 0;
 
     return sanitize_bogomips_value(get_bogomips_from_cpuinfo_buf(buf));
+}
+
+bool is_tsc_usable(void) {
+    uint32_t words[PAL_CPUID_WORD_NUM];
+
+    _DkCpuIdRetrieve(CPUID_LEAF_INVARIANT_TSC, 0, words);
+    return words[PAL_CPUID_WORD_EDX] & 1 << 8;
+}
+
+uint64_t get_tsc_hz(void) {
+    uint32_t words[PAL_CPUID_WORD_NUM];
+    uint64_t crys_hz;
+
+    _DkCpuIdRetrieve(CPUID_LEAF_TSC_FREQ, 0, words);
+    if (words[PAL_CPUID_WORD_EBX] > 0 && words[PAL_CPUID_WORD_EAX] > 0) {
+        /* nominal frequency of the core crystal clock in kHz */
+        crys_hz = words[PAL_CPUID_WORD_ECX];
+        if (crys_hz > 0) {
+            return crys_hz * words[PAL_CPUID_WORD_EBX] / words[PAL_CPUID_WORD_EAX];
+        }
+    }
+    return 0;
 }
 
 int _DkGetCPUInfo (PAL_CPU_INFO* ci) {
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -22,6 +22,7 @@
 #include <asm/ioctls.h>
 #include <asm/mman.h>
 #include <elf/elf.h>
+#include <stdint.h>
 #include <sysdeps/generic/ldsodefs.h>
 
 #include "ecall_types.h"
@@ -189,8 +190,14 @@ void pal_linux_main(char* uptr_args, uint64_t args_size, char* uptr_env, uint64_
      */
 
     PAL_HANDLE parent = NULL;
-    unsigned long start_time = _DkSystemTimeQuery();
     int rv;
+    uint64_t start_time;
+
+    rv = _DkSystemTimeQuery(&start_time);
+    if (rv < 0) {
+        SGX_DBG(DBG_E, "_DkSystemTimeQuery() failed: %d\n", rv);
+        ocall_exit(rv, /*is_exitgroup=*/true);
+    }
 
     struct pal_sec sec_info;
     if (!sgx_copy_to_enclave(&sec_info, sizeof(sec_info), uptr_sec_info, sizeof(sec_info))) {
@@ -290,6 +297,7 @@ void pal_linux_main(char* uptr_args, uint64_t args_size, char* uptr_env, uint64_
     init_enclave_key();
 
     init_cpuid();
+    init_tsc();
 
     /* now we can add a link map for PAL itself */
     setup_pal_map(&g_pal_map);
@@ -399,9 +407,14 @@ void pal_linux_main(char* uptr_args, uint64_t args_size, char* uptr_env, uint64_
     }
 
 #if PRINT_ENCLAVE_STAT == 1
-    printf("                >>>>>>>> "
-           "Enclave loading time =      %10ld milliseconds\n",
-           _DkSystemTimeQuery() - g_pal_sec.start_time);
+    uint64_t end_time;
+    rv = _DkSystemTimeQuery(&end_time);
+    if (rv < 0) {
+        SGX_DBG(DBG_E, "_DkSystemTimeQuery() failed: %d\n", rv);
+        ocall_exit(rv, /*is_exitgroup=*/true);
+    }
+    printf("                >>>>>>>> Enclave loading time =      %10ld milliseconds\n",
+           end_time - g_pal_sec.start_time);
 #endif
 
     /* set up thread handle */
--- a/Pal/src/host/Linux-SGX/db_misc.c
+++ b/Pal/src/host/Linux-SGX/db_misc.c
@@ -9,6 +9,7 @@
 
 #include <asm/fcntl.h>
 #include <linux/time.h>
+#include <stdint.h>
 
 #include "api.h"
 #include "pal.h"
@@ -28,14 +29,74 @@
 #include "sgx_api.h"
 #include "sgx_attest.h"
 
-unsigned long _DkSystemTimeQuery(void) {
-    unsigned long microsec;
-    int ret = ocall_gettime(&microsec);
-    if (ret)
-        return -PAL_ERROR_DENIED;
+#define TSC_REFINE_INIT_TIMEOUT_USECS 10000000
 
-    /* TODO: result comes from the untrusted host, introduce some schielding */
-    return microsec;
+static uint64_t g_tsc_hz = 0;
+static uint64_t g_start_tsc = 0;
+static uint64_t g_start_usec = 0;
+static PAL_LOCK g_tsc_lock = LOCK_INIT;
+
+/**
+ * Initialize the data structures used for date/time emulation using TSC
+ */
+void init_tsc(void) {
+    if (is_tsc_usable()) {
+        g_tsc_hz = get_tsc_hz();
+    }
+}
+
+/* TODO: result comes from the untrusted host, introduce some schielding */
+int _DkSystemTimeQuery(uint64_t* out_usec) {
+    uint64_t usec = 0;
+    uint64_t tsc_usec = 0, tsc_cyc1, tsc_cyc2, tsc_cyc, tsc_diff;
+    int ret;
+
+    if (g_tsc_hz > 0) {
+        _DkInternalLock(&g_tsc_lock);
+        if (g_start_tsc > 0 && g_start_usec > 0) {
+            /* calculate the TSC-based time */
+            tsc_diff = get_tsc() - g_start_tsc;
+            if (tsc_diff < INT64_MAX / 1000000) {
+                tsc_usec = g_start_usec + (tsc_diff * 1000000 / g_tsc_hz);
+                if (tsc_usec < g_start_usec + TSC_REFINE_INIT_TIMEOUT_USECS) {
+                    /* no need to refine yet */
+                    usec = tsc_usec;
+                }
+            }
+        }
+        _DkInternalUnlock(&g_tsc_lock);
+
+        if (!usec) {
+            /* refresh the baseline usec and TSC to contain the drift */
+            tsc_cyc1 = get_tsc();
+            ret = ocall_gettime(&usec);
+            tsc_cyc2 = get_tsc();
+            if (!ret) {
+                /* the ocall_gettime() is a time consuming operation.   *
+                 * it includes EENTER and EEXIT instructions, our best  *
+                 * estimation is the timestamp obtained in the middle   *
+                 * time point, therefore, the tsc_cyc as baseline will  *
+                 * be calibrated precisely in this way.                 */
+                tsc_cyc = ((tsc_cyc2 - tsc_cyc1) / 2) + tsc_cyc1;
+                _DkInternalLock(&g_tsc_lock);
+                /* refresh the baseline data if no other thread updated g_start_tsc */
+                if (g_start_tsc < tsc_cyc) {
+                    g_start_usec = usec;
+                    g_start_tsc = tsc_cyc;
+                }
+                _DkInternalUnlock(&g_tsc_lock);
+            } else {
+                return -PAL_ERROR_DENIED;
+            }
+        }
+    } else {
+        /* fallback to the slow ocall */
+        ret = ocall_gettime(&usec);
+        if (ret < 0)
+            return -PAL_ERROR_DENIED;
+    }
+    *out_usec = usec;
+    return 0;
 }
 
 int _DkInstructionCacheFlush(const void* addr, int size) {
--- a/Pal/src/host/Linux-SGX/enclave_ocalls.c
+++ b/Pal/src/host/Linux-SGX/enclave_ocalls.c
@@ -1233,10 +1233,9 @@ int ocall_shutdown (int sockfd, int how)
     return retval;
 }
 
-int ocall_gettime (unsigned long * microsec)
-{
+int ocall_gettime(uint64_t* microsec) {
     int retval = 0;
-    ms_ocall_gettime_t * ms;
+    ms_ocall_gettime_t* ms;
 
     void* old_ustack = sgx_prepare_ustack();
     ms = sgx_alloc_on_ustack_aligned(sizeof(*ms), alignof(*ms));
--- a/Pal/src/host/Linux-SGX/enclave_ocalls.h
+++ b/Pal/src/host/Linux-SGX/enclave_ocalls.h
@@ -75,7 +75,7 @@ int ocall_create_process(const char* uri, int nargs, const char** args, int* str
 
 int ocall_futex(uint32_t *uaddr, int op, int val, int64_t timeout_us);
 
-int ocall_gettime (unsigned long * microsec);
+int ocall_gettime(uint64_t* microsec);
 
 int ocall_sleep (unsigned long * microsec);
 
--- a/Pal/src/host/Linux-SGX/ocall_types.h
+++ b/Pal/src/host/Linux-SGX/ocall_types.h
@@ -243,7 +243,7 @@ typedef struct {
 } ms_ocall_shutdown_t;
 
 typedef struct {
-    unsigned long ms_microsec;
+    uint64_t ms_microsec;
 } ms_ocall_gettime_t;
 
 typedef struct {
--- a/Pal/src/host/Linux-SGX/pal_linux.h
+++ b/Pal/src/host/Linux-SGX/pal_linux.h
@@ -111,6 +111,10 @@ void _DkHandleExternalEvent(PAL_NUM event, sgx_cpu_context_t* uc, PAL_XREGS_STAT
 int init_trusted_files (void);
 void init_cpuid(void);
 
+bool is_tsc_usable(void);
+uint64_t get_tsc_hz(void);
+void init_tsc(void);
+
 /* Function: load_trusted_file
  * checks if the file to be opened is trusted or allowed,
  * according to the setting in manifest
--- a/Pal/src/host/Linux-SGX/sgx_enclave.c
+++ b/Pal/src/host/Linux-SGX/sgx_enclave.c
@@ -569,7 +569,7 @@ static long sgx_ocall_gettime(void * pms)
     ODEBUG(OCALL_GETTIME, ms);
     struct timeval tv;
     INLINE_SYSCALL(gettimeofday, 2, &tv, NULL);
-    ms->ms_microsec = tv.tv_sec * 1000000UL + tv.tv_usec;
+    ms->ms_microsec = tv.tv_sec * (uint64_t)1000000 + tv.tv_usec;
     return 0;
 }
 
--- a/Pal/src/host/Linux/db_misc.c
+++ b/Pal/src/host/Linux/db_misc.c
@@ -49,7 +49,7 @@ unsigned long _DkSystemTimeQueryEarly(void) {
 #endif
 }
 
-unsigned long _DkSystemTimeQuery(void) {
+int _DkSystemTimeQuery(uint64_t* out_usec) {
 #if USE_CLOCK_GETTIME == 1
     struct timespec time;
     int ret;
@@ -64,12 +64,12 @@ unsigned long _DkSystemTimeQuery(void) {
     }
 #endif
 
-    /* Come on, gettimeofday mostly never fails */
     if (IS_ERR(ret))
-        return 0;
+        return ret;
 
     /* in microseconds */
-    return 1000000ULL * time.tv_sec + time.tv_nsec / 1000;
+    *out_usec = 1000000 * (uint64_t)time.tv_sec + time.tv_nsec / 1000;
+    return 0;
 #else
     struct timeval time;
     int ret;
@@ -88,12 +88,12 @@ unsigned long _DkSystemTimeQuery(void) {
     }
 #endif
 
-    /* Come on, gettimeofday mostly never fails */
     if (IS_ERR(ret))
-        return 0;
+        return ret;
 
     /* in microseconds */
-    return 1000000ULL * time.tv_sec + time.tv_usec;
+    *out_usec = 1000000 * (uint64_t)time.tv_sec + time.tv_usec;
+    return 0;
 #endif
 }
 
--- a/Pal/src/host/Skeleton/db_misc.c
+++ b/Pal/src/host/Skeleton/db_misc.c
@@ -26,8 +26,8 @@ bool _DkInternalIsLocked(PAL_LOCK* lock) {
     __abort();
 }
 
-unsigned long _DkSystemTimeQuery(void) {
-    return 0;
+int _DkSystemTimeQuery(uint64_t* out_usec) {
+    return -PAL_ERROR_NOTIMPLEMENTED;
 }
 
 size_t _DkRandomBitsRead(void* buffer, size_t size) {
--- a/Pal/src/pal_internal.h
+++ b/Pal/src/pal_internal.h
@@ -279,7 +279,7 @@ void _DkExceptionReturn (void * event);
 void _DkInternalLock(PAL_LOCK* mut);
 void _DkInternalUnlock(PAL_LOCK* mut);
 bool _DkInternalIsLocked(PAL_LOCK* mut);
-unsigned long _DkSystemTimeQuery (void);
+int _DkSystemTimeQuery(uint64_t* out_usec);
 
 /*
  * Cryptographically secure random.


bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 41 at r1](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAceWvW1_ub1nnPyzWp:-MAcrn395MBjx25d3K0H:ba5n62a) ([raw file](https://github.com/oscarlab/graphene/blob/b620d560ef6d90c30034a187e291260715575b5a/Pal/src/host/Linux-SGX/db_misc.c#L41)):*

> ```
>  && true
> ```

something wrong.
</blockquote></details>

the term "&& true" has been removed, Done

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAccGDrCK4Xgt3qrbuy:-MAcrufKDqzpepGTAeMB:b-re7tnv) ([raw file](https://github.com/oscarlab/graphene/blob/b620d560ef6d90c30034a187e291260715575b5a/Pal/src/host/Linux-SGX/db_misc.c#L26)):*

Because those are specific to x86-64, those should be moved db_main-x86_64.c.
We keep the existing implementation as default.

@stefanburger any comments?


</blockquote></details>

Moved x86-64 related functions to db_main-x86_64.c. Done

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 76 at r1](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAcepcC9fMx9XpNzMKH:-MAcs68y1VH7ni0B_zF6:bqidss5) ([raw file](https://github.com/oscarlab/graphene/blob/b620d560ef6d90c30034a187e291260715575b5a/Pal/src/host/Linux-SGX/db_misc.c#L76)):*

Please drop tsc_hz as it's not modified.
Please just use g_tsc_hz.
</blockquote></details>

the tsc_hz dropped. Done

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 111 at r1](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAcbSoY5qoElqaco1Sl:-MAcsAFW7yuqXnvQvLAS:b-uzoi3h) ([raw file](https://github.com/oscarlab/graphene/blob/b620d560ef6d90c30034a187e291260715575b5a/Pal/src/host/Linux-SGX/db_misc.c#L111)):*

Because ocall is time consuming operation, it's a bad idea to invoke ocall with spin lock held.
Please unlock the g_tsc_lock before ocall and get the lock again after ocall.  You need to play with race.
</blockquote></details>

Updated according to your comment. Done

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 111 at r1](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAcbSoY5qoElqaco1Sl:-MAd68I-BOmOlDaV4ioX:btchu6z) ([raw file](https://github.com/oscarlab/graphene/blob/b620d560ef6d90c30034a187e291260715575b5a/Pal/src/host/Linux-SGX/db_misc.c#L111)):*

You've done wrongly.
the global variables are subject to racy access. So you need to protect those variables. 
</blockquote></details>

My code changes seems not get reflected here.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 111 at r1](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAcbSoY5qoElqaco1Sl:-MAd9Cly-gcfr9fMRDTF:b4wn769) ([raw file](https://github.com/oscarlab/graphene/blob/b620d560ef6d90c30034a187e291260715575b5a/Pal/src/host/Linux-SGX/db_misc.c#L111)):*
<details><summary><i>Previously, bigdata-memory (Wang, Gang) wrote…</i></summary><blockquote>

My code changes seems not get reflected here.
</blockquote></details>

I have re-submitted the PR w/ a fix for upper part of lock/unlock, but the bottom part of  lock/unlock still not shown correctly in Reviewable from my side.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 111 at r1](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAcbSoY5qoElqaco1Sl:-MAd9hDQ0yC0ixVxo3vj:b-sqz4rt) ([raw file](https://github.com/oscarlab/graphene/blob/b620d560ef6d90c30034a187e291260715575b5a/Pal/src/host/Linux-SGX/db_misc.c#L111)):*
<details><summary><i>Previously, bigdata-memory (Wang, Gang) wrote…</i></summary><blockquote>

I have re-submitted the PR w/ a fix for upper part of lock/unlock, but the bottom part of  lock/unlock still not shown correctly in Reviewable from my side. 
</blockquote></details>

Please referring to https://github.com/oscarlab/graphene/pull/1621/files
it shows correctly over there.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 52 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTe-QEj9AxVWBu85O:-MAgkHDe8rHjyparuXQf:be7tbkv) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L52)):*

> ```
> (double)
> ```

This cast should be avoided.
something like this.
now = get_tsc()
if (now - g_start_tsc < TSC_REFEIN_INIT_TIMEOUT_USEC * g_tsc_hz/1000000) 
...
</blockquote></details>

Why we need the new variable "now" ?
this expression is nothing about the timeout, 
it just used to calculate the TSC based clock time.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 72 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgS72I3XaBGY2vhdZ7:-MAhvDTA5RvwylQrjK9W:b-acn7pe) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L72)):*

@yamahata Isn't it benign? We'll have a redundant update but I don't see how it can break the logic.
</blockquote></details>

I have split the lock/unlock to two parts for for getting around the ocall, 
and also put it around those piece of codes where using the global variable, 
Any further changes I need to follow  ?

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 118 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAhtOJa48gugfYX55KH:-MAiEPCv2cyIncbmnnxW:b-kqu2ne) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_main-x86_64.c#L118)):*

You don't need to check for `words[PAL_CPUID_WORD_EAX] > 0` here; you already checked for it in the previous IF clause.
</blockquote></details>

I got it, I will updated it.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 86 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAhtzRpD8hHJa8TtMf2:-MAiIvHP9HUK721niGSD:bcwam46) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L86)):*
> PAL_ERROR_DENIED
The return type of function "unsigned long _DkSystemTimeQuery(void)"  is unsigned long, the value of -PAL_ERROR_DENIED is -6, so It cannot return it as -PAL_ERROR_DENIED as it will confuse the calling functions. that's why I use the _DkRaiseFailure(PAL_ERROR_DENIED).

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 79 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAhu9utA0X8e67AtHn4:-MAiJYJu7VEdWJiXalGE:bkv8wcw) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L79)):*
> The return type of function "unsigned long _DkSystemTimeQuery(void)" is unsigned long, the value of -PAL_ERROR_DENIED is -6, so It cannot return it as -PAL_ERROR_DENIED as it will confuse the calling functions. that's why I use the _DkRaiseFailure(PAL_ERROR_DENIED).

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 119 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTUos4eNBrnitk9HU:-MAiKCmx98QWItv7B8nO:birth2o) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_main-x86_64.c#L119)):*

> ```
> (double)
> ```

this cast doesn't make sense. remove this.
</blockquote></details>

Please refer to the following figure. 
it can make the code safer in case "crys_hz * words[PAL_CPUID_WORD_EBX]" got overflowed.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 72 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgS72I3XaBGY2vhdZ7:-MAisJ4P3XOWwi7-QMZf:b-qfjse4) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L72)):*

I had wallclock in mind.
If our time is older than g_start_{usec, tsc}, we don't want to update them. In such case, we also should return g_start_usec  (+ 1 or somehow to make it progress) as wall clock.
</blockquote></details>

You mean  TSC cycle counter could be go backward at runtime, right ?
if I understand you correctly, I'll add a condition to check this case beforehand.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 72 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgS72I3XaBGY2vhdZ7:-MAiu8Kg99OHyxozdNCI:b-oe2n13) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L72)):*
<details><summary><i>Previously, bigdata-memory (Wang, Gang) wrote…</i></summary><blockquote>

You mean  TSC cycle counter could be go backward at runtime, right ?
if I understand you correctly, I'll add a condition to check this case beforehand.
</blockquote></details>

But, should we detect if TSC cycle counter go backward ??
I recall the Linux Kernel code 
that doesn't have the logic code to detect TSC cycle will go backward.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 72 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgS72I3XaBGY2vhdZ7:-MAiwwP57mxmEYqjsWPM:b-qclw4t) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L72)):*
<details><summary><i>Previously, bigdata-memory (Wang, Gang) wrote…</i></summary><blockquote>

But, should we detect if TSC cycle counter go backward ??
I recall the Linux Kernel code 
that doesn't have the logic code to detect TSC cycle will go backward.
</blockquote></details>

In addition, If we cannot trust TSC cycle,  It would don't make sense to use TSC to calculate the wall clock because every time we use the RDTSC instruction, its counter number might be go backwards or always go backwards.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 72 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgS72I3XaBGY2vhdZ7:-MAizZN03mALz1CJLC13:bk5080b) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L72)):*

no. our thread can be preempted after getting tsc. and then other thread comes in to update them.
</blockquote></details>

But why I cannot protect the whole logic as one piece by locking the "g_tsc_lock" and unlock it in the end as my original code does ? 
the "g_tsc_lock" don't has the negative impact on other locks.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 72 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgS72I3XaBGY2vhdZ7:-MAj0mJ6-AQMdXkqjjDQ:b-814zkj) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L72)):*
<details><summary><i>Previously, bigdata-memory (Wang, Gang) wrote…</i></summary><blockquote>

But why I cannot protect the whole logic as one piece by locking the "g_tsc_lock" and unlock it in the end as my original code does ? 
the "g_tsc_lock" don't has the negative impact on other locks.
</blockquote></details>

I have restored my original code about the lock/unlock,
I think it could be more safer to protect the logic as a whole piece.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 118 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAhtOJa48gugfYX55KH:-MAiGora3kysXDa6cGQY:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_main-x86_64.c#L118)):*

I got it, I will updated it.
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 72 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgS72I3XaBGY2vhdZ7:-MB1LxvWB8dSJrsEnXHF:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L72)):*

I am Ok with this logic of one big lock/unlock. We only perform an OCALL once in 10 seconds, and this OCALL is gettimeofday() -- it never blocks and doesn't introduce any additional latency. So to me it's just easier to have the big lock and consider the OCALL case very rare.
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 111 at r1](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAcbSoY5qoElqaco1Sl:-MB1M20jDCvV2efJECdO:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b620d560ef6d90c30034a187e291260715575b5a/Pal/src/host/Linux-SGX/db_misc.c#L111)):*

Please referring to https://github.com/oscarlab/graphene/pull/1621/files
it shows correctly over there.
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 52 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTe-QEj9AxVWBu85O:-MBFIzzp58OdiY8PNlAR:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L52)):*

Why we need the new variable "now" ?
this expression is nothing about the timeout, 
it just used to calculate the TSC based clock time.
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 119 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTUos4eNBrnitk9HU:-MBFJ0_6-10EYtWsxJJs:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_main-x86_64.c#L119)):*

Please refer to the following figure. 
it can make the code safer in case "crys_hz * words[PAL_CPUID_WORD_EBX]" got overflowed.



</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 35 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAhtgap6dJ9TRM_lR0U:-MAiGaSD4h3guOxNtuvm:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L35)):*

This comment is incorrect. It must be `... used for gettimeofday() emulation` or simply `... used for date/time emulation` (whatever you prefer).
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 61 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAhu_Yk007zoqz7njYs:-MAiHHt87wonzi_pqkfq:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L61)):*

You don't refine TSC frequency here anymore. Let's replace with `/* refresh the baseline usec and TSC (to contain the drift) */`
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 119 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTUos4eNBrnitk9HU:-MBFzrqgEM4rGy7tkHxd:b-d0n4an) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_main-x86_64.c#L119)):*

Done.
</blockquote></details>

@yamahata  the double typecast needs to used because some CPU may provide very large value for A to make the result more precisely.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 119 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTUos4eNBrnitk9HU:-MBgInBEDR7n0PR1Dp3u:b-7voeuk) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_main-x86_64.c#L119)):*

@yamahata  the double typecast needs to used because some CPU may provide very large value for A to make the result more precisely. 
</blockquote></details>

the double cast removed.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 52 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTe-QEj9AxVWBu85O:-MBgIrbpEcAEbvPFSdQU:b-7voeuk) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L52)):*

Done.
</blockquote></details>

the double cast removed.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 72 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgS72I3XaBGY2vhdZ7:-MBgIv0q4v-WYoFVOZTx:b-skf7we) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L72)):*

Done.
</blockquote></details>

the lock has been separated .

dimakuv:
Jenkins, test this please

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 52 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTe-QEj9AxVWBu85O:-MBjbuZM9H1SBnvA71qe:bs1kdnl) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L52)):*

I’m suggesting the check the diff first and then calculate tsc_usec to avoid overflow.
</blockquote></details>

According to your instruction, I removed the double typecasting for this patch.
Do you ask me to provide the actual changing diff values or adding a "if" checking on this specific expression ? I didn't observe overflow from my tests. but if it is later, why not just using the double typecasting as my original code does ?

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 52 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTe-QEj9AxVWBu85O:-MBl9JLmElj19GTPp8TL:b-qen9sr) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L52)):*

The concrete code snippet will be something like

int64_t diff_tsc = get_tsc() - g_start_tsc
if (diff < TSC_REFINE_INIT_TIMEOUT_USEC * g_tsc_hz / 1000000)
    usec = g_start_usec + diff_tsc * 1000000 / g_tsc_hz

By checking if the diff is small, it is guaranteed that overflow won't happen.
</blockquote></details>

According to your suggested code snippet. I found that the diff_tsc still can be valid if the condition is not true because the diff_tsc contains the undetermined TSC drift at runtime. the TSC freq could have small changing when retrieved from cpuid according to my observation previously.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 52 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTe-QEj9AxVWBu85O:-MBlBNIz3oxdcc10TN94:bwv9yih) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L52)):*
> if (diff < TSC_REFINE_INIT_TIMEOUT_USEC * g_tsc_hz / 1000000)
In addition, this condition limits the tsc diff is only valid if smaller than the fixed value of "TSC_REFINE_INIT_TIMEOUT_USEC * g_tsc_hz / 1000000",  but actually tsc diff could be any valid value around it.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 52 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTe-QEj9AxVWBu85O:-MBlFPTF2Ys6Nd8tYQDj:be6bwby) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L52)):*

> if (diff < TSC_REFINE_INIT_TIMEOUT_USEC * g_tsc_hz / 1000000)
In addition, this condition limits the tsc diff is only valid if smaller than the fixed value of "TSC_REFINE_INIT_TIMEOUT_USEC * g_tsc_hz / 1000000",  but actually tsc diff could be any valid value around it.
</blockquote></details>

and Thanks for the code and what should we do in "else" case ?

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 52 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTe-QEj9AxVWBu85O:-MBoP2KLAjyZlEOur_iB:bhptn9w) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L52)):*

no else clause.
</blockquote></details>

Hi @yamahata , for Adding operation, the overflow may happen only if the signs of two numbers are the same and the sign of their sum is opposite to the signs of numbers, hence the diff_tsc alone cannot overflow. so if double type casting is prohibited (I don't know why since GSGX LibOS is using the double type and it shared the same user-space with application/workload), how about re-write the following code
(get_tsc() - g_start_tsc) * 1000000 / g_tsc_hz
---  to --->
(get_tsc() - g_start_tsc) * (1000000 / g_tsc_hz)  ?

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 52 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTe-QEj9AxVWBu85O:-MBp6TNxAQMlqcrgY7lo:bxbo9vt) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L52)):*

 Here is the exact code snippet I'm suggesting
```
int64_t diff_tsc = get_tsc() - g_start_tsc;
if (diff_tsc < TSC_REFINE_INIT_TIMEOUT_USEC * g_tsc_hz / 1000000) {
    // Here diff_tsc is small enough so that overflow by multiply won't happen
    usec = g_start_usec + diff_tsc * 1000000 / g_tsc_hz;
}
// "else" clause is not needed because usec == 0. ocall will be issued below and g_start_usec, g_start_tsc will be reset
//  now the variable, tsc_usec, is not used. please eliminate the variable.

```

regarding to
> (get_tsc() - g_start_tsc) * (1000000 / g_tsc_hz) ?

There is no point to lose preciseness where the precise calculation is possible as above.
</blockquote></details>

That's way some kernel code does, but anyway, do you think we don't need to considering the outband drift ignored by the if condition that I have explained my previous replies ?

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 52 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTe-QEj9AxVWBu85O:-MBqCzkK0huUGt9JZxVX:bpm9vog) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L52)):*

That's way some kernel code does, but anyway, do you think we don't need to considering the outband drift ignored by the if condition that I have explained my previous replies ?
</blockquote></details>

@yamahata  I turn to use the function __builtin_mul_overflow(...) to detect the overflow. please review the updated code.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 52 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTe-QEj9AxVWBu85O:-MBtaB9r3xr3rwVn0jOu:b-ilnz47) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L52)):*

There is no point to resort to use gcc extension when it can be written in plain normal way.
</blockquote></details>

@yamahata  why it is no point ? I found the function __builtin_mul_overflow(...) also been used in other place of GSGX LibOS.  I also identified the potential code logic issue in your suggested code snippet in my previous replies. please guide me how to choose ?

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 52 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTe-QEj9AxVWBu85O:-MBzJ6Y--JSpw_1hUxFL:b5e8d79) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L52)):*

@yamahata  why it is no point ? I found the function __builtin_mul_overflow(...) also been used in other place of GSGX LibOS.  I also identified the potential code logic issue in your suggested code snippet in my previous replies. please guide me how to choose ?
</blockquote></details>

If it is no point to use  __buildin_mul_overflow(...), the INT64_MAX looks good to fit into this case very well, and it also avoid the potential drift problem, please review it, thanks.

sudharkrish:
Hi Gary @bigdata-memory , I see that this PR has received 2 approvals so far. We are using this PR for redis benchmarking. Just wanted to check, if there is anything blocking, and whether this can get merged to master. Also CCing @dimakuv , @mkow @stefanberger.

bigdata-memory:
@sudharkrish This PR is expected to be merged this week along with the GSGX release v1.1, thanks.

sudharkrish:
@bigdata-memory I have been wondering about the accuracy of _DkSystemTimeQuery, now that we have this performance optimization. Although we have used this PR for redis benchmarking, we have not checked the accuracy of time retrieved with this patch. It would be good, if we can add a test case part of regression, that can be used to validate the accuracy of time obtained with this change. Also CCing @dimakuv , @yamahata , @mkow , to check if they have any suggestions for testing this change.

bigdata-memory:
@sudharkrish In this PR, it uses the a simple algorithm to calibrate the baseline TSC time along with the TSC cycles at fixed interval  of 10 secs, from my visual observation (longer than several 10 secs), and the max. drift is under 100 usecs. In my another PR #1599, it uses more complex way to calibrate and result in a smaller drift (> 40 usecs or so) and fewer calibration actions. I don't knew, how much drift is acceptable and it looks no way to expose the drift info w/o breaking the APIs. thanks.

bigdata-memory:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCBwGYlEa93raMJ6i0W:-MCFUF9w-1tvUbXsYE3p:bsa9nkw) (no related file):*

Why does it say "Fixes #1599" while it points to a PR, not an issue? Did you mistyped the number or this PR supersedes #1599? If the latter, please use "Closes #1599" (not "Fixes"), the current version is confusing.
</blockquote></details>

Sorry it should be Fixes #1598,  I have updated it, thanks.

---

bigdata-memory:
___
*[Pal/include/arch/x86_64/cpu.h, line 32 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MC9XxxkE8xKJUlEeonR:-MCFX8VH0cRxTzRgnOOj:bkl15kq) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/include/arch/x86_64/cpu.h#L32)):*

this should return `uint64_t`
</blockquote></details>

In this file, you can check the function cpuid(...), seems they uses the type "unsigned int" and no header file in this file cpu.h, so I suppose the code should minimize the dependency on other header files.

bigdata-memory:
___
*[Pal/include/arch/x86_64/cpu.h, line 35 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MC9YpH4BnDZviBs6cRa:-MCFYTD6EUMs_PWr5ukF:bgte2wf) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/include/arch/x86_64/cpu.h#L35)):*

> ```
> hi << 32
> ```
This is UB on 32-bit x86.
</blockquote></details>

This header file is under folder arch/x86_64/, so it suppose to compiled as 64 bits type.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 119 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTUos4eNBrnitk9HU:-MCF_c1S-zTGfTMKiTRF:bfum927) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_main-x86_64.c#L119)):*

late to the game, but my 2 cents: the overflow here was technically impossible regardless of what CPU returned, you do 64-bit multiplication on two 32-bit numbers here.
</blockquote></details>

My 2 cents, it makes the code more maintainable and robust, since the type might be changed by other contributors later and still not break it. I saw some codes, they add (double) typecasting as follows
"(double) A / B" to ensure the result doesn't get affected even the type of A has been changed to smaller one.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 79 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAhu9utA0X8e67AtHn4:-MCFbZZC8CYQyarOVxua:bvxm796) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L79)):*

Uh, I don't think this works. `_DkRaiseFailure` is not `noreturn` and it's (AFAIK) only meant to set pal errno on PAL API calls. You should return error here, not use this weird function ;) Ditto for other places in this function.

Ad signedness issue: C isn't a nice language ;) You have to return success/error via `int` and the actual result via pointer argument.
</blockquote></details>

Can I fix this issue as a whole in another PR ? since this PR is not expected to change the existing function signature. thanks.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 86 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCElTjNDcBl_giQokEE:-MCFcjbK6f4LIaSOMOI9:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_misc.c#L86)):*

> ```
> gettimeofday syscall
> ```
We're not on the Linux syscall layer here, ocall_gettime != gettimeofday. E.g. on Windows PAL (if ever created) this will be a call to some WinAPI. Let's just say "fallback to the slow ocall".
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 73 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEwNlb82pDLX0kTfc6:-MCFcmmk08J42tP3w4rE:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_misc.c#L73)):*

> ```
>  >> 1
> ```
Please use normal division, all compilers are smart enough to optimize this, no need for bit hackery here ;)
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 50 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEofAW9EmP6_yj3CKC:-MCFcoEl3Cz9pbr1yI8M:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_misc.c#L50)):*

`TSC based` -> `TSC-based`
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 42 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEt-pq09nep7aNru1X:-MCFcps65kJIUtqBX69v:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_misc.c#L42)):*

This function should return `uint64_t`, please fix this while changing the function's signature (see my other comment below).
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 34 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCElGid2Va1fBacUzjj:-MCFcrKF6KBwl5rTSZPb:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_misc.c#L34)):*

`for date/time emulation` -> `for date/time emulation using TSC`
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 26 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEzFpqB-yXwgXEExgG:-MCFcsTXE97R1-Ol0TFy:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_misc.c#L26)):*

Why `L` suffix?
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 120 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEiuFr3HRmVEgUSO6z:-MCFcuFV2ZVPPwa0YTTc:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_main-x86_64.c#L120)):*

This is simpler:
> ```C
> return crys_hz * words[PAL_CPUID_WORD_EBX] / words[PAL_CPUID_WORD_EAX];
> ```

</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 112 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCExj3mAg7fPAjO_crd:-MCFcvt_6c7VJ2u6Nqhe:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_main-x86_64.c#L112)):*

`uint64_t`
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 111 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEhes6DejBl8laNJ4C:-MCFcwrZ6Q98QGuCdkjq:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_main-x86_64.c#L111)):*

`uint32_t`
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 110 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEj3X54H6cvBaA3TE8:-MCFcxx5DsyayaeeI_Mg:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_main-x86_64.c#L110)):*

not needed
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 109 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEhQEZ54Es7iTQSb_H:-MCFczD8D7UfkP1m3ylD:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_main-x86_64.c#L109)):*

shouldn't this be unsigned?
ditto below
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 108 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEgeuk83jzbOY4WUHl:-MCFd-M9Bd1EaFn359YD:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_main-x86_64.c#L108)):*

ditto
If you want to highlight that this frequency is "baseline" then just add this to the function name, it will still be relatively short.
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 100 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEh8qj5LMZTzLACNH1:-MCFd0Nu27aii2dSBMWD:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_main-x86_64.c#L100)):*

`uint32_t`
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 99 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEhy4I5ZOZhXwNhs7T:-MCFd1Sk2bNnYNNxrAqS:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_main-x86_64.c#L99)):*

no need for this variable
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 97 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEgXwI2Gz57lw4YA94:-MCFd2bz0hQwLrIWpVk4:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/src/host/Linux-SGX/db_main-x86_64.c#L97)):*

this comment says exactly the same thing as the function name
</blockquote></details>

Done.

bigdata-memory:
___
*[Pal/include/arch/x86_64/cpu.h, line 31 at r7](https://reviewable.io/reviews/oscarlab/graphene/1621#-MC9Y4z71xEFe_x8-NDq:-MCFd5sW4zIg4HZLtf0Y:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b49e3e9e3432877c2167a35874889270679fae02/Pal/include/arch/x86_64/cpu.h#L31)):*

I'd remove this comment, it's already contained in the function's name.
</blockquote></details>

Done.

bigdata-memory:
@sudharkrish please note the observation was based on my machine w/ invariant TSC feature.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 119 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAgTUos4eNBrnitk9HU:-MCI4CQO98l0IRumlWtL:bgsg6h8) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_main-x86_64.c#L119)):*

>  "(double) A / B" to ensure the result doesn't get affected even the type of A has been changed to smaller one.

Such code is almost always used in C to turn integer division into float division, not to fix overflows. If you want to ensure this works on smaller types then you can cast to `uint64_t`, not to `double`.
</blockquote></details>

That expression result will be converted back a integer, something like
int C = (double) A / B
so from my understanding, the (double) type casting is used to make code robust.

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 79 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAhu9utA0X8e67AtHn4:-MCI5MgS5WgxrFQAZOIj:bnqr75k) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L79)):*

But you introduce a new bug in this PR, I'd prefer to fix it here. I think it shouldn't be too much work - this function is called in only 2 or 3 places and the change is simple.
</blockquote></details>

Actually, the PR fixed up the bug in the original code, because it tried to return -6 if failed, the _DkRaiseFailure(...) is also used in other places. if it is a weird function, we may need to consider a solution to fully correct it as a whole.

mkow:
Jenkins, retest this please

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 79 at r2](https://reviewable.io/reviews/oscarlab/graphene/1621#-MAhu9utA0X8e67AtHn4:-MCJ7s9r7IFTYfOD7EP8:b-naxt36) ([raw file](https://github.com/oscarlab/graphene/blob/56dc0f465e523fc3e1a481d090678e66073ba01f/Pal/src/host/Linux-SGX/db_misc.c#L79)):*

As agreed on priv, I implemented the fix and pushed to the PR branch.
</blockquote></details>

@mkow  I appreciate.

mkow:
Jenkins, retest this please

mkow:
Jenkins, retest this please

bigdata-memory:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1621#-MCEycVW-jNKkYJJxn_F:-MCNNtkvAR3z8dIdtHnw:b-paa8ih) (no related file):*

Done. I just ran `gettimeofday()` for different amounts of time (5 sec, 30 sec, and 5 min):
```
=========================================== MASTER

$ SGX=1 ~/graphene/Runtime/pal_loader ./multi_pthread 5
sec = 5   usec = 969
----- Total SGX stats for process 30070 -----
  # of EENTERs: 209
  # of EEXITs:  177
  # of AEXs:    194

$ SGX=1 ~/graphene/Runtime/pal_loader ./multi_pthread 30
sec = 30   usec = 5334
----- Total SGX stats for process 38543 -----
  # of EENTERs: 259
  # of EEXITs:  227
  # of AEXs:    194

$ SGX=1 ~/graphene/Runtime/pal_loader ./multi_pthread 300
sec = 300   usec = 53870
----- Total SGX stats for process 38589 -----
  # of EENTERs: 800
  # of EEXITs:  768
  # of AEXs:    200

=========================================== WITH PR

$ SGX=1 ~/graphene/Runtime/pal_loader ./multi_pthread 5
sec = 5   usec = 936
----- Total SGX stats for process 34315 -----
  # of EENTERs: 204
  # of EEXITs:  172
  # of AEXs:    204

$ SGX=1 ~/graphene/Runtime/pal_loader ./multi_pthread 30
sec = 30   usec = 5106
----- Total SGX stats for process 34341 -----
  # of EENTERs: 232
  # of EEXITs:  200
  # of AEXs:    196

$ SGX=1 ~/graphene/Runtime/pal_loader ./multi_pthread 300
sec = 300   usec = 50330
----- Total SGX stats for process 42831 -----
  # of EENTERs: 530
  # of EEXITs:  498
  # of AEXs:    203
```

As you can see, the drift in time is bound and very close to the previous OCALL-based logic. It is also visible from EENTERs/EEXITs that the number of OCALLs is less (the increase is due to printfs in my test case) -- this proves that gettime OCALL is replaced mostly by RDTSC.
</blockquote></details>

@dimakuv It is very valuable to confirm the improve from this PR and help me to improve my coming PRs, thank you very much!

---

sudharkrish:
@dimakuv , thanks for your efforts. It will be good to have your test code in the PR..for future reference.  Otherwise, just by looking at the output, it is not easy to find out, how you have tested this.

sudharkrish:
@bigdata-memory Given that you are re-calibrating the baseline TSC time along with the TSC cycles at fixed interval of 10 secs, I wonder, how it will vary between platforms. Ideally, it would have been good to add a regression test for this change. But I understand it is hard to add one for this. But anyways, thanks to @dimakuv 's testing, we do know that it's accuracy is on-par with the actual ocall for getimeofday, for the platforms on which it has been tested already.

Issue #1626, opened 2020-06-26
Please speed up the project building using all cpu cores. It is very slow.


bigdata-memory:
## Description of the problem
I have tried to build the clean project, it cost me about several more minutes on my powerful machine. I found it just uses 1 cpu core to fulfill the building. that is why it is so slow. 

## Steps to reproduce
make

## Expected results
Used up all cpu core to accelerate project compilation.

## Actual results
It just used 1 core to build this project.

dimakuv:
One can simply add `-j X` to parallelize the make build. In fact, this is what we do in Jenkins.

bigdata-memory:
Thanks.

PR #1627, opened 2020-06-26
[LIBOS] Speed up the project building about x[Cores] times.

bigdata-memory:
## Description of the changes 
This patch added flags to make command to speed up
the compilation of glibc buid. it automatically
detects the core number, it significant shorten
the compilation time.
related to issue:
https://github.com/oscarlab/graphene/issues/1626

## How to test this PR? 
time make

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1627)

### DIFF
--- a/LibOS/Makefile
+++ b/LibOS/Makefile
@@ -49,7 +49,7 @@ ifeq ($(findstring x86_64,$(SYS))$(findstring linux,$(SYS)),x86_64linux)
 
 $(BUILD_DIR)/Build.success: $(BUILD_DIR)/Makefile
 	@echo "Building glibc, may take a while to finish. Warning messages may show up. If this process terminates with failures, see \"$(BUILD_DIR)/build.log\" for more information."
-	($(MAKE) -C $(BUILD_DIR) 2>&1 >> build.log) && touch $@
+	(PARALLELMFLAGS="-j" $(MAKE) -C $(BUILD_DIR) 2>&1 >> build.log) && touch $@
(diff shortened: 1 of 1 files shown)


bigdata-memory:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1627#-MAmioib0OFB_oDI6Fjo:-MAmnn0Y0P9aRbWfdaHv:barr92b) (no related file):*

Our Jenkins already has something like this: `make -j 8 glibc-build/Build.success GLIBC_VERSION=2.23`. We pick `8` which is probably not the best number, but it works fine.

A user can do the same while building locally (by simply adding `-j X` to the make command).
</blockquote></details>

I think the user may just aware of using the make command at project level. 
At project level make -j 8, the "-j 8" looks not pass to the glibc-build make.

---

bigdata-memory:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1627#-MAmioib0OFB_oDI6Fjo:-MAmox4e-Xyzx67wEK7C:b-4iu6x6) (no related file):*
<details><summary><i>Previously, bigdata-memory (Wang, Gang) wrote…</i></summary><blockquote>

I think the user may just aware of using the make command at project level. 
At project level make -j 8, the "-j 8" looks not pass to the glibc-build make.
</blockquote></details>

Yes  you are right, I have tried, it can be passed to glibc-build, thanks.

---

PR #1630, opened 2020-06-29
PAL+LibOS: Set 'weak' parameter in atomics operations to 'false'

stefanberger:
## Description of the changes 

The 'weak' parameter may not have any influence on x86, but on ppc64 it
does cause 'spurious faults' due to a missing check on flags (missing
'bne' instruction). The weak parameter is described in the gcc atomics
operations documentation like this:

"[...] weak is true for weak compare_exchange, which may fail spuriously, and
false for the strong variation, which never fails spuriously."

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1630)


dimakuv:
Jenkins, test this please

dimakuv:
Jenkins, test this please

dimakuv:
Jenkins, test this please

PR #1635, opened 2020-06-30
[LibOS] Drop broken BOOKKEEP_INTERNAL_OBJ

mkow:
## Description of the changes 

Enabling this macro breaks the build due to a typo: `OBJECT_INTERNVAL`. `git blame` points to "The first official release" commit, which means that no one used it since at least 2014. We can safely get rid of it.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1635)


mkow:
Update: there are actually two typos in this line, lol.

PR #1633, opened 2020-06-30
[pal] fix parallel build

Mic92:
download: Found 'v1.7.12.tar.gz' (76068766...) in cache.
download: Found 'v1.7.12.tar.gz' (76068766...) in cache.
cp: cannot create regular file 'v1.7.12.tar.gz': Permission denied
make[5]: *** [Makefile:25: cJSON.c] Error 1
make[5]: *** Waiting for unfinished jobs....

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1633)

PR #1637, opened 2020-07-01
[Pal/lib] Replace old memory-handling functions with simpler versions

dimakuv:
##  Description of the changes 

Graphene used its own implementations of memory-handling functions like `memcmp()`, `memcpy()`, etc. Previously, these implementations were copied from Glibc and contained complicated code from year 2004. Modern HW and compilers do better job at optimizing simple C implementations of these functions. Thus, this commit replaces old implementations with simple ones taken from Musl libc and adapted to our code style.

One particular optimization is made to `memcpy()` on x86-64. `memcpy()` is heavily used in Linux-SGX PAL to copy data in/out of SGX enclave. Experiments with Redis 5.0 show perf improvement of using `rep movsb` at
3-5% for 4KB payloads.

## How to test this PR? 

All tests must pass.

Here are some numbers from a quick test with Redis 5.0 (+ exitless `sgx.rpc_thread_num = 5`):
```sh
# old hacky memcpy
$ ./src/src/redis-benchmark -t set,get -n 1000000 -r 1 -d 4096
====== SET ======
109878.04 requests per second
====== GET ======
112044.81 requests per second

# new rep movsb memcpy
$ ./src/src/redis-benchmark -t set,get -n 1000000 -r 1 -d 4096
====== SET ======
113558.94 requests per second
====== GET ======
117481.20 requests per second
```

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1637)


dimakuv:
Jenkins, restart Jenkins-SGX-18.04 please (`cJON` file seems to be incorrect, see below).

Jenkins shows this error:
```
download: Fetched 'v1.7.12.tar.gz' (76068766...) successfully.
tar -mxzf v1.7.12.tar.gz
cp cJSON-1.7.12/cJSON.c .
cp cJSON-1.7.12/cJSON.h .
attestation.c:13:10: error: cJSON.h is shorter than expected [-Werror]
 #include "cJSON.h"
          ^~~~~~~~~
gcc -Wall -std=c11 ... -c -o cJSON.o cJSON.c
```

Looks like cJSON was untared with errors (probably because of disk errors?).

dimakuv:
Jenkins, retest Jenkins-SGX-18.04 please

mkow:
@dimakuv I think the cJSON failures we see are caused by the bug fixed by #1633.

PR #1638, opened 2020-07-01
[Pal] don't depend on absolute PATH for tools

Mic92:
On NixOS/Guix tools are not installed into /usr/bin but come from
are composed as symlink trees from immutable packages in the /nix/store:

  i.e. /home/joerg/.nix-profile/bin/bash -> /nix/store/hgplspyw16amnbhlbi7qa2938w8i0rz6-bash-interactive-4.4-p23/bin/bash

Due it's immutable nature they are a perfect fit for graphene-sgx since
libraries store paths won't change unlike their traditional /usr/lib
equalivalents and hence don't break code signing.

One might argue that using for absolute paths makes it possible without a PATH
set, however this is not really true, since make in this case relies on other
tools. Also the whole script relies on bash to be in PATH for it's shebang.

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1638)

### DIFF
--- a/Runtime/pal_loader
+++ b/Runtime/pal_loader
@@ -17,9 +17,9 @@ do
 	shift
 done
 
-RUNTIME_DIR=$(/usr/bin/dirname "$(readlink -f "${BASH_SOURCE[0]}")")
+RUNTIME_DIR=$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")
 if [ -z "$PAL_HOST" ]; then
-    if [ ! -f /usr/bin/make ]; then
+    if ! command -v make >/dev/null; then
         libpal="$RUNTIME_DIR/libpal-*.so"
         libpal="$(echo -n "$libpal")"
         libpal="${libpal//$RUNTIME_DIR\//}"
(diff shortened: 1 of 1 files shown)


dimakuv:
Jenkins, test this please

Mic92:
> > _[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1638)_ status: :shipit: complete! all files reviewed, all discussions resolved
> 
> _[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1638#-MBB3jsEAvUqJYyz54t7:-MBB3jsEAvUqJYyz54t8:b-u46okb) (no related file):_
> 
> > Due it's immutable nature they are a perfect fit for graphene-sgx since
> > libraries store paths won't change unlike their traditional /usr/lib
> > equalivalents and hence don't break code signing.
> 
> @Mic92 Are you working on reproducible builds for Graphene? Sounds very interesting, hope to read the paper/blog soon :)

Not quite, right now I building some nix abstraction so I can benchmark this against Scone and my own implementation based on SGX-LKL. However as a result there will be some functions that allows to wrap packages defined in Nix and generate a signed Graphene policy for it. Since nix's dependencies are tracked more strict it should be harder to forget missing run-time dependencies.

dimakuv:
Jenkins, test this please

PR #1639, opened 2020-07-01
[Pal] invoke mbedtls with perl directly

Mic92:
Also scripts/config.pl has a sane shebang `/usr/bin/env perl`,
we don't provide `/usr/bin/env` in the [nix](https://nixos.org) sandbox.
Usually we have ways of patching those shebangs however in this case
it is hidden in a tarball therefore not so easy to patch.
For other platforms this patch is not necessary however it should not break
anything.

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1639)


dimakuv:
Jenkins, test this please

dimakuv:
Jenkins, test this please

mkow:
Jenkins, retest this please

PR #1666, opened 2020-07-10
[Tools/gsc] More resilient naming options for GSC

dayeol:
Currently, GSC generates two images: signed and non-signed,
whose names are determined automatically by adding the prefix "gsc-" and
the postfix "-unsigned", respectively.
This is less flexible, not very consistent (if you include tag
name), and also could not work in some cases if the image has a repository.
Instead of mixing prefix and postfix scheme, we add postfix to the image
name excluding the tag.
Also, we let users to set the name of the output image with '--name' or '-o' flag.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1666)

### DIFF
--- a/Tools/gsc/gsc.py
+++ b/Tools/gsc/gsc.py
@@ -14,11 +14,34 @@ import jinja2
 import docker
 import yaml
 
-def gsc_image_name(name):
-    return f'gsc-{name}'
+output_image_name = ''
+unsigned_image_name = ''
 
-def gsc_unsigned_image_name(name):
-    return f'gsc-{name}-unsigned'
+def split_image_tag(name):
+    image = ''
+    tag = ''
+
+    try:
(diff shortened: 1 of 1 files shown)


dimakuv:
Looks good to me, but summoning @vahldiek to review this PR.

dimakuv:
@vahldiek Do you think this PR is still valid? Or should we close it? I guess @dayeol won't be available to finish this PR...

vahldiek:
Its valid in the sense that there might be someone out there who prefers to name images. So far I've not gotten this request directly. Given that this PR is a bit old, I'm not sure how much sense it makes, since it needs serious rebase.

dimakuv:
Thank you @vahldiek. It looks like this PR is (a) quite outdated and (b) doesn't have a particular use case. Let's close it.

PR #1667, opened 2020-07-10
Protected env

### LINKED ISSUE #508: Provide a way to pass verified argv and environments

mkow:
## Description of the changes 

This is similar to #1562, but for environment variables. It implements a way to pass environment variables from a file, with intention to use it in conjunction with either trusted or protected files.

Fixes #508.

## How to test this PR? 

Two new tests were added for the new functionality.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1667)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -75,7 +75,8 @@ or
 
 If you want your application to use commandline arguments you need to either set
 ``loader.insecure__use_cmdline_argv`` (insecure in almost all cases) or point
-``loader.argv_src_file`` to a file containing output of :file:`Tools/argv_serializer`.
+``loader.argv_src_file`` to a file containing output of
+:file:`Tools/argv_serializer`.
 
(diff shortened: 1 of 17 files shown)


mkow:
Jenkins, retest Jenkins-SGX-18.04 please (`test_libos.TC_02_OpenMP.test_000_simple_for_loop` crashed, probably the same bug as the one fixed in #1663)

mkow:
Jenkins, retest Jenkins-Debug-18.04 please (`apps.LTP.select04` from LTP timed out, interesting, but probably unrelated. I'll try to reproduce this locally)

mkow:
Ok, `apps.LTP.select04` takes over 26s on my machine (it sleeps a lot, so it's not a Graphene-related issue), so no wonder it sometimes times out. I'll submit a PR raising the timeout for this test.

PR #1672, opened 2020-07-14
support application running in Graphene-SGX to process FIFO files which exist in host

AntonioDan:
This patch aims to support applications running in Graphene-SGX instance to process FIFO file which exists in host. 

Why do we make this change?
I have two applications which use FIFO to communicate with each other. The FIFO file has been created on the host OS using mknod. When trying to run the two applications in Graphene "Linux-SGX" PAL, I meet below two problems, which cause communication failure:
a. when application running in Graphene-SGX tries to open FIFO file (which exists in host) with O_WRONLY flag, Graphene-SGX instance would get hang. This is different behavior when we run the application in native environment or Graphen "Linux" PAL.  
b. when application running in Graphene-SGX tries to read a FIFO file, Graphene would do pread() with uri handle, however, pread() would return ESPIPE failure when the handle represents FIFO file. This causes failure to read the FIFO file. There is similar issue when trying to write a FIFO file.
When running the two applications in Graphene "Linux", there is not such issue and the communication can succeed.

After debugging, I found above issues are caused by below issues:
a. when trying to open file, Graphene "Linux-SGX" PAL would do ocall_open(uri, 0, 0) to get real file's stat. When the file is FIFO, this open() gets hung. According to Glibc document, this is because, if the file is FIFO, open() with READ flag would hold until the the FIFO is opened by another process with WRITE flag. However, in graphene-SGX, this means dead lock, see, the test application tries to open() the FIFO with O_WRONLY flag, but inside Linux-SGX PAL, it would call open(uri, 0, 0) and waits for FIFO write side is open...
b. Graphene use pread()/pwrite() to support read/write() operation, however, if it's FIFO file which doesn't support lseek(), pread/pwrite would return failure with ESPIPE, this causes read/write() failure.

Below is patch summary:
a. add an OCALL stat() to support querying file stat; modify Linux-SGX PAL to use ocall_stat() instead of ocall_open() following with ocall_fstat() to get file stat. This resolves Graphene-SGX hung issue when tring to open FIFO file.
b. Graphene would try read()/write() function when it finds pread()/pwrite() fail with error code ESPIPE.

How to test this PR?
There is a test application in https://github.com/AntonioDan/graphene.git, under Examples/fifotest.
After pulling the code, we can test like this:
a. go to "Examples/fifotest"
b. make SGX=1
c. after build pass, run it like this:
    (1) run "SGX=1 ./pal_loader readpipe testpipe"
    (2) in another console, go to same directory, run "SGX=1 ./pal_loader writepipe testpipe"
    When it run successfully, you can see the first console print out a "Hello, peer!" and exit.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1672)


dimakuv:
For reference: this PR was incorrectly created and thus closed. It was superseded by https://github.com/oscarlab/graphene/pull/1673.

PR #1673, opened 2020-07-14
support applications running in Graphene-SGX to process FIFO files which exist in host

AntonioDan:
This patch aims to support applications running in Graphene-SGX to process FIFO files which exist in host.

## Description of the changes 
I have two test applications which use FIFO to communicate with each other. The FIFO file has been created on the host OS with mknod. When trying to run the two applications in Graphene "Linux-SGX" PAL, I meet below problems, which cause communication failure:
a. when application running in Graphene-SGX ("Linux-SGX" PAL) tries to open FIFO file (which exists in host) with O_WRONLY access, Graphene-SGX instance would get hang. This is because PAL would try to open the file with all-zero flag so to get file stat, but this API would get hung when the file is FIFO.
b. when application running in Graphene-SGX tries to read on FIFO file, PAL would would call pread(), however, this API would return ESPIPE error, as a result, it fails to read the file. Same issue exists when the application tries to write FIFO file.
When running the two applications in Graphene "Linux" PAL, there is not such issue and the communication can succeed.

## How to test this PR? 
The test application is in https://github.com/AntonioDan/graphene.git, under "Examples/fifotest". After pulling the code, you can test like below:
a. go to "Example/fifotest", run "SGX=1 make"
b. after build is complete, run "SGX=1 ./pal_loader readpipe testpipe"
c. open a new console, go to same directory, run "SGX=1 ./pal_loader writepipe testpipe"
When it runs successfully, the first console would print "Hello, Peer!".

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1673)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_files.c
+++ b/Pal/src/host/Linux-SGX/db_files.c
@@ -52,6 +52,7 @@ static int file_open(PAL_HANDLE* handle, const char* type, const char* uri, int 
     hdl->file.realpath = (PAL_STR)path;
 
     struct protected_file* pf = get_protected_file(path);
+    struct stat st;
     /* whether to re-initialize the PF */
     bool pf_create = (create & PAL_CREATE_ALWAYS) || (create & PAL_CREATE_TRY);
 
@@ -66,6 +67,16 @@ static int file_open(PAL_HANDLE* handle, const char* type, const char* uri, int 
     }
 
     hdl->file.fd = fd;
+
+    /* check if the file is seekable and get real file size */
+    ret = ocall_fstat(fd, &st);
+    if (IS_ERR(ret)) {
+        SGX_DBG(DBG_E, "file_open(%s): fstat failed: %d\n", path, ret);
+        ret = unix_to_pal_error(ERRNO(ret));
+        goto out;
+    }
+
+    hdl->file.seekable = !S_ISFIFO(st.st_mode);
 
     if (pf) {
         pf_file_mode_t pf_mode = 0;
@@ -85,16 +96,14 @@ static int file_open(PAL_HANDLE* handle, const char* type, const char* uri, int 
             }
         }
 
-        /* get real file size */
-        struct stat st;
-        ret = ocall_fstat(fd, &st);
-        if (IS_ERR(ret)) {
-            SGX_DBG(DBG_E, "file_open(%s): fstat failed: %d\n", path, ret);
-            ret = unix_to_pal_error(ERRNO(ret));
+        ret = -PAL_ERROR_DENIED;
+
+        /* the protected files should be regular files (seekable) */
+        if (!hdl->file.seekable) {
+            SGX_DBG(DBG_E, "file_open(%s): disallowing non-seekable file handle\n", path);
             goto out;
         }
 
-        ret = -PAL_ERROR_DENIED;
         pf = load_protected_file(path, (int*)&hdl->file.fd, st.st_size, pf_mode, pf_create, pf);
         if (pf) {
             pf->refcount++;
@@ -111,10 +120,9 @@ static int file_open(PAL_HANDLE* handle, const char* type, const char* uri, int 
         void* umem;
         ret = load_trusted_file(hdl, &stubs, &total, create, &umem);
         if (ret < 0) {
-            SGX_DBG(DBG_E,
-                    "Accessing file:%s is denied. (%s) "
-                    "This file is not trusted or allowed.\n",
-                    hdl->file.realpath, pal_strerror(ret));
+            SGX_DBG(DBG_E, "Accessing file:%s is denied (%s). This file is not trusted or allowed."
+                    " Trusted files should be regular files (seekable).\n", hdl->file.realpath,
+                    pal_strerror(ret));
             goto out;
         }
 
@@ -173,9 +181,15 @@ static int64_t file_read(PAL_HANDLE handle, uint64_t offset, uint64_t count, voi
     sgx_stub_t* stubs = (sgx_stub_t*)handle->file.stubs;
 
     if (!stubs) {
-        ret = ocall_pread(handle->file.fd, buffer, count, offset);
+        if (handle->file.seekable) {
+            ret = ocall_pread(handle->file.fd, buffer, count, offset);
+        } else {
+            ret = ocall_read(handle->file.fd, buffer, count);
+        }
+
         if (IS_ERR(ret))
             return unix_to_pal_error(ERRNO(ret));
+
         return ret;
     }
 
@@ -230,9 +244,15 @@ static int64_t file_write(PAL_HANDLE handle, uint64_t offset, uint64_t count, co
     sgx_stub_t* stubs = (sgx_stub_t*)handle->file.stubs;
 
     if (!stubs) {
-        ret = ocall_pwrite(handle->file.fd, buffer, count, offset);
+        if (handle->file.seekable) {
+            ret = ocall_pwrite(handle->file.fd, buffer, count, offset);
+        } else {
+            ret = ocall_write(handle->file.fd, buffer, count);
+        }
+
         if (IS_ERR(ret))
             return unix_to_pal_error(ERRNO(ret));
+
         return ret;
     }
 
@@ -550,8 +570,9 @@ static int file_attrquery(const char* type, const char* uri, PAL_STREAM_ATTR* at
     if (strcmp_static(type, URI_TYPE_FILE) && strcmp_static(type, URI_TYPE_DIR))
         return -PAL_ERROR_INVAL;
 
-    /* try to do the real open */
-    int fd = ocall_open(uri, 0, 0);
+    /* open the file with O_NONBLOCK to avoid blocking the current thread if it is actually a FIFO pipe;
+     * O_NONBLOCK will be reset below if it is a regular file */
+    int fd = ocall_open(uri, O_NONBLOCK, 0);
     if (IS_ERR(fd))
         return unix_to_pal_error(ERRNO(fd));
 
@@ -576,8 +597,22 @@ static int file_attrquery(const char* type, const char* uri, PAL_STREAM_ATTR* at
 
     /* For protected files return the data size, not real FS size */
     struct protected_file* pf = get_protected_file(path);
-    if (pf && attr->handle_type != pal_type_dir)
+    if (pf && attr->handle_type != pal_type_dir) {
+        /* protected files should be regular files */
+        if (S_ISFIFO(stat_buf.st_mode)) {
+            ret = -PAL_ERROR_DENIED;
+            goto out;
+        }
+
+        /* reset O_NONBLOCK because pf_file_attrquery() may issue reads which don't expect non-blocking mode */
+        ret = ocall_fsetnonblock(fd, 0);
+        if (IS_ERR(ret)) {
+            ret = unix_to_pal_error(ERRNO(ret));
+            goto out;
+        }
+
         ret = pf_file_attrquery(pf, fd, path, stat_buf.st_size, attr);
+    }
     else
         ret = 0;
 
@@ -601,6 +636,10 @@ static int file_attrquerybyhdl(PAL_HANDLE handle, PAL_STREAM_ATTR* attr) {
         /* For protected files return the data size, not real FS size */
         struct protected_file* pf = find_protected_file_handle(handle);
         if (pf) {
+            /* protected files should be regular files (seekable) */
+            if (!handle->file.seekable)
+                return -PAL_ERROR_DENIED;
+
             uint64_t size;
             pf_status_t pfs = pf_get_size(pf->context, &size);
             __UNUSED(pfs);
--- a/Pal/src/host/Linux-SGX/enclave_framework.c
+++ b/Pal/src/host/Linux-SGX/enclave_framework.c
@@ -364,6 +364,10 @@ int load_trusted_file (PAL_HANDLE file, sgx_stub_t ** stubptr,
         return 0;
     }
 
+    /* trusted file must be a regular file (seekable) */
+    if (!file->file.seekable)
+        return -PAL_ERROR_DENIED;
+
     sgx_stub_t* stubs = NULL;
     /* mmap the whole trusted file in untrusted memory for future reads/writes; it is
      * caller's responsibility to unmap those areas after use */
--- a/Pal/src/host/Linux-SGX/pal_host.h
+++ b/Pal/src/host/Linux-SGX/pal_host.h
@@ -83,6 +83,7 @@ typedef struct pal_handle
             /* below fields are used only for trusted files */
             PAL_PTR stubs;    /* contains hashes of file chunks */
             PAL_PTR umem;     /* valid only when stubs != NULL */
+            PAL_BOL seekable; /* regular files are seekable, FIFO pipes are not */
         } file;
 
         struct {
--- a/Pal/src/host/Linux/db_files.c
+++ b/Pal/src/host/Linux/db_files.c
@@ -51,15 +51,35 @@ static int file_open(PAL_HANDLE* handle, const char* type, const char* uri, int 
     /* if try_create_path succeeded, prepare for the file handle */
     size_t uri_size = strlen(uri) + 1;
     PAL_HANDLE hdl = malloc(HANDLE_SIZE(file) + uri_size);
+    if (!hdl) {
+        INLINE_SYSCALL(close, 1, ret);
+        return -PAL_ERROR_NOMEM;
+    }
+
     SET_HANDLE_TYPE(hdl, file);
     HANDLE_HDR(hdl)->flags |= RFD(0)|WFD(0);
     hdl->file.fd = ret;
     hdl->file.map_start = NULL;
     char* path = (void*)hdl + HANDLE_SIZE(file);
     ret = get_norm_path(uri, path, &uri_size);
-    if (ret < 0)
+    if (ret < 0) {
+        INLINE_SYSCALL(close, 1, hdl->file.fd);
+        free(hdl);
         return ret;
+    }
+
     hdl->file.realpath = (PAL_STR)path;
+
+    struct stat st;
+    ret = INLINE_SYSCALL(fstat, 2, hdl->file.fd, &st);
+    if (IS_ERR(ret)) {
+        INLINE_SYSCALL(close, 1, hdl->file.fd);
+        free(hdl);
+        return unix_to_pal_error(ERRNO(ret));
+    }
+
+    hdl->file.seekable = !S_ISFIFO(st.st_mode);
+
     *handle = hdl;
     return 0;
 }
@@ -71,7 +91,11 @@ static int64_t file_read (PAL_HANDLE handle, uint64_t offset, uint64_t count,
     int fd = handle->file.fd;
     int64_t ret;
 
-    ret = INLINE_SYSCALL(pread64, 4, fd, buffer, count, offset);
+    if (handle->file.seekable) {
+        ret = INLINE_SYSCALL(pread64, 4, fd, buffer, count, offset);
+    } else {
+        ret = INLINE_SYSCALL(read, 3, fd, buffer, count);
+    }
 
     if (IS_ERR(ret))
         return unix_to_pal_error(ERRNO(ret));
@@ -86,7 +110,11 @@ static int64_t file_write (PAL_HANDLE handle, uint64_t offset, uint64_t count,
     int fd = handle->file.fd;
     int64_t ret;
 
-    ret = INLINE_SYSCALL(pwrite64, 4, fd, buffer, count, offset);
+    if (handle->file.seekable) {
+        ret = INLINE_SYSCALL(pwrite64, 4, fd, buffer, count, offset);
+    } else {
+        ret = INLINE_SYSCALL(write, 3, fd, buffer, count);
+    }
 
     if (IS_ERR(ret))
         return unix_to_pal_error(ERRNO(ret));
--- a/Pal/src/host/Linux/pal_host.h
+++ b/Pal/src/host/Linux/pal_host.h
@@ -73,6 +73,7 @@ typedef struct pal_handle
              * determined by parent process.
              */
             PAL_PTR map_start;
+            PAL_BOL seekable; /* regular files are seekable, FIFO pipes are not */
         } file;
 
         struct {


AntonioDan:
> > _[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1673)_ status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @AntonioDan, @dimakuv, and @mkow)
> 
> _[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCEDAoJClbFqiyCNRQ_:-MCEIKSZ4vPcv58XuZsz:bgwrnrl) (no related file):_
> 
> Previously, dimakuv (Dmitrii Kuvaiskii) wrote…
> The goal is to communicate non-graphene process (or another graphene process).
> Then, we can consider it yet another socket communication.
> 
> The current implementation seems assuming that file is a regular file. it needs to be audited.
> 
> _[Pal/src/host/Linux/db_files.c, line 85 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCED0l65l7wYb7i0lOy:-MCEJLWnDzlANjGfAll6:b5mnw3h) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux/db_files.c#L85)):_
> 
> Previously, dimakuv (Dmitrii Kuvaiskii) wrote…
> I would add seekable member to file handle. i.e. something like
> 
> ```
> if (file_handle->seekable)
>   pread()
> else
>   read()
> ```
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 72 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCEKwTO9mmtExsppD26:-MCEKwTO9mmtExsppD27:b-lw67au) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux-SGX/db_files.c#L72)):_
> 
> > ```c
> >     hdl->file.fd = fd;
> > ```
> 
> protected files or load_trusted_file() are assuming that the file is a regular file.
> new check needs to be added.
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 569 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCEByMtANjYgFHVoRVG:-MCEJnrT4zYkhQiKYCjk:bvwtzbl) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux-SGX/db_files.c#L569)):_
> 
> Previously, dimakuv (Dmitrii Kuvaiskii) wrote…
> > FYI: I like this fix. Even if we decide against this PR as a whole, I think we should have this fix -- just to be on the safe side when querying the attributes of a host file.
> 
> +1. some comment on it is desirable.

I agree it's better to check if the file is seekable or not. I have updated this patch accordingly, can you please review it again?
I implement below in this patch:
a. add a ocall_lseek() so we can check if the file is checkable. I don't find such an ocall in Linux-SGX PAL, so I add this.
b. in file_open(), use ocall_lseek() to check if the file is seekable.
c. add a PAL_BOL type variable "seekable" in pal_handle file.
d. in file_read(), we check if it's seekable, and call pread() / read() accordingly
e. in file_write(), we check if it's seekable and call pwrite() / write() accordingly

AntonioDan:
> > Reviewed 2 of 2 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1673)_ status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @AntonioDan)
> 
> _[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCEDAoJClbFqiyCNRQ_:-MCEDAoJClbFqiyCNRQa:b-diib5b) (no related file):_
> I like this PR! I think it doesn't introduce any security issues, and allows to use FIFO (named) pipes via `sgx.allowed_files = file:<my-fifo-pipe>`.
> 
> There is an issue of: why the user would want to communicate with the outside world using the FIFO pipe (and then be extremely careful to not leak any secrets on this pipe or encrypt everything at the application level). But on the other hand, we can treat FIFO pipes just like TCP/UDP socket communication.
> 
> @yamahata @mkow What do you think?
> 
> _[Pal/src/host/Linux/db_files.c, line 85 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCED0l65l7wYb7i0lOy:-MCED0l65l7wYb7i0lOz:b-signxt) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux/db_files.c#L85)):_
> 
> > ```c
> >             return unix_to_pal_error(ERRNO(ret));
> >         }
> >     }
> > ```
> 
> See my comments on security in the same logic in Linux-SGX. Here, I think it's totally reasonable to use this trick.
> 
> _[Pal/src/host/Linux/db_files.c, line 108 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCED8z-61yFFUxKPzlr:-MCED8z-61yFFUxKPzls:b-signxt) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux/db_files.c#L108)):_
> 
> > ```c
> >             return unix_to_pal_error(ERRNO(ret));
> >         }
> >     }
> > ```
> 
> See my comments on security in the same logic in Linux-SGX. Here, I think it's totally reasonable to use this trick.
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 188 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCECFp9Bt1-v1_A3of4:-MCECFp9Bt1-v1_A3of5:bhbps24) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux-SGX/db_files.c#L188)):_
> 
> > ```c
> >                 return unix_to_pal_error(ERRNO(ret));
> >             }
> >         }
> > ```
> 
> I generally like this trick. Note that we end up in this logic only for `sgx.allowed_files` (because Protected Files are handled above, and Trusted Files have `stubs != NULL` and are handled below). I think if the user marks some file as allowed, then the user kinda expects that security is non-existent or achieved by non-Graphene means. So even if the malicious host plays tricks in this new logic, who cares.
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 254 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCECuFsC0_5nPD5gGhB:-MCECuFsC0_5nPD5gGhC:btj9953) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux-SGX/db_files.c#L254)):_
> 
> > ```c
> >                 return unix_to_pal_error(ERRNO(ret));
> >             }
> >         }
> > ```
> 
> I like this trick, see my comment above.
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 569 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCEByMtANjYgFHVoRVG:-MCEByMtANjYgFHVoRVH:b-njzhnz) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux-SGX/db_files.c#L569)):_
> 
> > ```c
> >     /* try to do the real open */
> >     int fd = ocall_open(uri, O_NONBLOCK, 0);
> > ```
> 
> FYI: I like this fix. Even if we decide against this PR as a whole, I think we should have this fix -- just to be on the safe side when querying the attributes of a host file.

Yes, we can treat FIFO pipes just like TCP/UDP socket communication and avoid surprising application when it happens to use a FIFO file in host. The application should encrypt the content or be cautious to use named pipe.

I updated the patch to use check if the file is seekable instead of using 'ESPIPE' trick, can you please review the patch again? Please point out if you see anything else I need to update. I like Graphene and hope to make it even better. :)

AntonioDan:
> > _[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1673)_ status: 0 of 8 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @AntonioDan, @dimakuv, @mkow, and @yamahata)
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 569 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCEByMtANjYgFHVoRVG:-MCIHoD45nbh7DyIMBRM:b-avlv4) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux-SGX/db_files.c#L569)):_
> 
> Previously, AntonioDan wrote…
> I agree with adding `seekable`, but I vehemently dislike introducing the new OCALL `ocall_lseek()`.
> 
> We can check if this is a FIFO using the already existing `ocall_fstat()`: just check if the returned struct evaluates `S_ISFIFO(stat->st_mode) == true`. And please remove `ocall_lseek()` completely.

Accept. I have updated the patch to remove ocall_lseek() and use ocall_fstat() to check if it's FIFO, as you point out. Can you re-review the patch?

AntonioDan:
> protected files or load_trusted_file() are assuming that the file is a regular file.
> new check needs to be added.

I have updated patch to check protected files and trusted files should be regular file, can you re-review the patch? Thanks!

AntonioDan:
> > Reviewed 4 of 4 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1673)_ status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @AntonioDan and @mkow)
> 
> _[Pal/src/host/Linux/db_files.c, line 61 at r2](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCLCaVA60aEsoRzmgnE:-MCLCaVA60aEsoRzmgnF:b-wp75bg) ([raw file](https://github.com/oscarlab/graphene/blob/7f5cb2b8bc7be3941f909825a8737531f1f82710/Pal/src/host/Linux/db_files.c#L61)):_
> 
> > ```c
> >     ret = get_norm_path(uri, path, &uri_size);
> >     if (ret < 0) {
> >         INLINE_SYSCALL(close, 1, hdl->file.fd);
> > ```
> 
> Nice! Thanks for adding proper error handling :)
> 
> _[Pal/src/host/Linux/db_files.c, line 73 at r2](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCLCi2W0RFaSs4qCEbv:-MCLCi2W0RFaSs4qCEbw:b-55g04b) ([raw file](https://github.com/oscarlab/graphene/blob/7f5cb2b8bc7be3941f909825a8737531f1f82710/Pal/src/host/Linux/db_files.c#L73)):_
> 
> > ```c
> >         return unix_to_pal_error(ERRNO(ret));
> >     }
> >    
> > ```
> 
> Please remove trailing whitespaces from all files
> 
> _[Pal/src/host/Linux/pal_host.h, line 76 at r2](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCLCzqz6tE2emXepdME:-MCLCzqz6tE2emXepdMF:b-v31fyq) ([raw file](https://github.com/oscarlab/graphene/blob/7f5cb2b8bc7be3941f909825a8737531f1f82710/Pal/src/host/Linux/pal_host.h#L76)):_
> 
> > ```objc
> >              */
> >             PAL_PTR map_start;
> >             PAL_BOL seekable; /* true if the file is seekable */
> > ```
> 
> The comment is obvious. I suggest to change to: `/* regular files are seekable, FIFO pipes are not */`
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 569 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCEByMtANjYgFHVoRVG:-MCLEWmK2orV1T7Hv6BP:b-vkox7f) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux-SGX/db_files.c#L569)):_
> 
> Previously, AntonioDan wrote…
> I reviewed the patch. Thanks! One more iteration, and the PR will hopefully be merged.
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 84 at r2](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCLDc1e5g1kfIKl4n5X:-MCLDc1fETV_TzOPek2n:b9uoy40) ([raw file](https://github.com/oscarlab/graphene/blob/7f5cb2b8bc7be3941f909825a8737531f1f82710/Pal/src/host/Linux-SGX/db_files.c#L84)):_
> 
> > ```c
> >     hdl->file.seekable = S_ISFIFO(st.st_mode) ? false : true;
> > 
> >     if (pf) {
> > ```
> 
> This `if (pf)` means that the below logic is for Protected Files (transparently encrypted by Graphene). Please add a check here that `if (hdl->file.seekable) { <debug message and return PAL_ERROR_DENIED> }` because protected files cannot be FIFOs.
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 185 at r2](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCLE6KH1fGMqW56QKaG:-MCLE6KH1fGMqW56QKaH:b-uu4ra8) ([raw file](https://github.com/oscarlab/graphene/blob/7f5cb2b8bc7be3941f909825a8737531f1f82710/Pal/src/host/Linux-SGX/db_files.c#L185)):_
> 
> > ```c
> >             ret = ocall_read(handle->file.fd, buffer, count);
> >         }
> >         
> > ```
> 
> Please remove trailing whitespaces in all files



> > Reviewed 2 of 2 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1673)_ status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @AntonioDan)
> 
> _[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCEDAoJClbFqiyCNRQ_:-MCEDAoJClbFqiyCNRQa:b-diib5b) (no related file):_
> I like this PR! I think it doesn't introduce any security issues, and allows to use FIFO (named) pipes via `sgx.allowed_files = file:<my-fifo-pipe>`.
> 
> There is an issue of: why the user would want to communicate with the outside world using the FIFO pipe (and then be extremely careful to not leak any secrets on this pipe or encrypt everything at the application level). But on the other hand, we can treat FIFO pipes just like TCP/UDP socket communication.
> 
> @yamahata @mkow What do you think?
> 
> _[Pal/src/host/Linux/db_files.c, line 85 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCED0l65l7wYb7i0lOy:-MCED0l65l7wYb7i0lOz:b-signxt) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux/db_files.c#L85)):_
> 
> > ```c
> >             return unix_to_pal_error(ERRNO(ret));
> >         }
> >     }
> > ```
> 
> See my comments on security in the same logic in Linux-SGX. Here, I think it's totally reasonable to use this trick.
> 
> _[Pal/src/host/Linux/db_files.c, line 108 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCED8z-61yFFUxKPzlr:-MCED8z-61yFFUxKPzls:b-signxt) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux/db_files.c#L108)):_
> 
> > ```c
> >             return unix_to_pal_error(ERRNO(ret));
> >         }
> >     }
> > ```
> 
> See my comments on security in the same logic in Linux-SGX. Here, I think it's totally reasonable to use this trick.
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 188 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCECFp9Bt1-v1_A3of4:-MCECFp9Bt1-v1_A3of5:bhbps24) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux-SGX/db_files.c#L188)):_
> 
> > ```c
> >                 return unix_to_pal_error(ERRNO(ret));
> >             }
> >         }
> > ```
> 
> I generally like this trick. Note that we end up in this logic only for `sgx.allowed_files` (because Protected Files are handled above, and Trusted Files have `stubs != NULL` and are handled below). I think if the user marks some file as allowed, then the user kinda expects that security is non-existent or achieved by non-Graphene means. So even if the malicious host plays tricks in this new logic, who cares.
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 254 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCECuFsC0_5nPD5gGhB:-MCECuFsC0_5nPD5gGhC:btj9953) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux-SGX/db_files.c#L254)):_
> 
> > ```c
> >                 return unix_to_pal_error(ERRNO(ret));
> >             }
> >         }
> > ```
> 
> I like this trick, see my comment above.
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 569 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCEByMtANjYgFHVoRVG:-MCEByMtANjYgFHVoRVH:b-njzhnz) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux-SGX/db_files.c#L569)):_
> 
> > ```c
> >     /* try to do the real open */
> >     int fd = ocall_open(uri, O_NONBLOCK, 0);
> > ```
> 
> FYI: I like this fix. Even if we decide against this PR as a whole, I think we should have this fix -- just to be on the safe side when querying the attributes of a host file.

AntonioDan:
> > Reviewed 4 of 4 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1673)_ status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @AntonioDan and @mkow)
> 
> _[Pal/src/host/Linux/db_files.c, line 61 at r2](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCLCaVA60aEsoRzmgnE:-MCLCaVA60aEsoRzmgnF:b-wp75bg) ([raw file](https://github.com/oscarlab/graphene/blob/7f5cb2b8bc7be3941f909825a8737531f1f82710/Pal/src/host/Linux/db_files.c#L61)):_
> 
> > ```c
> >     ret = get_norm_path(uri, path, &uri_size);
> >     if (ret < 0) {
> >         INLINE_SYSCALL(close, 1, hdl->file.fd);
> > ```
> 
> Nice! Thanks for adding proper error handling :)
> 
> _[Pal/src/host/Linux/db_files.c, line 73 at r2](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCLCi2W0RFaSs4qCEbv:-MCLCi2W0RFaSs4qCEbw:b-55g04b) ([raw file](https://github.com/oscarlab/graphene/blob/7f5cb2b8bc7be3941f909825a8737531f1f82710/Pal/src/host/Linux/db_files.c#L73)):_
> 
> > ```c
> >         return unix_to_pal_error(ERRNO(ret));
> >     }
> >    
> > ```
> 
> Please remove trailing whitespaces from all files
> 
> _[Pal/src/host/Linux/pal_host.h, line 76 at r2](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCLCzqz6tE2emXepdME:-MCLCzqz6tE2emXepdMF:b-v31fyq) ([raw file](https://github.com/oscarlab/graphene/blob/7f5cb2b8bc7be3941f909825a8737531f1f82710/Pal/src/host/Linux/pal_host.h#L76)):_
> 
> > ```objc
> >              */
> >             PAL_PTR map_start;
> >             PAL_BOL seekable; /* true if the file is seekable */
> > ```
> 
> The comment is obvious. I suggest to change to: `/* regular files are seekable, FIFO pipes are not */`
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 569 at r1](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCEByMtANjYgFHVoRVG:-MCLEWmK2orV1T7Hv6BP:b-vkox7f) ([raw file](https://github.com/oscarlab/graphene/blob/2cca7215821134c8e0d67ff0ca13b024f07e4663/Pal/src/host/Linux-SGX/db_files.c#L569)):_
> 
> Previously, AntonioDan wrote…
> I reviewed the patch. Thanks! One more iteration, and the PR will hopefully be merged.
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 84 at r2](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCLDc1e5g1kfIKl4n5X:-MCLDc1fETV_TzOPek2n:b9uoy40) ([raw file](https://github.com/oscarlab/graphene/blob/7f5cb2b8bc7be3941f909825a8737531f1f82710/Pal/src/host/Linux-SGX/db_files.c#L84)):_
> 
> > ```c
> >     hdl->file.seekable = S_ISFIFO(st.st_mode) ? false : true;
> > 
> >     if (pf) {
> > ```
> 
> This `if (pf)` means that the below logic is for Protected Files (transparently encrypted by Graphene). Please add a check here that `if (hdl->file.seekable) { <debug message and return PAL_ERROR_DENIED> }` because protected files cannot be FIFOs.
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 185 at r2](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCLE6KH1fGMqW56QKaG:-MCLE6KH1fGMqW56QKaH:b-uu4ra8) ([raw file](https://github.com/oscarlab/graphene/blob/7f5cb2b8bc7be3941f909825a8737531f1f82710/Pal/src/host/Linux-SGX/db_files.c#L185)):_
> 
> > ```c
> >             ret = ocall_read(handle->file.fd, buffer, count);
> >         }
> >         
> > ```
> 
> Please remove trailing whitespaces in all files

I have updated the patch to remove trailing whitespace and update the comments. Please review latest patch. Thanks!

AntonioDan:
> > Reviewed 4 of 4 files at r3.
> > _[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1673)_ status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @AntonioDan and @mkow)
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 129 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCNz-y-4j-5APxX9QiL:-MCNz-y-4j-5APxX9QiM:b59jbfb) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux-SGX/db_files.c#L129)):_
> 
> > ```c
> >             SGX_DBG(DBG_E,
> >                     "Accessing file:%s is denied. (%s) "
> >                     "This file is not trusted or allowed. Trust files should be seekable.\n",
> > ```
> 
> `Trust files` -> `Trusted files`
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 628 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCNzXci9qz50RI_GWkR:-MCNzXci9qz50RI_GWkS:byrzc7) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux-SGX/db_files.c#L628)):_
> 
> > ```c
> >         /* for protected files, it should be regular file */
> >         if (!handle->file.seekable)
> >             return -PAL_ERROR_DENIED;
> > ```
> 
> This should be moved under `if (pf) { ...`
> 
> _[Pal/src/host/Linux-SGX/enclave_framework.c, line 367 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCNzl665hHHgbUIT97z:-MCNzl665hHHgbUIT98-:b1wliww) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux-SGX/enclave_framework.c#L367)):_
> 
> > ```c
> >     }
> > 
> >     /* assume the trusted file should be seekable */
> > ```
> 
> Why `assume`? Modify to `/* trusted file must be a regular file (seekable) */`

Accept. I have updated patch accordingly. Thanks for your comments!

AntonioDan:
> > Reviewed 1 of 4 files at r2, 4 of 4 files at r3.
> > _[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1673)_ status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @AntonioDan and @dimakuv)
> 
> _[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCEDAoJClbFqiyCNRQ_:-MCP53X6C2rOTlx8tFl2:b-fqmx73) (no related file):_
> As Isaku said, IMO there are many legitimate use-cases for this, although people may use it incorrectly (but they can do the same with sockets).
> 
> > The current implementation seems assuming that file is a regular file. it needs to be audited.
> 
> Yup, I think we have this incorrect assumption in tons of places in the old code. This is a part of "filesystem support rewrite" we need to do at some point.
> 
> _[Pal/src/host/Linux/db_files.c, line 61 at r2](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCLCaVA60aEsoRzmgnE:-MCP8Va0AXPM8mSDfHh_:bijouu6) ([raw file](https://github.com/oscarlab/graphene/blob/7f5cb2b8bc7be3941f909825a8737531f1f82710/Pal/src/host/Linux/db_files.c#L61)):_
> 
> Previously, dimakuv (Dmitrii Kuvaiskii) wrote…
> Actually, there are more wrongdoings in this function's error handling.
> 
> @AntonioDan While we're here, would you have time to fix the whole error handling here? Currently also `hdl` buffer leaks (same below). Also, malloc is not checked, but we can fix this on another occasion.
> 
> _[Pal/src/host/Linux/db_files.c, line 74 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCPB6JoEuI_01Jczyz4:-MCPB6JoEuI_01Jczyz5:b-a3u3uf) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux/db_files.c#L74)):_
> 
> > ```c
> >     }
> > 
> >     hdl->file.seekable = S_ISFIFO(st.st_mode) ? false : true;
> > ```
> 
> `S_ISFIFO(st.st_mode) ? false : true` -> `!S_ISFIFO(st.st_mode)`
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 82 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCPBCqo0upx2Apjrfz_:-MCPBCqo0upx2Apjrfza:b-a3u3uf) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux-SGX/db_files.c#L82)):_
> 
> > ```c
> >     }
> > 
> >     hdl->file.seekable = S_ISFIFO(st.st_mode) ? false : true;
> > ```
> 
> `S_ISFIFO(st.st_mode) ? false : true` -> `!S_ISFIFO(st.st_mode)`
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 102 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCPBMo-E5bWSJbRkkx0:-MCPBMo01SEUGKGVER7h:b38n1f5) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux-SGX/db_files.c#L102)):_
> 
> > ```c
> >         }
> > 
> >         /* the protected files should be regular file */
> > ```
> 
> `file` -> `files`
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 573 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCPCvxN30jeW9YLgzbj:-MCPCvxN30jeW9YLgzbk:bjmdwzw) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux-SGX/db_files.c#L573)):_
> 
> > ```c
> >     /* try to do the real open */
> >     int fd = ocall_open(uri, O_NONBLOCK, 0);
> > ```
> 
> Passing `O_NONBLOCK` to open has tricky semantics - now in addition to not blocking on `open()`, you left `fd` in non-blocking I/O mode. It can be further passed to `pf_file_attrquery()`, which may actually issue reads which don't expect the non-blocking mode.
> 
> _[Pal/src/host/Linux-SGX/enclave_framework.c, line 367 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCNzl665hHHgbUIT97z:-MCPE4h-6AZ6suTpe6zx:b-1su822) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux-SGX/enclave_framework.c#L367)):_
> 
> Previously, dimakuv (Dmitrii Kuvaiskii) wrote…
> But the `if` below actually checks if the file is seekable, not if it's a regular file. Not sure how to fix this without fixing (a.k.a. writing from scratch) the whole filesystem code in Graphene :)

I agree your comments about 'O_NONBLOCK'. We add this just to avoid blocking current thread when the uri is FIFO. I update patch to reset this flag when the uri is not FIFO, by this, it doesn't bring any runtime behavior change. Can you re-review the patch?

About "S_ISFIFO(st.st_mode) ? false : true` -> `!S_ISFIFO(st.st_mode)`", this is not intended behavior, the 'seekable' is true when the file is not FIFO. If I change the code as this, my test application would fail when trying to read a FIFO...

AntonioDan:
> > Reviewed 1 of 4 files at r2, 4 of 4 files at r3.
> > _[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1673)_ status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @AntonioDan and @dimakuv)
> 
> _[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCEDAoJClbFqiyCNRQ_:-MCP53X6C2rOTlx8tFl2:b-fqmx73) (no related file):_
> As Isaku said, IMO there are many legitimate use-cases for this, although people may use it incorrectly (but they can do the same with sockets).
> 
> > The current implementation seems assuming that file is a regular file. it needs to be audited.
> 
> Yup, I think we have this incorrect assumption in tons of places in the old code. This is a part of "filesystem support rewrite" we need to do at some point.
> 
> _[Pal/src/host/Linux/db_files.c, line 61 at r2](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCLCaVA60aEsoRzmgnE:-MCP8Va0AXPM8mSDfHh_:bijouu6) ([raw file](https://github.com/oscarlab/graphene/blob/7f5cb2b8bc7be3941f909825a8737531f1f82710/Pal/src/host/Linux/db_files.c#L61)):_
> 
> Previously, dimakuv (Dmitrii Kuvaiskii) wrote…
> Actually, there are more wrongdoings in this function's error handling.
> 
> @AntonioDan While we're here, would you have time to fix the whole error handling here? Currently also `hdl` buffer leaks (same below). Also, malloc is not checked, but we can fix this on another occasion.
> 
> _[Pal/src/host/Linux/db_files.c, line 74 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCPB6JoEuI_01Jczyz4:-MCPB6JoEuI_01Jczyz5:b-a3u3uf) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux/db_files.c#L74)):_
> 
> > ```c
> >     }
> > 
> >     hdl->file.seekable = S_ISFIFO(st.st_mode) ? false : true;
> > ```
> 
> `S_ISFIFO(st.st_mode) ? false : true` -> `!S_ISFIFO(st.st_mode)`
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 82 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCPBCqo0upx2Apjrfz_:-MCPBCqo0upx2Apjrfza:b-a3u3uf) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux-SGX/db_files.c#L82)):_
> 
> > ```c
> >     }
> > 
> >     hdl->file.seekable = S_ISFIFO(st.st_mode) ? false : true;
> > ```
> 
> `S_ISFIFO(st.st_mode) ? false : true` -> `!S_ISFIFO(st.st_mode)`
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 102 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCPBMo-E5bWSJbRkkx0:-MCPBMo01SEUGKGVER7h:b38n1f5) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux-SGX/db_files.c#L102)):_
> 
> > ```c
> >         }
> > 
> >         /* the protected files should be regular file */
> > ```
> 
> `file` -> `files`
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 573 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCPCvxN30jeW9YLgzbj:-MCPCvxN30jeW9YLgzbk:bjmdwzw) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux-SGX/db_files.c#L573)):_
> 
> > ```c
> >     /* try to do the real open */
> >     int fd = ocall_open(uri, O_NONBLOCK, 0);
> > ```
> 
> Passing `O_NONBLOCK` to open has tricky semantics - now in addition to not blocking on `open()`, you left `fd` in non-blocking I/O mode. It can be further passed to `pf_file_attrquery()`, which may actually issue reads which don't expect the non-blocking mode.
> 
> _[Pal/src/host/Linux-SGX/enclave_framework.c, line 367 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCNzl665hHHgbUIT97z:-MCPE4h-6AZ6suTpe6zx:b-1su822) ([raw file](https://github.com/oscarlab/graphene/blob/6f32c56538ab196b85416457f4dfce1e1da73412/Pal/src/host/Linux-SGX/enclave_framework.c#L367)):_
> 
> Previously, dimakuv (Dmitrii Kuvaiskii) wrote…
> But the `if` below actually checks if the file is seekable, not if it's a regular file. Not sure how to fix this without fixing (a.k.a. writing from scratch) the whole filesystem code in Graphene :)

I have updated patch to fix error handling issue in that function. My pleasure :)

AntonioDan:
> `S_ISFIFO(st.st_mode) ? false : true` -> `!S_ISFIFO(st.st_mode)`

I updated patch about "!S_ISFIFO(st.mode)". Can you re-review the patch? Thanks!

AntonioDan:
> > Reviewed 3 of 3 files at r4.
> > _[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1673)_ status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @AntonioDan)
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 574 at r4](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCSakWX-pp47uDIJJeW:-MCSakWY4-gGz96emxqU:b1iiy7z) ([raw file](https://github.com/oscarlab/graphene/blob/b575439578250655de6222120c124e955818638c/Pal/src/host/Linux-SGX/db_files.c#L574)):_
> 
> > ```c
> >     /* try to do the real open.
> >      * it sets O_NONBLOCK flag temporarily so to avoid blocking current thread if the uri is a FIFO file;
> >      * if the uri is actually regular file, we reset O_NONBLOCK flag */
> > ```
> 
> I would improve this whole comment as follows:
> 
> ```
> /* open the file with O_NONBLOCK to avoid blocking the current thread if it is actually a FIFO pipe;
>  * O_NONBLOCK will be reset below if it is a regular (protected) file */
> ```
> 
> _[Pal/src/host/Linux-SGX/db_files.c, line 607 at r4](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCSbKIM2BfrtSaslE0e:-MCSbKIM2BfrtSaslE0f:b-thrrt3) ([raw file](https://github.com/oscarlab/graphene/blob/b575439578250655de6222120c124e955818638c/Pal/src/host/Linux-SGX/db_files.c#L607)):_
> 
> > ```c
> >         }
> > 
> >         /* reset O_NONBLOCK flag since it's regular file */
> > ```
> 
> I would improve this comment like this:
> 
> ```
> /* reset O_NONBLOCK because pf_file_attrquery() may issue reads which don't expect non-blocking mode */
> ```
> 
> _[Pal/src/host/Linux-SGX/enclave_framework.c, line 367 at r3](https://reviewable.io/reviews/oscarlab/graphene/1673#-MCNzl665hHHgbUIT97z:-MCSbe7m7erTqMtJpFxN:b-oxumbk) ([raw file](https://github.com/oscarlab/graphene/blob/4aa56db6603688a69b50440625206bdec048a439/Pal/src/host/Linux-SGX/enclave_framework.c#L367)):_
> Previously, AntonioDan wrote…
> 
> I guess your text about `!IS_FIFO(...)` is outdated because you actually implemented this in your code? I guess you didn't notice the `!` (flip the result value) when you wrote the comment. Is my understanding correct?

Accept. I have updated the comments as you point out. Can you please re-review the patch?

I misunderstood your "S_ISFIFO" comment as "!S_ISFIFO(st.st_mode) ? false : true" when I wrote the comment, I realized I was wrong, then I updated code accordingly and added a new comment. :)

AntonioDan:
Hi mkow, dimakuv,

I have updated the patch according to your review comments. Could you
re-review the patch?

Hi mkow,
Yes, I should reply in the corresponding thread. I hope this time
it's better.

AntonioDan:
@dimakuv I have updated the patch according to your comments. Can you
re-review the patch? Thanks a lot for your comments!

@mkow I get your point to use Reviewable.io and would follow it, however,
when I click the icon, I meet Reviewable OFFLINE issue with Reason "no
connection to Firebase", I don't know why this occurs and need to try to
address this... I think I have resolved the review comments. Thanks!

AntonioDan:
@mkow,
I can finally access the Reviewable.io and have marked the review comments
as Done. Can you please review the patch and comment status? Please let me
know if I need to resolve anything else. Thanks!

mkow:
Jenkins, retest this please

mkow:
Jenkins, retest Jenkins-Debug please (`truncate01` and `truncate01_64` timed out)

PR #1674, opened 2020-07-14
[Pal/Linux-SGX] Provision wrap key for protected files at runtime

dimakuv:
## Description of the changes 

This commit adds the ability to provision the wrap (master) key for protected files at runtime (in contrast to previous approach of hard-coding `protected_files_key` in the manifest as a temporary solution). This is achieved as follows:

- New PAL API `DkSetProtectedFilesKey()` is added.
- New writable pseudo-file `/dev/attestation/protected_files_key` is added. It calls `DkSetProtectedFilesKey()` after it was written to.
- New `SECRET_PROVISION_SET_PF_KEY` option is added to the Secret Provisioning library. If it is set, the library assumes that the first provisioned secret is the wrap key for PF and write it into the new pseudo-file.

## How to test this PR? 

The Secret Provisioning example `ra-tls-secret-prov` is updated to include the new protected-files client. This client receives the wrap key for PF via secret provisioning and reads & outputs the protected file `files/input.txt`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1674)


Examples/ra-tls-secret-prov/secret_prov_pf_client.manifest.template
```diff
@@ -0,0 +1,75 @@
...
+
+loader.env.LD_LIBRARY_PATH = /lib:/lib/x86_64-linux-gnu
+
+# Secret Provisioning library (client-side) is preloaded
+loader.env.LD_PRELOAD = libsecret_prov_attest.so
+
+# Read application arguments directly from the command line. Don't use this on production!
+loader.insecure__use_cmdline_argv = 1
+
+# dummy key, will be replaced by the provisioned at runtime key
```
sudharkrish, line 17: Looks like you have a typo in this sentence.

PR #1686, opened 2020-07-17
[Pal/Linux-SGX] Fix bugs in protected files that broke large files

dimakuv:
## Description of the changes 

This PR fixes two bugs in Protected Files:
- mbedTLS source of entropy (`mbedtls_entropy_context`) was allocated on the stack and thus destroyed after return from init function; in reality, this object must live for the whole duration of execution (this bug led to spurious segfaults due to overwritten stack).
- MHT nodes of protected files must be updated and flushed to disk in reverse order of sorting, from bottom layers to upper layers (this bug led to outdated MHT key/MAC values in resulting encrypted file, so the file could not be correctly decrypted).

Note that both bugs crept into our port of Intel SGX SDK. The first bug is absent from SDK because it doesn't use mbedTLS. The second bug is actually a blunder of porting (we even had the comment to sort in reverse order!), and SDK does the reverse sort correctly: https://github.com/intel/linux-sgx/blob/d1029a5821acffae31d66efde06589617a3df9c6/sdk/protected_fs/sgx_tprotected_fs/file_flush.cpp#L369.

## How to test this PR? 

It only happens on rather big files (in our case, 250MB). So just test manually:
```bash
graphene/Pal/src/host/Linux-SGX/tools/pf_crypt/pf_crypt encrypt -w key -i big-file -o big-file.enc
graphene/Pal/src/host/Linux-SGX/tools/pf_crypt/pf_crypt decrypt -w key -i big-file.enc -o big-file.dec
md5sum big-file*  # original and dec must be the same
```

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1686)

### DIFF
--- a/Pal/src/host/Linux-SGX/protected-files/protected_files.c
+++ b/Pal/src/host/Linux-SGX/protected-files/protected_files.c
@@ -550,13 +550,12 @@ static bool ipf_update_all_data_and_mht_nodes(pf_context_t* pf) {
         data = lruc_get_next(pf->cache);
     }
 
-    // sort the list from the last node to the first (bottom layers first)
     if (dirty_count > 0)
         sort_nodes(mht_array, 0, dirty_count - 1);
 
-    // update the gmacs in the parents
-    for (dirty_idx = 0; dirty_idx < dirty_count; dirty_idx++) {
-        file_mht_node = mht_array[dirty_idx];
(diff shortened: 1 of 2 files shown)


mkow:
Jenkins, retest Jenkins-Debug-18.04 please (some weird internal Jenkins fault?)

PR #1695, opened 2020-07-19
 [LibOS] test/ltp: Try to deobfuscate Makefiles 

mkow:
## Description of the changes 

Our LTP makefiles are a huge and complicated mess. I tried to fix them, but gave up at some point and did just a few changes.

Changes:
- `make all` previously rebuilt everything each time, now it doesn't.
- Using out-of-tree build.
- Unfortunately we have to build the whole LTP now (minus modules), so the build is slower.
- But: `make -jX` works, previously was broken because of missing dependencies.
- Not using symlinks as targets anymore - Make is broken and checks the timestamp of the *destination*, not the symlink itself. Previously the links were recreated on each `make` call.
- `make regression` printed `/bin/sh: 1: test: Illegal number: ` and now it doesn't.

I'll create a follow-up PR which adds more parallelism to our Jenkinsfiles (but only to building tests, not running). This should generate a significant speed-up, e.g. building and running LTP tests as done in our Jenkinsfiles on SGX previously took 7m 24s, after this PR + the follow-up with parallelism it takes 5m 01s.

## How to test this PR? 

CI + play manually with LTP

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1695)

### DIFF
--- a/.ci/run-pylint
+++ b/.ci/run-pylint
@@ -14,8 +14,9 @@ find . -name \*.py \
     -and -not -path ./Pal/lib/crypto/mbedtls/\* \
     -and -not -path ./LibOS/glibc-build/\* \
     -and -not -path ./LibOS/glibc-\?.\?\?/\* \
-    -and -not -path ./LibOS/shim/test/ltp/opt/\* \
     -and -not -path ./LibOS/shim/test/ltp/src/\* \
+    -and -not -path ./LibOS/shim/test/ltp/build/\* \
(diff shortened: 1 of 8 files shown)


mkow:
Jenkins, retest Jenkins-SGX please (`exec_same` from LibOS regression crashes, unrelated to this PR, but interesting...)

mkow:
> exec_same from LibOS regression crashes, unrelated to this PR, but interesting...

For bookkeeping: This issue was debugged and fixed in #1698.

PR #1681, opened 2020-07-16
Examples: Upgrade to busybox 1.32.0 (latest) to avoid linker issues

stefanberger:
## Description of the changes 

Using busybox 1.32.0 on more recent distros (Fedora 32) avoids linker
issues like the following ones:

/usr/bin/ld: coreutils/lib.a(mktemp.o): in function `mktemp_main':
mktemp.c:(.text.mktemp_main+0xd0): warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
/usr/bin/ld: util-linux/lib.a(rdate.o): in function `rdate_main':
rdate.c:(.text.rdate_main+0x150): undefined reference to `stime'
/usr/bin/ld: coreutils/lib.a(date.o): in function `date_main':
date.c:(.text.date_main+0x2a4): undefined reference to `stime'
collect2: error: ld returned 1 exit status
Note: if build needs additional libraries, put them in CONFIG_EXTRA_LDLIBS.
Example: CONFIG_EXTRA_LDLIBS="pthread dl tirpc audit pam"
make[1]: *** [Makefile:718: busybox_unstripped] Error 1

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1681)

PR #1700, opened 2020-07-21
Updated redis-server manifest template file with additional dependancies for  redis-server version 6+.

sudharkrish:
## Description of the changes 
When I tried the steps in the Examples/redis/README file,  launch of redis-server failed, due to additional
dependencies that were not listed as trusted files in the manifest template.

Reason being this change->https://github.com/oscarlab/graphene/pull/1681,  updated, the version of redis-server in the Makefile, which has additional dependencies.

## How to test this PR? 
With this updated redis-server manifest file, able to launch redis-server and run benchmarks on Ubuntu 18.04 system.
Need to check if redis-server gets tested as part of Jenkins.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1700)

### DIFF
--- a/Examples/redis/redis-server.manifest.template
+++ b/Examples/redis/redis-server.manifest.template
@@ -37,7 +37,7 @@ loader.insecure__use_cmdline_argv = 1
 # In case of Redis:
 # - /lib is searched for Glibc libraries (ld, libc, libpthread)
 # - $(ARCH_LIBDIR) is searched for Name Service Switch (NSS) libraries
-loader.env.LD_LIBRARY_PATH = /lib:$(ARCH_LIBDIR)
+loader.env.LD_LIBRARY_PATH = /lib:$(ARCH_LIBDIR):/usr/$(ARCH_LIBDIR)
 
 ################################# MOUNT FS  ###################################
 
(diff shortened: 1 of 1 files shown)


dimakuv:
Jenkins, test this please

dimakuv:
Jenkins, test this please

sudharkrish:
@dimakuv that is quite strange. In my case, when I tried to launch redis-server, it fails to launch without these libraries. 
I did wait for some time, and checked if process came up(ps -aux | grep pal-Linux-SGX), but it didnt show up.

And this was the error, I was getting.
SGX=1 ./pal_loader redis-server --save '' --protected-mode no
sgx.protected_fileskey not found in manifest, protected files will not be available
WARNING: Using insecure argv source. Don't use this configuration in production!
Accessing file:/lib/x86_64-linux-gnu/libsystemd.so.0 is denied. (Unknown error) This file is not trusted or allowed.
redis-server: error while loading shared libraries: libsystemd.so.0: cannot open shared object file: No such file or directory

sudharkrish:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1700#-MCmCJ2cBusqf-AgjNmR:-MCmZTrNERor7iePNbIH:b-njedyd) (no related file):*

I just verified that Redis 6 worked before this patch but also works with this patch. Kinda surprising, but indeed Redis is looking for these libraries -- it just looks like they are not always needed (and Redis can continue without them, even if opening these libs failed).
</blockquote></details>

@dimakuv that is quite strange. In my case, when I tried to launch redis-server, it fails to launch without these libraries.
I did wait for some time, and checked if process came up(ps -aux | grep pal-Linux-SGX), but it didnt show up.

And this was the error, I was getting.
SGX=1 ./pal_loader redis-server --save '' --protected-mode no
sgx.protected_fileskey not found in manifest, protected files will not be available
WARNING: Using insecure argv source. Don't use this configuration in production!
Accessing file:/lib/x86_64-linux-gnu/libsystemd.so.0 is denied. (Unknown error) This file is not trusted or allowed.
redis-server: error while loading shared libraries: libsystemd.so.0: cannot open shared object file: No such file or directory

---

PR #1705, opened 2020-07-22
[LibOS] Increase timeout params of Async helper thread

dimakuv:
## Description of the changes 

Async helper thread self-destructs after some idle time (when it is unused). The timeout was previously set to too-low values so that Async helper thread sporadically died even when it was supposed to stay alive. This PR increases the timeout values slightly.

## How to test this PR? 

One can check when the Async helper thread terminates manually when looking at the debug output of `multi_pthread` LibOS test.

### DIFF
--- a/LibOS/shim/src/shim_async.c
+++ b/LibOS/shim/src/shim_async.c
@@ -13,8 +13,8 @@
 #include <shim_thread.h>
 #include <shim_utils.h>
 
-#define IDLE_SLEEP_TIME 1000
-#define MAX_IDLE_CYCLES 100
+#define IDLE_SLEEP_TIME 1000000
+#define MAX_IDLE_CYCLES 10000
 
 DEFINE_LIST(async_event);
 struct async_event {


sudharkrish:
@dimakuv Just to be clear, so you are increasing these timeouts, to prevent dump of AEX stats for async helper thread, which would add un-neccessary noise to analysis of data. Is this a correct assumption. If correct, having this for longer than needed, would it add un-neccessary latency. Also as an alternate option, would it be possible to add a name in the output, to indicate this thread is async helper thread, so folks can ignore analysis for that thread.

sudharkrish:
@dimakuv thanks Dmitrii, makes sense. It will be good (maybe in the future) to add a naming feature, to distinguish threads created by LibOS. For now, this is good enough.

mkow:
Jenkins, retest Jenkins-SGX-18.04 please (Jenkins was restarted and the job killed)

PR #1714, opened 2020-07-27
do getsockname in bind when binding on port 0.

### LINKED ISSUE #1713: [Graphene socket] incorrect results returned from getsockname after binding on port 0
## Description of the problem

*getsockname* returns port ipaddr:0 when a program binds on an arbitrary port.

The sample code is as follows:

```C
#include <ifaddrs.h>
#include <errno.h>
#include <iostream>
#include <stdlib.h>
#include <stdio.h>
#include <string.h>
#include <sys/socket.h>
#include <arpa/inet.h>
#include <unistd.h>
int main() {

  int sockfd = socket(AF_INET, SOCK_STREAM, 0);
  struct sockaddr bind_addr;
  struct sockaddr_in *addr = (struct sockaddr_in*)&bind_addr;
  addr->sin_family = AF_INET;
  addr->sin_addr.s_addr = inet_addr("127.0.0.1");
  addr->sin_port = 0;

  if (bind(sockfd, (struct sockaddr *)addr, sizeof(bind_addr)) < 0) {
        std::cout << "error in bind " << errno << "\n";
        return -1;
  }

  if (listen(sockfd, 3) < 0) {
        std::cout << "error in listen\n";
        return -1;
  }

  struct sockaddr connect_addr;
  struct sockaddr_in *addrr = (struct sockaddr_in*)&connect_addr;
  socklen_t len = sizeof(struct sockaddr);
  int ret = 0;
  if ((ret = getsockname(sockfd, &connect_addr, &len)) < 0) {
        std::cout << "error in getsockname " << ret << "\n";
        return -1;
  }

  std::cout << "Local IP address is: " << inet_ntoa(addrr->sin_addr) << ":" << ntohs(addrr->sin_port) << "\n";

  close(sockfd);
  printf("success\n");
  return 0;
}
```

## Steps to reproduce
Tested on Graphene (commit 0c7472cbbcd18dba7a9be4d2f47256e97dd07f5d). Graphene-SGX seems not to be affected, but I have not tested.

## Expected results
```
127.0.0.1:port(some port other than 0)
```

## Actual results
```
127.0.0.1:0
```

### Bug traces

I tried to debug it and found out that in *Pal/src/host/Linux/db_sockets.c*, the port information is filled in using *getsockname* only when both address and port are 0 (i.e. 0.0.0.0:0). However, this is also required when the port is not assigned (e.g., 127.0.0.1:0).

```C
    static bool check_any_addr(struct sockaddr* addr) {
    if (addr->sa_family == AF_INET) {
        struct sockaddr_in* addr_in = (struct sockaddr_in*)addr;

        return addr_in->sin_port == 0 && check_zero(&addr_in->sin_addr, sizeof(addr_in->sin_addr));
    } else if (addr->sa_family == AF_INET6) {
        struct sockaddr_in6* addr_in6 = (struct sockaddr_in6*)addr;

        return addr_in6->sin6_port == 0 &&
               check_zero(&addr_in6->sin6_addr, sizeof(addr_in6->sin6_addr));
    }

    return false;
    }

    if (check_any_addr(bind_addr)) {
        /* call getsockname to get socket address */
        if ((ret = INLINE_SYSCALL(getsockname, 3, fd, bind_addr, &bind_addrlen)) < 0)
            goto failed;
    }
```

I think changing *&&* to *||* and this resolved the problem.

I also looked into the Graphene-SGX code, which does *getsockname* when the address pointer is not null(Pal/src/host/Linux-SGX/sgx_enclave.c). Therefore, I think this bug does not affect Graphene-SGX.

jianyu-m:
## Description of the changes

See issue #1713. Fixes #1713.

## How to test this PR? 

Run the test program in #1713.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1714)


jianyu-m:
@yamahata @dimakuv I agreed that these seems redundant in the current code base. My concerns are that whether always doing **getsockname** in **bind** produces some performance overhead. This may be minor as I do not think an application would do bind heavily.

If you agree that such performance overhead is minor, I would have this patch done in a couple of days.

dimakuv:
@jianyu-m Please go ahead and prepare the new patch :) 

Socket creation/bind/connect is a rare operation (in well-written software and sane workloads), and so one additional small syscall like `getsockname()` shouldn't affect performance.

jianyu-m:
Hi @dimakuv , can you help check the code and merge?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1714)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @jianyu-m)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1714#-MGQ-sKM5VK2UukbsVMZ:-MGQ-sKM5VK2UukbsVM_:b-prtz50) (no related file):*
I'm taking over this PR since the original author disappeared.

I will re-create this PR in my own branch (because this PR was done in the author's master branch, and I don't want to mess with it).

---

PR #1725, opened 2020-07-30
Fix GSC testing Makefile and build template

vahldiek:
This PR fixes two issues.

1. A recent change which made all Runtime folder Symlinks relative results in `gsc build` command failing. As part of `gsc build` command all symlinks in /graphene/Runtime are converted to real files by iterating over all symlinks and copying them over. When the change was made to make the symlinks relative, the copy command no longer finds the file to copy and hence, the Docker `build` fails.
2. The failure of 1. was not discovered, since the Jenkins test ran on the master branch instead of the PR. The test makefile of GSC includes a `sed` command to create the required GSC `config.yaml`. This `sed` command was ignoring the Graphene branch input which is set in the Jenkins script.

## Description of the changes 

To fix 1.:
Changed `Tools/gsc/templates/Dockerfile.build.ubuntu18.04.template` to read the symlink and add the prefix path `/graphene/Runtime` to it. 

To fix 2.:
Changed `Tools/gsc/test/Makefile` to correctly update the configuration file with the Graphene branch.

## How to test this PR? 

Run Jenkins GSC pipeline and check the output includes the following line:
```
RUN cd /graphene && git fetch origin <commit #> && git checkout <commit #> ...
```

The correct commit hash can be found via the PR and should match a shell variable which is exported (look for `export COMMIT=`)

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1725)

### DIFF
--- a/Tools/gsc/templates/Dockerfile.ubuntu18.04.build.template
+++ b/Tools/gsc/templates/Dockerfile.ubuntu18.04.build.template
@@ -36,7 +36,7 @@ RUN cd /graphene && ISGX_DRIVER_PATH=/graphene/Pal/src/host/Linux-SGX/sgx-driver
     {% if linux %} && make -s -j4 WERROR=1 {% if debug %} DEBUG=1 {% endif %} {% else %} && true {%endif %}
 
 # Translate runtime symlinks to files
-RUN for f in $(find /graphene/Runtime -type l); do cp --remove-destination $(readlink $f) $f; done
+RUN for f in $(find /graphene/Runtime -type l); do cp --remove-destination $(realpath $f) $f; done
 
 # Finished building Graphene
 
--- a/Tools/gsc/templates/Dockerfile.ubuntu18.04.build.template
+++ b/Tools/gsc/templates/Dockerfile.ubuntu18.04.build.template
@@ -21,7 +21,7 @@ RUN git clone {{Graphene.Repository}} /graphene
 # Init submodules
 RUN cd /graphene \
     && git fetch origin {{Graphene.Branch}} \
-    && git checkout origin/{{Graphene.Branch}} \
+    && git checkout {{Graphene.Branch}} \
     && git submodule update --init -- Pal/src/host/Linux-SGX/sgx-driver/
 
 # Create SGX driver for header files
--- a/Tools/gsc/test/Makefile
+++ b/Tools/gsc/test/Makefile
@@ -1,5 +1,6 @@
 TESTCASES ?= python3 hello-world nodejs bash numpy pytorch
 DISTRIBUTIONS ?= ubuntu18.04
+GRAPHENE_BRANCH ?= master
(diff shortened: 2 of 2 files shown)


mkow:
Jenkins, test Jenkins-SGX-gsc please

dimakuv:
Jenkins, test this please

bigdata-memory:
___
*[Tools/gsc/templates/Dockerfile.ubuntu18.04.build.template, line 39 at r2](https://reviewable.io/reviews/oscarlab/graphene/1725#-MD_ox6m5tLxrFloLDMK:-MD_ox6m5tLxrFloLDML:b8gnvqq) ([raw file](https://github.com/oscarlab/graphene/blob/25a548c676e7823b2303a95a061c77ea79a1b9b8/Tools/gsc/templates/Dockerfile.ubuntu18.04.build.template#L39)):*
> ```
> 
> # Translate runtime symlinks to files
> RUN for f in $(find /graphene/Runtime -type l); do cp --remove-destination $(realpath $f) $f; done
> ```

In some corner cases, it is better to using the following ways to materialize the symlinks.

1) in case, the file names that contain newlines or other types of white space
    `find / graphene/Runtime -type l -print0`

2) replacing the `for` with pipeline operation to simplify it further
    `find / graphene/Runtime -type l -print0 | xargs -0 -n1 -I{} do cp --remove-destination $(realpath {}) {}`

3) I prefer to keep original folder rather than in place copying to avoid half baked modification in case of error occurred also using the rsync to do the dirty works and take care of symlink of sub-dirs.
```bash
    mv /graphene/Runtime /graphene/Runtime.original && 
    rsync -a -L -r -k /graphene/Runtime.original /graphene &&
    rm -rf  /graphene/Runtime.original
```

mkow:
Jenkins, retest this please

bigdata-memory:
___
*[Tools/gsc/templates/Dockerfile.ubuntu18.04.build.template, line 39 at r2](https://reviewable.io/reviews/oscarlab/graphene/1725#-MD_ox6m5tLxrFloLDMK:-MDf2VXv6v18E7FskGKw:bpgee2z) ([raw file](https://github.com/oscarlab/graphene/blob/25a548c676e7823b2303a95a061c77ea79a1b9b8/Tools/gsc/templates/Dockerfile.ubuntu18.04.build.template#L39)):*

1. We should actually test such paths in CI, this is a more broad issue in Graphene's shell glue in various places. @bigdata-memory Could you create an issue about this?
2. Tbh I find the current implementation much more readable (with `for`).
3. Hmm, this whole thing happens inside Docker, on failure we just throw away the mis-built image. I don't think we care about its state after a failure. Or maybe I'm missing something?

Anyways, all issues listed here were already in the code before this PR and we need to merge it soon because without it CI is failing.
</blockquote></details>

@mkow  I have created a new Issue according to your advice, #1734 , thanks.

PR #1735, opened 2020-08-01
[LibOS] use addq/subq $-128 instead of subq/addq $128

yamahata:
They produce shorter opecode because the range of 8bit immediate is
[-128, 127].

addq $-128, %rsp
subq $-128, %rsp
48 83 c4 80                  addq   $0xffffffffffffff80,%rsp
48 83 ec 80                  subq   $0xffffffffffffff80,%rsp

subq $128, %rsp
addq $128, %rsp
48 81 ec 80 00 00 00         subq   $0x80,%rsp
48 81 c4 80 00 00 00         addq   $0x80,%rsp

Signed-off-by: Isaku Yamahata <isaku.yamahata@intel.com>

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1735)


bigdata-memory:
My 2 cents, the readability is more important than the size of instruction, the size of 6 bytes is negligible compared to expressing the intention precisely.

PR #1740, opened 2020-08-04
[Pal/Linux-SGX] Remove redundant memset in _DkVirtualMemoryAlloc()

dimakuv:
## Description of the changes 

There is already one memset in `get_enclave_pages()` controlled by manifest option `sgx.zero_heap_on_demand`. So the explicit memset in `_DkVirtualMemoryAlloc()` is not needed. Observed perf improvement is 30% on a worst-case micro-benchmark with `mmap()`.

## How to test this PR? 

This is a perf improvement, so I didn't add any tests to CI. However, this tiny micro-benchmark is the worst-case for the bug fixed in this PR:
```
for (size_t i = 0; i < 3; i++) {
    addr = mmap(NULL, 1024 * 1024 * 256, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
    *addr = 0;
}
```

Running it on a Skylake machine with 128MB of EPC gave me:
```
# without this PR (with redundant memset)
        User time (seconds): 3.77
        Minor (reclaiming a frame) page faults: 707887

# with this PR (no redundant memset)
        User time (seconds): 2.62
        Minor (reclaiming a frame) page faults: 511154
```

I also did some quick tests with Redis. I observed no performance change (I assume because my Redis workload didn't create/re-create too many memory mappings).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1740)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_memory.c
+++ b/Pal/src/host/Linux-SGX/db_memory.c
@@ -54,8 +54,6 @@ int _DkVirtualMemoryAlloc(void** paddr, uint64_t size, int alloc_type, int prot)
     if (!mem)
         return addr ? -PAL_ERROR_DENIED : -PAL_ERROR_NOMEM;
 
-    memset(mem, 0, size);
-
     *paddr = mem;
     return 0;
 }


dimakuv:
This turned out to be wrong. Many applications re-use the same memory pages and expect them to be zero-initialized (via `mmap(MAP_ANONYMOUS)`). The previous PR to have `sgx.zero_heap_on_demand` was incorrect because it didn't identify this additional `memset()` in this PR.

I'm closing this PR and will create a correct one that removes `sgx.zero_heap_on_demand` (because the heap must always be zeroed-out on demand) and also removes the heap zero-out on enclave initialization (because the pages are anyhow zeroed-out on demand).

PR #1640, opened 2020-07-02
[Pal/Linux-SGX] Lazily zero-out the heap instead of proactively on init

dimakuv:
## Description of the changes 

Previously, Linux-SGX PAL initialized the whole heap to zero from within the enclave at startup. This led to significant slowdown on large-sized enclaves. This commit removes the old proactive logic and instead zeroes-out the heap lazily, when enclave pages are actually requested.

This does not change the security guarantees because the whole heap is anyway RWX on SGX v1, and the attacker may exploit app vulnerabilities (if any) by supplying malicious data within the enclave during runtime.

## How to test this PR? 

All tests must pass. Below are some of my quick tests. Enclave sizes: 8GB for `Bootstrap6`, 2x 256MB for `fork_and_exec`, and 1GB for `Redis`.

### Using older Skylake system and non-DCAP Intel SGX driver

The CPU has 128MB of EPC. The SGX driver adds *one* page per IOCTL.

- Without this PR:
```
        Command being timed: "/home/dimakuv/graphene/Runtime/pal_loader ./Bootstrap6.manifest.sgx"
        User time (seconds): 11.79
        System time (seconds): 58.32
        Elapsed (wall clock) time (h:mm:ss or m:ss): 1:10.91
        Minor (reclaiming a frame) page faults: 2081992

        Command being timed: "./pal_loader ./fork_and_exec"
        User time (seconds): 0.51
        System time (seconds): 2.36
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:08.59
        Minor (reclaiming a frame) page faults: 66759

        Command being timed: "./pal_loader redis-server --help"
        User time (seconds): 1.51
        System time (seconds): 7.90
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:10.05
        Minor (reclaiming a frame) page faults: 252220
```

- With this PR:
```
        Command being timed: "/home/dimakuv/graphene/Runtime/pal_loader ./Bootstrap6.manifest.sgx"
        User time (seconds): 0.17
        System time (seconds): 30.13
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:31.14
        Minor (reclaiming a frame) page faults: 2265

        Command being timed: "./pal_loader ./fork_and_exec"
        User time (seconds): 0.23
        System time (seconds): 1.78
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:05.97
        Minor (reclaiming a frame) page faults: 25047

[...]

PR #1741, opened 2020-08-04
[Pal/Linux-SGX] Remove unneeded manifest option `sgx.zero_heap_on_demand`

dimakuv:
## Description of the changes 

Previously, we introduced `sgx.zero_heap_on_demand` in Linux-SGX as a knob to trade off runtime degradation on memory allocations for faster enclave start-up times. This was an incorrect fix because Linux-SGX's `_DkVirtualMemoryAlloc()` always zeros the requested memory region, so there was a double-zero of the heap at runtime. Note that POSIX requires `_DkVirtualMemoryAlloc()` to zero-out the memory, and many applications rely on this (Apache, Blender in my experiments). Thus, this commit keeps the zero-out in `_DkVirtualMemoryAlloc()` and removes zero-outs on enclave init and in `get_enclave_pages()`.

This renders `sgx.zero_heap_on_demand` useless, so this manifest option is also removed. Also note that this commit doesn't introduce any performance degradation (in fact, now Graphene behaves as if `sgx.zero_heap_on_demand = 1` always).

See the following PRs for the context and history:

- https://github.com/oscarlab/graphene/pull/1640
- https://github.com/oscarlab/graphene/pull/1668
- https://github.com/oscarlab/graphene/pull/1689
- https://github.com/oscarlab/graphene/pull/1740

(This PR basically undoes the first three.)

## How to test this PR? 

All tests must pass. There must be no performance degradation.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1741)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -445,18 +445,3 @@ This syntax specifies whether to enable SGX enclave-specific statistics:
 *Note:* this option is insecure and cannot be used with production enclaves
 (``sgx.debug = 0``). If the production enclave is started with this option set,
 Graphene will fail initialization of the enclave.
-
-Zero out heap on demand vs during enclave init
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-::
-
-    sgx.zero_heap_on_demand=[1|0]
-    (Default: 0)
-
-This syntax specifies whether to zero out the heap on demand (when new enclave
-pages are requested) or during enclave initialization. If this option is set to
-``1``, then the initial heap space is uninitialized; this improves start-up
-performance but worsens run-time performance. If this option is set to ``0``,
-then the whole heap is zeroed out before enclave starts app execution; this
-worsens start-up performance but improves run-time performance.
--- a/Documentation/perf-practices.rst
+++ b/Documentation/perf-practices.rst
@@ -119,45 +119,6 @@ If you need additional statistics, you can check this unofficial patch:
 
 * https://github.com/oscarlab/graphene/tree/dimakuv/DONTMERGE-more-perf-stats-tweaks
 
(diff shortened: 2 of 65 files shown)


dimakuv:
As requested by @mkow, a quick experiment (same as for #1740).

I take `native/helloworld.c` and add this:
```c
for (size_t i = 0; i < 3; i++) {
    addr = mmap(NULL, 1024 * 1024 * 256, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
    *addr = 0;
}
```

the corresponding manifest change is in `native/manifest.template`:
```sh
-loader.debug_type = inline
+loader.debug_type = none         # to not account for logs
+loader.insecure__disable_aslr=1  # to ignore ASLR enclave-size issues
+sgx.enclave_size=2G              # need big enough enclave
```

I run this: `SGX=1 /usr/bin/time --verbose ~/graphene/Runtime/pal_loader ./helloworld`.

- On master (before this PR):
```
        Command being timed: "/home/dimakuv/graphene/Runtime/pal_loader ./helloworld"
        User time (seconds): 3.80
        System time (seconds): 17.84
        Minor (reclaiming a frame) page faults: 707423
```

- With this PR:
```
        Command being timed: "/home/dimakuv/graphene/Runtime/pal_loader ./helloworld"
        User time (seconds): 1.14
        System time (seconds): 10.69
        Minor (reclaiming a frame) page faults: 200216
```

Main cause of perf improvement is that on master, Helloworld manifest didn't have `sgx.zero_heap_on_demand` so it defaulted to zero-out of whole heap on enclave init (plus Graphene also did memset during runtime). With this PR, there is no zero-out on enclave init.

Also recall that this example is the *worst-case*, since I test on a Skylake machine with 128MB of EPC and thus a lot of EPC paging.

PR #1746, opened 2020-08-05
[Pal/Linux-SGX] Allow starting enclave address space with non-zero

dimakuv:
## Description of the changes 

Previously, Graphene-SGX required `vm.mmap_min_addr` equal to `0x0` for non-PIE executables (this is due to the SGX hardware requirement of enclave base being aligned on enclave size, coupled with non-PIE requirement of starting the code section at address `0x400000`).

The legacy Intel SGX driver required a device-file-backed mmap starting at address `0x0`. The newer DCAP/in-kernel SGX drivers don't require this, so Graphene-SGX can create the enclave at non-zero starting address and deprecate the hack of `vm.mmap_min_addr == 0x0`.

## How to test this PR? 

1. Use a machine with DCAP/in-kernel driver (DCAP v1.6+ and in-kernel v1.32+).

2. Reset `mmap_min_addr` like this: `echo 65535 | sudo tee /proc/sys/vm/mmap_min_addr`.

3. Manually build and run one of the LibOS regression tests (they are built with `sgx.static_address = 1` so their enclaves will be forced to have `enclave_base = 0x0`).

4. Do strace and you'll see stuff like this:
```
mmap(0x10000, 268369920, PROT_NONE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x10000
```

This means that the enclave address space starts with `0x10000 = 65535` (even though enclave base is `0x0`). And the enclave still works! Now you don't need the hack of `echo 0 | sudo tee /proc/sys/vm/mmap_min_addr`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1746)

### DIFF
--- a/Documentation/cloud-deployment.rst
+++ b/Documentation/cloud-deployment.rst
@@ -63,13 +63,6 @@ Build and Test
             ISGX_DRIVER_PATH=/usr/src/linux-azure-headers-`uname -r`/arch/x86/ \
                   make SGX=1
 
-#. Set ``vm.mmap_min_addr=0`` in the system:
-
-      .. code-block:: sh
-
-            # WARNING: read "Security Implications" section before running this command
-            sudo sysctl vm.mmap_min_addr=0
-
 #. Build and Run :program:`helloworld`:
 
       .. code-block:: sh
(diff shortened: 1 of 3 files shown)


dimakuv:
Jenkins, retest Jenkins please (`kill03` LTP test failed, known issue)

bigdata-memory:
___
*[Pal/src/host/Linux-SGX/sgx_framework.c, line 157 at r3](https://reviewable.io/reviews/oscarlab/graphene/1746#-ME3eRvF6i9fBNK6iHD0:-ME3eRvF6i9fBNK6iHD1:bs0jmi9) ([raw file](https://github.com/oscarlab/graphene/blob/87891ce7ecc2aeb4909b8627887969d33908d466/Pal/src/host/Linux-SGX/sgx_framework.c#L157)):*
> ```C
>     /* newer DCAP/in-kernel SGX drivers allow starting enclave address space with non-zero;
>      * the below trick to start from DEFAULT_HEAP_MIN is to avoid vm.mmap_min_addr==0 issue */
>     if (request_mmap_addr < DEFAULT_HEAP_MIN) {
> ```

I think it is better to add a debug message here for prompting user/dev. that the enclave mapping does not honor the configuration of secs. 
Furthermore, my 2 cents, I favor to move this adjustment ahead as possible, the sec->base can reflect this changes. thanks.

dimakuv:
___
*[Pal/src/host/Linux-SGX/sgx_framework.c, line 157 at r3](https://reviewable.io/reviews/oscarlab/graphene/1746#-ME3eRvF6i9fBNK6iHD0:-ME48oc_9LzWkjD3O3eb:b-eiv6ud) ([raw file](https://github.com/oscarlab/graphene/blob/87891ce7ecc2aeb4909b8627887969d33908d466/Pal/src/host/Linux-SGX/sgx_framework.c#L157)):*

I think it is better to add a debug message here for prompting user/dev. that the enclave mapping does not honor the configuration of secs. 
Furthermore, my 2 cents, I favor to move this adjustment ahead as possible, the sec->base can reflect this changes. thanks.
</blockquote></details>

We could add a debug message in a separate PR since this is already merged.

But we cannot adjust `secs->base` since it must be aligned to the `secs->size` and must be printed as such.

PR #1754, opened 2020-08-07
[LibOS] Fix setting O_NONBLOCK flag in pipe2

### LINKED ISSUE #1752: O_NONBLOCK ignored in pipe2
## Description of the problem
`O_NONBLOCK` flag is ignored in `pipe2`

## Steps to reproduce
```
#define _GNU_SOURCE
#include <err.h>
#include <fcntl.h>
#include <stdio.h>
#include <unistd.h>

int main(void) {
    int p[2];
    if (pipe2(p, O_NONBLOCK | O_CLOEXEC) < 0) {
        err(1, "pipe2");
    }

    if (write(p[1], "a", 1) != 1) {
        err(1, "write");
    }

    char c;
    if (read(p[0], &c, 1) != 1) {
        err(1, "read");
    }

    ssize_t x = read(p[0], &c, 1);
    printf("read: %zd %m\n", x);

    return 0;
}
```

## Expected results
`read: -1 Resource temporarily unavailable`

## Actual results
Graphene hangs on `read`

boryspoplawski:
## Description of the changes 
Specifying O_NONBLOCK flag in the call to pipe2 applied it only to one
end of the pipe.

Fixes #1752

## How to test this PR? 
New regression test added

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1754)

### DIFF
--- a/LibOS/shim/include/shim_handle.h
+++ b/LibOS/shim/include/shim_handle.h
@@ -348,6 +348,9 @@ void flush_handle(struct shim_handle* hdl);
 void get_handle(struct shim_handle* hdl);
 void put_handle(struct shim_handle* hdl);
 
+/* Set handle in non-blocking mode. */
+int set_handle_nonblocking(struct shim_handle* hdl);
+
 /* file descriptor table */
 struct shim_fd_handle {
     FDTYPE vfd; /* virtual file descriptor */
--- a/LibOS/shim/src/sys/shim_fcntl.c
+++ b/LibOS/shim/src/sys/shim_fcntl.c
@@ -18,6 +18,26 @@
 #include <shim_thread.h>
 #include <shim_utils.h>
 
+#define FCNTL_SETFL_MASK (O_APPEND | O_NONBLOCK)
+
+static int _set_handle_flags(struct shim_handle* hdl, unsigned long arg) {
+    if (hdl->fs && hdl->fs->fs_ops && hdl->fs->fs_ops->setflags) {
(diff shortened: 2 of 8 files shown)


bigdata-memory:
___
*[LibOS/shim/src/sys/shim_fcntl.c, line 34 at r2](https://reviewable.io/reviews/oscarlab/graphene/1754#-METse9mCD95iFTBex09:-METse9n1eDXZYvCxRn2:b-xr03sd) ([raw file](https://github.com/oscarlab/graphene/blob/74b75c0b014c6457085727c43c9fd5fc24891c98/LibOS/shim/src/sys/shim_fcntl.c#L34)):*
> ```C
> }
> 
> int set_handle_nonblocking(struct shim_handle* hdl) {
> ```

I suggest this public utility function needs to check the hdl pointer to prevent from dereferencing NULL pointer.

dimakuv:
___
*[LibOS/shim/src/sys/shim_fcntl.c, line 30 at r3](https://reviewable.io/reviews/oscarlab/graphene/1754#-MEYX3TC3oksn3Qb_0Fp:-MEYcAkNAgp3tqE_gLyR:bx0uds6) ([raw file](https://github.com/oscarlab/graphene/blob/2abb2f325e7431c4726d207a0a335e16f4db7f0e/LibOS/shim/src/sys/shim_fcntl.c#L30)):*

not `=` instead of `|=`?
</blockquote></details>

Oops, I think Michal found a bug here. It seems benign because we never "go back" from say O_APPEND to no O_APPEND, but it is a bug. @boryspoplawski ?

PR #1511, opened 2020-05-15
[LibOS,Pal] Add MSG_DONTWAIT and SCM_RIGHTS for recvmsg/sendmsg

dimakuv:
## Description of the changes 

This commit improves the emulation of recvfrom/sendfrom, recvmsg/sendmsg, and recvmmsg/sendmmsg system calls. In particular, MSG_DONTWAIT flag is allowed though not really emulated (benign in most cases). Also, it is possible now to send/receive FDs via SCM_RIGHTS on a UNIX domain socket (only send/recv of pipes and UNIX domain sockets is currently supported). 

## How to test this PR? 

Corresponding LibOS test `scm_rights` is added.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1511)

PR #1766, opened 2020-08-17
[LibOS] Allow but ignore MSG_DONTWAIT flag on recvfrom/sendto

### LINKED ISSUE #752: [LibOS] sendmsg ignores flags
The current implementation of do_sendmsg() does not look at the flags argument.  This is an omission we should address, or at least warn on behaviors not supported.

### LINKED ISSUE #1764: [LibOS]Apache flink using JDK7 not work due to socket flag checking
## Description of the problem
When enabling https://github.com/apache/flink , Java exception `java.net.SocketException: Operation not supported (Read failed) ` is returned because Graphene socket syscall `shim_do_recvfrom` rejects socket flags `(MSG_DONTWAIT | MSG_PEEK)` .
## Steps to reproduce
Run flink using Graphene-SGX.
flink Java code
[int response = is.read();](https://github.com/apache/flink/blob/c5915cf87f96e1c7ebd84ad00f7eabade7e7fe37/flink-runtime/src/main/java/org/apache/flink/runtime/blob/BlobOutputStream.java#L151)
calls JDK code
[n = socketRead0(fd, b, off, length, timeout);](https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/java/net/SocketInputStream.java#L150)
which is a native function `java.base/java.net.SocketInputStream.socketRead0(Native Method) `
Finally, 
` ssize_t recvfrom(int sockfd, void *buf, size_t len, int flags,
                        struct sockaddr *src_addr, socklen_t *addrlen);`
 with flags `(MSG_DONTWAIT | MSG_PEEK)` is called.

## Expected results
recvmsg() with flags (MSG_DONTWAIT | MSG_PEEK) succeeds.
The above Java code succeeds.
## Actual results
recvfrom() with flags (MSG_DONTWAIT | MSG_PEEK) fails with debug log
```
[   105] recvmsg()/recvmmsg()/recvfrom(): unknown flag (only MSG_PEEK is supported).
[   105] ---- shim_recvfrom(204,0xc3d407de0,1,64,0x0,0x0) = -95 
```
The log comes from checking 
https://github.com/oscarlab/graphene/blob/2093d64ce22c430d599da1dd9ba0ae5802be56d4/LibOS/shim/src/sys/shim_socket.c#L1216-L1220

Java exception `java.net.SocketException: Operation not supported (Read failed) ` is caught.
## Additional information
After I comment these lines to bypass the checking, flink can work. Currently I use this change as workaround. 
Need to evaluate  whether updating flag bit combine checking is enough or adding logic to deal with `MSG_DONTWAIT` is required.

dimakuv:
## Description of the changes 

This PR adds dummy emulation (simply ignoring) the flag `MSG_DONTWAIT` in `recv/recvmsg/recvmmsg`. This logic is partially taken from #1511 .

It is much more complicated (probably impossible?) to add the real emulation of `MSG_DONTWAIT` in Graphene. So for now we ignore this flag to make applications like Erlang and Apache Flink to work.

This PR also adds checks for supported flags in sendto/sendmsg and allows only MSG_DONTWAIT and MSG_NOSIGNAL.

Fixes #1764. See this issue for more context.

Fixes #752.

## How to test this PR? 

`tcp_msg_peek` LibOS test was augmented with this flag, just for testing.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1766)

### DIFF
--- a/LibOS/shim/include/shim_types.h
+++ b/LibOS/shim/include/shim_types.h
@@ -219,10 +219,14 @@ struct __kernel_ustat
 /* bits/socket.h */
 enum
 {
-    MSG_OOB  = 0x01, /* Process out-of-band data. */
-    MSG_PEEK = 0x02, /* Peek at incoming messages. */
+    MSG_OOB      = 0x01,   /* Process out-of-band data. */
+    MSG_PEEK     = 0x02,   /* Peek at incoming messages. */
+    MSG_DONTWAIT = 0x40,   /* Nonblocking IO. */
+    MSG_NOSIGNAL = 0x4000, /* Do not generate SIGPIPE. */
 #define MSG_OOB MSG_OOB
 #define MSG_PEEK MSG_PEEK
+#define MSG_DONTWAIT MSG_DONTWAIT
+#define MSG_NOSIGNAL MSG_NOSIGNAL
 };
 
 struct msghdr {
--- a/LibOS/shim/src/sys/shim_socket.c
+++ b/LibOS/shim/src/sys/shim_socket.c
@@ -1004,9 +1004,6 @@ int shim_do_accept4(int fd, struct sockaddr* addr, int* addrlen, int flags) {
 
 static ssize_t do_sendmsg(int fd, struct iovec* bufs, int nbufs, int flags,
                           const struct sockaddr* addr, int addrlen) {
-    // Issue #752 - https://github.com/oscarlab/graphene/issues/752
-    __UNUSED(flags);
-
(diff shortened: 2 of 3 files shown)


dimakuv:
Jenkins, retest Jenkins-SGX please (`pipe_nonblocking` LibOS test failed, cannot reproduce locally)

PR #1813, opened 2020-09-12
[Pal/Linux-SGX] Don't cast off `const` in sgx_copy_to_enclave

### LINKED ISSUE #646: [PAL/Linux-SGX] Structs in ocall_types.h do not need const
Some of the `ms_*` structs have `const void* ptr` and some have `void* ptr` fields (see `ocall_types.h`). These structs are only used to pass ocall arguments so if something shouldn't be changed it needs to be enforced inside the enclave and a compiler `const` hint doesn't help.

This cleanup would satisfy the compiler on such functions as `copy_outside_enclave()` in `enclave_ocalls.c`. (See #622.)

mkow:
## Description of the changes 

See #646. Turns out that there's no need to drop consts from the structs, only `sgx_copy_to_enclave()` needs to be fixed. I guess the ticket was created when this wasn't possible, but our code has changed a lot since then.

Fixes #646.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1813)

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_framework.c
+++ b/Pal/src/host/Linux-SGX/enclave_framework.c
@@ -88,13 +88,13 @@ bool sgx_copy_ptr_to_enclave(void** ptr, void* uptr, size_t size) {
     return true;
 }
 
-bool sgx_copy_to_enclave(const void* ptr, size_t maxsize, const void* uptr, size_t usize) {
+bool sgx_copy_to_enclave(void* ptr, size_t maxsize, const void* uptr, size_t usize) {
     if (usize > maxsize ||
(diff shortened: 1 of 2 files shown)


yamahata (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1813)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1813)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

Issue #1809, opened 2020-09-10
[LibOS] Implement waitid() system call


dimakuv:
We need to implement `waitid()` system call. We already have `wait()` and `waitpid()`, so this should be pretty straight-forward. We found an application that uses this syscall.

thiagomacieira:
Please note that the Linux `waitid()` syscall has an extra parameter not present in POSIX.

POSIX: `int waitid(idtype_t idtype, id_t id, siginfo_t *infop, int options);`

Linux: `int waitid(int which, pid_t pid, siginfo_t *infop, int options, struct rusage *ru)`

See https://code.woboq.org/linux/linux/kernel/exit.c.html#kernel_waitid

With this extra parameter, you can implement all of the other wait functions with `waitid`:
```c
pid_t wait4(pid_t pid, int *wstatus, int options, struct rusage *rusage)
{
    siginfo_t si;
    idtype_t which = P_PID;
    if (pid < -1) {
        which = P_PGID;
        pid = -pid;
    } else if (pid == -1) {
        which = P_ANY;
        pid = 0;
    }
    pid_t r = waitid(which, pid, &si, options, rusage);
    if (r > 0 && wstatus) {
        /* converting siginfo_t si_code and si_status to wstatus left as an exercise for the reader */
    }
    return r;
}
pid_t wait3(pid_t pid, int *wstatus, struct rusage *rusage)
{
    return wait4(pid, wstatus, 0, rusage);
}
pid_t waitpid(pid_t pid, int *wstatus, int options)
{
    return wait4(pid, wstatus, options, NULL);
}
pid_t wait(int *wstatus)
{
    return wait4(-1, wstatus, 0, NULL);
}
```

mkow:
@dimakuv I could take a look at this, but I'd need an app which uses this to test the implementation. Worst case I can just use the two tests LTP has, but would be better to test it also on the thing you're using.

thiagomacieira:
Very simple testcase:
```c
#include <stdlib.h>
#include <stdio.h>
#include <string.h>
#include <sys/resource.h>
#include <sys/syscall.h>
#include <sys/types.h>
#include <sys/wait.h>
#include <unistd.h>

static int sys_waitid(int which, int pid_or_pidfd, siginfo_t *infop, int options,
                      struct rusage *ru)
{
    /* use the waitid raw system call, which has an extra parameter that glibc
     * doesn't offer to us */
[...]

PR #1821, opened 2020-09-15
Implement waitid

pwmarcz:
## Description of the changes 

Fixes https://github.com/oscarlab/graphene/issues/1809.

Supports WEXITED and WNOWAIT (no WSTOPPED and WCONTINUED).

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)


LibOS/shim/src/sys/shim_wait.c
```diff
@@ -117,15 +132,53 @@ pid_t shim_do_wait4(pid_t pid, int* status, int option, struct __kernel_rusage*
...
+        infop->si_pid = thread->tid;
+        infop->si_uid = thread->uid;
+        infop->si_signo = SIGCHLD;
+
         /* Bits 0--7 are for the signal, if any.
          * Bits 8--15 are for the exit code */
-        *status = thread->term_signal;
-        *status |= ((thread->exit_code & 0xff) << 8);
+        infop->si_code = thread->term_signal;
+        infop->si_code |= ((thread->exit_code & 0xff) << 8);
```
thiagomacieira, line 143: This is not correct. Unlike the other `wait` functions, `si_code` contains the actual exit code unshifted if `si_status` is `CLD_EXITED`

LibOS/shim/src/sys/shim_wait.c
```diff
@@ -117,15 +132,53 @@ pid_t shim_do_wait4(pid_t pid, int* status, int option, struct __kernel_rusage*
...
+        id = pid;
+    }
+
+    info.si_pid = 0;
+    info.si_code = 0;
+    int ret = shim_do_waitid(which, id, &info, option, ru);
+    if (ret >= 0) {
+        ret = info.si_pid;
+        if (status)
+            *status = info.si_code;
```
thiagomacieira, line 181: When the above line is fixed, this line will become wrong.

thiagomacieira (changes requested):

LibOS/shim/src/sys/shim_wait.c
thiagomacieira: Looks like I also got code and status inverted in the bug report.

`si_code` should be one of the `CLD_` constants and `si_status` should be the signal number or exit code (unshifted).

yamahata (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz)

___
*[LibOS/shim/src/shim_syscalls.c, line 848 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHnzb54qyv741i5fkf:-MHHnzb54qyv741i5fkg:b-x2bevn) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/shim_syscalls.c#L848)):*
> ```
> SHIM_SYSCALL_RETURN_ENOSYS(waitid, 5, int, int, which, pid_t, pid, siginfo_t*, infop, int, options,
>                            struct __kernel_rusage*, ru)
> ```

Please remove this line instead of commenting out.
___
*[LibOS/shim/src/sys/shim_wait.c, line 27 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHpREn3Wc_K04vnrco:-MHHpREo7TCMTp9e8udR:bppvhpn) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/sys/shim_wait.c#L27)):*
> ```C
>     __UNUSED(ru);
> 
>     if (option & ~(WNOHANG | WUNTRACED | WEXITED | WCONTINUED | __WNOTHREAD | __WCLONE | __WALL)) {
> ```

Can you please add explicitly comment that  WSTOPPED, WCONTINUED, WNOWAIT aren't supported?
___
*[LibOS/shim/src/sys/shim_wait.c, line 50 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHrbTR5UGVeb5ho4at:-MHHrbTR5UGVeb5ho4au:b-1kz0c1) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/sys/shim_wait.c#L50)):*
> ```C
>             unlock(&thread->lock);
>             if (!(option & WNOHANG))
>                 goto block_pid;
> ```

Hmm this godo is ugly. But it can be addressed as follow up.
___
*[LibOS/shim/src/sys/shim_wait.c, line 105 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHro4IEsB-pjNWuDPk:-MHHro4J8fuMxvQrVVUT:bf0nvnd) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/sys/shim_wait.c#L105)):*
> ```C
> 
>         if (!(option & WNOHANG))
>             goto block;
> ```

This goto is ugly.
___
*[LibOS/shim/src/sys/shim_wait.c, line 107 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHrNOX3JakpZJZJNCp:-MHHrNOX3JakpZJZJNCq:b74h1ul) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/sys/shim_wait.c#L107)):*
> ```C
>             goto block;
>     } else {
>         assert (which == P_ALL);
> ```

This if clause reaches when which is P_ALL or P_PGID. Now this else clause is unreached.
something wrong.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: 4 of 6 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @thiagomacieira and @yamahata)

___
*[LibOS/shim/src/shim_syscalls.c, line 848 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHnzb54qyv741i5fkf:-MHLFtfo3gmuoyoXOKkc:b859rnp) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/shim_syscalls.c#L848)):*

> ```
> SHIM_SYSCALL_RETURN_ENOSYS(waitid, 5, int, int, which, pid_t, pid, siginfo_t*, infop, int, options,
>                            struct __kernel_rusage*, ru)
> ```

Please remove this line instead of commenting out.
</blockquote></details>

Done. Out of curiosity, why are some other calls in this file commented out instead of removed?
___
*[LibOS/shim/src/sys/shim_wait.c, line 27 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHpREn3Wc_K04vnrco:-MHLGNDXCtXI6t0xHlgK:b-de56ml) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/sys/shim_wait.c#L27)):*

Can you please add explicitly comment that  WSTOPPED, WCONTINUED, WNOWAIT aren't supported?
</blockquote></details>

Done.

(Not sure why we permit the exact set of flags checked here, so I did not change it except for adding WEXITED).
___
*[LibOS/shim/src/sys/shim_wait.c, line 107 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHrNOX3JakpZJZJNCp:-MHLIjzCBoML70FFxozk:b-7pin8o) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/sys/shim_wait.c#L107)):*

This if clause reaches when which is P_ALL or P_PGID. Now this else clause is unreached.
something wrong.
</blockquote></details>

Right. This assert is correct, but the `if` should say `if (which == P_PGID)`. Fixed.
___
*[LibOS/shim/src/sys/shim_wait.c, line 143 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHH3_WdwDpN73ibaqMJ-r1-143:-MHLMXRX3MfRDj_xSkK_:b7q2js3) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/sys/shim_wait.c#L143)):*
<details><summary><i>Previously, thiagomacieira (Thiago Macieira) wrote…</i></summary><blockquote>

Looks like I also got code and status inverted in the bug report.

`si_code` should be one of the `CLD_` constants and `si_status` should be the signal number or exit code (unshifted).
</blockquote></details>

I see. Fixed.
___
*[LibOS/shim/src/sys/shim_wait.c, line 181 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHH3_WdwDpN73ibaqMJ-r1-181:-MHLM_IM-bNrd7ygIaJI:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/sys/shim_wait.c#L181)):*
<details><summary><i>Previously, thiagomacieira (Thiago Macieira) wrote…</i></summary><blockquote>

When the above line is fixed, this line will become wrong.
</blockquote></details>

Fixed.

mkow (changes requested):
> Reviewed 4 of 6 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz, @thiagomacieira, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHLwX3O2Dt2nS_DBrrv:-MHLwX3O2Dt2nS_DBrrw:bsg1ovr) (no related file):*
Adding a blocking note to not forget about this later: if @thiagomacieira confirms that this PR fixes #1809, then we need to add "Fixes #1809" to the PR description to autoclose the issue.

---
___
*[LibOS/shim/include/shim_table.h, line 390 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM-NXm52xzje1UqPgL:-MHM-NXm52xzje1UqPgM:b-cttl2k) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/include/shim_table.h#L390)):*
> ```objc
> int shim_do_execve(const char* file, const char** argv, const char** envp);
> noreturn int shim_do_exit(int error_code);
> int shim_do_waitid(int which, pid_t id, siginfo_t* infop, int option, struct __kernel_rusage* ru);
> ```

This syscall in Linux returns `long`, not `int`.
___
*[LibOS/shim/include/shim_table.h, line 390 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM05Vy1mDiS55MYt3e:-MHM05Vy1mDiS55MYt3f:b-w0j6rw) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/include/shim_table.h#L390)):*
> ```objc
> int shim_do_execve(const char* file, const char** argv, const char** envp);
> noreturn int shim_do_exit(int error_code);
> int shim_do_waitid(int which, pid_t id, siginfo_t* infop, int option, struct __kernel_rusage* ru);
> ```

`option` -> `options` (ditto for other places in this PR)
___
*[LibOS/shim/src/shim_parser.c, line 323 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM0xW99i57D4ReeGs9:-MHM0xW99i57D4ReeGsA:buhytfh) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/shim_parser.c#L323)):*
> ```C
>         [__NR_mq_getsetattr]   = {.slow = 0, .parser = {NULL}},
>         [__NR_kexec_load]      = {.slow = 0, .parser = {NULL}},
>         [__NR_waitid]      = {.slow = 1, .parser = {NULL, NULL, NULL, &parse_wait_option, NULL}},
> ```

`parse_wait_option` recognizes only one of the flags, but more alarmingly, it silently skips all other bits. Could you add support for all possible flags to it? Also, if there are any unknown bits they should be emitted as `| 0x1234` instead of being skipped.
___
*[LibOS/shim/src/shim_syscalls.c, line 848 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHnzb54qyv741i5fkf:-MHM2P5LC3rO6Kx7HnDr:bh33tud) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/shim_syscalls.c#L848)):*

Done. Out of curiosity, why are some other calls in this file commented out instead of removed?
</blockquote></details>

No idea, I think all the commented lines come from the old codebase and it's just that no one so far came to clean them up.
___
*[LibOS/shim/src/shim_syscalls.c, line 311 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM2IBEC8JDd_VXHI9c:-MHM2IBFCSbGpn1jDGPI:b6hu6e3) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/shim_syscalls.c#L311)):*
> ```C
> 
> /* waitid: sys/shim_wait.c */
> DEFINE_SHIM_SYSCALL(waitid, 5, shim_do_waitid, int, int, which, pid_t, id, siginfo_t*, infop,
> ```

ditto, retval should be `long`
___
*[LibOS/shim/src/sys/shim_wait.c, line 27 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHpREn3Wc_K04vnrco:-MHM5Pqb6LXJzA4CztC8:b-6nv2kl) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/sys/shim_wait.c#L27)):*

Done.

(Not sure why we permit the exact set of flags checked here, so I did not change it except for adding WEXITED).
</blockquote></details>

I guess this was copied from the kernel, but it seems that waitid and wait4 have distinct sets of supported options. See:
waitid: https://elixir.bootlin.com/linux/v5.8.9/source/kernel/exit.c#L1502
wait4: https://elixir.bootlin.com/linux/v5.8.9/source/kernel/exit.c#L1594
I think we have to adapt the check for this, somehow.
___
*[LibOS/shim/src/sys/shim_wait.c, line 32 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM619zEuFOqXJ00Gj6:-MHM619zEuFOqXJ00Gj7:b6i6w92) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/sys/shim_wait.c#L32)):*
> ```C
>         return -EINVAL;
>     }
> ```

waitid has an additional check which wasn't present in wait4:
```C
if (!(options & (WEXITED | WSTOPPED | WCONTINUED)))
    return -EINVAL;
```
___
*[LibOS/shim/src/sys/shim_wait.c, line 156 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM8Z-m5HqdNs-wQWFC:-MHM8Z-m5HqdNs-wQWFD:b-bh8yoo) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/sys/shim_wait.c#L156)):*
> ```C
> }
> 
> pid_t shim_do_wait4(pid_t pid, int* status, int option, struct __kernel_rusage* ru) {
> ```

While we're here, wait4 syscall returns `long`, not `pid_t`. Could you fix this?
___
*[LibOS/shim/src/sys/shim_wait.c, line 156 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM8n9tAPi6YI5ABthB:-MHM8n9tAPi6YI5ABthC:bouwy03) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/sys/shim_wait.c#L156)):*
> ```C
> }
> 
> pid_t shim_do_wait4(pid_t pid, int* status, int option, struct __kernel_rusage* ru) {
> ```

`option` -> `options`
___
*[LibOS/shim/src/sys/shim_wait.c, line 180 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM9c4R36l1R0kJeHgl:-MHM9c4R36l1R0kJeHgm:b-ig4gud) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/sys/shim_wait.c#L180)):*
> ```C
>         ret = info.si_pid;
>         if (ret > 0 && status) {
>             if (info.si_code == CLD_EXITED)
> ```

At this point kernel has an additional check for the pointer provided by the user, we should emulate this, because LTP likes to check for such things.
Should be something like:
```C
if (test_user_memory(status, sizeof(*status), /*write=*/true))
    return -EFAULT;
```

`shim_waitid()` also needs such a check, see https://elixir.bootlin.com/linux/v5.8.9/source/kernel/exit.c#L1570.
___
*[LibOS/shim/test/ltp/ltp.cfg, line 3167 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHMBhPy41kmBg4ORq1K:-MHMBhPy41kmBg4ORq1L:b-7fhjs2) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/test/ltp/ltp.cfg#L3167)):*
> ```INI
> skip = yes
> 
> [wait401]
> ```

While we're here, could you take a quick look why this and the test below fail? No need to spend too much time debugging them though, I'm just curious how far we're from fixing them.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow, @pwmarcz, @thiagomacieira, and @yamahata)

___
*[LibOS/shim/include/shim_table.h, line 390 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM-NXm52xzje1UqPgL:-MHMWCFbBQVqzE2i-j5Q:b-49d1z5) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/include/shim_table.h#L390)):*

This syscall in Linux returns `long`, not `int`.
</blockquote></details>

Done. Changed both `shim_do_waitid` and `shim_do_wait4` to return `long` (but not `shim_waitid` and `shim_wait4`, I assume they're the user-mode wrappers).
___
*[LibOS/shim/include/shim_table.h, line 390 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM05Vy1mDiS55MYt3e:-MHMWRmG8MvIsCWQvc4d:b-62gx31) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/include/shim_table.h#L390)):*

`option` -> `options` (ditto for other places in this PR)
</blockquote></details>

Done, for `wait4` as well.
___
*[LibOS/shim/src/shim_parser.c, line 323 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM0xW99i57D4ReeGs9:-MHMd1GIB20uvCJfOCBZ:b-19hffd) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/shim_parser.c#L323)):*

`parse_wait_option` recognizes only one of the flags, but more alarmingly, it silently skips all other bits. Could you add support for all possible flags to it? Also, if there are any unknown bits they should be emitted as `| 0x1234` instead of being skipped.
</blockquote></details>

Done. I see the flag printing could be unified, but I didn't want to make it too big, so I reused the table-driven machinery for `clone()` flags.

How do I test it, though? I tried setting `loader.debug_type = inline` in the manifest for LTP and running the `wait401` test, but it seems to segfault on startup, even when I compile the code from master...
___
*[LibOS/shim/src/shim_syscalls.c, line 311 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM2IBEC8JDd_VXHI9c:-MHMdV7SAkuUL6e-mRhJ:bqdhtik) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/shim_syscalls.c#L311)):*

ditto, retval should be `long`
</blockquote></details>

Are you sure? This macro seems to control return code of `shim_waitid`, not `shim_do_waitid`. Should that also return `long`?
___
*[LibOS/shim/src/sys/shim_wait.c, line 27 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHpREn3Wc_K04vnrco:-MHMdhsu6NrCxM96QcXt:bintunv) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/sys/shim_wait.c#L27)):*

I guess this was copied from the kernel, but it seems that waitid and wait4 have distinct sets of supported options. See:
waitid: https://elixir.bootlin.com/linux/v5.8.9/source/kernel/exit.c#L1502
wait4: https://elixir.bootlin.com/linux/v5.8.9/source/kernel/exit.c#L1594
I think we have to adapt the check for this, somehow.
</blockquote></details>

I see. The previous check seems to have been copied directly from kernel. Keeping with that, I made sure both `wait4` and `waitid` accept the same flags as in kernel, even though they ignore most of them.
___
*[LibOS/shim/src/sys/shim_wait.c, line 32 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM619zEuFOqXJ00Gj6:-MHMduLS-9D8MstNdor4:b-1vd9af) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/sys/shim_wait.c#L32)):*

waitid has an additional check which wasn't present in wait4:
```C
if (!(options & (WEXITED | WSTOPPED | WCONTINUED)))
    return -EINVAL;
```
</blockquote></details>

Done, see above.
___
*[LibOS/shim/src/sys/shim_wait.c, line 156 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM8Z-m5HqdNs-wQWFC:-MHMdzDd8Y09kr_rINbL:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/sys/shim_wait.c#L156)):*

While we're here, wait4 syscall returns `long`, not `pid_t`. Could you fix this?
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_wait.c, line 156 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM8n9tAPi6YI5ABthB:-MHMe-0L3oDjl2-aUhii:b-896fiz) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/sys/shim_wait.c#L156)):*

`option` -> `options`
</blockquote></details>

Done,
___
*[LibOS/shim/src/sys/shim_wait.c, line 180 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM9c4R36l1R0kJeHgl:-MHMoxCE2Ds4V57G4HO7:b-myka8h) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/sys/shim_wait.c#L180)):*

At this point kernel has an additional check for the pointer provided by the user, we should emulate this, because LTP likes to check for such things.
Should be something like:
```C
if (test_user_memory(status, sizeof(*status), /*write=*/true))
    return -EFAULT;
```

`shim_waitid()` also needs such a check, see https://elixir.bootlin.com/linux/v5.8.9/source/kernel/exit.c#L1570.
</blockquote></details>

Done. I added the checks for `infop` and for `status`, just before writing to them.
___
*[LibOS/shim/test/ltp/ltp.cfg, line 3167 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHMBhPy41kmBg4ORq1K:-MHMiXA_3CGA7QuufVVp:b-cp06fw) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/test/ltp/ltp.cfg#L3167)):*

While we're here, could you take a quick look why this and the test below fail? No need to spend too much time debugging them though, I'm just curious how far we're from fixing them.
</blockquote></details>

`wait401` does fork, and the forked process does `exit(0)`. For some reason we get exit code 2 from `wait`, though - maybe the fork is disallowed for some reason?

`wait402` fails with "Failed to open FILE '/proc/sys/kernel/pid_max'".

bigdata-memory (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: 2 of 6 files reviewed, 15 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow, @thiagomacieira, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHLwX3O2Dt2nS_DBrrv:-MHN7YCNEbMGO4pbnOyI:bmf4a3k) (no related file):*

Adding a blocking note to not forget about this later: if @thiagomacieira confirms that this PR fixes #1809, then we need to add "Fixes #1809" to the PR description to autoclose the issue.
</blockquote></details>

Might be better to have a section in PR template to remind contributors about this.

---
___
*[LibOS/shim/src/shim_parser.c, line 458 at r3](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHN9kGq4CalfkZJADsh:-MHN9kGq4CalfkZJADsi:byrlk48) ([raw file](https://github.com/oscarlab/graphene/blob/e078444ec9c0b9e28eca2cb0980a5bcf718622d7/LibOS/shim/src/shim_parser.c#L458)):*
> ```C
> };
> 
> static int parse_flags(int flags, const struct flag_table* all_flags, size_t count) {
> ```

I think it would be better to separate the printing flags from parsing the flags instead of mixing them up.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz, @thiagomacieira, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHLwX3O2Dt2nS_DBrrv:-MHNEUaZCrR0lhaGeGpa:bszerm3) (no related file):*

Might be better to have a section in PR template to remind contributors about this.
</blockquote></details>

I like this idea, let me do a quick PR adding this.

---
___
*[LibOS/shim/include/shim_table.h, line 390 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM-NXm52xzje1UqPgL:-MHMr0pYERbHMPp59LZz:bkxu0wx) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/include/shim_table.h#L390)):*

Done. Changed both `shim_do_waitid` and `shim_do_wait4` to return `long` (but not `shim_waitid` and `shim_wait4`, I assume they're the user-mode wrappers).
</blockquote></details>

Not exactly - they come from some old and unused feature which I deleted yesterday :) (or at least created a PR which will do so - #1822)
___
*[LibOS/shim/src/shim_parser.c, line 323 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM0xW99i57D4ReeGs9:-MHMrZwi-pa9v3tBjCoo:b-8b1yml) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/shim_parser.c#L323)):*

Done. I see the flag printing could be unified, but I didn't want to make it too big, so I reused the table-driven machinery for `clone()` flags.

How do I test it, though? I tried setting `loader.debug_type = inline` in the manifest for LTP and running the `wait401` test, but it seems to segfault on startup, even when I compile the code from master...
</blockquote></details>

That's interesting. For me it prints `BROK: Failed to open FILE '/proc/2/stat' for reading`.
To have debug logs AFAIR you need both `loader.debug_type = inline` and a build with `DEBUG=1`, maybe you're missing `DEBUG=1`?
___
*[LibOS/shim/src/shim_parser.c, line 458 at r3](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHN9kGq4CalfkZJADsh:-MHNE83dEyejDON8hzjp:bcnk2hd) ([raw file](https://github.com/oscarlab/graphene/blob/e078444ec9c0b9e28eca2cb0980a5bcf718622d7/LibOS/shim/src/shim_parser.c#L458)):*

I think it would be better to separate the printing flags from parsing the flags instead of mixing them up.
</blockquote></details>

What do you mean? Parsing here is quite simple (just ANDing with a value), and printing is actually delegated to those `PUTS` and `PUTCH` macros.
___
*[LibOS/shim/src/shim_parser.c, line 699 at r3](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHN6cuh-OtrCk0Z60F_:-MHN6cuiA8BFldsR1Z4b:b7oy6or) ([raw file](https://github.com/oscarlab/graphene/blob/e078444ec9c0b9e28eca2cb0980a5bcf718622d7/LibOS/shim/src/shim_parser.c#L699)):*
> ```C
> #undef FLG
> 
>     flags = parse_flags(flags, all_flags, ARRAY_SIZE(all_flags));
> ```

Nice cleanup!
___
*[LibOS/shim/src/shim_syscalls.c, line 311 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM2IBEC8JDd_VXHI9c:-MHN81YQ8NBJBQWqR8Ok:b4nblqx) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/shim_syscalls.c#L311)):*

Are you sure? This macro seems to control return code of `shim_waitid`, not `shim_do_waitid`. Should that also return `long`?
</blockquote></details>

It controls both. This should get slightly less confusing after #1822 which removes this weird feature. Anyways, even with it in place, the `shim_<sc_name>` wrappers weren't usermode/"glibc-like" versions, they were just exports for binaries linked directly against LibOS (see LibOS/shim/test/inline directory).
___
*[LibOS/shim/src/sys/shim_wait.c, line 36 at r3](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHN8nt2BYA9dIoNRe_m:-MHN8nt2BYA9dIoNRe_n:b-lb9p82) ([raw file](https://github.com/oscarlab/graphene/blob/e078444ec9c0b9e28eca2cb0980a5bcf718622d7/LibOS/shim/src/sys/shim_wait.c#L36)):*
> ```C
> 
> 	if (!(options & (WEXITED | WSTOPPED | WCONTINUED)))
> 		return -EINVAL;
> ```

tabs -> spaces
If you're interested, we have recommended Vim and Emacs config files in https://graphene.readthedocs.io/en/latest/devel/setup.html to prevent such mistakes :)
___
*[LibOS/shim/test/ltp/ltp.cfg, line 3167 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHMBhPy41kmBg4ORq1K:-MHND6Jl7fpy5ks5h9VW:bix1epr) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/test/ltp/ltp.cfg#L3167)):*

`wait401` does fork, and the forked process does `exit(0)`. For some reason we get exit code 2 from `wait`, though - maybe the fork is disallowed for some reason?

`wait402` fails with "Failed to open FILE '/proc/sys/kernel/pid_max'".
</blockquote></details>

I checked `wait401` and for me it prints `Failed to open FILE '/proc/2/stat' for reading`. Could you place a comment before both tests stating why they are failing?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz, @thiagomacieira, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHLwX3O2Dt2nS_DBrrv:-MHNG_dr5RlkIpUaR91k:b-wh5oqc) (no related file):*

I like this idea, let me do a quick PR adding this.
</blockquote></details>

Done, see https://github.com/oscarlab/graphene/pull/1824.

---

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz and @thiagomacieira)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: 2 of 6 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow, @thiagomacieira, and @yamahata)

___
*[LibOS/shim/include/shim_table.h, line 390 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM-NXm52xzje1UqPgL:-MHQR3Z154-m8lqWZBgG:b-6gjg7n) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/include/shim_table.h#L390)):*

Not exactly - they come from some old and unused feature which I deleted yesterday :) (or at least created a PR which will do so - #1822)
</blockquote></details>

Okay. Since it's not merged yet, I updated the return types there as well.
___
*[LibOS/shim/src/shim_parser.c, line 323 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM0xW99i57D4ReeGs9:-MHQSyBj6K15oEc4iw4Q:b-onj8yi) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/shim_parser.c#L323)):*

That's interesting. For me it prints `BROK: Failed to open FILE '/proc/2/stat' for reading`.
To have debug logs AFAIR you need both `loader.debug_type = inline` and a build with `DEBUG=1`, maybe you're missing `DEBUG=1`?
</blockquote></details>

Okay, that was a local setup issue (I had a `wait401.manifest` file laying around from trying things out). Now it works for me.

Anyway, the parser seems to work fine, except that 0 is printed as an empty string (e.g. `shim_wait4(2,0xabcd,,0x0)`). But it was like this before, and I'd rather leave it this way now, than add a special-case to this function.

(Maybe I will come back later and tidy the whole file up... I looked briefly at what strace does but it seems way too complex for us).
___
*[LibOS/shim/src/shim_syscalls.c, line 311 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM2IBEC8JDd_VXHI9c:-MHQWP517JpbS2L9HsXU:begkwoo) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/shim_syscalls.c#L311)):*

It controls both. This should get slightly less confusing after #1822 which removes this weird feature. Anyways, even with it in place, the `shim_<sc_name>` wrappers weren't usermode/"glibc-like" versions, they were just exports for binaries linked directly against LibOS (see LibOS/shim/test/inline directory).
</blockquote></details>

Okay, I updated that (along with the declarations for these).
___
*[LibOS/shim/src/sys/shim_wait.c, line 36 at r3](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHN8nt2BYA9dIoNRe_m:-MHQW_a46YviAzqkIOuV:b-tm2qde) ([raw file](https://github.com/oscarlab/graphene/blob/e078444ec9c0b9e28eca2cb0980a5bcf718622d7/LibOS/shim/src/sys/shim_wait.c#L36)):*

tabs -> spaces
If you're interested, we have recommended Vim and Emacs config files in https://graphene.readthedocs.io/en/latest/devel/setup.html to prevent such mistakes :)
</blockquote></details>

Ouch. That was a copy-paste from kernel sources :) Fixed.
___
*[LibOS/shim/test/ltp/ltp.cfg, line 3167 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHMBhPy41kmBg4ORq1K:-MHQY322EjLXyPldItOC:bwkwb41) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/test/ltp/ltp.cfg#L3167)):*

I checked `wait401` and for me it prints `Failed to open FILE '/proc/2/stat' for reading`. Could you place a comment before both tests stating why they are failing?
</blockquote></details>

Oh, I see now. Added comments to both tests.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @thiagomacieira)

___
*[LibOS/shim/src/shim_parser.c, line 323 at r2](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHM0xW99i57D4ReeGs9:-MHRy3Lp-2dqDKA2751V:bje0u05) ([raw file](https://github.com/oscarlab/graphene/blob/c1462208fd8adef6512cca860afee8ee8673349f/LibOS/shim/src/shim_parser.c#L323)):*

Okay, that was a local setup issue (I had a `wait401.manifest` file laying around from trying things out). Now it works for me.

Anyway, the parser seems to work fine, except that 0 is printed as an empty string (e.g. `shim_wait4(2,0xabcd,,0x0)`). But it was like this before, and I'd rather leave it this way now, than add a special-case to this function.

(Maybe I will come back later and tidy the whole file up... I looked briefly at what strace does but it seems way too complex for us).
</blockquote></details>

Sure, we can fix this later.

thiagomacieira:
Simple unit-test for this functionality:

```c
    pid_t pid = fork();
    if (pid == 0) {
        _exit(0);
    } else if (pid > 0) {
        siginfo_t info;
        waitid(P_ALL, 0, &info, WNOWAIT | WEXITED);
        waitid_works = (info.si_pid != 0);
        info.si_pid = 0;

        // now really reap the child
        waitid(P_PID, pid, &info, WEXITED);
        waitid_works = waitid_works && (info.si_pid != 0);
    }
```

mkow:
@thiagomacieira Our main question is whether the workload mentioned in #1809 started to work, it may contain more complicated uses of `waitid()`. The snippet you pasted uses `WNOWAIT` flag, is it used by this workload? I mean, we want to implement the minimal viable set of features for now, as some of them are hard to implement.

Also, the most standard uses should already be covered by the LTP test enabled in this PR ([`waitid01`](https://github.com/linux-test-project/ltp/blob/ac33ef87f0b9dc56533026435be3ea8c67ce86d2/testcases/kernel/syscalls/waitid/waitid01.c)).

thiagomacieira:
> @thiagomacieira Our main question is whether the workload mentioned in #1809 started to work, it may contain more complicated uses of `waitid()`. The snippet you pasted uses `WNOWAIT` flag, is it used by this workload? I mean, we want to implement the minimal viable set of features for now, as some of them are hard to implement.

It did not. This specific code section (including the WNOWAIT) crashed Graphene. I've asked the team to get a backtrace, but I haven't got it yet.

Note that the child side crashed.

thiagomacieira:
I did get a log file:
```
[P6544:T1:testapp] ---- shim_clone(CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|[SIGCHLD],0x0,0x0,0x3d6ed50,0x0) ... clone
[P6544:T1:testapp] clone without CLONE_FS is not yet implemented
[P6544:T1:testapp] clone without CLONE_SIGHAND is not yet implemented
[P6544:T1:testapp] clone without CLONE_SYSVSEM is not yet implemented
[...]
[P6565:T4:testapp] ---- shim_set_robust_list(0x3d6ed60,24) = 0
[P6565:T4:testapp] memory fault at 0x00000000 (IP = 0x022fec59)
[P6565:T4:testapp] signal 11 handled
[P6565:T4:testapp] run signal handler 0xf63259f0 (11, 0xfbbe2e60, 0xfbbe2ee8)
[P6565:T4:testapp] killed by signal 11
```

mkow:
@thiagomacieira I think the test you pasted is bugged, not our implementation - it doesn't check `waitid` result and then uses uninitialized `info` variable (which in your run resulted in a NULL deref).

thiagomacieira:
> @thiagomacieira I think the test you pasted is bugged, not our implementation - it doesn't check `waitid` result and then uses uninitialized `info` variable (which in your run resulted in a NULL deref).

I disagree. The parameters are organised in such a way that the call cannot fail on a compliant implementation.
```c
        siginfo_t info;
        waitid(P_ALL, 0, &info, WNOWAIT | WEXITED);
        waitid_works = (info.si_pid != 0);
```

Don't confuse `WNOWAIT` with `WNOHANG`. This test "hangs" in the sense that it will block until the child has exited. But once it has, it won't actually `wait` on it, so the child can be waited upon again. The call cannot fail, because:
# we have a child (we `fork`ed above)
# we passed `P_ALL` so we will get the state of any child that has exited (and will block if there isn't one yet)

And since this call succeeds, `info.si_pid` will be the PID of that child.
```c 
        info.si_pid = 0;

        // now really reap the child
        waitid(P_PID, pid, &info, WEXITED);
        waitid_works = waitid_works && (info.si_pid != 0);
```

Because the first call was `WNOWAIT`, we can execute the call again and this time we specified `P_PID` with our actual child. So this call must not block (because the child has already exited, even if the call doesn't have `WNOHANG`) and must set `info.si_pid` again.

Even if the call failed for some reason (out of resources, ENOSYS, etc.) then `info.si_pid` will remain zero. So checking that it changed is an effective verification that the call did succeed.

mkow:
> The parameters are organised in such a way that the call cannot fail on a compliant implementation.

I don't think you can assume this, you should always check for syscalls' return values. E.g. the kernel you're running on may not support this syscall or all the flags you passed. In this case you cannot assume that the result returned in output arguments is valid.

> Even if the call failed for some reason (out of resources, ENOSYS, etc.) then info.si_pid will remain zero.

What do you mean by "remain" here? `info` is a stack variable which was not initialized anywhere before being passed to `waitid()`, so it wasn't previously zero.

thiagomacieira:
> > Even if the call failed for some reason (out of resources, ENOSYS, etc.) then info.si_pid will remain zero.
> 
> What do you mean by "remain" here? `info` is a stack variable which was not initialized anywhere before being passed to `waitid()`, so it wasn't previously zero.

The first call forgot to initialise the variable, but the second did. So if the system call succeeded (and it has to, if it's implemented), the `si.si_pid` value must now be equal to `pid`. The other places in my code where `waitid` is called do intialise it properly:

```c
static int isChildReady(pid_t pid, siginfo_t *info)
{
    info->si_pid = 0;
    return waitid(P_PID, pid, info, WEXITED | WNOHANG | WNOWAIT) == 0 && info->si_pid == pid;
}
```

`` c
        /* we have waitid(2), which gets us some payload values on some systems */
        siginfo_t info;
        info.si_pid = 0;
        int ret = waitid(P_PID, pid, &info, WEXITED | WNOHANG) == 0 && info.si_pid == pid;
        if (!ret)
            return ret;
```

```c
        /* be optimistic: try to see if we can get the child that exited */
search_next_child:
        /* waitid returns -1 ECHILD if there are no further children at all;
         * it returns 0 and sets si_pid to 0 if there are children but they are not ready
         * to be waited (we're passing WNOHANG). We should not get EINTR because
         * we're passing WNOHANG and we should definitely not get EINVAL or anything else.
         * That means we can actually ignore the return code and only inspect si_pid.
         */
        info.si_pid = 0;
        waitid(P_ALL, 0, &info, WNOHANG | WNOWAIT | WEXITED);
        if (info.si_pid == 0) {
            /* there are no further un-waited-for children, so we can just exit.
             */
            return;
        }
```

thiagomacieira:
Either way, even if this code is wrong, it can't crash. And most assuredly, the **child** can't crash. And yet it did.

Awaiting more info from the team that ran this code.

pwmarcz:
Thanks @thiagomacieira for the information, I think it's clear that we do need `WNOWAIT` to work - I'm looking into that now.

As for the crash, so far I wasn't able to reproduce it - [the code snippet](https://github.com/oscarlab/graphene/pull/1821#issuecomment-696230904) runs under SGX for me, and both threads exit normally.

This looks like the relevant part of the log. @mkow, any ideas?

```
[P6565:T4:testapp] ---- shim_set_robust_list(0x3d6ed60,24) = 0
[P6565:T4:testapp] memory fault at 0x00000000 (IP = 0x022fec59)
[P6565:T4:testapp] signal 11 handled
[P6565:T4:testapp] run signal handler 0xf63259f0 (11, 0xfbbe2e60, 0xfbbe2ee8)
[P6565:T4:testapp] killed by signal 11
```

pwmarcz:
Alright, I pushed out an implementation for `WNOWAIT`, and also (if I read this correctly) fixed a reference counting bug in the code path for `P_PGID` or `P_ALL`. That might have been responsible for the crash of the child thread.

mkow:
> So if the system call succeeded (and it has to, if it's implemented)

I don't think you can rely on this, syscalls may fail due to various reasons and not all of them are well documented. Also, it may be a case that the specific kernel you're using doesn't support all the flags you passed (e.g. someone running on an older kernel).

> The other places in my code where waitid is called do intialise it properly:

Yeah, but they also check the return value, except the third listing, which is IMO incorrect - you shouldn't rely on the state of `info` after a failed call (it may work on the current kernel but break in the future).

> Either way, even if this code is wrong, it can't crash

Yup, now that I'm looking at it, the use of uninitialized memory in the test you pasted shouldn't cause any NULL derefs. I guess it just triggered an unrelated bug in Graphene (unrelated to this PR).


Anyways, Paweł implemented this additional flag and fixed one refcounting bug, so maybe it will start to work now.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz and @thiagomacieira)

___
*[LibOS/shim/src/sys/shim_wait.c, line 154 at r5](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHqldrx1eNBd4HaPTvo:-MHqldrx1eNBd4HaPTvp:bf50fh0) ([raw file](https://github.com/oscarlab/graphene/blob/4d37c139a1b1fab20677ed2d72d2ad7c01c10752/LibOS/shim/src/sys/shim_wait.c#L154)):*
> ```C
>             infop->si_code = CLD_EXITED;
>             infop->si_status = thread->exit_code;
>         } else  {
> ```

Here we also need:
```
else if (thread->term_signal & 0x80) {
    infop->si_code = CLD_DUMPED;
    infop->si_status = thread->term_signal ^ 0x80;
} else {
```
Actually the `0x80` constant should be introduced to some header, currently it's only defined in `LibOS/shim/src/bookkeep/shim_signal.c`
___
*[LibOS/shim/src/sys/shim_wait.c, line 209 at r5](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHqmV1R77lCY3WGdGKi:-MHqmV1R77lCY3WGdGKj:b-g2forh) ([raw file](https://github.com/oscarlab/graphene/blob/4d37c139a1b1fab20677ed2d72d2ad7c01c10752/LibOS/shim/src/sys/shim_wait.c#L209)):*
> ```C
>             return -EFAULT;
> 
>         if (info.si_code == CLD_EXITED)
> ```

Need to handle `CLD_DUMPED`, see the comment above.

boryspoplawski (changes requested):
AD return values: syscalls can always fail and not checking the return value doesn't make sense to me, you don't even gain any speedups with it ...
BUT, that being said, this could should work in practice and Graphene definitely has a bug.
Also Graphene should (currently does not) zero out `si_pid` filed, if there are no exited children and `WNOHANG` was specified. Note that this only applies to a successful call, it it fails, then it sgould **not** zero out those fields.

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821#-:-MHqpZIb9jSMhdyM7qrP:b-98dfu0)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz and @thiagomacieira)

___
*[LibOS/shim/src/sys/shim_wait.c, line 58 at r5](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHqoJTOC7X4nyD3WXp0:-MHqoJTP5i3c8P5PRHc6:b-iz6elt) ([raw file](https://github.com/oscarlab/graphene/blob/4d37c139a1b1fab20677ed2d72d2ad7c01c10752/LibOS/shim/src/sys/shim_wait.c#L58)):*
> ```C
>                 goto block_pid;
>             put_thread(thread);
>             return 0;
> ```

We should zero out `si_pid` and `si_signo` here. For more info `man waitid(2)`
___
*[LibOS/shim/src/sys/shim_wait.c, line 125 at r5](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHqophC3XHTCL3jBqtc:-MHqophC3XHTCL3jBqtd:bltxl6b) ([raw file](https://github.com/oscarlab/graphene/blob/4d37c139a1b1fab20677ed2d72d2ad7c01c10752/LibOS/shim/src/sys/shim_wait.c#L125)):*
> ```C
> 
>     unlock(&cur->lock);
>     return 0;
> ```

ditto (zero out fields)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: 5 of 8 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @thiagomacieira)

___
*[LibOS/shim/src/sys/shim_wait.c, line 50 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHrbTR5UGVeb5ho4at:-MHuHs182InToVjNaCWy:b-byjdjn) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/sys/shim_wait.c#L50)):*

Hmm this godo is ugly. But it can be addressed as follow up.
</blockquote></details>

Agreed but I'd rather not change too much in this PR.
___
*[LibOS/shim/src/sys/shim_wait.c, line 105 at r1](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHHro4IEsB-pjNWuDPk:-MHuHwF_1sI7DwLxuhvQ:bdgzt23) ([raw file](https://github.com/oscarlab/graphene/blob/9cbb45067229b9034051787fbbee5af5bfceb482/LibOS/shim/src/sys/shim_wait.c#L105)):*

This goto is ugly.
</blockquote></details>

As above, I agree but I'd rather not change too much at the same time.
___
*[LibOS/shim/src/sys/shim_wait.c, line 58 at r5](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHqoJTOC7X4nyD3WXp0:-MHuHnaq9D_oahgIsidH:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/4d37c139a1b1fab20677ed2d72d2ad7c01c10752/LibOS/shim/src/sys/shim_wait.c#L58)):*

We should zero out `si_pid` and `si_signo` here. For more info `man waitid(2)`
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_wait.c, line 125 at r5](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHqophC3XHTCL3jBqtc:-MHuHq8E2s9tBkkyt_4E:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/4d37c139a1b1fab20677ed2d72d2ad7c01c10752/LibOS/shim/src/sys/shim_wait.c#L125)):*

ditto (zero out fields)
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_wait.c, line 154 at r5](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHqldrx1eNBd4HaPTvo:-MHuJ4-O6wUR2-P2rBsZ:bxoirs6) ([raw file](https://github.com/oscarlab/graphene/blob/4d37c139a1b1fab20677ed2d72d2ad7c01c10752/LibOS/shim/src/sys/shim_wait.c#L154)):*

Here we also need:
```
else if (thread->term_signal & 0x80) {
    infop->si_code = CLD_DUMPED;
    infop->si_status = thread->term_signal ^ 0x80;
} else {
```
Actually the `0x80` constant should be introduced to some header, currently it's only defined in `LibOS/shim/src/bookkeep/shim_signal.c`
</blockquote></details>

Done, the constant (`__WCOREDUMP_BIT`) was defined there and I moved it to `shim_signal.h`.
___
*[LibOS/shim/src/sys/shim_wait.c, line 209 at r5](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHqmV1R77lCY3WGdGKi:-MHuLMJW2AYu6JDpTMZj:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/4d37c139a1b1fab20677ed2d72d2ad7c01c10752/LibOS/shim/src/sys/shim_wait.c#L209)):*

Need to handle `CLD_DUMPED`, see the comment above.
</blockquote></details>

Done.

mkow (changes requested):
> Reviewed 2 of 2 files at r5, 3 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski, @pwmarcz, and @thiagomacieira)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHujuBj6zE5dtpgAbE1:-MHujuBk5BJobmyuLGRY:b-5zcke9) (no related file):*
Making a blocking comment for the final rebase: We'll need to update the commit message when rebasing, because `WNOWAIT` is now implemented.

---
___
*[LibOS/shim/src/shim_parser.c, line 323 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHul7mT2XH1B-4lZ02G:-MHul7mT2XH1B-4lZ02H:b-o2pla) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/shim_parser.c#L323)):*
> ```C
>         [__NR_mq_getsetattr]   = {.slow = 0, .parser = {NULL}},
>         [__NR_kexec_load]      = {.slow = 0, .parser = {NULL}},
>         [__NR_waitid]      = {.slow = 1, .parser = {NULL, NULL, NULL, &parse_wait_options, NULL}},
> ```

Could you also pretty-print `which`? I was just looking at some logs and it would be quite useful to also pretty-print this argument.
When doing so, please also parse `P_PIDFD` option (https://elixir.bootlin.com/linux/v5.8.10/source/include/uapi/linux/wait.h#L20), but I'm not sure if we need and can support it in the implementation.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz and @thiagomacieira)

___
*[LibOS/shim/src/sys/shim_wait.c, line 62 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHuyDb305E_bPLTz2E2:-MHuyDb305E_bPLTz2E3:bcn7t3w) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/sys/shim_wait.c#L62)):*
> ```C
> 
>             if (infop) {
>                 if (test_user_memory(infop, sizeof(*infop), /*write=*/true))
> ```

Just add `test_user_memory` at the beginning of this function and remove from all write-sites.
___
*[LibOS/shim/src/sys/shim_wait.c, line 136 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHuykLu2PjQPrzmn59H:-MHuykLv09tgb06EilSC:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/sys/shim_wait.c#L136)):*
> ```C
> 
>     if (infop) {
>         if (test_user_memory(infop, sizeof(*infop), /*write=*/true))
> ```

ditto
___
*[LibOS/shim/src/sys/shim_wait.c, line 158 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHuys0F9XZAC9pTGgvX:-MHuys0F9XZAC9pTGgvY:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/sys/shim_wait.c#L158)):*
> ```C
>     ret = 0;
>     if (infop) {
>         if (test_user_memory(infop, sizeof(*infop), /*write=*/true)) {
> ```

ditto
___
*[LibOS/shim/src/sys/shim_wait.c, line 228 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHuzYb61vIsCYxdhd4b:-MHuzYb61vIsCYxdhd4c:b-wpntlo) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/sys/shim_wait.c#L228)):*
> ```C
>             return -EFAULT;
> 
>         if (info.si_code == CLD_EXITED)
> ```

nitpick/personal opinion (not blocking) I really dislike mulit-clause if-else statements without `{` `}` blocks

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz and @thiagomacieira)

___
*[LibOS/shim/src/sys/shim_wait.c, line 14 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHv0rGW8jqvVcDEtW9s:-MHv0rGW8jqvVcDEtW9t:bimz3ch) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/sys/shim_wait.c#L14)):*
> ```C
> #include <sys/mman.h>
> #include <sys/syscall.h>
> #include <sys/wait.h>
> ```

Do not include headers from `sys/`, they might contain conflicting definitions of functions etc.
Use only those in `linux/` (they contain only structs, typedefs and defines). I believe, those few structs that cannot be found there, we have defined manually in Graphene internal headers.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski, @mkow, @pwmarcz, and @thiagomacieira)

___
*[LibOS/shim/src/shim_parser.c, line 323 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHul7mT2XH1B-4lZ02G:-MHv7enVAoMqiq10BP_u:btoxn9t) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/shim_parser.c#L323)):*

Could you also pretty-print `which`? I was just looking at some logs and it would be quite useful to also pretty-print this argument.
When doing so, please also parse `P_PIDFD` option (https://elixir.bootlin.com/linux/v5.8.10/source/include/uapi/linux/wait.h#L20), but I'm not sure if we need and can support it in the implementation.
</blockquote></details>

Done. Looks like P_PIDFD is newer than the kernel for Ubuntu 18 (4.15.0), so I added it under `#ifdef`.
___
*[LibOS/shim/src/sys/shim_wait.c, line 14 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHv0rGW8jqvVcDEtW9s:-MHv8gTA3WhICPitQcFC:br5z526) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/sys/shim_wait.c#L14)):*

Do not include headers from `sys/`, they might contain conflicting definitions of functions etc.
Use only those in `linux/` (they contain only structs, typedefs and defines). I believe, those few structs that cannot be found there, we have defined manually in Graphene internal headers.
</blockquote></details>

I see the code in `Pal` including `sys/wait.h`, does this apply only to LibOS?

Anyway - it looks like the kernel uses a raw `0x80`. I added a `#define WCOREFLAG` at the beginning of this file; unless you can think of a better place?
___
*[LibOS/shim/src/sys/shim_wait.c, line 62 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHuyDb305E_bPLTz2E2:-MHvB-0s167IBsc4uccW:bm3yqt3) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/sys/shim_wait.c#L62)):*

Just add `test_user_memory` at the beginning of this function and remove from all write-sites.
</blockquote></details>

OK, done.
___
*[LibOS/shim/src/sys/shim_wait.c, line 136 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHuykLu2PjQPrzmn59H:-MHvB0G_1ZDdH9nY3eCs:bm3yqt3) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/sys/shim_wait.c#L136)):*

ditto
</blockquote></details>

OK, done.
___
*[LibOS/shim/src/sys/shim_wait.c, line 158 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHuys0F9XZAC9pTGgvX:-MHvB0msDFN4Z6-I0YF1:bm3yqt3) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/sys/shim_wait.c#L158)):*

ditto
</blockquote></details>

OK, done.
___
*[LibOS/shim/src/sys/shim_wait.c, line 228 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHuzYb61vIsCYxdhd4b:-MHvBEgPEQdG6doTt5EU:bgckuqw) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/sys/shim_wait.c#L228)):*

nitpick/personal opinion (not blocking) I really dislike mulit-clause if-else statements without `{` `}` blocks
</blockquote></details>

OK, updated.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @pwmarcz, and @thiagomacieira)

___
*[LibOS/shim/src/shim_parser.c, line 323 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHul7mT2XH1B-4lZ02G:-MHvEuDU1IxccsQEE4dp:b8bgcqy) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/shim_parser.c#L323)):*

Done. Looks like P_PIDFD is newer than the kernel for Ubuntu 18 (4.15.0), so I added it under `#ifdef`.
</blockquote></details>

Ideally, we should have our own copies of all required system defines - we aren't emulating the Linux which is present on the build machine, but some specific version. Anyways, this is an unrelated issue we need to finally address.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz and @thiagomacieira)

___
*[LibOS/shim/src/sys/shim_wait.c, line 14 at r6](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHv0rGW8jqvVcDEtW9s:-MHvQd8L46PUFwtSDjyV:b-ysk549) ([raw file](https://github.com/oscarlab/graphene/blob/b0d94efdf33292e41c2fdc69ac3e15306eba1450/LibOS/shim/src/sys/shim_wait.c#L14)):*

I see the code in `Pal` including `sys/wait.h`, does this apply only to LibOS?

Anyway - it looks like the kernel uses a raw `0x80`. I added a `#define WCOREFLAG` at the beginning of this file; unless you can think of a better place?
</blockquote></details>

Well, a lot of code here is far from perfect :)
AFAIK this flag is exactly the same as `___WCOREDUMP_BIT` (i.e. it's not just a coincidence the value is the same, it's really the same flag), but I also think this file is soon to be rewritten anyway, so maybe live it as it is to merge this PR faster.

mkow (approved):
> Reviewed 6 of 6 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz and @thiagomacieira)

mkow:
@thiagomacieira I prepared this PR for merging, as it looks ready. Could you take a final look at it and resolve the review discussions you still have open?
Ad P_PIDFD: We can either implement it as a separate PR or look at the clone flags which are needed as an alternative. But I'd like to merge this PR without adding any more features to it.

boryspoplawski (approved):
> Reviewed 1 of 2 files at r7, 1 of 6 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 3 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz and @thiagomacieira)

thiagomacieira:
PIDFD support is not required for our workload to work. If you do implement `P_PIDFD` for `waitid`, please make sure it's also supported in `poll` and in `clone` (`CLONE_PIDFD`). Don't break the assumptions from https://github.com/qt/qtbase/blob/dev/src/3rdparty/forkfd/forkfd_linux.c#L100-L123

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 3 unresolved discussions (waiting on @pwmarcz and @thiagomacieira)

___
*[LibOS/shim/src/sys/shim_wait.c, line 216 at r8](https://reviewable.io/reviews/oscarlab/graphene/1821#-MHxisg12mLgTM9wH_Zc:-MHxisg12mLgTM9wH_Zd:b-9lmo0r) ([raw file](https://github.com/oscarlab/graphene/blob/cbb3bc2958a1f086a38d546bf93c910f2a448f72/LibOS/shim/src/sys/shim_wait.c#L216)):*
> ```C
> 
>     info.si_pid = 0;
>     int ret = shim_do_waitid(which, id, &info, options, ru);
> ```

This should be refactored. But for now, it's okay.

yamahata (approved):
> Reviewed 1 of 3 files at r6, 1 of 2 files at r7, 6 of 6 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: all files reviewed, 3 unresolved discussions (waiting on @pwmarcz and @thiagomacieira)

mkow (approved):
> Dismissed @thiagomacieira from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1821)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1852, opened 2020-09-29
[LibOS, Pal/Linux-SGX] Signal handling bug fixes

dimakuv:
## Description of the changes 

This PR has two bug fixes:

1. [LibOS] Wake up parent thread on SIGCHLD from child. Previously, Graphene had a bug of trying to wake up the (remote) child thread on SIGCHLD. This was a nop. It also meant that the parent thread, who was waiting on the child, wasn't interrupted and continued hanging instead of handling SIGCHLD.

2. [Pal/Linux-SGX] Save XSAVE area of normal app context on sync signals. Previously, when a synchronous signal like SIGFPE/SIGILL happened in the application (e.g., due to div-by-zero or SGX-prohibited SYSCALL instruction), Linux-SGX assembly routine incorrectly saved the XSAVE area (containing among other things FP/SSE/AVX registers) of the signal context. This commit fixes this bug by extracting the XSAVE area of the normal app context, i.e., XSAVE of SSA 0.

This was detected on a large statically built application. Kudos to @thiagomacieira for help in debugging and analyzing the issues.

## How to test this PR? 

New LibOS regression test is added to test bug fix 2. The test is x64-only and assumes an SSE2 enabled CPU.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)

### DIFF
--- a/LibOS/shim/src/sys/shim_exit.c
+++ b/LibOS/shim/src/sys/shim_exit.c
@@ -67,8 +67,8 @@ int thread_destroy(struct shim_thread* thread, bool send_ipc) {
             info.si_status = (exit_code & 0xff) << 8;
 
             if (append_signal(parent, &info) >= 0) {
-                thread_wakeup(thread);
-                DkThreadResume(thread->pal_handle);
+                thread_wakeup(parent);
+                DkThreadResume(parent->pal_handle);
             }
         }
 
--- a/LibOS/shim/test/regression/.gitignore
+++ b/LibOS/shim/test/regression/.gitignore
@@ -68,6 +68,7 @@
 /shared_object
 /sigaction_per_process
 /sigaltstack
+/sighandler_divbyzero
 /sighandler_reset
 /sighandler_sigpipe
 /signal_multithread
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -3,7 +3,8 @@ include ../../../../Scripts/Makefile.configs
 c_executables-x86_64 = \
 	attestation \
 	cpuid \
-	rdtsc
+	rdtsc \
+	sighandler_divbyzero
 
 c_executables = \
 	abort \
--- a/LibOS/shim/test/regression/sighandler_divbyzero.c
+++ b/LibOS/shim/test/regression/sighandler_divbyzero.c
@@ -1,0 +1,62 @@
+#define _GNU_SOURCE
+#include <emmintrin.h> // Intel SSE2 (XMM) intrinsics
+#include <err.h>
+#include <errno.h>
+#include <signal.h>
+#include <stdatomic.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <ucontext.h>
+
+static atomic_int sigfpe_ctr = 0;
+
+static void sigfpe_handler(int signum, siginfo_t* si, void* uc) {
+    printf("Got signal %d\n", signum);
+    if (signum == SIGFPE) {
+        ((ucontext_t*)uc)->uc_mcontext.gregs[REG_RBX] = 1; /* fix divisor */
+        sigfpe_ctr++;
+    }
+}
+
+int main(int argc, char** argv) {
+    int ret;
+
+    const struct sigaction act = {
+        .sa_sigaction = sigfpe_handler,
+        .sa_flags = SA_SIGINFO,
+    };
+
+    ret = sigaction(SIGFPE, &act, NULL);
+    if (ret < 0) {
+        err(EXIT_FAILURE, "sigaction failed");
+    }
+
+    int16_t in[8] __attribute__((aligned(16))) = {0, 1, 2, 3, 4, 5, 6, 7};
+    int16_t out[8] __attribute__((aligned(16))) = {0};
+
+    __asm__ volatile("pxor %%xmm0, %%xmm0\n" /* clear xmm0 for sanity */
+                     "movdqa %1, %%xmm0\n"   /* move `in` into xmm0 */
+                     "movq $1, %%rax\n"      /* force div-by-zero exception via 1/0 */
+                     "cqo\n"
+                     "movq $0, %%rbx\n"
+                     "divq %%rbx\n"
+                     "movdqa %%xmm0, %0\n"   /* move xmm0 into `out` */
+                     "pxor %%xmm0, %%xmm0\n" /* clear xmm0 for sanity */
+                     :"=m"(out) : "m"(in) : "xmm0", "rax", "rbx", "rdx", "cc", "memory");
+
+    for (int i = 0; i < 8; i++) {
+        if (out[i] != in[i]) {
+            errx(EXIT_FAILURE, "XSAVE state (XMM registers' values) was lost!");
+        }
+    }
+
+    if (sigfpe_ctr != 1) {
+        errx(EXIT_FAILURE, "Expected exactly 1 SIGFPE signal but received %d!", sigfpe_ctr);
+    }
+
+    printf("Got %d SIGFPE signal(s)\n", sigfpe_ctr);
+    puts("TEST OK");
+    exit(EXIT_SUCCESS);
+}
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -1,9 +1,10 @@
 #!/usr/bin/env python3
 
 import os
-import unittest
 import shutil
+import signal
 import subprocess
+import unittest
 
 from regression import (
     HAS_SGX,
@@ -440,13 +441,13 @@ class TC_30_Syscall(RegressionTestCase):
         # Sigaltstack Test
         self.assertIn('OK on sigaltstack in main thread before alarm', stdout)
         self.assertIn('&act == 0x', stdout)
-        self.assertIn('sig 14 count 1 goes off with sp=0x', stdout)
+        self.assertIn('sig %d count 1 goes off with sp=0x' % signal.SIGALRM, stdout)
         self.assertIn('OK on signal stack', stdout)
         self.assertIn('OK on sigaltstack in handler', stdout)
-        self.assertIn('sig 14 count 2 goes off with sp=0x', stdout)
+        self.assertIn('sig %d count 2 goes off with sp=0x' % signal.SIGALRM, stdout)
         self.assertIn('OK on signal stack', stdout)
         self.assertIn('OK on sigaltstack in handler', stdout)
-        self.assertIn('sig 14 count 3 goes off with sp=0x', stdout)
+        self.assertIn('sig %d count 3 goes off with sp=0x' % signal.SIGALRM, stdout)
         self.assertIn('OK on signal stack', stdout)
         self.assertIn('OK on sigaltstack in handler', stdout)
         self.assertIn('OK on sigaltstack in main thread', stdout)
@@ -468,7 +469,7 @@ class TC_30_Syscall(RegressionTestCase):
 
     def test_090_sighandler_reset(self):
         stdout, _ = self.run_binary(['sighandler_reset'])
-        self.assertIn('Got signal 17', stdout)
+        self.assertIn('Got signal %d' % signal.SIGCHLD, stdout)
         self.assertIn('Handler was invoked 1 time(s).', stdout)
 
     def test_091_sigaction_per_process(self):
@@ -482,16 +483,16 @@ class TC_30_Syscall(RegressionTestCase):
         except subprocess.CalledProcessError as e:
             # FIXME: It's unclear what Graphene process should return when the app
             # inside dies due to a signal.
-            self.assertTrue(e.returncode in [13, 141])
+            self.assertTrue(e.returncode in [signal.SIGPIPE, 128 + signal.SIGPIPE])
             stdout = e.stdout.decode()
-            self.assertIn('Got signal 13', stdout)
+            self.assertIn('Got signal %d' % signal.SIGPIPE, stdout)
             self.assertIn('Got 1 SIGPIPE signal(s)', stdout)
             self.assertIn('Could not write to pipe: Broken pipe', stdout)
 
     @unittest.skipUnless(ON_X86, "x86-specific")
     def test_093_sighandler_divbyzero(self):
         stdout, _ = self.run_binary(['sighandler_divbyzero'])
-        self.assertIn('Got signal 8', stdout)
+        self.assertIn('Got signal %d' % signal.SIGFPE, stdout)
         self.assertIn('Got 1 SIGFPE signal(s)', stdout)
         self.assertIn('TEST OK', stdout)
 
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -488,7 +488,14 @@ class TC_30_Syscall(RegressionTestCase):
             self.assertIn('Got 1 SIGPIPE signal(s)', stdout)
             self.assertIn('Could not write to pipe: Broken pipe', stdout)
 
-    def test_093_signal_multithread(self):
+    @unittest.skipUnless(ON_X86, "x86-specific")
+    def test_093_sighandler_divbyzero(self):
+        stdout, _ = self.run_binary(['sighandler_divbyzero'])
+        self.assertIn('Got signal 8', stdout)
+        self.assertIn('Got 1 SIGFPE signal(s)', stdout)
+        self.assertIn('TEST OK', stdout)
+
+    def test_094_signal_multithread(self):
         stdout, _ = self.run_binary(['signal_multithread'])
         self.assertIn('TEST OK', stdout)
 
--- a/Pal/src/host/Linux-SGX/enclave_entry.S
+++ b/Pal/src/host/Linux-SGX/enclave_entry.S
@@ -466,11 +466,18 @@ enclave_entry:
 	leaq _DkExceptionHandler(%rip), %rdi
 	movq %rdi, SGX_GPR_RIP(%rbx)
 
+	# dump the XSAVE region (XMM/YMM/etc part of context); SGX saves it at the
+	# very beginning of the SSA frame; note that __restore_xregs / __save_xregs
+	# clobber RDX so need to stash it in RBX
 	movq %rdx, %rbx
-	leaq SGX_CPU_CONTEXT_SIZE + 8(%rsi), %rdi
+	movq %gs:SGX_SSA, %rdi
 	leaq 1f(%rip), %r11
-	jmp __save_xregs
+	jmp __restore_xregs
 1:
+	leaq SGX_CPU_CONTEXT_SIZE + 8(%rsi), %rdi
+	leaq 2f(%rip), %r11
+	jmp __save_xregs
+2:
 	movq %rbx, %rdx
 
 .Leexit_exception:


mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski)

boryspoplawski (changes requested):
> Reviewed 5 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @dimakuv)

___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP1ylLCHa-YCF2fUui:-MIP1ylM80DKYj1PkAyn:brz0xp0) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L26)):*
> ```C
> 
>     const struct sigaction act = {
>         .sa_sigaction = sigfpe_handler,
> ```

We need to set `.sa_flags` here as well, at least to `SA_SIGINFO` (potentially also: `SA_ONSTACK`).
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 31 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP2LYtC_u7z4w83tBi:-MIP2LYtC_u7z4w83tBj:b-ddt7ov) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L31)):*
> ```C
>     ret = sigaction(SIGFPE, &act, NULL);
>     if (ret < 0) {
>         fprintf(stderr, "sigaction failed: %s\n", strerror(errno));
> ```

Not blocking, just for the future: `err` is a nice shorthand for these two calls.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 36 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP5ILQ6tEReR78_Mkj:-MIP5ILQ6tEReR78_Mkk:bg4guwn) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L36)):*
> ```C
> 
>     /* populate array with 0,1,2,3... via argc (this test expects no command-line options and thus
>      * argc == 1); this is to prevent the compiler from optimizing out the XMM logic */
> ```

If `argc` stays, please add a check that it's really `1`.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 37 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP3qMOBEMfgaVcTltK:-MIP3qMOBEMfgaVcTltL:b-m1gtqm) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L37)):*
> ```C
>     /* populate array with 0,1,2,3... via argc (this test expects no command-line options and thus
>      * argc == 1); this is to prevent the compiler from optimizing out the XMM logic */
>     int16_t a[8];
> ```

This needs to be 16-byte aligned.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 41 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP40UcDIlnMF0OzU10:-MIP40UcDIlnMF0OzU11:bd6fx8l) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L41)):*
> ```C
>         a[i] = argc - 1 + i;
>     __m128i xmm_reg = _mm_load_si128((__m128i*)a);
>     __asm__ volatile("" ::: "memory");
> ```

Since we are using inline assembly anyway, why not just use `movdqa` directly? We wouldn't need to worry about compiler optimizing out memory writes and use `argc` trick.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP3mpT2hPGYcV4xFv3:-MIP3mpT2hPGYcV4xFv4:b-o94lvj) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L49)):*
> ```C
>                      ::: "rax", "rbx", "rdx", "cc", "memory");
> 
>     if (_mm_extract_epi16(xmm_reg, 0) != 0 || _mm_extract_epi16(xmm_reg, 1) != 1 ||
> ```

This looks like a good place to use `for` loop :)
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP4rlY4GyiEfSGsngW:-MIP4rlZ6D0WNZ-ztMw7:b85dsr9) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L49)):*
> ```C
>                      ::: "rax", "rbx", "rdx", "cc", "memory");
> 
>     if (_mm_extract_epi16(xmm_reg, 0) != 0 || _mm_extract_epi16(xmm_reg, 1) != 1 ||
> ```

I'm not sure how "smart" is gcc around these areas, but can't it optimize away these comparisons? `xmm_reg` is not memory location and there there is no dependency on it in the inline asm.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 58 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP5NcF8Ln9tynYffHv:-MIP5NcF8Ln9tynYffHw:b-sbqft9) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L58)):*
> ```C
> 
>     __asm__ volatile("" ::: "memory");
>     printf("Got %d SIGFPE signal(s)\n", sigfpe_ctr);
> ```

Can we add a check for an expected `sigfpe_ctr` value here?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: 5 of 6 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP1ylLCHa-YCF2fUui:-MIPDGiO-RHo_vIhzbW-:b-7cmmoe) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L26)):*

We need to set `.sa_flags` here as well, at least to `SA_SIGINFO` (potentially also: `SA_ONSTACK`).
</blockquote></details>

Done. We hit the segfault even without the use of alternate stack, so didn't add `SA_ONSTACK`.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 31 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP2LYtC_u7z4w83tBi:-MIPDj2LDPMjPLnv0oRK:b-qefv0h) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L31)):*

Not blocking, just for the future: `err` is a nice shorthand for these two calls.
</blockquote></details>

I don't like this because in some cases we don't have any errno to report (like below). And we end up using two different ways of reporting failures in the same program. A matter of taste.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 36 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP5ILQ6tEReR78_Mkj:-MIPEW9c6SP-tB2I-9cL:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L36)):*

If `argc` stays, please add a check that it's really `1`.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 37 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP3qMOBEMfgaVcTltK:-MIPEL6q9BmWVmN72HJ7:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L37)):*

This needs to be 16-byte aligned.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 41 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP40UcDIlnMF0OzU10:-MIPEeSn7dhTpeu8Dtrw:b-l1uzdz) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L41)):*

Since we are using inline assembly anyway, why not just use `movdqa` directly? We wouldn't need to worry about compiler optimizing out memory writes and use `argc` trick.
</blockquote></details>

Mostly because it would result in a lot of not-so-readable assembly code. I have no particular preference but using intrinsics felt right here. However, I agree that the `argc` trick is ugly, is there a simpler way to init an XMM register to some constant without the fear of optimize-out?
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP4rlY4GyiEfSGsngW:-MIPFMDC8KN75_VORE-h:bg5rqdo) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L49)):*

I'm not sure how "smart" is gcc around these areas, but can't it optimize away these comparisons? `xmm_reg` is not memory location and there there is no dependency on it in the inline asm.
</blockquote></details>

Not sure what you mean. How could GCC optimize this away? We take the `argc` precaution so that GCC doesn't know the actual XMM reg values.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP3mpT2hPGYcV4xFv3:-MIPGlCL76JNI8zdZ5wc:b-eu1ggl) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L49)):*

This looks like a good place to use `for` loop :)
</blockquote></details>

GCC is not that smart... `error: selector must be an integer constant in the range 0..7`. I don't want to fight against GCC, so I'll leave it as is.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 58 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP5NcF8Ln9tynYffHv:-MIPFcCW8dz5hxvKhFil:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L58)):*

Can we add a check for an expected `sigfpe_ctr` value here?
</blockquote></details>

Done.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 31 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP2LYtC_u7z4w83tBi:-MIPVrtR6RLUJa6QFZ60:b-dzng1f) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L31)):*

I don't like this because in some cases we don't have any errno to report (like below). And we end up using two different ways of reporting failures in the same program. A matter of taste.
</blockquote></details>

There's `errx`, which does not append `strerror(errno)`.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 41 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP40UcDIlnMF0OzU10:-MIPa1ja5V0gbedxGSBw:b4e8raq) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L41)):*

Mostly because it would result in a lot of not-so-readable assembly code. I have no particular preference but using intrinsics felt right here. However, I agree that the `argc` trick is ugly, is there a simpler way to init an XMM register to some constant without the fear of optimize-out?
</blockquote></details>

Actually, this is just the same discussion about gcc optimizations, resolving this, let's just stick to the one thread below.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP3mpT2hPGYcV4xFv3:-MIPX98mE901TgV7_a7E:b-xjtrfk) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L49)):*

GCC is not that smart... `error: selector must be an integer constant in the range 0..7`. I don't want to fight against GCC, so I'll leave it as is.
</blockquote></details>

Oh, ok.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP4rlY4GyiEfSGsngW:-MIPXIicCwGnyCTfexNp:bovcduv) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L49)):*

Not sure what you mean. How could GCC optimize this away? We take the `argc` precaution so that GCC doesn't know the actual XMM reg values.
</blockquote></details>

1) Now we have a check for `argc`, which I think gcc is smart enough to notice.
2) I meant that gcc might be smart enough to notice that this xmm regiser is not changing, so it could skip using xmm at all, just replace those checks with something like `argc - 1 + 0 != 0` etc ...

Generally speaking, I'm not sure if we can express all we want to do here in C or C mixed with inline asm (setting a xmm regiser, doing div, then checking the reg value).

yamahata (approved):
> Reviewed 5 of 6 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/src/sys/shim_exit.c, line 70 at r2](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIPPLgKCUxKGfzwXxV9:-MIPPLgKCUxKGfzwXxVA:bl2exw) ([raw file](https://github.com/oscarlab/graphene/blob/0a7ec1cba7f55235ad13a05a2cb4752f0cdad1db/LibOS/shim/src/sys/shim_exit.c#L70)):*
> ```C
> 
>             if (append_signal(parent, &info) >= 0) {
>                 thread_wakeup(parent);
> ```

nice catch. Why has it hidden for so long?
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP4rlY4GyiEfSGsngW:-MIPcISX550PHN07KEkE:bf5tb5f) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L49)):*

1) Now we have a check for `argc`, which I think gcc is smart enough to notice.
2) I meant that gcc might be smart enough to notice that this xmm regiser is not changing, so it could skip using xmm at all, just replace those checks with something like `argc - 1 + 0 != 0` etc ...

Generally speaking, I'm not sure if we can express all we want to do here in C or C mixed with inline asm (setting a xmm regiser, doing div, then checking the reg value).
</blockquote></details>

Although I haven't tested it, how about 

register __m128 xmm0 asm("xmm0")?
__asm__ volatile (" <set xmm0>": "=v"(xmm0));
...
___
*[LibOS/shim/test/regression/test_libos.py, line 494 at r2](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIPTlA23RI65AosfJD1:-MIPTlA23RI65AosfJD2:b-qi9ii) ([raw file](https://github.com/oscarlab/graphene/blob/0a7ec1cba7f55235ad13a05a2cb4752f0cdad1db/LibOS/shim/test/regression/test_libos.py#L494)):*
> ```
> 8
> ```

nitpick: signal.SIGFPE can be used instead of magic number.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @yamahata)

___
*[LibOS/shim/src/sys/shim_exit.c, line 70 at r2](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIPPLgKCUxKGfzwXxV9:-MIPhAo8EGbGsii8vUTu:b-h3425o) ([raw file](https://github.com/oscarlab/graphene/blob/0a7ec1cba7f55235ad13a05a2cb4752f0cdad1db/LibOS/shim/src/sys/shim_exit.c#L70)):*

nice catch. Why has it hidden for so long?
</blockquote></details>

I doubt this code is run often - it's the case of destroying a remote thread (thread in another process), whatever that would mean.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP4rlY4GyiEfSGsngW:-MIPhSqz9iGoEdsGP1W3:b-hdiq2b) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L49)):*

Although I haven't tested it, how about 

register __m128 xmm0 asm("xmm0")?
__asm__ volatile (" <set xmm0>": "=v"(xmm0));
...


</blockquote></details>

The problem is that gcc might use `xmm0` in between asm blocks - we cannot stop it from doing so ...

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: 4 of 6 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @yamahata)

___
*[LibOS/shim/src/sys/shim_exit.c, line 70 at r2](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIPPLgKCUxKGfzwXxV9:-MIT9rh98pCq4DSQ4A9x:b-8j6gwd) ([raw file](https://github.com/oscarlab/graphene/blob/0a7ec1cba7f55235ad13a05a2cb4752f0cdad1db/LibOS/shim/src/sys/shim_exit.c#L70)):*

I doubt this code is run often - it's the case of destroying a remote thread (thread in another process), whatever that would mean.
</blockquote></details>

Well, this scenario doesn't happen often: the parent must be blocked on a non-waitpid syscall for this particular code to become a bug. Parent processes typically wait via `waitpid()` or similar, but my particular program (where this bug manifests) actually used `poll()` to wait.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 31 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP2LYtC_u7z4w83tBi:-MITBL0e9ya6s_Fwogd9:b-uqmyfz) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L31)):*

There's `errx`, which does not append `strerror(errno)`.
</blockquote></details>

Done. I learned something.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 41 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP40UcDIlnMF0OzU10:-MITKNLb8Y4mNuUr5mDf:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L41)):*

Actually, this is just the same discussion about gcc optimizations, resolving this, let's just stick to the one thread below.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP4rlY4GyiEfSGsngW:-MITKPuvEinRQ7wIA-eJ:b-uspc4x) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L49)):*

The problem is that gcc might use `xmm0` in between asm blocks - we cannot stop it from doing so ...
</blockquote></details>

Done. Ok, you pressed me enough to re-write everything in inline asm. Turned out to be extremely simple :) I wrote three separate asm blocks for readability, not sure if you agree.
___
*[LibOS/shim/test/regression/test_libos.py, line 494 at r2](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIPTlA23RI65AosfJD1:-MITNZ1235LjCjLYkj0n:bat1g4g) ([raw file](https://github.com/oscarlab/graphene/blob/0a7ec1cba7f55235ad13a05a2cb4752f0cdad1db/LibOS/shim/test/regression/test_libos.py#L494)):*

> ```
> 8
> ```

nitpick: signal.SIGFPE can be used instead of magic number.

</blockquote></details>

Done. In other places as well (all in a separate commit).

boryspoplawski (changes requested):
> Reviewed 1 of 6 files at r1, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @yamahata)

___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP4rlY4GyiEfSGsngW:-MIUXf8HAr3aDHTA1eD2:bskhfsc) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L49)):*

Done. Ok, you pressed me enough to re-write everything in inline asm. Turned out to be extremely simple :) I wrote three separate asm blocks for readability, not sure if you agree.
</blockquote></details>

Good idea with a out memory buffer!
Unfortunately I do not agree, this should be one asm block. There's nothing blocking gcc from changing xmm0 in between these blocks :)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: 5 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIP4rlY4GyiEfSGsngW:-MIUhuCf19ubfrj7RWrc:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3d10b9c30c577e23bd8648b207e3a9c9580ec88a/LibOS/shim/test/regression/sighandler_divbyzero.c#L49)):*

Good idea with a out memory buffer!
Unfortunately I do not agree, this should be one asm block. There's nothing blocking gcc from changing xmm0 in between these blocks :)
</blockquote></details>

Done.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 1 of 2 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 50 at r4](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIUqyRH5X-k9ep-qgmW:-MIUqyRI0oHIhSo6qPiA:bc4icfe) ([raw file](https://github.com/oscarlab/graphene/blob/894976c290ad7cb0363c1404c140713ff9091b3a/LibOS/shim/test/regression/sighandler_divbyzero.c#L50)):*
> ```C
> 
>     for (int i = 0; i < 8; i++) {
>         if (out[i] != i) {
> ```

I'd prefer `out[i] != in[i]`

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

yamahata (approved):
> Reviewed 1 of 2 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: 1 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski, @mkow, and @yamahata)

___
*[LibOS/shim/test/regression/sighandler_divbyzero.c, line 50 at r4](https://reviewable.io/reviews/oscarlab/graphene/1852#-MIUqyRH5X-k9ep-qgmW:-MIVgrLHAxwfDkkK2uvp:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/894976c290ad7cb0363c1404c140713ff9091b3a/LibOS/shim/test/regression/sighandler_divbyzero.c#L50)):*

I'd prefer `out[i] != in[i]`
</blockquote></details>

Done.

mkow (approved):
> Reviewed 5 of 5 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

yamahata (approved):
> Reviewed 5 of 5 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1852)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #1814, opened 2020-09-13
Rewrite GDB integration


mkow:
## Description of the problem

Current GDB integration has a lot of bugs, especially the Linux-SGX part.

## Known bugs/issues

- [x] Starting the debuggee causes a spam of logs about SIGILL, which triggers GDB pagination and requires manual scrolling through them. Fixed by #1828.
- [x] On Linux PAL there's an annoying prompt on each GDB start generated by one of our scripts. Fixed by #1828.
- [x] Attaching to an already running process on Linux-SGX misses symbols for all modules loaded so far. Fixed by #1853.
- [x] Breakpoints in Linux-SGX doesn't work until one breaks (e.g. using Ctrl+C) the execution while inside enclave. Fixed by #1853.
- [x] ~#794 (unloading of modules isn't communicated back to GDB, only loading)~ (I think I misunderstood this one, seems to be a mem leak in LibOS `link_map` handling, not directly related to GDB integration)
- [x] ~GDB crashes from time to time with our SGX integration, but it's hard to say if it's our bug or GDB's.~ Haven't reproduced recently, removing this point for now.
- [x] Stack walking is not always correct, probably some of our .cfi directives are slightly off (I think they are wrong in `__morestack` function, but I'm not sure). Linux PAL fixed by #1864. Linux-SGX fixed by #1868.
- [x] Extended register support - looking at sgx_gdb.c it seems that extended registers are not supported at all, so most likely it's not possible to debug enclave code which uses e.g. floating point registers. Fixed by #1874.
- [x] Symbols seem to have incorrect filenames assigned, which makes it hard to use ones which are defined multiple time, like `main()`, e.g.: (see the addresses)
  ```
  (gdb) p 'sgx_main.c'::main
  $3 = {int (int, char **, char **)} 0x556ac9dd36a0 <main>
  (gdb) p 'bootstrap.c'::main
  $4 = {int (int, char **, char **)} 0x556ac9dd36a0 <main>
  ```
  Seems to be a bug in older GDBs. Works correctly on GDB 9.2 (see @pwmarcz comment below).
- [ ] Remove glibc modification of `register_library()` (https://github.com/oscarlab/graphene/blob/v1.1/LibOS/glibc-patches/glibc-2.31.patch#L104)
[...]

PR #1853, opened 2020-09-29
[Pal/Linux-SGX] Rewrite GDB symbol file updates

pwmarcz:
See https://github.com/oscarlab/graphene/issues/1814.

The old integration was based on directly executing GDB commands.
Instead, we maintain a linked list of debug sections, and notify
GDB using a breakpointed function.

In addition, make sure to read the structure when we attach to a
new process, by hooking the GDB stop event.

From my testing, this also solves the problem with breakpoints not
working until you break (Ctrl-C) inside an enclave. Adding a
breakpoint to a specific file (`break foo.c:main`) also seems to
work correctly, even before the process is started (the breakpoint
is conditional on a file load).

## How to test this PR?

Starting a new process:
* Compile Graphene with `SGX=1 DEBUG=1`
* Compile LibOS/test/regression with `SGX=1 DEBUG=1`
* Start GDB: `GDB=1 SGX=1 ./pal_loader bootstrap`
* Add a breakpoint: `break bootstrap.c:main`, confirm that it's conditional on library load
* Run: `run`
* Verify that the debug information is there: `list`, `info files`
* Continue: `continue`, check if it exits cleanly

Attaching to an existing process:
* Compile as above
* Run a program that pauses (I'm adding a `getchar()` to `bootstrap.c`)
* Check the PID with `pgrep pal-Linux-SGX`
* Start GDB: `GDB=1 SGX=1 ./pal_loader bootstrap`
* Attach to existing process: `attach <pid>`
* Verify that the debug information is there: `list bootstrap.c:main`, `info files`

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_rtld.c
+++ b/Pal/src/host/Linux-SGX/db_rtld.c
@@ -10,6 +10,8 @@
  * Library.
  */
 
+#include <assert.h>
+
 #include "api.h"
 #include "elf-x86_64.h"
 #include "elf/elf.h"
@@ -22,7 +24,108 @@
 #include "pal_linux_defs.h"
 #include "pal_rtld.h"
 #include "pal_security.h"
+#include "sgx_rtld.h"
+#include "spinlock.h"
 #include "sysdeps/generic/ldsodefs.h"
+
+/* Global debug map. To simplify setup, the pointer to g_debug_map is passed outside with
+ * ocall_update_debugger().
+ * (Note that we pass a pointer to g_debug_map, not its current value, to avoid race conditions). */
+static struct debug_map* _Atomic g_debug_map = NULL;
+
+/* Lock for modifying g_debug_map on our end. Even though the list can be read at any
+ * time, we need to prevent concurrent modification. */
+static spinlock_t g_debug_map_lock = INIT_SPINLOCK_UNLOCKED;
+
+static struct debug_map* debug_map_alloc(const char* file_name, void* load_addr) {
+    struct debug_map* map;
+
+    if (!(map = malloc(sizeof(*map))))
+        return NULL;
+
+    if (!(map->file_name = strdup(file_name))) {
+        free(map);
+        return NULL;
+    }
+
+    map->load_addr = load_addr;
+    map->section = NULL;
+    map->next = NULL;
+    return map;
+}
+
+static struct debug_section* debug_map_add_section(struct debug_map* map, const char* section_name,
+                                                   void* addr) {
+    struct debug_section* section;
+
+    if (!(section = malloc(sizeof(*section))))
+        return NULL;
+
+    if (!(section->name = strdup(section_name))) {
+        free(section);
+        return NULL;
+    }
+
+    section->addr = addr;
+    section->next = map->section;
+    map->section = section;
+    return section;
+}
+
+static void debug_map_free(struct debug_map* map) {
+    struct debug_section* section = map->section;
+    while (section) {
+        struct debug_section* next = section->next;
+        free(section->name);
+        free(section);
+        section = next;
+    }
+    free(map->file_name);
+    free(map);
+}
+
+static void debug_map_add(struct debug_map* map) {
+    spinlock_lock(&g_debug_map_lock);
+
+    map->next = g_debug_map;
+    g_debug_map = map;
+
+    spinlock_unlock(&g_debug_map_lock);
+
+    ocall_update_debugger(&g_debug_map);
+}
+
+static bool debug_map_del(void* load_addr) {
+    assert(g_debug_map);
+
+    spinlock_lock(&g_debug_map_lock);
+
+    struct debug_map* prev = NULL;
+    struct debug_map* map = g_debug_map;
+    while (map) {
+        if (map->load_addr == load_addr)
+            break;
+        prev = map;
+        map = map->next;
+    }
+
+    if (!map) {
+        spinlock_unlock(&g_debug_map_lock);
+        return false;
+    }
+
+    if (prev == NULL)
+        g_debug_map = map->next;
+    else
+        prev->next = map->next;
+
+    spinlock_unlock(&g_debug_map_lock);
+
+    debug_map_free(map);
+
+    ocall_update_debugger(&g_debug_map);
+    return true;
+}
 
 void _DkDebugAddMap(struct link_map* map) {
     const ElfW(Ehdr)* ehdr = (void*)map->l_map_start;
@@ -76,47 +179,33 @@ void _DkDebugAddMap(struct link_map* map) {
 
     ocall_close(fd);
 
-    ElfW(Addr) text_addr = 0;
-    for (ElfW(Shdr)* s = shdr; s < shdrend; s++)
-        if (!strcmp_static(shstrtab + s->sh_name, ".text")) {
-            text_addr = map->l_addr + s->sh_addr;
-            break;
-        }
-
-    if (!text_addr)
+    struct debug_map* debug_map = debug_map_alloc(map->l_name, (void*)map->l_addr);
+    if (!debug_map) {
+        SGX_DBG(DBG_E, "_DkDebugAddMap: error allocating new map\n");
         return;
-
-#define BUFFER_LENGTH 4096
-
-    char buffer[BUFFER_LENGTH];
-    char* ptr = buffer;
-
-    snprintf(ptr, BUFFER_LENGTH - (ptr - buffer), "add-symbol-file %s 0x%016lx -readnow",
-             map->l_name, text_addr);
-    ptr += strlen(ptr);
+    }
 
     for (ElfW(Shdr)* s = shdr; s < shdrend; s++) {
         if (!s->sh_name || !s->sh_addr)
-            continue;
-        if (!strcmp_static(shstrtab + s->sh_name, ".text"))
             continue;
         if (s->sh_type == SHT_NULL)
             continue;
         if (strstartswith_static(shstrtab + s->sh_name, ".debug_"))
             continue;
 
-        snprintf(ptr, BUFFER_LENGTH - (ptr - buffer), " -s %s 0x%016lx", shstrtab + s->sh_name,
-                 map->l_addr + s->sh_addr);
-        ptr += strlen(ptr);
+        if (!debug_map_add_section(debug_map, shstrtab + s->sh_name,
+                                   (void*)(map->l_addr + s->sh_addr))) {
+            SGX_DBG(DBG_E, "_DkDebugAddMap: error allocating new section\n");
+            debug_map_free(debug_map);
+            return;
+        }
     }
 
-    ocall_load_debug(buffer);
+    debug_map_add(debug_map);
 }
 
 void _DkDebugDelMap(struct link_map* map) {
-    char buffer[BUFFER_LENGTH];
-    snprintf(buffer, BUFFER_LENGTH, "remove-symbol-file %s", map->l_name);
-    ocall_load_debug(buffer);
+    debug_map_del((void*)map->l_addr);
 }
 
 extern void* g_section_text;
@@ -135,14 +224,31 @@ void setup_pal_map(struct link_map* pal_map) {
     pal_map->l_phnum   = header->e_phnum;
     setup_elf_hash(pal_map);
 
-    char buffer[BUFFER_LENGTH];
-    snprintf(buffer, BUFFER_LENGTH,
-             "add-symbol-file %s %p -readnow -s .rodata %p "
-             "-s .dynamic %p -s .data %p -s .bss %p",
-             pal_map->l_name, &g_section_text, &g_section_rodata, &g_section_dynamic,
-             &g_section_data, &g_section_bss);
-
-    ocall_load_debug(buffer);
     pal_map->l_prev = pal_map->l_next = NULL;
     g_loaded_maps = pal_map;
+
+    struct debug_map* debug_map = debug_map_alloc(pal_map->l_name, &g_section_text);
+    if (!debug_map) {
+        SGX_DBG(DBG_E, "setup_pal_map: error allocating new map\n");
+        return;
+    }
+
+    if (!debug_map_add_section(debug_map, ".rodata", &g_section_rodata))
+        goto fail;
+
+    if (!debug_map_add_section(debug_map, ".dynamic", &g_section_dynamic))
+        goto fail;
+
+    if (!debug_map_add_section(debug_map, ".data", &g_section_data))
+        goto fail;
+
+    if (!debug_map_add_section(debug_map, ".bss", &g_section_bss))
+        goto fail;
+
+    debug_map_add(debug_map);
+    return;
+
+fail:
+    SGX_DBG(DBG_E, "setup_pal_map: error allocating new section\n");
+    debug_map_free(debug_map);
 }
--- a/Pal/src/host/Linux-SGX/enclave_ocalls.c
+++ b/Pal/src/host/Linux-SGX/enclave_ocalls.c
@@ -1354,18 +1354,19 @@ int ocall_delete(const char* pathname) {
     return retval;
 }
 
-int ocall_load_debug(const char* command) {
+int ocall_update_debugger(struct debug_map* _Atomic* debug_map) {
     int retval = 0;
-    int len = strlen(command) + 1;
+    ms_ocall_update_debugger_t* ms;
 
     void* old_ustack = sgx_prepare_ustack();
-    const char* ms = sgx_copy_to_ustack(command, len);
+    ms = sgx_alloc_on_ustack_aligned(sizeof(*ms), alignof(*ms));
     if (!ms) {
         sgx_reset_ustack(old_ustack);
         return -EPERM;
     }
 
-    retval = sgx_exitless_ocall(OCALL_LOAD_DEBUG, (void*)ms);
+    WRITE_ONCE(ms->ms_debug_map, debug_map);
+    retval = sgx_exitless_ocall(OCALL_UPDATE_DEBUGGER, ms);
 
     sgx_reset_ustack(old_ustack);
     return retval;
--- a/Pal/src/host/Linux-SGX/enclave_ocalls.h
+++ b/Pal/src/host/Linux-SGX/enclave_ocalls.h
@@ -10,6 +10,7 @@
 #include "linux_types.h"
 #include "pal_linux.h"
 #include "sgx_attest.h"
+#include "sgx_rtld.h"
 
 noreturn void ocall_exit(int exitcode, int is_exitgroup);
 
@@ -87,7 +88,7 @@ int ocall_rename(const char* oldpath, const char* newpath);
 
 int ocall_delete(const char* pathname);
 
-int ocall_load_debug(const char* command);
+int ocall_update_debugger(struct debug_map* _Atomic* debug_map);
 
 int ocall_eventfd(unsigned int initval, int flags);
 
--- a/Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx.gdb
+++ b/Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx.gdb
@@ -35,7 +35,6 @@ set follow-exec-mode same
 set follow-fork-mode child
 set displaced-stepping off
 
-
 # CPUID/RDTSC SIGILL skipping. See [2] above.
 
 catch signal SIGILL
--- a/Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py
+++ b/Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later */
 # Copyright (C) 2020 Intel Corporation
 #                    Michał Kowalczyk <mkow@invisiblethingslab.com>
+#                    Paweł Marczewski <pawel@invisiblethingslab.com>
 
 import os
 
@@ -9,6 +10,94 @@ import gdb # pylint: disable=import-error
 # pylint: disable=no-self-use,too-few-public-methods
 
 _g_paginations = []
+
+
+def retrieve_debug_maps():
+    '''
+    Retrieve the debug_map structure from the inferior process. The result is a dict with the
+    following structure:
+
+    {load_addr: (file_name, {name: addr})}
+    '''
+
+    if int(gdb.parse_and_eval('g_pal_enclave.debug_map')) == 0:
+        # Not initialized yet
+        return {}
+
+    debug_maps = {}
+    val_map = gdb.parse_and_eval('*g_pal_enclave.debug_map')
+    while int(val_map) != 0:
+        file_name = val_map['file_name'].string()
+        file_name = os.path.abspath(file_name)
+
+        load_addr = int(val_map['load_addr'])
+
+        sections = {}
+        val_section = val_map['section']
+        while int(val_section) != 0:
+            name = val_section['name'].string()
+            addr = int(val_section['addr'])
+
+            sections[name] = addr
+            val_section = val_section['next']
+
+        # We need the text_addr to use add-symbol-file (at least until GDB 8.2).
+        if '.text' in sections:
+            debug_maps[load_addr] = (file_name, sections)
+
+        val_map = val_map['next']
+
+    return debug_maps
+
+
+class UpdateDebugMaps(gdb.Command):
+    """Update debug maps for the inferior process."""
+
+    def __init__(self):
+        super().__init__('update-debug-maps', gdb.COMMAND_USER)
+
+    def invoke(self, arg, _from_tty):
+        self.dont_repeat()
+        assert arg == ''
+
+        # Store the currently loaded maps inside the Progspace object, so that we can compare
+        # old and new states. See:
+        # https://sourceware.org/gdb/current/onlinedocs/gdb/Progspaces-In-Python.html
+        progspace = gdb.current_progspace()
+        if not hasattr(progspace, 'sgx_debug_maps'):
+            progspace.sgx_debug_maps = {}
+
+        old = progspace.sgx_debug_maps
+        new = retrieve_debug_maps()
+        for load_addr in set(old) | set(new):
+            # Skip unload/reload if the map is unchanged
+            if old.get(load_addr) == new.get(load_addr):
+                continue
+
+            # Note that this doesn't escape the file names.
+
+            if load_addr in old:
+                # Log the removing, because remove-symbol-file itself doesn't produce helpful output
+                # on errors.
+                file_name, sections = old[load_addr]
+                print("Removing symbol file (was {}) from addr: 0x{:x}".format(
+                    file_name, load_addr))
+                try:
+                    gdb.execute('remove-symbol-file -a 0x{:x}'.format(load_addr))
+                except gdb.error:
+                    print('warning: failed to remove symbol file')
+
+            if load_addr in new:
+                file_name, sections = new[load_addr]
+                text_addr = sections['.text']
+                cmd = 'add-symbol-file {} 0x{:x} '.format(file_name, text_addr)
+                cmd += ' '.join('-s {} 0x{:x}'.format(name, addr)
+                                for name, addr in sections.items()
+                                if name != '.text')
+                gdb.execute(cmd)
+
+        progspace.sgx_debug_maps = new
+
 
 class PushPagination(gdb.Command):
     """Temporarily changing pagination and saving the old state.
@@ -46,18 +135,34 @@ class PopPagination(gdb.Command):
         gdb.execute('set pagination ' + ('on' if pagination else 'off'))
 
 
-class LoadCommandBreakpoint(gdb.Breakpoint):
+class UpdateBreakpoint(gdb.Breakpoint):
     def __init__(self):
-        gdb.Breakpoint.__init__(self, spec="execute_gdb_command", internal=1)
+        gdb.Breakpoint.__init__(self, spec="update_debugger", internal=1)
 
     def stop(self):
-        command = gdb.parse_and_eval("(const char*)$rdi").string()
-        gdb.execute(command)
+        gdb.execute('update-debug-maps')
         return False
+
+
+def stop_handler(_event):
+    # Make sure we handle connecting to a new process correctly:
+    # update the debug maps if we never did it before.
+    progspace = gdb.current_progspace()
+    if not hasattr(progspace, 'sgx_debug_maps'):
+        gdb.execute('update-debug-maps')
+
+
+def clear_objfiles_handler(event):
+    # Record that symbol files has been cleared on GDB's side (e.g. on program exit), so that we do
+    # not try to remove them again.
+    if hasattr(event.progspace, 'sgx_debug_maps'):
+        delattr(event.progspace, 'sgx_debug_maps')
+
 
 def main():
     PushPagination()
     PopPagination()
+    UpdateDebugMaps()
 
     # Some of the things we want to do can't be done using gdb Python API, we need to fall back to a
     # standard gdb script.
@@ -65,7 +170,9 @@ def main():
     print("[%s] Loading %s..." % (os.path.basename(__file__), gdb_script))
     gdb.execute("source " + gdb_script)
 
-    LoadCommandBreakpoint()
+    UpdateBreakpoint()
+    gdb.events.stop.connect(stop_handler)
+    gdb.events.clear_objfiles.connect(clear_objfiles_handler)
 
 if __name__ == "__main__":
     main()
--- a/Pal/src/host/Linux-SGX/ocall_types.h
+++ b/Pal/src/host/Linux-SGX/ocall_types.h
@@ -10,6 +10,7 @@
 #include "pal.h"
 #include "sgx_arch.h"
 #include "sgx_attest.h"
+#include "sgx_rtld.h"
 
 /*
  * GCC's structure padding may cause leaking from uninialized
@@ -57,7 +58,7 @@ enum {
     OCALL_POLL,
     OCALL_RENAME,
     OCALL_DELETE,
-    OCALL_LOAD_DEBUG,
+    OCALL_UPDATE_DEBUGGER,
     OCALL_EVENTFD,
     OCALL_GET_QUOTE,
     OCALL_NR,
@@ -265,6 +266,10 @@ typedef struct {
 typedef struct {
     const char* ms_pathname;
 } ms_ocall_delete_t;
+
+typedef struct {
+    struct debug_map* _Atomic* ms_debug_map;
+} ms_ocall_update_debugger_t;
 
 typedef struct {
     unsigned int ms_initval;
--- a/Pal/src/host/Linux-SGX/sgx_enclave.c
+++ b/Pal/src/host/Linux-SGX/sgx_enclave.c
@@ -629,10 +629,16 @@ static long sgx_ocall_eventfd(void* pms) {
     return ret;
 }
 
-static long sgx_ocall_load_debug(void* pms) {
-    const char* command = (const char*)pms;
-    ODEBUG(OCALL_LOAD_DEBUG, (void*)command);
-    execute_gdb_command(command);
+static long sgx_ocall_update_debugger(void* pms) {
+    ms_ocall_update_debugger_t* ms = (ms_ocall_update_debugger_t*)pms;
+    ODEBUG(OCALL_UPDATE_DEBUGGER, ms);
+
+#ifdef DEBUG
+    g_pal_enclave.debug_map = ms->ms_debug_map;
+    update_debugger();
+#else
+    __UNUSED(ms);
+#endif
     return 0;
 }
 
@@ -679,7 +685,7 @@ sgx_ocall_fn_t ocall_table[OCALL_NR] = {
     [OCALL_POLL]             = sgx_ocall_poll,
     [OCALL_RENAME]           = sgx_ocall_rename,
     [OCALL_DELETE]           = sgx_ocall_delete,
-    [OCALL_LOAD_DEBUG]       = sgx_ocall_load_debug,
+    [OCALL_UPDATE_DEBUGGER]  = sgx_ocall_update_debugger,
     [OCALL_EVENTFD]          = sgx_ocall_eventfd,
     [OCALL_GET_QUOTE]        = sgx_ocall_get_quote,
 };
--- a/Pal/src/host/Linux-SGX/sgx_gdb_info.c
+++ b/Pal/src/host/Linux-SGX/sgx_gdb_info.c
@@ -7,8 +7,7 @@
 #include "sgx_internal.h"
 
 /* This function is hooked by our gdb integration script and should be left as is. */
-__attribute__((__noinline__)) void execute_gdb_command(const char* command) {
-    __UNUSED(command);
+__attribute__((__noinline__)) void update_debugger(void) {
     __asm__ volatile(""); // Required in addition to __noinline__ to prevent deleting this function.
                           // See GCC docs.
 }
--- a/Pal/src/host/Linux-SGX/sgx_internal.h
+++ b/Pal/src/host/Linux-SGX/sgx_internal.h
@@ -16,6 +16,7 @@
 #include "api.h"
 #include "pal_linux.h"
 #include "pal_security.h"
+#include "sgx_rtld.h"
 #include "sysdep-arch.h"
 
 #define IS_ERR   INTERNAL_SYSCALL_ERROR
@@ -68,6 +69,12 @@ extern struct pal_enclave {
 
     /* Path to the PAL binary */
     char* libpal_uri;
+
+#ifdef DEBUG
+    /* Pointer to information for GDB inside the enclave (see sgx_rtld.h).
+     * Set up using update_debugger() ocall. */
+    struct debug_map* _Atomic* debug_map;
+#endif
 
     /* security information */
     struct pal_sec pal_sec;
@@ -138,6 +145,6 @@ int sgx_signal_setup(void);
 int block_signals(bool block, const int* sigs, int nsig);
 int block_async_signals(bool block);
 
-void execute_gdb_command(const char* command);
+void update_debugger(void);
 
 #endif
--- a/Pal/src/host/Linux-SGX/sgx_main.c
+++ b/Pal/src/host/Linux-SGX/sgx_main.c
@@ -783,6 +783,8 @@ static int load_enclave(struct pal_enclave* enclave, int manifest_fd, char* mani
 
         env_i += strnlen(&env[env_i], env_size - env_i) + 1;
     }
+
+    enclave->debug_map = NULL;
 #endif
 
     enclave->manifest = manifest_fd;
--- a/Pal/src/host/Linux-SGX/sgx_rtld.h
+++ b/Pal/src/host/Linux-SGX/sgx_rtld.h
@@ -1,0 +1,48 @@
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
+/* Copyright (C) 2020 Intel Corporation
+ *                    Paweł Marczewski <pawel@invisiblethingslab.com>
+ */
+
+/*
+ * Internal debug maps, used for SGX to communicate with debugger. We maintain it so that it is in a
+ * consistent state any time the process is stopped (any add/delete is an atomic modification of one
+ * pointer).
+ *
+ * The debug map is maintained inside the enclave, and the debugger is notified using
+ * ocall_update_debugger().
+ */
+
+#ifndef SGX_RTLD_H
+#define SGX_RTLD_H
+
+/*
+ * TODO: (GDB 8.2)
+ *
+ * To add the files in GDB, we use the 'add-symbol-file' command. In the GDB versions we support,
+ * that command requires specifying a text section address (and, apparently, all the other
+ * sections). In GDB 8.2, the 'text_addr' parameter is optional, and there is a new '-o offset'
+ * option which allows to just specify load address for the whole file, and that's enough to load
+ * all the sections.
+ *
+ * Once we are able to rely on newer GDB, we can get rid the section list (struct debug_section).
+ * It's also possible that we will be able to use the r_debug structure instead, so that the same
+ * mechanism is
+ * used in Linux and Linux-SGX (even though in Linux-SGX we parse the structure manually in Python).
+ */
+
+struct debug_section {
+    char* name;
+    void* addr;
+
+    struct debug_section* next;
+};
+
+struct debug_map {
+    char* file_name;
+    void* load_addr;
+    struct debug_section* section;
+
+    struct debug_map* _Atomic next;
+};
+
+#endif /* SGX_RTLD_H */


yamahata (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: 0 of 13 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz)

___
*[Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py, line 65 at r1](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIPZgCa2MYdG1MpS3Yc:-MIPZgCbDDgjcYzHJJwi:bcpxk48) ([raw file](https://github.com/oscarlab/graphene/blob/a5616cbe346f0cdab7c31e92d60eb2109dc6f1c7/Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py#L65)):*
<details><summary><i alt="summary">Quoted 4 lines of code…</i></summary>

> ```
>         for file_name in set(old) | set(new):
>             # Skip unload/reload if the map is unchanged
>             if old.get(file_name) == new.get(file_name):
>                 continue
> ```
</details>


set(old) ^ set(new)

yamahata (dismissed):
> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @pwmarcz)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @yamahata)

___
*[Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py, line 65 at r1](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIPZgCa2MYdG1MpS3Yc:-MIPhhHy3yMBDwASJt9U:btg8aq0) ([raw file](https://github.com/oscarlab/graphene/blob/a5616cbe346f0cdab7c31e92d60eb2109dc6f1c7/Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py#L65)):*

> ```
>         for file_name in set(old) | set(new):
>             # Skip unload/reload if the map is unchanged
>             if old.get(file_name) == new.get(file_name):
>                 continue
> ```

set(old) ^ set(new)
</blockquote></details>

I do want to check the files that are in both sets, in case the file is the same, but there has been a change in the list of sections. That's why I compare the values in next line.

(I think it's an unlikely case, but might be possible if we skip some updates, e.g. detach then attach to a process after some time).

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: 12 of 13 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @yamahata)

___
*[Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py, line 65 at r1](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIPZgCa2MYdG1MpS3Yc:-MIQIcPx76X1jcc_LTf0:b2ni9bh) ([raw file](https://github.com/oscarlab/graphene/blob/a5616cbe346f0cdab7c31e92d60eb2109dc6f1c7/Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py#L65)):*

I do want to check the files that are in both sets, in case the file is the same, but there has been a change in the list of sections. That's why I compare the values in next line.

(I think it's an unlikely case, but might be possible if we skip some updates, e.g. detach then attach to a process after some time).
</blockquote></details>

ok.

dimakuv (changes requested):
> Reviewed 12 of 13 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIT_ptsA5RtlQocjdJD:-MIT_ptsA5RtlQocjdJE:b-hgxs65) (no related file):*
This is awesome work! Thank you

---
___
*[Pal/src/pal_rtld.h, line 176 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIT_gJR4Sslp-WzD5J7:-MIT_gJR4Sslp-WzD5J8:b-ixw0n) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/pal_rtld.h#L176)):*
> ```objc
>     struct debug_section* section;
> 
>     struct debug_map* _Atomic next;
> ```

Why Atomic here? Looks like all `debug_map->next` operations are under a global spinlock always.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 252 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITUL83EZ_ZtYpNAGTq:-MITUL83EZ_ZtYpNAGTr:bexw8wt) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_main.c#L252)):*
> ```C
> #ifdef DEBUG
>     g_pal_sec.in_gdb = sec_info.in_gdb;
>     g_pal_sec.debug_map = sec_info.debug_map;
> ```

This looks like a useless assignment? From inside the enclave, we seem to only operate on `g_debug_map` which is propagated by Graphene's untrusted bootstrapping code through `sec_info.debug_map` pointer. And this `g_pal_sec.debug_map` seems never used.

Or is it simply for sanity, to have the same values in both untrusted `sec_info` and trusted `g_pal_sec`?
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 33 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITV-YA0xuapL16Qj0F:-MITV-YA0xuapL16Qj0G:bnicozn) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L33)):*
> ```C
> /* Lock for modifying the debug_map linked list on our end. Even though the list can be read at any
>  * time, we need to prevent concurrent modification. */
> static spinlock_t g_debug_map_lock = INIT_SPINLOCK_UNLOCKED;
> ```

Please add an explicit `#include "spinlock.h"`
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 36 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITVCnL0U3IKpnhFP9G:-MITVCnL0U3IKpnhFP9H:b-t3weik) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L36)):*
> ```C
> 
> static struct debug_map* debug_map_alloc(const char* file_name, void* text_addr) {
>     struct debug_map* map = malloc(sizeof(struct debug_map));
> ```

We prefer `sizeof(*map)` but not blocking on this
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 53 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITVIEcEaEEchoYYLoo:-MITVIEcEaEEchoYYLop:b-hkars5) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L53)):*
> ```C
> static struct debug_section* debug_map_add_section(struct debug_map* map, const char* section_name,
>                                                    void* addr) {
>     struct debug_section* section = malloc(sizeof(struct debug_section));
> ```

We prefer `sizeof(*section)` but not blocking on this
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 90 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITWUoW5tH6lDfDMfFV:-MITWUoW5tH6lDfDMfFW:bhbhozv) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L90)):*
> ```C
> }
> 
> static bool debug_map_del(const char* file_name) {
> ```

Is it possible that there are two `map`s with the same file name? In that case, we would delete only one of them. I see no code that checks for duplicates in `debug_map_add()` or its caller, but maybe it just never happens in real life?
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 105 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITVv3bBoQK1V31Nk9N:-MITVv3bBoQK1V31Nk9O:b-mz3ggf) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L105)):*
> ```C
> 
>     if (!map)
>         return false;
> ```

Forgot to unlock
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 241 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITXZYk9YGKBijsWT76:-MITXZYl4Krwc1GLjeYL:bo1z9gq) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L241)):*
> ```C
>     debug_map_add_section(debug_map, ".dynamic", &g_section_dynamic);
>     debug_map_add_section(debug_map, ".data", &g_section_data);
>     debug_map_add_section(debug_map, ".bss", &g_section_bss);
> ```

Each of these four can fail. Please add error handling.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 822 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITXptr8NI4xYTfuY2o:-MITXptr8NI4xYTfuY2p:bjn2q6e) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/sgx_main.c#L822)):*
> ```C
> 
>     enclave->debug_map = NULL;
>     pal_sec->debug_map = &enclave->debug_map;
> ```

BTW, I hate these historic names... I still don't understand why we have a separate `pal_sec` and what exactly this means.

mkow (changes requested):
> Reviewed 12 of 13 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIT_ptsA5RtlQocjdJD:-MITh8TpAhqlJQA3pe4i:b-z7iz35) (no related file):*

This is awesome work! Thank you
</blockquote></details>

+1, this was really needed :)

---
___
*[Pal/src/pal_rtld.h, line 162 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITUuwADaEJwABFLq5H:-MITUuwB0hdf0o8Cq1lo:b-eobm5b) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/pal_rtld.h#L162)):*
> ```objc
>  * Internal debug maps, used for SGX to communicate with debugger. We maintain it so that it is in a
>  * consistent state any time the process is stopped (any add/delete is an atomic modification of one
>  * pointer).
> ```

If this is SGX-specific, shouldn't it be in host/Linux-SGX?
___
*[Pal/src/pal_rtld.h, line 173 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITWMf80vT2xPIl7mKk:-MITWMf80vT2xPIl7mKl:biuoriy) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/pal_rtld.h#L173)):*
> ```objc
> struct debug_map {
>     char* file_name;
>     void* text_addr;
> ```

`text` section doesn't necessarily have to go first, I'd call this either `addr`, `img_base`, `loading_addr` or something similar. I see that GDB uses the same name and I'm really confused... What's going on there? :)
___
*[Pal/src/pal_rtld.h, line 176 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIT_gJR4Sslp-WzD5J7:-MITntUpCpD-trTk-8Av:b-2igtlj) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/pal_rtld.h#L176)):*

Why Atomic here? Looks like all `debug_map->next` operations are under a global spinlock always.
</blockquote></details>

But GDB can't use the lock. But if GDB is only reading, then this whole construction is correct and safe.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 69 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITabO3-c_paPs6df92:-MITabO41XR_eByd-8H9:bbe9i5) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L69)):*
> ```C
> 
> static void debug_map_free(struct debug_map* map) {
>     for (struct debug_section* section = map->section; section; section = section->next) {
> ```

This is use-after-free: You read `section->next`, but `section` was already freed.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 90 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITWUoW5tH6lDfDMfFV:-MITblfF2xmL5XSGMoVu:b-6layph) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L90)):*

Is it possible that there are two `map`s with the same file name? In that case, we would delete only one of them. I see no code that checks for duplicates in `debug_map_add()` or its caller, but maybe it just never happens in real life?
</blockquote></details>

I think this may happen. I'd actually prefer to always use the loading address as a lookup key, not the name (it's unique + comparing integers is faster).
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 96 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITdKz38GWw4FZtq1xZ:-MITdKz38GWw4FZtq1x_:b-tieqsm) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L96)):*
> ```C
> 
>     struct debug_map* prev = NULL;
>     struct debug_map* map = *g_debug_map;
> ```

The untrusted world can change the pointer and manipulate us to into corrupting enclave memory. But I think this is not really a problem, as we use this only in debug builds (in this case a comment here would be good - that this is insecure, but enabling GDB integration is inherently insecure and not for production use and is enabled only for debug builds).

Btw. We could add an additional switch `INSECURE__ENABLE_GDB` to force release builds to also emit the integration, sometimes it's useful. But that's something for another PR.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 230 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITgjefEsutxrj-wL4p:-MITgjefEsutxrj-wL4q:b-pj8z1m) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L230)):*
> ```C
> 
>     if (!g_debug_map)
>         return;
> ```

Doesn't this indicate a bug in our code? If so, this should rather be an assert.
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1359 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITi2UU8hcqyFAGLNkw:-MITi2UU8hcqyFAGLNkx:bgucw5g) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1359)):*
> ```C
> int ocall_update_debugger(void) {
>     void *dummy = NULL;
>     return sgx_exitless_ocall(OCALL_UPDATE_DEBUGGER, dummy);
> ```

Why not just passing NULL directly? I guess this was for readability?
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 822 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITXptr8NI4xYTfuY2o:-MITimdw5pYikVwiKWRY:b-ajndd2) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/sgx_main.c#L822)):*

BTW, I hate these historic names... I still don't understand why we have a separate `pal_sec` and what exactly this means.
</blockquote></details>

Neither I know the meaning. We have to refactor this finally.
___
*[Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py, line 3 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITjW6Q3CTtgp2pH0F_:-MITjW6Q3CTtgp2pH0Fa:b-19rp64) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py#L3)):*
> ```Python
> # SPDX-License-Identifier: LGPL-3.0-or-later */
> # Copyright (C) 2020 Intel Corporation
> #                    Michał Kowalczyk <mkow@invisiblethingslab.com>
> ```

please insert yourself here :)
___
*[Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py, line 72 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITneTR-8wK-s9kbjrx:-MITneTT9E3Amal8twpF:blthy3l) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py#L72)):*
> ```Python
> 
>             if file_name in old:
>                 # Remove the file by address, not by name, because:
> ```

Why not using the address as a key in the map then?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: 8 of 14 files reviewed, 14 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, @pwmarcz, and @yamahata)

___
*[Pal/src/pal_rtld.h, line 162 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITUuwADaEJwABFLq5H:-MIU9dhT-ipDWY5yK1O2:b-l0cydn) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/pal_rtld.h#L162)):*

If this is SGX-specific, shouldn't it be in host/Linux-SGX?
</blockquote></details>

Makes sense. Moved to its own header (`sgx_rtld.h`).
___
*[Pal/src/pal_rtld.h, line 176 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIT_gJR4Sslp-WzD5J7:-MIUA0PqBVfsrg7oJY8c:btjr4ou) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/pal_rtld.h#L176)):*

But GDB can't use the lock. But if GDB is only reading, then this whole construction is correct and safe.
</blockquote></details>

Yes, that's why the atomic is necessary.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 252 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITUL83EZ_ZtYpNAGTq:-MIUANkC1GiEuy4w16XP:b5teo4g) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_main.c#L252)):*

This looks like a useless assignment? From inside the enclave, we seem to only operate on `g_debug_map` which is propagated by Graphene's untrusted bootstrapping code through `sec_info.debug_map` pointer. And this `g_pal_sec.debug_map` seems never used.

Or is it simply for sanity, to have the same values in both untrusted `sec_info` and trusted `g_pal_sec`?
</blockquote></details>

For sanity. This is never used, but I would rather copy the whole structure without omitting the fields. 

I thought about just using `g_pal_sec.debug_map`, on one hand it's a bit cleaner, but on the other, it introduces additional dependency (`db_rtld.c` on `g_pal_sec). I'm open to changing it, though.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 33 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITV-YA0xuapL16Qj0F:-MIUBRQC6Df8jepx5gdL:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L33)):*

Please add an explicit `#include "spinlock.h"`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 36 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITVCnL0U3IKpnhFP9G:-MIUBkbN6ZHVtywU9txz:bi5ypcf) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L36)):*

We prefer `sizeof(*map)` but not blocking on this
</blockquote></details>

OK, changed.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 53 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITVIEcEaEEchoYYLoo:-MIUBm_q-cW-FwxLyXIz:b-y3gzqm) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L53)):*

We prefer `sizeof(*section)` but not blocking on this
</blockquote></details>

Same as above, changed.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 69 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITabO3-c_paPs6df92:-MIUC0KE2K7fYaewnQAP:bfoih03) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L69)):*

This is use-after-free: You read `section->next`, but `section` was already freed.
</blockquote></details>

Nice one :) Fixed.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 90 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITWUoW5tH6lDfDMfFV:-MIUD66a2gx_whkIhC-t:bz16i1i) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L90)):*

I think this may happen. I'd actually prefer to always use the loading address as a lookup key, not the name (it's unique + comparing integers is faster).
</blockquote></details>

OK, I can modify the code so that we operate on loading address, and look for `.text` on the Python side since it seems to be a GDB-specific quirk.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 96 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITdKz38GWw4FZtq1xZ:-MIUEf1k9o8IUmi8kKBd:b-f8sf1p) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L96)):*

The untrusted world can change the pointer and manipulate us to into corrupting enclave memory. But I think this is not really a problem, as we use this only in debug builds (in this case a comment here would be good - that this is insecure, but enabling GDB integration is inherently insecure and not for production use and is enabled only for debug builds).

Btw. We could add an additional switch `INSECURE__ENABLE_GDB` to force release builds to also emit the integration, sometimes it's useful. But that's something for another PR.
</blockquote></details>

Added a comment.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 105 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITVv3bBoQK1V31Nk9N:-MIUEku738shSfgQ7m2l:bj3eoea) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L105)):*

Forgot to unlock
</blockquote></details>

Fixed, thanks.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 230 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITgjefEsutxrj-wL4p:-MIUEyFz6tohGKW0tNhy:b-wsg4i3) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L230)):*

Doesn't this indicate a bug in our code? If so, this should rather be an assert.
</blockquote></details>

Actually no, it's being called in `db_main`.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 241 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITXZYk9YGKBijsWT76:-MIUFHkG2FkSkTKqSLjy:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L241)):*

Each of these four can fail. Please add error handling.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1359 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITi2UU8hcqyFAGLNkw:-MIUFn3u7DHPjy5dGb-1:bmqseyp) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1359)):*

Why not just passing NULL directly? I guess this was for readability?
</blockquote></details>

No idea, I followed what the other calls did.
___
*[Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py, line 3 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITjW6Q3CTtgp2pH0F_:-MIUGSpw19ZiCRp0VgiS:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py#L3)):*

please insert yourself here :)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py, line 72 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITneTR-8wK-s9kbjrx:-MIUG2h15u095-pKQExZ:b-4wawb2) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py#L72)):*

Why not using the address as a key in the map then?
</blockquote></details>

Makes sense, I'll do that.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: 8 of 14 files reviewed, 14 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, @pwmarcz, and @yamahata)

___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 96 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITdKz38GWw4FZtq1xZ:-MIU_7o1ESm0CNp80-gD:by032uv) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L96)):*

Added a comment.
</blockquote></details>

Thinking about it some more... I think it would make sense to do it the *other* way around, with the list head inside the enclave:
```
// inside enclave:
struct debug_map* _Atomic g_debug_map;

// during process setup (indirectly):
pal_enclave.debug_map = &g_debug_map;
```

That way, everything is maintained here, and we never need to read the memory outside of the enclave. What do you think? (also paging @mkow)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: 8 of 14 files reviewed, 14 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, @pwmarcz, and @yamahata)

___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 96 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITdKz38GWw4FZtq1xZ:-MIUcS5p5an2jukdtz1I:bacd1at) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L96)):*

Thinking about it some more... I think it would make sense to do it the *other* way around, with the list head inside the enclave:
```
// inside enclave:
struct debug_map* _Atomic g_debug_map;

// during process setup (indirectly):
pal_enclave.debug_map = &g_debug_map;
```

That way, everything is maintained here, and we never need to read the memory outside of the enclave. What do you think? (also paging @mkow)
</blockquote></details>

Or, another idea: instead of accessing the external memory directly, maybe pass the pointer using the `update_debugger` ocall? I.e. do `ocall_update_debugger(debug_map);` which would 1. update the external pointer, 2. notify GDB (if it is attached). That would allow me to remove the whole part with process setup (`pal_security` etc.)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 69 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITabO3-c_paPs6df92:-MIUlnaQBHI340i0Sfsa:b-nc3vrs) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L69)):*

Nice one :) Fixed.
</blockquote></details>

Yeah, welcome to brutal C world :) We really need to enable sanitizers in CI, but in our case this won't be trivial, unfortunately.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 96 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITdKz38GWw4FZtq1xZ:-MIUmHi58nLe640hVB9O:b-afx6h) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L96)):*

Or, another idea: instead of accessing the external memory directly, maybe pass the pointer using the `update_debugger` ocall? I.e. do `ocall_update_debugger(debug_map);` which would 1. update the external pointer, 2. notify GDB (if it is attached). That would allow me to remove the whole part with process setup (`pal_security` etc.)
</blockquote></details>

I like both ideas, but the latter (with update_debugger parameter) seems the best option :)
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 230 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITgjefEsutxrj-wL4p:-MIUmfyB6CFi_OFAdRmJ:beba5z4) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L230)):*

Actually no, it's being called in `db_main`.
</blockquote></details>

I mean, is it really possible that we call "setup_pal_map" while not having `g_debug_map` initialized? If so, this sounds like a bug.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 273 at r3](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIUnLG8AOUIbXihUsST:-MIUnLG8AOUIbXihUsSU:b3s07ht) ([raw file](https://github.com/oscarlab/graphene/blob/57d60da589c4882118abe395295c2fac3ee76c44/Pal/src/host/Linux-SGX/db_rtld.c#L273)):*
> ```C
> 
> fail:
>     SGX_DBG(DBG_E, "setup_pal_map: error allocating new section");
> ```

`SGX_DBG` doesn't append implicit newline at the end
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1359 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITi2UU8hcqyFAGLNkw:-MIUnT0iEL2I7bkTrprI:b-v3cxn0) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1359)):*

No idea, I followed what the other calls did.
</blockquote></details>

Then I'd just pass NULL, I'm not sure why other places do it like this.
___
*[Pal/src/host/Linux-SGX/sgx_rtld.h, line 3 at r3](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIUklVL8jBBkqW2NnXT:-MIUklVM3P870o9J8MmD:bp5c8vr) ([raw file](https://github.com/oscarlab/graphene/blob/57d60da589c4882118abe395295c2fac3ee76c44/Pal/src/host/Linux-SGX/sgx_rtld.h#L3)):*
> ```objc
> /* Copyright (C) 2020 Intel Corporation
>  *                    Paweł Marczewski <pawel@invisiblethingslab.com> */
> ```

Could you move `*/` to the next line, for consistency with other files? It will also reduce diff when inserting more lines here.
___
*[Pal/src/host/Linux-SGX/sgx_rtld.h, line 5 at r3](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIUkxTs8CPv_Ke_ipfl:-MIUkxTs8CPv_Ke_ipfm:b11q04v) ([raw file](https://github.com/oscarlab/graphene/blob/57d60da589c4882118abe395295c2fac3ee76c44/Pal/src/host/Linux-SGX/sgx_rtld.h#L5)):*
> ```objc
>  *                    Paweł Marczewski <pawel@invisiblethingslab.com> */
> 
> /* sgx_rtld.h
> ```

I know that some other parts of the codebase do this, but I don't like this convention with naming the file inside itself. Seems really useless to me + as proven in the recent past, this often gets out of sync with the real file name.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 252 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITUL83EZ_ZtYpNAGTq:-MIUwups6GBfZXPa7CrK:b-1uysjv) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_main.c#L252)):*

For sanity. This is never used, but I would rather copy the whole structure without omitting the fields. 

I thought about just using `g_pal_sec.debug_map`, on one hand it's a bit cleaner, but on the other, it introduces additional dependency (`db_rtld.c` on `g_pal_sec). I'm open to changing it, though.
</blockquote></details>

OK. Resolving, though seems like this will be changed anyway.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 273 at r3](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIUnLG8AOUIbXihUsST:-MIUwcwh3hFLbk5neN76:b-xmhphb) ([raw file](https://github.com/oscarlab/graphene/blob/57d60da589c4882118abe395295c2fac3ee76c44/Pal/src/host/Linux-SGX/db_rtld.c#L273)):*

`SGX_DBG` doesn't append implicit newline at the end
</blockquote></details>

+1, here and everywhere else in your PR
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1359 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITi2UU8hcqyFAGLNkw:-MIUwSLuBwT1Ecj2ILty:b-4orqjj) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1359)):*

Then I'd just pass NULL, I'm not sure why other places do it like this.
</blockquote></details>

+1, I think it should work.

yamahata (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIV5pRpEjtWCLFHTD5b:-MIV5pRqENqAd_7bRhDJ:b2vlkjf) (no related file):*
Just I'm thinking loudly.

With this direction (moving actual work into the python script, pal only notifies necessary info to the python script), glibc modification of register_library() can be removed. 

https://github.com/oscarlab/graphene/blob/master/LibOS/glibc-patches/glibc-2.31.patch#L94

pal can notify main executable and interpreter(ld.so dynamic linker) in enclave to the python script on startup. Then  python script can monitor _r_debug of ld.so in enclave, and register loaded library on behalf of ld.so in enclave to gdb. (Similar change is needed for Pal/Linux.)

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, @pwmarcz, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIV5pRpEjtWCLFHTD5b:-MIXwe-u5TOU481f0oU2:b-ba09m1) (no related file):*

Just I'm thinking loudly.

With this direction (moving actual work into the python script, pal only notifies necessary info to the python script), glibc modification of register_library() can be removed. 

https://github.com/oscarlab/graphene/blob/master/LibOS/glibc-patches/glibc-2.31.patch#L94

pal can notify main executable and interpreter(ld.so dynamic linker) in enclave to the python script on startup. Then  python script can monitor _r_debug of ld.so in enclave, and register loaded library on behalf of ld.so in enclave to gdb. (Similar change is needed for Pal/Linux.)

</blockquote></details>

+1. Would be good to remove this patch from Glibc. Also, I'd remove `glibc_version()` there as well (it is unrelated but a useless feature anyway).

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: 2 of 11 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @yamahata)

___
*[Pal/src/pal_rtld.h, line 173 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITWMf80vT2xPIl7mKk:-MIZRBgH2iIhBV6-cW5h:bwbyk76) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/pal_rtld.h#L173)):*

`text` section doesn't necessarily have to go first, I'd call this either `addr`, `img_base`, `loading_addr` or something similar. I see that GDB uses the same name and I'm really confused... What's going on there? :)
</blockquote></details>

That's changed now. We pass `load_addr`, and look for `.text` section on Python's side (because that's what GDB wants).
___
*[Pal/src/host/Linux-SGX/db_main.c, line 252 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITUL83EZ_ZtYpNAGTq:-MIZQzR07Cn48VnRdJ8d:bd2z3j6) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_main.c#L252)):*

OK. Resolving, though seems like this will be changed anyway.
</blockquote></details>

This turned out to be not necessary after all, the new version doesn't use `pal_security`.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 90 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITWUoW5tH6lDfDMfFV:-MIZRMqpDitTDw_TUFOZ:bgg1l5q) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L90)):*

OK, I can modify the code so that we operate on loading address, and look for `.text` on the Python side since it seems to be a GDB-specific quirk.
</blockquote></details>

Done. The structure contains `load_attr`, and we use it for deleting.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 96 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITdKz38GWw4FZtq1xZ:-MIZRiF48fteej6McnPQ:bk9y9r1) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L96)):*

I like both ideas, but the latter (with update_debugger parameter) seems the best option :)
</blockquote></details>

Implemented the last idea, with `ocall_update_debugger(&debug_map)` call.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 230 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITgjefEsutxrj-wL4p:-MIZRoFlCrOGpCD8buOd:b-qrk7w1) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L230)):*

I mean, is it really possible that we call "setup_pal_map" while not having `g_debug_map` initialized? If so, this sounds like a bug.
</blockquote></details>

Well, `setup_pal_map` runs is called even in the non-DEBUG version...
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 273 at r3](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIUnLG8AOUIbXihUsST:-MIZSFyF17VyR-BVJEcE:boc80cq) ([raw file](https://github.com/oscarlab/graphene/blob/57d60da589c4882118abe395295c2fac3ee76c44/Pal/src/host/Linux-SGX/db_rtld.c#L273)):*

+1, here and everywhere else in your PR
</blockquote></details>

Added newlines.
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1359 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITi2UU8hcqyFAGLNkw:-MIZSI0FERVlCQpU1EJb:b-v2l1uw) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1359)):*

+1, I think it should work.
</blockquote></details>

The call has a parameter again.
___
*[Pal/src/host/Linux-SGX/sgx_rtld.h, line 3 at r3](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIUklVL8jBBkqW2NnXT:-MIZSWCi4YfY-kDkV3hH:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/57d60da589c4882118abe395295c2fac3ee76c44/Pal/src/host/Linux-SGX/sgx_rtld.h#L3)):*

Could you move `*/` to the next line, for consistency with other files? It will also reduce diff when inserting more lines here.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_rtld.h, line 5 at r3](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIUkxTs8CPv_Ke_ipfl:-MIZSX-RE-a0-dycjpHi:b-lupfy4) ([raw file](https://github.com/oscarlab/graphene/blob/57d60da589c4882118abe395295c2fac3ee76c44/Pal/src/host/Linux-SGX/sgx_rtld.h#L5)):*

I know that some other parts of the codebase do this, but I don't like this convention with naming the file inside itself. Seems really useless to me + as proven in the recent past, this often gets out of sync with the real file name.
</blockquote></details>

Removed file name.
___
*[Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py, line 72 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITneTR-8wK-s9kbjrx:-MIZSZ88DSmmEw2QCb9T:b62u9xv) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/gdb_integration/graphene_sgx_gdb.py#L72)):*

Makes sense, I'll do that.
</blockquote></details>

Done. We now use the load address as key.

I also fixed the case of restarting a program: GDB flushes the symbols then, and we need to notice that (by hooking the `clear_objfiles` event).

mkow (changes requested):
> Reviewed 11 of 12 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @pwmarcz, and @yamahata)

___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 230 at r2](https://reviewable.io/reviews/oscarlab/graphene/1853#-MITgjefEsutxrj-wL4p:-MI_-yqt5uTcf6DsDfuF:b-gx3bo5) ([raw file](https://github.com/oscarlab/graphene/blob/53aeba2bab6264e780499aba467aadc9cc2fda40/Pal/src/host/Linux-SGX/db_rtld.c#L230)):*

Well, `setup_pal_map` runs is called even in the non-DEBUG version...
</blockquote></details>

Huh, weird.
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1359 at r4](https://reviewable.io/reviews/oscarlab/graphene/1853#-MI_0rOA3qlAPeziwXpm:-MI_0rOA3qlAPeziwXpn:b4y3qj7) ([raw file](https://github.com/oscarlab/graphene/blob/c0bb70de0852ebc770aca5cf5008681b40b11fa0/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1359)):*
> ```C
> int ocall_update_debugger(struct debug_map* _Atomic* debug_map) {
>     int retval = 0;
>     ms_ocall_update_debugger_t *ms;
> ```

please move `*` to the left

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: 10 of 11 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @yamahata)

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1359 at r4](https://reviewable.io/reviews/oscarlab/graphene/1853#-MI_0rOA3qlAPeziwXpm:-MI_7k5mBSbx7jSXOIE7:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/c0bb70de0852ebc770aca5cf5008681b40b11fa0/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1359)):*

please move `*` to the left
</blockquote></details>

Done.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @yamahata)

dimakuv (approved):
> Reviewed 10 of 12 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIcVqgvAD_ihCwTmJsd:-MIcVqgvAD_ihCwTmJse:b-c7b643) (no related file):*
Do I understand correctly that we still want to merge this PR, and then Pawel will work on the GDB 8.2 simplifications in a separate PR? (Seems that everyone agreed on requiring GDB 8.2.) I will put this PR in the merge queue for now.

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIcVqgvAD_ihCwTmJsd:-MIcjuruBf9KhS1sXNNQ:b-fmtspz) (no related file):*

Do I understand correctly that we still want to merge this PR, and then Pawel will work on the GDB 8.2 simplifications in a separate PR? (Seems that everyone agreed on requiring GDB 8.2.) I will put this PR in the merge queue for now.
</blockquote></details>

Yes, please. The simplification itself is easy, but I'd rather do it together with adding GDB to the build, and some regression tests, so it's going to take some time.

---

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIV5pRpEjtWCLFHTD5b:-MIdAyrQ0WlQJayPGqC5:bln07k3) (no related file):*

+1. Would be good to remove this patch from Glibc. Also, I'd remove `glibc_version()` there as well (it is unrelated but a useless feature anyway).
</blockquote></details>

Also something for another PR. Maybe you could move the discussion to #1814?

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIcVqgvAD_ihCwTmJsd:-MIdAmFO-3CHxCkQE2N3:bzhchsr) (no related file):*

Yes, please. The simplification itself is easy, but I'd rather do it together with adding GDB to the build, and some regression tests, so it's going to take some time.
</blockquote></details>

Yes, I also think this should go into a separate PR.

---

mkow (approved):
> Reviewed 1 of 12 files at r4, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1853#-MIV5pRpEjtWCLFHTD5b:-MI_Mjv_3671HrHUbgkA:b804rrq) (no related file):*

+1. Would be good to remove this patch from Glibc. Also, I'd remove `glibc_version()` there as well (it is unrelated but a useless feature anyway).
</blockquote></details>

Added a item for this to #1814 

---

yamahata (approved):
> Reviewed 7 of 12 files at r4, 1 of 1 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1853)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1875, opened 2020-10-09
[LibOS] Add XSAVE area handling and save/restore it on child thread creation

dimakuv:
## Description of the changes 

Previously, LibOS (shim) layer of Graphene didn't support XSAVE area.

The XSAVE area stores FP, XMM, YMM, ZMM, etc. registers and control states and is handled via FXSAVE/XSAVE and FXRSTOR/XRSTOR x86-64 instructions.

This PR is the first step towards adding full-fledged XSAVE support to LibOS. It adds XSAVE related structs and functions to LibOS code, and propagates XSAVE regs/states from parent to child on thread creation via `clone()`.

This PR is a partial recreation of #1218. My PR borrows the XSAVE structs and functions (but renames them for uniformity) from #1218. But XSAVE on signal-handling is not yet implemented; this is for the future.

Future commits will add XSAVE handling on syscall transitions and arriving signals.

## How to test this PR? 

New test `fp_multithread` is added to LibOS regression suite.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)

### DIFF
--- a/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h
+++ b/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h
@@ -26,6 +26,85 @@ struct shim_regs {
     uint64_t rip;
 };
 
+/* adopt Linux x86-64 structs for FP layout: self-contained definition is needed for LibOS, so
+ * define the exact same layout with `shim_` prefix; taken from
+ * https://elixir.bootlin.com/linux/v5.9/source/arch/x86/include/uapi/asm/sigcontext.h */
+#define SHIM_FP_XSTATE_MAGIC1      0x46505853U
+#define SHIM_FP_XSTATE_MAGIC2      0x46505845U
+#define SHIM_FP_XSTATE_MAGIC2_SIZE (sizeof(SHIM_FP_XSTATE_MAGIC2))
+
+#define SHIM_XSTATE_ALIGN 64
+
+enum SHIM_XFEATURE {
+    SHIM_XFEATURE_FP,
+    SHIM_XFEATURE_SSE,
+    SHIM_XFEATURE_YMM,
+    SHIM_XFEATURE_BNDREGS,
+    SHIM_XFEATURE_BNDCSR,
+    SHIM_XFEATURE_OPMASK,
+    SHIM_XFEATURE_ZMM_Hi256,
+    SHIM_XFEATURE_Hi16_ZMM,
+    SHIM_XFEATURE_PT,
+    SHIM_XFEATURE_PKRU,
+};
+
+#define SHIM_XFEATURE_MASK_FP        (1UL << SHIM_XFEATURE_FP)
+#define SHIM_XFEATURE_MASK_SSE       (1UL << SHIM_XFEATURE_SSE)
+#define SHIM_XFEATURE_MASK_YMM       (1UL << SHIM_XFEATURE_YMM)
+#define SHIM_XFEATURE_MASK_BNDREGS   (1UL << SHIM_XFEATURE_BNDREGS)
+#define SHIM_XFEATURE_MASK_BNDCSR    (1UL << SHIM_XFEATURE_BNDCSR)
+#define SHIM_XFEATURE_MASK_OPMASK    (1UL << SHIM_XFEATURE_OPMASK)
+#define SHIM_XFEATURE_MASK_ZMM_Hi256 (1UL << SHIM_XFEATURE_ZMM_Hi256)
+#define SHIM_XFEATURE_MASK_Hi16_ZMM  (1UL << SHIM_XFEATURE_Hi16_ZMM)
+#define SHIM_XFEATURE_MASK_PT        (1UL << SHIM_XFEATURE_PT)
+#define SHIM_XFEATURE_MASK_PKRU      (1UL << SHIM_XFEATURE_PKRU)
+
+#define SHIM_XFEATURE_MASK_FPSSE     (SHIM_XFEATURE_MASK_FP | SHIM_XFEATURE_MASK_SSE)
+#define SHIM_XFEATURE_MASK_AVX512    (SHIM_XFEATURE_MASK_OPMASK | SHIM_XFEATURE_MASK_ZMM_Hi256 \
+                                      | SHIM_XFEATURE_MASK_Hi16_ZMM)
+
+/* Bytes 464..511 in the 512B layout of the FXSAVE/FXRSTOR frame are reserved for SW usage. On
+ * CPUs supporting XSAVE/XRSTOR, these bytes are used to extend the fpstate pointer in the
+ * sigcontext, which includes the extended state information along with fpstate information. */
+struct shim_fpx_sw_bytes {
+    uint32_t magic1;        /*!< SHIM_FP_XSTATE_MAGIC1 (it is an xstate context) */
+    uint32_t extended_size; /*!< g_shim_xsave_size + SHIM_FP_STATE_MAGIC2_SIZE */
+    uint64_t xfeatures;     /*!< XSAVE features (feature bit mask, including FP/SSE/extended) */
+    uint32_t xstate_size;   /*!< g_shim_xsave_size (XSAVE area size as reported by CPUID) */
+    uint32_t padding[7];    /*!< for future use */
+};
+
+/* 64-bit FPU frame (FXSAVE format, 512B total size) */
+struct shim_fpstate {
+    uint16_t cwd;
+    uint16_t swd;
+    uint16_t twd;
+    uint16_t fop;
+    uint64_t rip;
+    uint64_t rdp;
+    uint32_t mxcsr;
+    uint32_t mxcr_mask;
+    uint32_t st_space[32];  /*  8x  FP registers, 16 bytes each */
+    uint32_t xmm_space[64]; /* 16x XMM registers, 16 bytes each */
+    uint32_t reserved2[12];
+    union {
+        uint32_t reserved3[12];
+        struct shim_fpx_sw_bytes sw_reserved; /* potential extended state is encoded here */
+    };
+};
+
+struct shim_xstate_header {
+    uint64_t xfeatures;
+    uint64_t reserved1[2];
+    uint64_t reserved2[5];
+};
+
+struct shim_xstate {
+    struct shim_fpstate fpstate;          /* 512B legacy FXSAVE/FXRSTOR frame (FP and XMM regs) */
+    struct shim_xstate_header xstate_hdr; /* 64B header for newer XSAVE/XRSTOR frame */
+    /* rest is filled with extended regs (YMM, ZMM, ...) by HW + 4B of MAGIC2 value by SW */
+} __attribute__((aligned(SHIM_XSTATE_ALIGN)));
+
 static inline uint64_t shim_regs_get_sp(struct shim_regs* sr) {
     return sr->rsp;
 }
--- a/LibOS/shim/include/shim_context.h
+++ b/LibOS/shim/include/shim_context.h
@@ -9,6 +9,15 @@
 
 #include "shim_tcb.h"
 
+extern bool     g_shim_xsave_enabled;
+extern uint64_t g_shim_xsave_features;
+extern uint32_t g_shim_xsave_size;
+
+void shim_xstate_init(void);
+void shim_xstate_save(void* xstate_extended);
+void shim_xstate_restore(const void* xstate_extended);
+void shim_xstate_reset(void);
+
 void restore_child_context_after_clone(struct shim_context* context);
 void fixup_child_context(struct shim_regs* regs);
 
--- a/LibOS/shim/include/shim_thread.h
+++ b/LibOS/shim/include/shim_thread.h
@@ -313,6 +313,7 @@ struct shim_clone_args {
     struct shim_thread* thread;
     void* stack;
     unsigned long fs_base;
+    void* xstate_extended;
 };
 
 void* allocate_stack(size_t size, size_t protect_size, bool user);
--- a/LibOS/shim/src/Makefile
+++ b/LibOS/shim/src/Makefile
@@ -127,6 +127,10 @@ all_objs = \
 graphene_lib = .lib/graphene-lib.a
 pal_lib = $(RUNTIME_DIR)/libpal-$(PAL_HOST).so
 
+ifeq ($(ARCH),x86_64)
+	CFLAGS += -mfxsr -mxsave
+endif
+
 ifeq ($(findstring x86_64,$(SYS))$(findstring linux,$(SYS)),x86_64linux)
 all: $(files_to_build) $(files_to_install)
 else
--- a/LibOS/shim/src/shim_context-x86_64.c
+++ b/LibOS/shim/src/shim_context-x86_64.c
@@ -1,13 +1,137 @@
 /* SPDX-License-Identifier: LGPL-3.0-or-later */
 
 /*
- * This file contains code for x86-specific CPU context manipulation.
+ * This file contains code for x86_64-specific CPU context manipulation.
  */
 
 #include "shim_context.h"
 
 #include "asm-offsets.h"
+#include "immintrin.h"
+#include "pal.h"
 #include "shim_internal.h"
+
+/* 512 for legacy regs, 64 for xsave header */
+#define XSTATE_RESET_SIZE (512 + 64)
+
+bool     g_shim_xsave_enabled  = false;
+uint64_t g_shim_xsave_features = 0;
+uint32_t g_shim_xsave_size     = 0;
+
+const uint32_t g_shim_xstate_reset_state[XSTATE_RESET_SIZE / sizeof(uint32_t)]
+__attribute__((aligned(SHIM_XSTATE_ALIGN))) = {
+    0x037F, 0, 0, 0, 0, 0, 0x1F80,     0xFFFF, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0,      0, 0, 0, 0, 0, 0,          0,      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0,      0, 0, 0, 0, 0, 0,          0,      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0,      0, 0, 0, 0, 0, 0,          0,      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0,      0, 0, 0, 0, 0, 0,          0,      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0,      0, 0, 0, 0, 0, 0x80000000, 0,      0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    // XCOMP_BV[63] = 1, compaction mode
+};
+
+#define CPUID_FEATURE_XSAVE   (1UL << 26)
+#define CPUID_FEATURE_OSXSAVE (1UL << 27)
+
+#define CPUID_LEAF_PROCINFO 0x00000001
+#define CPUID_LEAF_XSAVE 0x0000000d
+
+void shim_xstate_init(void) {
+    /* by default, fall back to old-style FXSAVE (if cannot deduce from CPUID below) */
+    g_shim_xsave_enabled  = false;
+    g_shim_xsave_features = SHIM_XFEATURE_MASK_FPSSE;
+    g_shim_xsave_size     = XSTATE_RESET_SIZE;
+
+    unsigned int value[4];
+    if (!DkCpuIdRetrieve(CPUID_LEAF_PROCINFO, 0, value))
+        goto out;
+
+    if (!(value[PAL_CPUID_WORD_ECX] & CPUID_FEATURE_XSAVE) ||
+        !(value[PAL_CPUID_WORD_ECX] & CPUID_FEATURE_OSXSAVE))
+        goto out;
+
+    if (!DkCpuIdRetrieve(CPUID_LEAF_XSAVE, 0, value))
+        goto out;
+
+    uint32_t xsavesize = value[PAL_CPUID_WORD_ECX];
+    uint64_t xfeatures = value[PAL_CPUID_WORD_EAX] |
+                         ((uint64_t)value[PAL_CPUID_WORD_EDX] << 32);
+    if (!xsavesize || !xfeatures) {
+        /* could not read xfeatures; fall back to old-style FXSAVE */
+        goto out;
+    }
+
+    if (xfeatures & ~SHIM_XFEATURE_MASK_FPSSE) {
+        /* support more than just FP and SSE, can use XSAVE (it was introduced with AVX) */
+        g_shim_xsave_enabled = true;
+    }
+
+    g_shim_xsave_features  = xfeatures;
+    g_shim_xsave_size      = xsavesize;
+
+out:
+    debug("LibOS xsave_enabled %d, xsave_size 0x%x(%u), xsave_features 0x%lx\n",
+          g_shim_xsave_enabled, g_shim_xsave_size, g_shim_xsave_size, g_shim_xsave_features);
+}
+
+void shim_xstate_save(void* xstate_extended) {
+    assert(IS_ALIGNED_PTR(xstate_extended, SHIM_XSTATE_ALIGN));
+
+    struct shim_xstate* xstate = (struct shim_xstate*)xstate_extended;
+    char* bytes_after_xstate   = (char*)xstate_extended + g_shim_xsave_size;
+
+    if (g_shim_xsave_enabled) {
+        memset(&xstate->xstate_hdr, 0, sizeof(xstate->xstate_hdr));
+        __builtin_ia32_xsave64(xstate, /*mask=*/-1LL);
+    } else {
+        __builtin_ia32_fxsave64(xstate);
+    }
+
+    /* Emulate software format for bytes 464..511 in the 512-byte layout of the FXSAVE/FXRSTOR
+     * frame that Linux uses for x86-64:
+     *   https://elixir.bootlin.com/linux/v5.9/source/arch/x86/kernel/fpu/signal.c#L86
+     *   https://elixir.bootlin.com/linux/v5.9/source/arch/x86/kernel/fpu/signal.c#L517
+     *
+     * This format is assumed by Glibc; this will also be useful if we implement checks in LibOS
+     * similar to Linux's check_for_xstate(). Note that we don't care about CPUs older than
+     * FXSAVE-enabled (so-called "legacy frames"), therefore we always use MAGIC1 and MAGIC2. */
+    struct shim_fpx_sw_bytes* fpx_sw = &xstate->fpstate.sw_reserved;
+    fpx_sw->magic1        = SHIM_FP_XSTATE_MAGIC1;
+    fpx_sw->extended_size = g_shim_xsave_size + SHIM_FP_XSTATE_MAGIC2_SIZE;
+    fpx_sw->xfeatures     = g_shim_xsave_features;
+    fpx_sw->xstate_size   = g_shim_xsave_size;
+    memset(&fpx_sw->padding, 0, sizeof(fpx_sw->padding));
+
+    /* the last 32-bit word of the extended FXSAVE/XSAVE area (at the xstate + extended_size
+     * - FP_XSTATE_MAGIC2_SIZE address) is set to FP_XSTATE_MAGIC2 so that app/Graphene can sanity
+     * check FXSAVE/XSAVE size calculations */
+    *((__typeof__(SHIM_FP_XSTATE_MAGIC2)*)bytes_after_xstate) = SHIM_FP_XSTATE_MAGIC2;
+}
+
+void shim_xstate_restore(const void* xstate_extended) {
+    assert(IS_ALIGNED_PTR(xstate_extended, SHIM_XSTATE_ALIGN));
+
+    struct shim_xstate* xstate = (struct shim_xstate*)xstate_extended;
+    char* bytes_after_xstate   = (char*)xstate_extended + g_shim_xsave_size;
+
+    struct shim_fpx_sw_bytes* fpx_sw = &xstate->fpstate.sw_reserved;
+    assert(fpx_sw->magic1 == SHIM_FP_XSTATE_MAGIC1);
+    assert(fpx_sw->extended_size == g_shim_xsave_size + SHIM_FP_XSTATE_MAGIC2_SIZE);
+    assert(fpx_sw->xfeatures == g_shim_xsave_features);
+    assert(fpx_sw->xstate_size = g_shim_xsave_size);
+    assert(*((__typeof__(SHIM_FP_XSTATE_MAGIC2)*)bytes_after_xstate) == SHIM_FP_XSTATE_MAGIC2);
+
+    __UNUSED(bytes_after_xstate);
+    __UNUSED(fpx_sw);
+
+    if (g_shim_xsave_enabled)
+        __builtin_ia32_xrstor64(xstate, /*mask=*/-1LL);
+    else
+        __builtin_ia32_fxrstor64(xstate);
+}
+
+void shim_xstate_reset(void) {
+    shim_xstate_restore(g_shim_xstate_reset_state);
+}
 
 void restore_child_context_after_clone(struct shim_context* context) {
     assert(context->regs);
--- a/LibOS/shim/src/shim_init.c
+++ b/LibOS/shim/src/shim_init.c
@@ -459,6 +459,8 @@ noreturn void* shim_init(int argc, void* args) {
         shim_clean_and_exit(-EINVAL);
     }
 
+    shim_xstate_init();
+
     if (!create_lock(&__master_lock)) {
         debug("shim_init(): error: failed to allocate __master_lock\n");
         shim_clean_and_exit(-ENOMEM);
--- a/LibOS/shim/src/sys/shim_clone.c
+++ b/LibOS/shim/src/sys/shim_clone.c
@@ -74,6 +74,13 @@ static int clone_implementation_wrapper(struct shim_clone_args* arg) {
     debug("set fs_base to 0x%lx\n", tcb->context.fs_base);
 
     struct shim_regs regs = *arg->parent->shim_tcb->context.regs;
+
+    /* FIXME: The below XSAVE area restore is not really correct but rather a dummy and will be
+     * fixed later. Now it restores the extended state from within LibOS rather than the app. In
+     * reality, XSAVE area should be part of shim_regs, and XRSTOR should happen during
+     * restore_context(). */
+    shim_xstate_restore(arg->xstate_extended);
+
     if (my_thread->set_child_tid) {
         *(my_thread->set_child_tid) = my_thread->tid;
         my_thread->set_child_tid = NULL;
@@ -383,6 +390,16 @@ long shim_do_clone(unsigned long flags, unsigned long user_stack_addr, int* pare
     new_args.parent  = self;
     new_args.stack   = (void*)(user_stack_addr ?: shim_context_get_sp(&self->shim_tcb->context));
     new_args.fs_base = fs_base;
+
+    /* FIXME: The below XSAVE area save is not really correct but rather a dummy and will be fixed
+     * later. Now it saves the extended state from within LibOS rather than the app. In reality,
+     * XSAVE area should be part of shim_regs, and XSAVE should happen during syscalldb().
+     * Also note that we require up to 4KB of stack space for XSAVE -- this is wrong for e.g. Go
+     * because its goroutines start with 2KB stack size; but we'll remove XSAVE here anyway. */
+    size_t xstate_extended_size = g_shim_xsave_size + SHIM_FP_XSTATE_MAGIC2_SIZE;
+    new_args.xstate_extended    = ALIGN_DOWN_PTR(new_args.stack - xstate_extended_size,
+                                                 SHIM_XSTATE_ALIGN);
+    shim_xstate_save(new_args.xstate_extended);
 
     // Invoke DkThreadCreate to spawn off a child process using the actual
     // "clone" system call. DkThreadCreate allocates a stack for the child
--- a/LibOS/shim/test/regression/.gitignore
+++ b/LibOS/shim/test/regression/.gitignore
@@ -36,6 +36,7 @@
 /file_size
 /fopen_cornercases
 /fork_and_exec
+/fp_multithread
 /fstat_cwd
 /futex
 /futex_bitset
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -29,6 +29,7 @@ c_executables = \
 	file_size \
 	fopen_cornercases \
 	fork_and_exec \
+	fp_multithread \
 	fstat_cwd \
 	futex_bitset \
 	futex_requeue \
@@ -165,6 +166,9 @@ CFLAGS-attestation += -I$(PALDIR)/../lib/crypto/mbedtls/crypto/include \
                       -I$(PALDIR)/host/Linux-SGX \
                       -I$(PALDIR)/../include/pal
 LDLIBS-attestation += $(PALDIR)/../lib/crypto/mbedtls/install/lib/libmbedcrypto.a
+
+CFLAGS-fp_multithread += -pthread -fno-builtin  # see comment in the test's source
+LDLIBS-fp_multithread += -lm
 
 # Handle multiple tests which share one binary (Graphene requires 1-1 manifest-executable
 # correspondence).
--- a/LibOS/shim/test/regression/fp_multithread.c
+++ b/LibOS/shim/test/regression/fp_multithread.c
@@ -1,0 +1,52 @@
+/* use `-fno-builtin` compiler option to prevent nearbyint and fesetround being optimized out */
+
+#include <err.h>
+#include <fenv.h>
+#include <math.h>
+#include <pthread.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+static int rounding_modes[] = {FE_TONEAREST, FE_UPWARD, FE_DOWNWARD, FE_TOWARDZERO};
+
+static char* get_rounding_mode(int mode) {
+    switch (mode) {
+        case FE_TONEAREST:  return "FE_TONEAREST ";
+        case FE_UPWARD:     return "FE_UPWARD    ";
+        case FE_DOWNWARD:   return "FE_DOWNWARD  ";
+        case FE_TOWARDZERO: return "FE_TOWARDZERO";
+    }
+    return "UNKNOWN";
+}
+
+static void* thread_fp(void* arg) {
+    int mode = fegetround();
+    printf("%s  child: 42.5 = %.1f, -42.5 = %.1f\n", get_rounding_mode(mode),
+           nearbyint(42.5), nearbyint(-42.5));
+    return arg;
+}
+
+int main(int argc, char* argv[]) {
+    int ret;
+
+    for (int i = 0; i < sizeof(rounding_modes)/sizeof(rounding_modes[0]); i++) {
+        int mode = rounding_modes[i];
+        ret = fesetround(mode);
+        if (ret)
+            err(EXIT_FAILURE, "fesetround failed");
+
+        pthread_t thread;
+        ret = pthread_create(&thread, NULL, thread_fp, NULL);
+        if (ret)
+            err(EXIT_FAILURE, "pthread_create failed");
+
+        ret = pthread_join(thread, NULL);
+        if (ret)
+            err(EXIT_FAILURE, "pthread_join failed");
+
+        printf("%s parent: 42.5 = %.1f, -42.5 = %.1f\n", get_rounding_mode(mode),
+               nearbyint(42.5), nearbyint(-42.5));
+    }
+
+    return 0;
+}
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -234,6 +234,17 @@ class TC_01_Bootstrap(RegressionTestCase):
         # Multiple thread creation
         self.assertIn('128 Threads Created', stdout)
 
+    def test_602_fp_multithread(self):
+        stdout, _ = self.run_binary(['fp_multithread'])
+        self.assertIn('FE_TONEAREST   child: 42.5 = 42.0, -42.5 = -42.0', stdout)
+        self.assertIn('FE_TONEAREST  parent: 42.5 = 42.0, -42.5 = -42.0', stdout)
+        self.assertIn('FE_UPWARD      child: 42.5 = 43.0, -42.5 = -42.0', stdout)
+        self.assertIn('FE_UPWARD     parent: 42.5 = 43.0, -42.5 = -42.0', stdout)
+        self.assertIn('FE_DOWNWARD    child: 42.5 = 42.0, -42.5 = -43.0', stdout)
+        self.assertIn('FE_DOWNWARD   parent: 42.5 = 42.0, -42.5 = -43.0', stdout)
+        self.assertIn('FE_TOWARDZERO  child: 42.5 = 42.0, -42.5 = -42.0', stdout)
+        self.assertIn('FE_TOWARDZERO parent: 42.5 = 42.0, -42.5 = -42.0', stdout)
+
 @unittest.skipUnless(HAS_SGX,
     'This test is only meaningful on SGX PAL because only SGX catches raw '
     'syscalls and redirects to Graphene\'s LibOS. If we will add seccomp to '


yamahata (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: 0 of 10 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[LibOS/shim/src/shim_context-x86_64.c, line 30 at r1](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJD6PCO4doW4SeIMkTK:-MJD6PCP3HzCzKF1UONy:b8v83z5) ([raw file](https://github.com/oscarlab/graphene/blob/84ddd104135966cf3bdbff7ad68b9f5cbd8e84de/LibOS/shim/src/shim_context-x86_64.c#L30)):*
<details><summary><i alt="summary">Quoted 6 lines of code…</i></summary>

> ```
> enum SHIM_CPUID_WORD {
>     SHIM_CPUID_WORD_EAX = 0,
>     SHIM_CPUID_WORD_EBX = 1,
>     SHIM_CPUID_WORD_ECX = 2,
>     SHIM_CPUID_WORD_EDX = 3,
>     SHIM_CPUID_WORD_NUM = 4,
> ```
</details>


Why doesn't PAL_CPU_WORD_* work?
___
*[LibOS/shim/src/sys/shim_clone.c, line 106 at r1](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJD8GzH-bCIeL176Q9o:-MJD8GzH-bCIeL176Q9p:b-p0vld8) ([raw file](https://github.com/oscarlab/graphene/blob/84ddd104135966cf3bdbff7ad68b9f5cbd8e84de/LibOS/shim/src/sys/shim_clone.c#L106)):*
> ```C
> 
>     /* TODO: XSAVE area should be part of shim_regs, and restore_context() should XRSTOR */
>     shim_xsave_restore(arg->xregs_state);
> ```

Because currently LibOS is complied with FP code enabled, it doesn't make sense to restore FP register here.
It must be done in restore_context().
or Disable FP code on the execution path.
___
*[LibOS/shim/src/sys/shim_clone.c, line 397 at r1](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJD7OCZBR0YXAe4gSW1:-MJD7OCZBR0YXAe4gSW2:b-yomejz) ([raw file](https://github.com/oscarlab/graphene/blob/84ddd104135966cf3bdbff7ad68b9f5cbd8e84de/LibOS/shim/src/sys/shim_clone.c#L397)):*
> ```C
> 
>     struct shim_vma_info vma_info;
>     if (lookup_vma(ALLOC_ALIGN_DOWN_PTR(user_stack_addr), &vma_info) < 0) {
> ```

nice catch.
___
*[LibOS/shim/src/sys/shim_clone.c, line 431 at r1](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJD8sZJ6u_9wuzPaR6_:-MJD8sZJ6u_9wuzPaR6a:b-scbef9) ([raw file](https://github.com/oscarlab/graphene/blob/84ddd104135966cf3bdbff7ad68b9f5cbd8e84de/LibOS/shim/src/sys/shim_clone.c#L431)):*
> ```C
>      * NOTE: this assumes the user-provided stack is at least ~2KB (for AVX512-enabled CPU) */
>     new_args.xregs_state = ALIGN_DOWN_PTR(user_stack_addr - g_shim_xsave_size, SHIM_XSTATE_ALIGN);
>     shim_xsave_save(new_args.xregs_state);
> ```

similar here.
FP control words are callee save, and the newly interoduced test only tests control word. I think that's the reason why the new test passes.

yamahata (commented):
We know the handling of FP register is heavily broken, and this is only a starting point.
So I'm fine with merging inperfect PR at this point.

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875#-:-MJDAlv6CxDNcL--stJn:b-wsi8co)* status: 0 of 10 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

dimakuv (commented):
Isaku is right. This PR is just a first step to adding FP/XSAVE support to the LibOS layer. The main purpose of this PR is to introduce the required structs and primitives (`xregs_state`, init/save/restore functions). The poor-man's emulation of XSAVE on `clone()` is just to test that it works, and also to fix one particular use case of Graphene. Next PRs will start adding proper emulation of FP/XSAVE.

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875#-:-MJQxhdV91tpaiovVmM6:b-r3wkcv)* status: 0 of 10 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @yamahata)

___
*[LibOS/shim/src/shim_context-x86_64.c, line 30 at r1](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJD6PCO4doW4SeIMkTK:-MJQz4dqEB4vSgf-bHk6:bbwdgvz) ([raw file](https://github.com/oscarlab/graphene/blob/84ddd104135966cf3bdbff7ad68b9f5cbd8e84de/LibOS/shim/src/shim_context-x86_64.c#L30)):*

> ```
> enum SHIM_CPUID_WORD {
>     SHIM_CPUID_WORD_EAX = 0,
>     SHIM_CPUID_WORD_EBX = 1,
>     SHIM_CPUID_WORD_ECX = 2,
>     SHIM_CPUID_WORD_EDX = 3,
>     SHIM_CPUID_WORD_NUM = 4,
> ```

Why doesn't PAL_CPU_WORD_* work?
</blockquote></details>

Done. I first wanted it to be self-contained, but now I see that these `PAL_CPUID_...` constants are a part of the PAL interface, so perfectly legitimate to use.
___
*[LibOS/shim/src/sys/shim_clone.c, line 106 at r1](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJD8GzH-bCIeL176Q9o:-MJQzGcI9fGUkvh3xb8q:bix2opg) ([raw file](https://github.com/oscarlab/graphene/blob/84ddd104135966cf3bdbff7ad68b9f5cbd8e84de/LibOS/shim/src/sys/shim_clone.c#L106)):*

Because currently LibOS is complied with FP code enabled, it doesn't make sense to restore FP register here.
It must be done in restore_context().
or Disable FP code on the execution path.
</blockquote></details>

Indeed it doesn't really make sense. This is a "dummy" to test that XSAVE functionality works in general. This is what my TODO comment refers to. Next PRs will introduce a more correct handling of xstate and fix this. Changed to FIXME to make it more explicit.
___
*[LibOS/shim/src/sys/shim_clone.c, line 431 at r1](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJD8sZJ6u_9wuzPaR6_:-MJQzbqv2pzO29Mv4q7M:b-1pu603) ([raw file](https://github.com/oscarlab/graphene/blob/84ddd104135966cf3bdbff7ad68b9f5cbd8e84de/LibOS/shim/src/sys/shim_clone.c#L431)):*

similar here.
FP control words are callee save, and the newly interoduced test only tests control word. I think that's the reason why the new test passes.
</blockquote></details>

Indeed, see my other reply.

yamahata (approved):
> Reviewed 8 of 10 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

dimakuv:
FYI: I had to rebase this PR to test Jenkins (previously we had issues with Jenkins which were merged in the meantime).

dimakuv:
Jenkins, retest Jenkins please (`fork11` LTP test failed, known and unrelated)

mkow (changes requested):
> Reviewed 5 of 10 files at r1, 1 of 2 files at r2, 3 of 3 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/include/arch/x86_64/shim_tcb-arch.h, line 29 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJxQxCb4TM3bE1EpDoX:-MJxQxCb4TM3bE1EpDoY:b-ewg1nl) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h#L29)):*
> ```
> Linux-style FP layout (_libc_fpstate of glibc)
> ```
So, Linux- or glibc- style? Why not using Linux structs?
___
*[LibOS/shim/include/arch/x86_64/shim_tcb-arch.h, line 48 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJy6DZDBYG5DLiT_Vw1:-MJy6DZDBYG5DLiT_Vw2:byh5aci) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h#L48)):*
> ```
> SHIM_XFEATURE_MAX,
> ```

Looks unused.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 33 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJy8sGe4HO4WkUhBL-7:-MJy8sGe4HO4WkUhBL-8:b1fb736) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L33)):*
> ```C
> #define ECX_XSAVE   (1UL << 26)
> #define ECX_OSXSAVE (1UL << 27)
> #define XSAVE_CPUID 0x0000000d
> ```

In some other place we used the following format: `CPUID_LEAF_<leaf_name>`
___
*[LibOS/shim/src/shim_context-x86_64.c, line 42 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJy90y38NdRmNSGcH38:-MJy90y45BjAzx9hEtst:b-4jggk) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L42)):*
> ```C
> 
>     unsigned int value[4];
>     if (!DkCpuIdRetrieve(0x1, 0, value))
> ```

Could you change this 0x1 into a `CPUID_LEAF_<leaf_name>` define?
___
*[LibOS/shim/src/shim_context-x86_64.c, line 53 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-e-hUETTVJOVxwjf1:-MK-e-hVC-fw1KOFb52K:b-xjm3u9) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L53)):*
> ```C
>     uint32_t xsavesize = value[PAL_CPUID_WORD_ECX];
>     uint64_t xfeatures = value[PAL_CPUID_WORD_EAX] |
>                          (((uint64_t)value[PAL_CPUID_WORD_EDX]) << 32);
> ```

No need for parenthesis around the cast, casting is much higher in priority than the bit shifting (so, I don't think people would find this code confusing without them).
___
*[LibOS/shim/src/shim_context-x86_64.c, line 55 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-i8mr1XFUg-1-4XN1:-MK-i8mr1XFUg-1-4XN2:b-2ge3yy) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L55)):*
> ```C
>                          (((uint64_t)value[PAL_CPUID_WORD_EDX]) << 32);
>     if (!xfeatures)
>         goto out;
> ```

Why do we have this `if` here if we check this variable again, right below?
___
*[LibOS/shim/src/shim_context-x86_64.c, line 75 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-iQoH1bJlPeyfE-h2:-MK-iQoH1bJlPeyfE-h3:bu7vpst) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L75)):*
> ```C
>     if (g_shim_xsave_enabled) {
>         long lmask = -1;
>         long hmask = -1;
> ```

please use `uint64_t`
___
*[LibOS/shim/src/shim_context-x86_64.c, line 84 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-jQe51AjWk2V9fkbn:-MK-jQe51AjWk2V9fkbo:b-bbt0ow) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L84)):*
> ```C
>     /* Emulate format for FP registers that Linux uses:
>      * https://elixir.bootlin.com/linux/v5.4.13/source/arch/x86/kernel/fpu/signal.c#L86
>      * https://elixir.bootlin.com/linux/v5.4.13/source/arch/x86/kernel/fpu/signal.c#L459 */
> ```

Actually, why are we emulating Linux format? Is any syscall using it? Or is this actually an x86 format?
___
*[LibOS/shim/src/shim_context-x86_64.c, line 90 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-kI0x5er5Yj103ow0:-MK-kI0x5er5Yj103ow1:b-v079ro) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L90)):*
> ```C
>     fpx_sw->xfeatures     = g_shim_xsave_features;
>     fpx_sw->xstate_size   = g_shim_xsave_size;
>     memset(fpx_sw->padding, 0, sizeof(fpx_sw->padding));
> ```

Please add `&` to the first arg (just for consistency)
___
*[LibOS/shim/src/shim_context-x86_64.c, line 93 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4dsTh8axy0TUBSpYJ:-MK4dsTi1Kzs0w3MZ9nj:b-8uuip3) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L93)):*
> ```C
>     if (g_shim_xsave_enabled) {
>         *((__typeof__(SHIM_FP_XSTATE_MAGIC2)*)((char*)xregs_state + fpx_sw->xstate_size)) =
>             SHIM_FP_XSTATE_MAGIC2;
> ```

How does this work? Why `shim_xregs_state` is actually a structure, if its fields are shifted depending on some runtime config?
___
*[LibOS/shim/src/shim_context-x86_64.c, line 101 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4kDjR4by5C4lbZMFe:-MK4kDjS-R1NTzxctB0i:bbg0lw8) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L101)):*
> ```C
>     if (g_shim_xsave_enabled) {
>         long lmask = -1;
>         long hmask = -1;
> ```

`uint64_t`
___
*[LibOS/shim/src/sys/shim_clone.c, line 393 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4lL1_1UbYk8JtFyBy:-MK4lL1_1UbYk8JtFyBz:b-6ffijx) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/sys/shim_clone.c#L393)):*
> ```C
>     enable_locking();
> 
>     /* when cloning a new thread, user must provide stack space for this new thread */
> ```

/summon @boryspoplawski
There was a discussion about this case under #1902. Basically, there will be a lot of collisions between both PRs in this function. In this case Borys says we handle `user_stack_addr == 0` correctly (which is a valid argument for clone).
___
*[LibOS/shim/test/regression/fp_multithread.c, line 24 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4mpGe8Dr-ggoMaueD:-MK4mpGe8Dr-ggoMaueE:b-sxm00z) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/test/regression/fp_multithread.c#L24)):*
> ```C
>         mode = atoi(argv[1]);
>         if (mode > sizeof(rounding_modes)/sizeof(rounding_modes[0]) - 1) {
>             errx(EXIT_FAILURE, "run with single argument <rounding mode: one of 0..%lu>",
> ```

`with single argument` -> `with a single argument`
___
*[LibOS/shim/test/regression/Makefile, line 165 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJstVFkAlR7xyXJkjhx:-MJstVFl3aP8rpBYJq1K:b3ei4fd) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/test/regression/Makefile#L165)):*
> ```Makefile
> LDLIBS-attestation += $(PALDIR)/../lib/crypto/mbedtls/install/lib/libmbedcrypto.a
> 
> CFLAGS-fp_multithread = -pthread -fno-builtin
> ```

Please explain in a comment why `-fno-builtin` is needed (something like "this is needed, see the comment in the test's source" is enough)
___
*[LibOS/shim/test/regression/multi_pthread.manifest.template, line 3 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4omPS5dpOa37m5_5Q:-MK4omPS5dpOa37m5_5R:b-n0lbir) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/test/regression/multi_pthread.manifest.template#L3)):*
> ```
> loader.preload = file:../../src/libsysdb.so
> loader.env.LD_LIBRARY_PATH = /lib
> loader.debug_type = inline
> ```

debugging leftover?
___
*[LibOS/shim/test/regression/test_libos.py, line 254 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4p31t5FDQmklMvPxJ:-MK4p31t5FDQmklMvPxK:b3inydc) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/test/regression/test_libos.py#L254)):*
> ```Python
>         stdout, _ = self.run_binary(['fp_multithread', '2'])
>         self.assertIn('child: 42.5 = 42.0, -42.5 = -43.0', stdout)
>         self.assertIn('parent: 42.5 = 42.0, -42.5 = -43.0', stdout)
> ```

Instead of 3 binary invocations (which take some considerable time on SGX), could we have just one which just loops over all possible settings?

mkow (changes requested):
> Reviewed 5 of 10 files at r1, 1 of 2 files at r2, 3 of 3 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: 1 of 10 files reviewed, 16 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @yamahata)

___
*[LibOS/shim/include/arch/x86_64/shim_tcb-arch.h, line 29 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJxQxCb4TM3bE1EpDoX:-MK57Cj274p5-obCjB8f:b-itg2xa) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h#L29)):*

> ```
> Linux-style FP layout (_libc_fpstate of glibc)
> ```
So, Linux- or glibc- style? Why not using Linux structs?
</blockquote></details>

Done. Linux and Glibc for x86-64 have the same format (they have to be ABI compatible). But you are right, since we're emulating Linux, we should use the style used in Linux sources, not a mix with Glibc.

Please note that I also changed the names from Glibc-like to Linux-like. The most important change is `xregs_state` -> `xstate`
___
*[LibOS/shim/include/arch/x86_64/shim_tcb-arch.h, line 48 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJy6DZDBYG5DLiT_Vw1:-MK57SSY05NlSHO-yo2Q:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h#L48)):*

> ```
> SHIM_XFEATURE_MAX,
> ```

Looks unused.
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 33 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJy8sGe4HO4WkUhBL-7:-MK58xW5-IgbwdxBOniJ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L33)):*

In some other place we used the following format: `CPUID_LEAF_<leaf_name>`
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 42 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJy90y38NdRmNSGcH38:-MK5903Q5Kv6UX-aQinP:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L42)):*

Could you change this 0x1 into a `CPUID_LEAF_<leaf_name>` define?
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 53 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-e-hUETTVJOVxwjf1:-MK597vTADvPQR0wBCuI:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L53)):*

No need for parenthesis around the cast, casting is much higher in priority than the bit shifting (so, I don't think people would find this code confusing without them).
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 55 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-i8mr1XFUg-1-4XN1:-MK5A79t4a9ASiKMm_HX:b-lwevzz) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L55)):*

Why do we have this `if` here if we check this variable again, right below?
</blockquote></details>

But this is a different check: here we check if we were able to retrieve the values for size and bitmask of available XSAVE features. If not, we fall back to FXSAVE. I added more explanations and an additional sanity check.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 75 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-iQoH1bJlPeyfE-h2:-MK5APet7cXrPbvCTAin:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L75)):*

please use `uint64_t`
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 84 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-jQe51AjWk2V9fkbn:-MK5HGe62ETT6f3S4p-n:b6hmoc2) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L84)):*

Actually, why are we emulating Linux format? Is any syscall using it? Or is this actually an x86 format?
</blockquote></details>

This is a good question. Glibc and Musl (and generally any libc) assume the same format. This allows to do checks in user-space that the user didn't corrupt XSTATE (which can be in legacy FXSAVE format or in newer XSAVE format). Though I don't see Glibc performing any checks. Also, Linux actually checks this format and the magic values on context switches, especially on signal-handling returns from user land. See https://elixir.bootlin.com/linux/v5.9/source/arch/x86/kernel/fpu/signal.c#L24. We may want to implement something like this in LibOS one day.

It is *not* an x86 format. This is pure SW convention, to be able to check for state corruptions (HW just gives bytes 464..511 of FXSAVE/XSAVE area for SW to do whatever).
___
*[LibOS/shim/src/shim_context-x86_64.c, line 90 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-kI0x5er5Yj103ow0:-MK5BD2U91xHx44vSbIM:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L90)):*

Please add `&` to the first arg (just for consistency)
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 93 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4dsTh8axy0TUBSpYJ:-MK5IXsQ2CCO7e8uDnDd:bqryzg0) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L93)):*
This explains it: https://elixir.bootlin.com/linux/v5.9/source/arch/x86/kernel/fpu/signal.c#L42.

Basically, there are two HW ways of saving FP/extended state: legacy FXSAVE and newer XSAVE:
- FXSAVE has a fixed size of 512B. It is enough to have some header fields, then x87 and SSE reg values, and a dozen of bytes for random SW purposes.
- XSAVE has an arbitrary size of >512B. This allows for extensibility: new XMM, YMM, ZMM, bounds-checking registers may require 2-4KB of memory to save all values. This XSAVE feature started with AVX technology, it's rather new.

Now `xregs_state` (which I renamed to `xstate` to be closer to Linux sources) has the `xstate_size` which is 512B for FXSAVE or more for XSAVE (detected via CPUID at runtime). But its "real total size" is `xstate_size + FP_XSTATE_MAGIC2_SIZE` so there are 4 bytes at the very end of `state` object that contain the magic2 value. And Linux puts this magic2 value to make sure the calculations in the app and kernel sources are correct at all times. (This is checked in e.g. `check_for_xstate()` https://elixir.bootlin.com/linux/v5.9/source/arch/x86/kernel/fpu/signal.c#L24)

> if its fields are shifted depending on some runtime config

I didn't understand this comment. There is no shifting of fields. We only add a magic 4B value to the very end of this struct (actually, even past the "xsave-area" struct size).
___
*[LibOS/shim/src/shim_context-x86_64.c, line 101 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4kDjR4by5C4lbZMFe:-MK5BPhx6yunbjlnzSdd:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L101)):*

`uint64_t`
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_clone.c, line 393 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4lL1_1UbYk8JtFyBy:-MK5LmRa3IUz6nIkIsan:baknf5u) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/sys/shim_clone.c#L393)):*

/summon @boryspoplawski
There was a discussion about this case under #1902. Basically, there will be a lot of collisions between both PRs in this function. In this case Borys says we handle `user_stack_addr == 0` correctly (which is a valid argument for clone).
</blockquote></details>

Yes, my PR wants the user to specify `user_stack_addr != 0`. Because I use the new-thread stack to copy the XSAVE area (SSE/XMM/... regs) from parent thread to child thread. But I could also use `malloc()`/`free()` if there is a chance that a new thread may have no stack...
___
*[LibOS/shim/test/regression/fp_multithread.c, line 24 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4mpGe8Dr-ggoMaueD:-MK5Ni8A-6A0RP_WBpgq:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/test/regression/fp_multithread.c#L24)):*

`with single argument` -> `with a single argument`
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/Makefile, line 165 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MJstVFkAlR7xyXJkjhx:-MK5NrWq0ZyjUlD4wHfX:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/test/regression/Makefile#L165)):*

Please explain in a comment why `-fno-builtin` is needed (something like "this is needed, see the comment in the test's source" is enough)
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/multi_pthread.manifest.template, line 3 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4omPS5dpOa37m5_5Q:-MK5NtQYAmOGjvQwQobg:b-qsx7u6) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/test/regression/multi_pthread.manifest.template#L3)):*

debugging leftover?
</blockquote></details>

Done. Oops.
___
*[LibOS/shim/test/regression/test_libos.py, line 254 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4p31t5FDQmklMvPxJ:-MK5O686-oHWCCFU1IYP:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/test/regression/test_libos.py#L254)):*

Instead of 3 binary invocations (which take some considerable time on SGX), could we have just one which just loops over all possible settings?
</blockquote></details>

Done.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: 1 of 10 files reviewed, 16 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @yamahata)

___
*[LibOS/shim/src/sys/shim_clone.c, line 393 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4lL1_1UbYk8JtFyBy:-MK5ergl3WTcLWLxZD2_:b-q27me0) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/sys/shim_clone.c#L393)):*

Yes, my PR wants the user to specify `user_stack_addr != 0`. Because I use the new-thread stack to copy the XSAVE area (SSE/XMM/... regs) from parent thread to child thread. But I could also use `malloc()`/`free()` if there is a chance that a new thread may have no stack...
</blockquote></details>

1. I think it would be a good idea if you rebased this PR on top of 1902 once it gets merged (once I debug a test failing).
2. If `user_stack_addr == 0` then we use parent stack. I know it sounds crazy and in normal cases it won't work, but:
    a) that's what Linux does,
    b) it's dead easy to implement (one if),
    c) it might work in some weird apps switching stack in parent, not using it at all, or some other craziness.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: 1 of 10 files reviewed, 16 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @yamahata)

___
*[LibOS/shim/src/sys/shim_clone.c, line 393 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4lL1_1UbYk8JtFyBy:-MK6SrDJDwqe1KS2eF_i:bwh6hqs) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/sys/shim_clone.c#L393)):*

1. I think it would be a good idea if you rebased this PR on top of 1902 once it gets merged (once I debug a test failing).
2. If `user_stack_addr == 0` then we use parent stack. I know it sounds crazy and in normal cases it won't work, but:
    a) that's what Linux does,
    b) it's dead easy to implement (one if),
    c) it might work in some weird apps switching stack in parent, not using it at all, or some other craziness.
</blockquote></details>

1902 was merged

mkow (changes requested):
> Reviewed 10 of 10 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/src/shim_context-x86_64.c, line 55 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-i8mr1XFUg-1-4XN1:-MK5s7tV8pEdLyyiDQu_:b-1t8z23) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L55)):*

But this is a different check: here we check if we were able to retrieve the values for size and bitmask of available XSAVE features. If not, we fall back to FXSAVE. I added more explanations and an additional sanity check.
</blockquote></details>

Ok, but then I think it would be more readable and safer to write `xfeatures & ~SHIM_XFEATURE_MASK_FPSSE`, this actually checks what you wanted to check (without any additional assumptions).
___
*[LibOS/shim/src/shim_context-x86_64.c, line 84 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-jQe51AjWk2V9fkbn:-MK5q3qw1D8Nr-eVjgcU:bl427kw) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L84)):*

This is a good question. Glibc and Musl (and generally any libc) assume the same format. This allows to do checks in user-space that the user didn't corrupt XSTATE (which can be in legacy FXSAVE format or in newer XSAVE format). Though I don't see Glibc performing any checks. Also, Linux actually checks this format and the magic values on context switches, especially on signal-handling returns from user land. See https://elixir.bootlin.com/linux/v5.9/source/arch/x86/kernel/fpu/signal.c#L24. We may want to implement something like this in LibOS one day.

It is *not* an x86 format. This is pure SW convention, to be able to check for state corruptions (HW just gives bytes 464..511 of FXSAVE/XSAVE area for SW to do whatever).
</blockquote></details>

Thanks for the explanation, it's much clearer to me now.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 93 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4dsTh8axy0TUBSpYJ:-MK5tGSz0Z21OJ6e6Gm0:b-63ezjs) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L93)):*

This explains it: https://elixir.bootlin.com/linux/v5.9/source/arch/x86/kernel/fpu/signal.c#L42.

Basically, there are two HW ways of saving FP/extended state: legacy FXSAVE and newer XSAVE:
- FXSAVE has a fixed size of 512B. It is enough to have some header fields, then x87 and SSE reg values, and a dozen of bytes for random SW purposes.
- XSAVE has an arbitrary size of >512B. This allows for extensibility: new XMM, YMM, ZMM, bounds-checking registers may require 2-4KB of memory to save all values. This XSAVE feature started with AVX technology, it's rather new.

Now `xregs_state` (which I renamed to `xstate` to be closer to Linux sources) has the `xstate_size` which is 512B for FXSAVE or more for XSAVE (detected via CPUID at runtime). But its "real total size" is `xstate_size + FP_XSTATE_MAGIC2_SIZE` so there are 4 bytes at the very end of `state` object that contain the magic2 value. And Linux puts this magic2 value to make sure the calculations in the app and kernel sources are correct at all times. (This is checked in e.g. `check_for_xstate()` https://elixir.bootlin.com/linux/v5.9/source/arch/x86/kernel/fpu/signal.c#L24)

> if its fields are shifted depending on some runtime config

I didn't understand this comment. There is no shifting of fields. We only add a magic 4B value to the very end of this struct (actually, even past the "xsave-area" struct size).
</blockquote></details>

Ah, what confused me is that you use `struct shim_xstate`, but it actually has trailing data afterwards. Could you either comment this stuct or maybe pass it as void* and unpack here to header + data? Current version was very confusing to me.
___
*[LibOS/shim/test/regression/fp_multithread.c, line 12 at r5](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK75utgAjm-p3Lr7b7A:-MK75uth73vJW61KXXvm:bc6izqj) ([raw file](https://github.com/oscarlab/graphene/blob/f1cfee9c0a7884933dca7a54bca1ebb4180c2dea/LibOS/shim/test/regression/fp_multithread.c#L12)):*
> ```C
> static int rounding_modes[] = {FE_TONEAREST, FE_UPWARD, FE_DOWNWARD, FE_TOWARDZERO};
> 
> static char* print_rounding_mode(int mode) {
> ```

`print` -> `get`: it's not actually printing anything
___
*[LibOS/shim/test/regression/fp_multithread.c, line 19 at r5](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK6LgnS1LkgqQThE-gs:-MK6LgnS1LkgqQThE-gt:b-1368dj) ([raw file](https://github.com/oscarlab/graphene/blob/f1cfee9c0a7884933dca7a54bca1ebb4180c2dea/LibOS/shim/test/regression/fp_multithread.c#L19)):*
> ```
> UNKOWN
> ```

Something is not right here :)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: 0 of 10 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKAW2ZYBrOdsujQj9V8:-MKAW2ZYBrOdsujQj9V9:b-v44vdf) (no related file):*
NOTE: Squashed and rebased to the current master because #1902 was merged.

---
___
*[LibOS/shim/src/shim_context-x86_64.c, line 55 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK-i8mr1XFUg-1-4XN1:-MKACrEsCcBVUI7L1FvT:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L55)):*

Ok, but then I think it would be more readable and safer to write `xfeatures & ~SHIM_XFEATURE_MASK_FPSSE`, this actually checks what you wanted to check (without any additional assumptions).
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 93 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4dsTh8axy0TUBSpYJ:-MKAGM5k7f0ys1ypwqy0:b-x6aq1d) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L93)):*

Ah, what confused me is that you use `struct shim_xstate`, but it actually has trailing data afterwards. Could you either comment this stuct or maybe pass it as void* and unpack here to header + data? Current version was very confusing to me.
</blockquote></details>

Done. Actually found a couple bugs while doing this. Now the code should be much clearer.
___
*[LibOS/shim/src/sys/shim_clone.c, line 393 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4lL1_1UbYk8JtFyBy:-MKAWKgn8S2WTDikwVRz:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/sys/shim_clone.c#L393)):*

1902 was merged
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/fp_multithread.c, line 12 at r5](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK75utgAjm-p3Lr7b7A:-MKA6uWM7hT52gtxa5UL:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f1cfee9c0a7884933dca7a54bca1ebb4180c2dea/LibOS/shim/test/regression/fp_multithread.c#L12)):*

`print` -> `get`: it's not actually printing anything
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/fp_multithread.c, line 19 at r5](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK6LgnS1LkgqQThE-gs:-MKA6tjKCSi5AyzvqOw3:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f1cfee9c0a7884933dca7a54bca1ebb4180c2dea/LibOS/shim/test/regression/fp_multithread.c#L19)):*

> ```
> UNKOWN
> ```

Something is not right here :)
</blockquote></details>

Done.

dimakuv:
Jenkins, retest Jenkins please (`truncate01` and `writev07` LTP tests failed, cannot reproduce)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: 0 of 10 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/src/sys/shim_clone.c, line 402 at r6](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKAdcMk1DSAr4921dTR:-MKAdcMl5qXPOOJZn74A:bx6t6pi) ([raw file](https://github.com/oscarlab/graphene/blob/b4d21052387297453ef876f2f2910829010ce329/LibOS/shim/src/sys/shim_clone.c#L402)):*
> ```C
>     new_args.xstate_extended    = ALIGN_DOWN_PTR(new_args.stack - xstate_extended_size,
>                                                  SHIM_XSTATE_ALIGN);
>     shim_xstate_save(new_args.xstate_extended);
> ```

This touches memory pointed by `new_args.stack` before validity of that argument is checked (`lookup_vma` in `clone_implementaton_wrapper`).

mkow (changes requested):
> Reviewed 10 of 10 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/src/shim_context-x86_64.c, line 93 at r4](https://reviewable.io/reviews/oscarlab/graphene/1875#-MK4dsTh8axy0TUBSpYJ:-MKAc05W9CZRY64gAVRV:bwxjn8z) ([raw file](https://github.com/oscarlab/graphene/blob/196fe0818fb05fcf7a448bd504501c11e2dda967/LibOS/shim/src/shim_context-x86_64.c#L93)):*

Done. Actually found a couple bugs while doing this. Now the code should be much clearer.
</blockquote></details>

Yeah, it's much better now :)
___
*[LibOS/shim/src/sys/shim_clone.c, line 401 at r6](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKAdodiBfAtHUXG3xHu:-MKAdodiBfAtHUXG3xHv:b-bg8s47) ([raw file](https://github.com/oscarlab/graphene/blob/b4d21052387297453ef876f2f2910829010ce329/LibOS/shim/src/sys/shim_clone.c#L401)):*
> ```C
>     size_t xstate_extended_size = g_shim_xsave_size + SHIM_FP_XSTATE_MAGIC2_SIZE;
>     new_args.xstate_extended    = ALIGN_DOWN_PTR(new_args.stack - xstate_extended_size,
>                                                  SHIM_XSTATE_ALIGN);
> ```

Shouldn't we fix up `new_args.stack` after this alignment?
___
*[LibOS/shim/test/regression/Makefile, line 168 at r6](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKAe8HI3o5CWTWLP6b_:-MKAe8HJ5ME5H4BCR-af:b-mabwp1) ([raw file](https://github.com/oscarlab/graphene/blob/b4d21052387297453ef876f2f2910829010ce329/LibOS/shim/test/regression/Makefile#L168)):*
> ```Makefile
> LDLIBS-attestation += $(PALDIR)/../lib/crypto/mbedtls/install/lib/libmbedcrypto.a
> 
> CFLAGS-fp_multithread = -pthread -fno-builtin  # see comment in the test's source
> ```

Just noticed: why it's `=`, not `+=`?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[LibOS/shim/src/sys/shim_clone.c, line 401 at r6](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKAdodiBfAtHUXG3xHu:-MKAfNmL71NvUMgZMXi-:b9a6jsd) ([raw file](https://github.com/oscarlab/graphene/blob/b4d21052387297453ef876f2f2910829010ce329/LibOS/shim/src/sys/shim_clone.c#L401)):*

Shouldn't we fix up `new_args.stack` after this alignment?
</blockquote></details>

No, because we kinda abuse this stack to dump XSAVE registers (so that the child thread can restore this state). See the FIXME comment; it's only a hack for now. Should be removed in future PRs...
___
*[LibOS/shim/src/sys/shim_clone.c, line 402 at r6](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKAdcMk1DSAr4921dTR:-MKAf_Tt3BtTOtolvj-c:b-6xgbwa) ([raw file](https://github.com/oscarlab/graphene/blob/b4d21052387297453ef876f2f2910829010ce329/LibOS/shim/src/sys/shim_clone.c#L402)):*

This touches memory pointed by `new_args.stack` before validity of that argument is checked (`lookup_vma` in `clone_implementaton_wrapper`).
</blockquote></details>

Yes, see the FIXME comment. I didn't want to introduce additional checks of correctness of this memory region, because this is wrong anyway and will be fixed in future PRs.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: 9 of 10 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/test/regression/Makefile, line 168 at r6](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKAe8HI3o5CWTWLP6b_:-MKiu6C-BjNGhmDaVxAa:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b4d21052387297453ef876f2f2910829010ce329/LibOS/shim/test/regression/Makefile#L168)):*

Just noticed: why it's `=`, not `+=`?
</blockquote></details>

Done.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

boryspoplawski (commented):
> Reviewed 9 of 10 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/src/shim_context-x86_64.c, line 130 at r7](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKjpTKL-CvUMf5UcBtt:-MKjpTKM4R5f2IpCSgVk:b-vkg63) ([raw file](https://github.com/oscarlab/graphene/blob/afc1f446f8a1699659fb522795ea20e1836d63f5/LibOS/shim/src/shim_context-x86_64.c#L130)):*
> ```C
>         uint64_t lmask = (uint64_t)-1;
>         uint64_t hmask = (uint64_t)-1;
>         __asm__ volatile("xrstor64 (%0)" :: "r"(xstate), "a"(lmask), "d"(hmask) : "memory");
> ```

We need to add all changed regisers (like `xmm0` to clobber list here or use gcc builtins for `xrstor`.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 132 at r7](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKjpeGBC4V9DBLsGJid:-MKjpeGBC4V9DBLsGJie:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/afc1f446f8a1699659fb522795ea20e1836d63f5/LibOS/shim/src/shim_context-x86_64.c#L132)):*
> ```C
>         __asm__ volatile("xrstor64 (%0)" :: "r"(xstate), "a"(lmask), "d"(hmask) : "memory");
>     } else {
>         __asm__ volatile("fxrstor64 (%0)" :: "r"(xstate) : "memory");
> ```

ditto
___
*[LibOS/shim/test/regression/fp_multithread.c, line 22 at r7](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKjq6ldB2dax7eqIerD:-MKjq6ldB2dax7eqIerE:bioep0f) ([raw file](https://github.com/oscarlab/graphene/blob/afc1f446f8a1699659fb522795ea20e1836d63f5/LibOS/shim/test/regression/fp_multithread.c#L22)):*
> ```C
> }
> 
> static void* thread_fp(void* arg) {
> ```

`arg` is not used (which should cause a compiler warrning), either add `(void)arg;` or simply return it from this function.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: 8 of 11 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/src/shim_context-x86_64.c, line 130 at r7](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKjpTKL-CvUMf5UcBtt:-MKnZEj57gWznyhUcAeU:bw886pq) ([raw file](https://github.com/oscarlab/graphene/blob/afc1f446f8a1699659fb522795ea20e1836d63f5/LibOS/shim/src/shim_context-x86_64.c#L130)):*

We need to add all changed regisers (like `xmm0` to clobber list here or use gcc builtins for `xrstor`.
</blockquote></details>

Done. Good catch. Couldn't find a good way to specify that all "extended state" registers (not only XMM/YMM/ZMM but also exotic ones like BND0..BND3) are clobbered in inline asm, so switched to GCC builtins. This topic seems to be a gray area, couldn't find any assurance that the GCC builtins correctly clobber all possible "extended state" regs...

But anyway, these xsave/xrtor will only be used during app <-> Graphene transitions, so shouldn't be a problem in the future when we move them to appropriate places.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 132 at r7](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKjpeGBC4V9DBLsGJid:-MKn_-YO0fGy-gkkgQzn:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/afc1f446f8a1699659fb522795ea20e1836d63f5/LibOS/shim/src/shim_context-x86_64.c#L132)):*

ditto
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/fp_multithread.c, line 22 at r7](https://reviewable.io/reviews/oscarlab/graphene/1875#-MKjq6ldB2dax7eqIerD:-MKjshgP4MH4e53iuW-a:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/afc1f446f8a1699659fb522795ea20e1836d63f5/LibOS/shim/test/regression/fp_multithread.c#L22)):*

`arg` is not used (which should cause a compiler warrning), either add `(void)arg;` or simply return it from this function.
</blockquote></details>

Done.

dimakuv:
Jenkins, retest Jenkins-18.04 please (`fork11` LTP test failed, known issue)

dimakuv:
Jenkins, retest Jenkins please (`fork11` LTP test failed, known issue)

dimakuv:
Jenkins, retest Jenkins-SGX-apps please (landed on a dying Jenkins worker)

boryspoplawski (dismissed):
> Reviewed 3 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

boryspoplawski (approved):
> Reviewed 4 of 4 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv:
Jenkins, retest Jenkins-SGX-apps please (landed on dying Jenkins worker)

dimakuv:
Jenkins, retest Jenkins-SGX please (landed on a dying Jenkins worker, seems to have messed with output)

mkow (approved):
> Reviewed 3 of 3 files at r8, 4 of 4 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: :shipit: complete! all files reviewed, all discussions resolved

yamahata (approved):
> Reviewed 4 of 10 files at r6, 3 of 3 files at r8, 4 of 4 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1875)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #895, opened 2019-08-05
Reorganize the repository


boryspoplawski:
Currently this repository has a bit messy layout. Proposed changes:
1. Move .h files into "include" directories (now there are symlinks pointing to src/.h)
2. Move shared code (between Pal and LibOS) out of the Pal directory
   ...
Please suggest more

donporter:
3. Consistent naming in the SGX PAL for code that runs in and out of the enclave
4. PAL/common/unix for common code to any unix host.

dimakuv:
5. LibOS/shim/test/regression should be moved to LibOS/regression (similar to Pal/regression)

yamahata:
6. Pal/src/start.S should be moved to appropriate place. It's not a part of PAL. but startup routine for Pal/{test, regression}
7. sort out compile environment.
   Pal is compiled for Pal target. i.e. Linux, Linux-SGX, FreeBSD
   LibOS is always compiled for Linux.
   Currently it's intermixed.

woju:
8. Rename top-level directories lower-case first letter. Capital-cased should be only "meta" things, like Makefile, README, Documentation, LICENCE etc. That way it is sorted correctly: meta together and actual code together.

boryspoplawski:
8. Use appropriate types, e.g. size_t for sizes (instead of int). If a function needs to return size or an error, use output parameter.

mkow:
> 3. Consistent naming in the SGX PAL for code that runs in and out of the enclave

Including both function and file names. An example of confusing naming: `enclave_entry.S` contains code which runs inside enclave, `sgx_entry.S` contains code which runs outside of it.

> 5. LibOS/shim/test/regression should be moved to LibOS/regression (similar to Pal/regression)

I'd prefer `LibOS/test/regression` (and moving all other tests under `LibOS/test`).

9. Rename `regression` tests to just `tests` (or move them one dir up, or find a better name). Currently it seems that all tests which were written by us just land there, regardless of their real type, so this name is misleading.

10. Why our LibOS code resides in `shim` directory? Doesn't `shim` mean something different than "the whole LibOS implementation"?

11. Why `Scripts` and `Tools` are different folders? Also, `Tools` just contains GSCE, without even a separate directory inside.

boryspoplawski:
12. Change name of chroot filesystem

woachk:
13. Using Clang + LLD with -target, this will allow building the LibOS directly on macOS or FreeBSD.

[...]

PR #1876, opened 2020-10-09
Basic benchmarking

woju:
## Description of the changes 

Add basic benchmarking (for now, `helloworld` and writing some megabytes to `/dev/null`) and framework for them.
Including this in Jenkins is left for another PR.

`asv` framework is capable of following changes between commits, but this is currently disabled, because `asv` is strongly tied to Python (i.e., builds projects using `setuptools` and installs with `pip` in `virtualenv`s, or, even worse, `conda`). I believe there is a possiblity to eventually get this working, but this could be done after this as well. 

## How to test this PR? 

Run the benchmarks as described in documentation (also included).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)


woju:
@mkow

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRepiN3lMnteAP-Vsl:-MJRepiO-gU_-RncJMhS:bowiwpi) (no related file):*
General comments and questions:
1. ASV seems to be tailored for Python apps. I couldn't find any info on support for compiled apps. For example, `asv profile` seems to be Python-only. Can we assume that ASV won't be too tightly-coupled with Python in the future?
2. Currently your ASV scripts are for SGX only. Can we also have runs for non-SGX Graphene?
3. How will ASV be incorporated into Jenkins? We'll need to run ASV on exactly the same machines, with exactly the same performance characteristics? Otherwise ASV-detected regressions will be false positives.
4. I don't like `graphene-root/benchmarks/` directory. We currently have LibOS/Linux programs tests under `graphene-root/LibOS/shim/test/` which is also not a particularly subdir. But we should either move the new `benchmarks` there, or move the current tests to an upper-level dir like `graphene-root/test`.

---
___
*[asv.conf.json, line 34 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRgu2aDsBfq-KKm4Ei:-MJRgu2aDsBfq-KKm4Ej:b-e4ezxf) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/asv.conf.json#L34)):*
> ```JSON
>         "openssl genrsa -3 -out Pal/src/host/Linux-SGX/signer/enclave-key.pem 3072",
>         "make -j8 SGX=1 WERROR=1 ISGX_DRIVER_PATH=/opt/intel/linux-sgx-driver",
>         "make -j8 SGX=1 WERROR=1 ISGX_DRIVER_PATH=/opt/intel/linux-sgx-driver sgx-tokens",
> ```

Can ASV read from env variables? Hard-coding `/opt/intel/linux-sgx-driver` will lead to discrepancies in the future.
___
*[benchmarks/benchmarks.py, line 7 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRh5l81-PyS45j4S5n:-MJRh5l81-PyS45j4S5o:b-wqwngm) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/benchmarks.py#L7)):*
> ```Python
> 
> def setup():
>     os.environ['SGX'] = '1'
> ```

It would be good to test both non-SGX Graphene and SGX Graphene.
___
*[benchmarks/helloworld.c, line 3 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRh9MW7Gjr1dDQJ1TC:-MJRh9MW7Gjr1dDQJ1TD:bq3b07r) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/helloworld.c#L3)):*
> ```C
> int main()
> {
> 	return 0;
> ```

In your C files, you mix tabs with spaces. We only use 4-space indents, could you fix in all your C files please?
___
*[benchmarks/Makefile, line 8 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRhGCF3y9XNBWOhLYq:-MJRhGCF3y9XNBWOhLYr:b-2wil35) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/Makefile#L8)):*
> ```Makefile
> 	 write_1e7
> GRAPHENEDIR = $(abspath $(dir $(lastword $(MAKEFILE_LIST)))/.. )
> signer = ../Pal/src/host/Linux-SGX/signer
> ```

Why relpaths instead of `$(GRAPHENEDIR)/Pal/src ...`?
___
*[benchmarks/manifest.in, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRhSCLA5uBOVZATZbn:-MJRhSCLA5uBOVZATZbo:bgkkvjy) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/manifest.in#L1)):*
> ```
> #sgx.enable_stats = 1
> ```

We typically call such files `manifest.template`. I'd stick to our current naming scheme.
___
*[benchmarks/manifest.in, line 3 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRhXsr8lhZlsY2Yoz6:-MJRhXsr8lhZlsY2Yoz7:b-t9qfgc) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/manifest.in#L3)):*
> ```
> #sgx.enable_stats = 1
> 
> loader.preload = file:@GRAPHENEDIR@/Runtime/libsysdb.so
> ```

We typically use the bash-like syntax for variables: `$(GRAPHENEDIR)/...`
___
*[benchmarks/write_pages.c, line 14 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRhv3J2igw4eFt14N2:-MJRhv3J2igw4eFt14N3:b1jtzrg) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/write_pages.c#L14)):*
> ```C
> 
> int main(int argc, char *argv[])
> {
> ```

We have slightly different rules for C files as you use here: https://graphene.readthedocs.io/en/latest/devel/coding-style.html#code-formatting
___
*[benchmarks/write_pages.c, line 20 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRiIbi539vP6oywbOV:-MJRiIbi539vP6oywbOW:bhj2aqb) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/write_pages.c#L20)):*
> ```C
> 	int fd, i, ret;
> 
>     ret = -1;
> ```

We typically return `1` instead of `-1/255`

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: 7 of 11 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRepiN3lMnteAP-Vsl:-MJSMryWE9MJBRVFweR-:bdsdxb3) (no related file):*

General comments and questions:
1. ASV seems to be tailored for Python apps. I couldn't find any info on support for compiled apps. For example, `asv profile` seems to be Python-only. Can we assume that ASV won't be too tightly-coupled with Python in the future?
2. Currently your ASV scripts are for SGX only. Can we also have runs for non-SGX Graphene?
3. How will ASV be incorporated into Jenkins? We'll need to run ASV on exactly the same machines, with exactly the same performance characteristics? Otherwise ASV-detected regressions will be false positives.
4. I don't like `graphene-root/benchmarks/` directory. We currently have LibOS/Linux programs tests under `graphene-root/LibOS/shim/test/` which is also not a particularly subdir. But we should either move the new `benchmarks` there, or move the current tests to an upper-level dir like `graphene-root/test`.
</blockquote></details>

1. Yes, this is a significant problem, but could be overcome. Here is an example where it was adapted to installing arbitrary nix packages: https://github.com/Warbo/asv-nix/blob/master/python/asv_nix/__init__.py and I intend to steal from this example left and right. This is also the reason why the build functionality is currently disabled (env set to "existing" and python set to "same"), because otherwise this wants to setup new virtualenv and everything breaks.
Yes, `asv profile` is probably python-specific, as are `mem_` functions and probably other features. I don't think we need all of them.
If you know of better tool, I'm open to suggestions. For me, the killer features are graphing and automatic bisecting: https://asv.readthedocs.io/en/stable/using.html#finding-a-commit-that-produces-a-large-regression
2. This would mean we would have to build both of them at the same time, which is incompatible with how we run Jenkins right now.
3. For the reasons you stated I don't believe we should run this on Jenkins for other reasons than to show that the script won't error out. Theoretically this could be done with very specific setup (dedicated node with semaphore so multiple jobs won't run together), but in practice I don't think this is really viable. But I think it would still be convenient to include the benchmarks in the repo, so people can run the same benchmarks, all of them on their respective machines.
4. This is a matter of setting benchmark_dir in `asv.conf.json`, but I'd have to research if the conf file itself could be moved. This would probably have to wait until I come up with the solution to building, which will probably affect the outcome.

---
___
*[asv.conf.json, line 34 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRgu2aDsBfq-KKm4Ei:-MJSPEJjDtT99lVZ8Flp:b-ffy5il) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/asv.conf.json#L34)):*

Can ASV read from env variables? Hard-coding `/opt/intel/linux-sgx-driver` will lead to discrepancies in the future.
</blockquote></details>

I believe it could, but then you would have to remember to always put that variable (or we could even write in shell: `${ISGX_DRIVER_PATH:-/opt/intel/linux-sgx-driver}`, but this is already hardcoded somewhere, (or `enclave-key.pem` path, I don't remember), so I wouldn't want to address this now, and instead as part of the meson rewrite. In fact, this part is what pushed me to revive that buildsystem effort.
___
*[benchmarks/benchmarks.py, line 7 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRh5l81-PyS45j4S5n:-MJSS3UbBzdYKDQhzuKX:bb0z0gs) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/benchmarks.py#L7)):*

It would be good to test both non-SGX Graphene and SGX Graphene.
</blockquote></details>

Eventually yes, but then we would have to build them both. Maybe this could be arranged, but not as part of this PR.
___
*[benchmarks/helloworld.c, line 3 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRh9MW7Gjr1dDQJ1TC:-MJSVw5b5o76b6fomj1n:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/helloworld.c#L3)):*

In your C files, you mix tabs with spaces. We only use 4-space indents, could you fix in all your C files please?
</blockquote></details>

Done.
___
*[benchmarks/Makefile, line 8 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRhGCF3y9XNBWOhLYq:-MJSVyUhDmXK4M5ynf2Q:b51qgik) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/Makefile#L8)):*

Why relpaths instead of `$(GRAPHENEDIR)/Pal/src ...`?
</blockquote></details>

No specific reason, and `..` is shorter. Here it can be used, in manifest not, because those are run from a temporary directory in `/tmp`.
___
*[benchmarks/write_pages.c, line 14 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRhv3J2igw4eFt14N2:-MJSWfOX2VBKqwDIrUaF:bsi8v3l) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/write_pages.c#L14)):*

We have slightly different rules for C files as you use here: https://graphene.readthedocs.io/en/latest/devel/coding-style.html#code-formatting
</blockquote></details>

Sorry, my bad.
___
*[benchmarks/write_pages.c, line 20 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRiIbi539vP6oywbOV:-MJSWnYF0pgJI3-xgO1o:bftq442) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/write_pages.c#L20)):*

We typically return `1` instead of `-1/255`
</blockquote></details>

Be my guest.
___
*[benchmarks/manifest.in, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRhSCLA5uBOVZATZbn:-MJSWQ_x8h78pfPza5Gn:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/manifest.in#L1)):*

We typically call such files `manifest.template`. I'd stick to our current naming scheme.
</blockquote></details>

Done.
___
*[benchmarks/manifest.in, line 3 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRhXsr8lhZlsY2Yoz6:-MJSWTI95omt8EaqqUXB:bd33n33) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/manifest.in#L3)):*

We typically use the bash-like syntax for variables: `$(GRAPHENEDIR)/...`
</blockquote></details>

Can this remain as-is? `$()` sub is kind of hard to do in Makefile.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRepiN3lMnteAP-Vsl:-MJVn5-CC3gXUedwwyWi:b-3ymy1v) (no related file):*

1. Yes, this is a significant problem, but could be overcome. Here is an example where it was adapted to installing arbitrary nix packages: https://github.com/Warbo/asv-nix/blob/master/python/asv_nix/__init__.py and I intend to steal from this example left and right. This is also the reason why the build functionality is currently disabled (env set to "existing" and python set to "same"), because otherwise this wants to setup new virtualenv and everything breaks.
Yes, `asv profile` is probably python-specific, as are `mem_` functions and probably other features. I don't think we need all of them.
If you know of better tool, I'm open to suggestions. For me, the killer features are graphing and automatic bisecting: https://asv.readthedocs.io/en/stable/using.html#finding-a-commit-that-produces-a-large-regression
2. This would mean we would have to build both of them at the same time, which is incompatible with how we run Jenkins right now.
3. For the reasons you stated I don't believe we should run this on Jenkins for other reasons than to show that the script won't error out. Theoretically this could be done with very specific setup (dedicated node with semaphore so multiple jobs won't run together), but in practice I don't think this is really viable. But I think it would still be convenient to include the benchmarks in the repo, so people can run the same benchmarks, all of them on their respective machines.
4. This is a matter of setting benchmark_dir in `asv.conf.json`, but I'd have to research if the conf file itself could be moved. This would probably have to wait until I come up with the solution to building, which will probably affect the outcome.
</blockquote></details>

Thanks for replies! I'll leave this comment blocking until you resolve (4).

---
___
*[benchmarks/write_pages.c, line 13 at r2](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJVmVZd38dhEtAUw1yb:-MJVmVZd38dhEtAUw1yc:by8pkvu) ([raw file](https://github.com/oscarlab/graphene/blob/cfc000020bb194626176e23adfc8b4e5228fcf28/benchmarks/write_pages.c#L13)):*
> ```C
> #include <unistd.h>
> 
> int main(int argc, char *argv[]) {
> ```

Nitpick: we also use stick-to-left asterisks: `char* argv[]`
___
*[benchmarks/manifest.in, line 3 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRhXsr8lhZlsY2Yoz6:-MJVlrm29ffMga2U0HRx:b-y3bgsu) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/manifest.in#L3)):*

Can this remain as-is? `$()` sub is kind of hard to do in Makefile.
</blockquote></details>

OK, now I see that it's also used in Meson.
___
*[benchmarks/manifest.template, line 18 at r2](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJVmOXOCGjfM2ugjFEj:-MJVmOXOCGjfM2ugjFEk:bewegue) ([raw file](https://github.com/oscarlab/graphene/blob/cfc000020bb194626176e23adfc8b4e5228fcf28/benchmarks/manifest.template#L18)):*
> ```
> sgx.trusted_files.libm = file:@GRAPHENEDIR@/Runtime/libm.so.6
> sgx.trusted_files.libpthread = file:@GRAPHENEDIR@/Runtime/libpthread.so.0
> #sgx.trusted_files.libgcc_s = file:$(ARCH_LIBDIR)/libgcc_s.so.1
> ```

Can we remove these two lines? They use the old-style string replacement.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: 2 of 11 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRepiN3lMnteAP-Vsl:-MJWdYtoEk-z3Emo8N7W:b-jlj4l2) (no related file):*

Thanks for replies! I'll leave this comment blocking until you resolve (4).
</blockquote></details>

OK, I moved everything under `tests/, however if this wouldn't work out with building, I'll move `asv.conf.json` into root directory again.

Also fixed some other minor problems. I think this is ready for another round.

---
___
*[asv.conf.json, line 34 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRgu2aDsBfq-KKm4Ei:-MJY1uA6DG-EawrBFLnc:bmam9nu) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/asv.conf.json#L34)):*

I believe it could, but then you would have to remember to always put that variable (or we could even write in shell: `${ISGX_DRIVER_PATH:-/opt/intel/linux-sgx-driver}`, but this is already hardcoded somewhere, (or `enclave-key.pem` path, I don't remember), so I wouldn't want to address this now, and instead as part of the meson rewrite. In fact, this part is what pushed me to revive that buildsystem effort.
</blockquote></details>

I removed all this, because with `'existing'` env this is no-op.
___
*[benchmarks/Makefile, line 8 at r1](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRhGCF3y9XNBWOhLYq:-MJY2Aq43_1AKkJnxIUZ:bxcx82z) ([raw file](https://github.com/oscarlab/graphene/blob/540ca5de34d60903aec891e06f0f66f864fc7735/benchmarks/Makefile#L8)):*

No specific reason, and `..` is shorter. Here it can be used, in manifest not, because those are run from a temporary directory in `/tmp`.
</blockquote></details>

Did `$(GRAPHENEDIR)` anyway when moving to `benchmarks/`.
___
*[benchmarks/manifest.template, line 18 at r2](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJVmOXOCGjfM2ugjFEj:-MJWdfYG53Wep7BteT1v:b-dx8gw) ([raw file](https://github.com/oscarlab/graphene/blob/cfc000020bb194626176e23adfc8b4e5228fcf28/benchmarks/manifest.template#L18)):*

Can we remove these two lines? They use the old-style string replacement.
</blockquote></details>

Yeah, too much copy-paste and not enough attention.
___
*[benchmarks/write_pages.c, line 13 at r2](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJVmVZd38dhEtAUw1yb:-MJWdlZi0jQ2KcBo2-ee:b-q16c27) ([raw file](https://github.com/oscarlab/graphene/blob/cfc000020bb194626176e23adfc8b4e5228fcf28/benchmarks/write_pages.c#L13)):*

Nitpick: we also use stick-to-left asterisks: `char* argv[]`
</blockquote></details>

I'm happy to oblige.

woju (commented):
I rebased this on top of current master and made some changes which were necessary after loader rework and esp. removal of `loader.exec` option. Now I'm running the suite again, for the commits between old and new merge-base. I'll report if something is broken.

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876#-:-MKez1gu-sgvXFNQBJYs:bapys1z)* status: 0 of 14 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: 0 of 17 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "WIP" found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1876#-MJRepiN3lMnteAP-Vsl:-MKkvLuy8xVDx-I6nZJB:b-lwrrmi) (no related file):*

OK, I moved everything under `tests/, however if this wouldn't work out with building, I'll move `asv.conf.json` into root directory again.

Also fixed some other minor problems. I think this is ready for another round.
</blockquote></details>

OK, another week, another round. I rebased this (notably on top of loader rework). Added nginx benchmarks.

---

mkow (changes requested):
> Reviewed 4 of 11 files at r3, 16 of 16 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "WIP" found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[tests/benchmarks/__init__.py, line 1 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0GWlt5FxQSvrwlJuA:-ML0GWlt5FxQSvrwlJuB:b-70y4tt) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/__init__.py#L1)):*
> ```Python
> # Copyright (c) 2020 Intel Corporation, Wojtek Porczyk <woju@invisiblethingslab.com>
> ```

Please include SPDX + reformat the copyright to match the format in other our files.
Ditto for other sources in this PR.
___
*[tests/benchmarks/http-root/Makefile, line 3 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0PhyD5vS_Rcarkznv:-ML0PhyEAO3iwqlDDM08:b-r82ci0) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/http-root/Makefile#L3)):*
> ```Makefile
> .PHONY: all
> all: 10K.1.html
> 10K.1.html:
> ```

please add a newline above this line
___
*[tests/benchmarks/http-root/Makefile, line 5 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0Pjy78y3QfCljqon-:-ML0Pjy78y3QfCljqon0:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/http-root/Makefile#L5)):*
> ```Makefile
> 10K.1.html:
> 	dd if=/dev/urandom of=$@ bs=1000 count=10
> .PHONY: clean
> ```

ditto
___
*[tests/asv.conf.json, line 40 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0DZaf0tAAfAOUvsXm:-ML0DZagCowY7jp5ynqW:b-ff6z9g) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/asv.conf.json#L40)):*
> ```JSON
>     // (for git) or "default" (for mercurial).
>     // "branches": ["master"], // for git
>     // "branches": ["default"],    // for mercurial
> ```

weird alignment of the comment
___
*[tests/asv.conf.json, line 42 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0EQ6C8qr62vJUvOV6:-ML0EQ6C8qr62vJUvOV7:b3svwc) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/asv.conf.json#L42)):*
> ```JSON
>     // "branches": ["default"],    // for mercurial
> 
>     // The tool to use to create environments.  May be "conda",
> ```

We decided in some other place to not use double spaces between sentences. Ditto for this whole file.
OTOH, was this just copied from somewhere? If so, could you comment about this at the top?
___
*[tests/asv.conf.json, line 54 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0F64V0qlWR61dXeug:-ML0F64V0qlWR61dXeuh:bc5g2gf) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/asv.conf.json#L54)):*
> ```JSON
> 
>     // the base URL to show a commit for the project.
>     "show_commit_url": "http://github.com/oscarlab/graphene/commit/",
> ```

http -> https
___
*[tests/benchmarks/.gitignore, line 7 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0GR7D3ZB-AohvRCY2:-ML0GR7D3ZB-AohvRCY3:b-2rwqjx) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/.gitignore#L7)):*
> ```ignore
> *.token
> helloworld
> write_1e[4567]
> ```

here and previous line: should start with `/`
___
*[tests/benchmarks/basic.manifest.template, line 19 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0Jif40oN1uI_RFFoH:-ML0Jif52O9h9iseXX9K:bjh70dp) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/basic.manifest.template#L19)):*
> ```
> sgx.trusted_files.libpthread = file:@GRAPHENEDIR@/Runtime/libpthread.so.0
> 
> #sgx.allow_file_creation = 1
> ```

This option was removed recently
___
*[tests/benchmarks/write_pages.c, line 21 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0O9gW3BoasISSUFMo:-ML0O9gW3BoasISSUFMp:b-w7suwk) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/write_pages.c#L21)):*
> ```C
> 
> int main(int argc, char* argv[]) {
>     void *buf;
> ```

`void *` -> `void* `

dimakuv (changes requested):
> Reviewed 3 of 11 files at r3, 14 of 16 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: all files reviewed, 20 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "WIP" found in commit messages' one-liners (waiting on @mkow and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7FB0Z5wUYyhmtzl5C:-ML7FB0Z5wUYyhmtzl5D:b-gbcd3v) (no related file):*
We have the convention (at least at the top level of our repo) to start directory names with a capital letter. Could you rename `tests/` to `Tests/`? I know this convention is rather arbitrary, but let's stick to it for now.

---
___
*[tests/benchmarks/basic.py, line 22 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7IEx2EplHJYpn7uJd:-ML7HiICA8KCbV87dV9v:b-pl3otc) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/basic.py#L22)):*
> ```Python
>     write_pages = Exec('write_pages', manifest_template='basic.manifest.template')
>     params = [10000, 100000, 1000000, 10000000]
>     param_namess = ['pagecount']
> ```

Typo: must be `param_names`. Why does it even work in ASV?
___
*[tests/benchmarks/makeenv.py, line 41 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7JKkj4RjjCYHN9OXo:-ML7JKkk2jQHONV77gNz:b-i08iz7) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/makeenv.py#L41)):*
> ```Python
>     def _run(self, *args, env=None, **kwds):
>         # when we don't need asv.util.check_*, do not use, because it may deadlock
>         # (those function have problems with their poor reimplementation of .communicate()
> ```

`function` -> `functions`. Plus add a closing bracket.

To be honest, I don't understand what happens here :) Well, as long as it works
___
*[tests/benchmarks/makeenv.py, line 61 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7K5hO8Nx82HER99zC:-ML7K5hO8Nx82HER99zD:bwlrn65) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/makeenv.py#L61)):*
> ```Python
>         for cmd in self._build_command:
>             self._run(cmd, shell=True, cwd=build_dir, check=True,
>                 timeout=self._install_timeout)
> ```

Is it meaningful to set an "install" timeout here during the build? Builds may take much longer than installs.
___
*[tests/benchmarks/makeenv.py, line 70 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7JiCe7e8nHahDJxBk:-ML7JiCe7e8nHahDJxBl:bdn14eu) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/makeenv.py#L70)):*
> ```Python
>             return
> 
>         for cmd in self._uninstall_command:
> ```

I guess it should be `self._install_command`
___
*[tests/benchmarks/nginx.py, line 1 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7LR0sDJBaLyDpa31t:-ML7LR0t2CqcDlrbec7I:bkp2ttm) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/nginx.py#L1)):*
> ```Python
> import os
> ```

Could you add the same header as for `redis.py`? And add the descriptions of the available envvars here: `AB` and `NGINX`.
___
*[tests/benchmarks/redis.py, line 11 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7LjGK2CuOBAXkiDy6:-ML7LjGK2CuOBAXkiDy7:b-14731m) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/redis.py#L11)):*
> ```Python
> 
> .. envvar:: REDIS_SERVER_ARGS
>     Arguments to redis-server. Defaults to ``--save '' --protected-mode no``.
> ```

I don't see `REDIS_SERVER_ARGS` in your PR. Please remove this and update description of `REDIS_SERVER`.
___
*[tests/benchmarks/redis.py, line 12 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7LLdLD45KxLKtLlX9:-ML7LLdLD45KxLKtLlXA:b-ytesgd) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/redis.py#L12)):*
> ```Python
> .. envvar:: REDIS_SERVER_ARGS
>     Arguments to redis-server. Defaults to ``--save '' --protected-mode no``.
> '''
> ```

What about `REDIS_BENCHMARK`? Do you want to add it in this description?
___
*[tests/benchmarks/redis.py, line 44 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7MFQA8yzyTi5k7t8q:-ML7MFQA8yzyTi5k7t8r:b-lturzs) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/redis.py#L44)):*
> ```Python
>         'RPOP', 'SADD', 'HSET', 'SPOP', 'LPUSH', 'LRANGE_100', 'LRANGE_300',
>         'LRANGE_500', 'LRANGE_600', 'MSET',
>     ]
> ```

Don't you want to add `param_names = ('test')` here and use `test` instead of `t`?
___
*[tests/benchmarks/redis.py, line 45 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7MClJAtDtnQOKL1iB:-ML7MClJAtDtnQOKL1iC:bxerphf) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/redis.py#L45)):*
> ```Python
>         'LRANGE_500', 'LRANGE_600', 'MSET',
>     ]
>     unit = 'requests/s'
> ```

This is not used
___
*[tests/benchmarks/.gitignore, line 5 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7Ffu51bSMWEJOT2fP:-ML7Ffu51bSMWEJOT2fQ:b5ucgz9) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/.gitignore#L5)):*
> ```ignore
> *.manifest.sgx
> *.sig
> *.token
> ```

Why do we need `*.manifest.sgx`, `*.sig`, `*.token` lines here? We already have them in the top-level gitignore.
___
*[tests/benchmarks/basic.manifest.template, line 19 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0Jif40oN1uI_RFFoH:-ML7Ifu0DYzSxINVBWW1:b-bj9fca) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/basic.manifest.template#L19)):*

This option was removed recently
</blockquote></details>

+1. Please remove it completely.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: 7 of 17 files reviewed, 20 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7FB0Z5wUYyhmtzl5C:-ML9M-yt6uWiXHk49Tlu:b-9n45bh) (no related file):*

We have the convention (at least at the top level of our repo) to start directory names with a capital letter. Could you rename `tests/` to `Tests/`? I know this convention is rather arbitrary, but let's stick to it for now.
</blockquote></details>

Those should be lowercase, which is a better convention, because meta-things (like Documentation, Makefile etc) should be sorted together and things like code (tests) should be separate.

https://github.com/oscarlab/graphene/issues/895#issuecomment-518721538

I wanted to do this right on the first take, because this will be used by other teams, so I don't want to change this later.

---
___
*[tests/benchmarks/__init__.py, line 1 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0GWlt5FxQSvrwlJuA:-ML9Q9oUBBalL_ioNlP1:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/__init__.py#L1)):*

Please include SPDX + reformat the copyright to match the format in other our files.
Ditto for other sources in this PR.
</blockquote></details>

Done.
___
*[tests/benchmarks/basic.py, line 22 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7IEx2EplHJYpn7uJd:-ML9QKiB8xMLxCfKbIMD:b-8b4g3p) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/basic.py#L22)):*

Typo: must be `param_names`. Why does it even work in ASV?
</blockquote></details>

Because this is optional.
___
*[tests/benchmarks/makeenv.py, line 41 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7JKkj4RjjCYHN9OXo:-ML9QyWw-G_RCJsTvm6M:b-q4pcqv) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/makeenv.py#L41)):*

`function` -> `functions`. Plus add a closing bracket.

To be honest, I don't understand what happens here :) Well, as long as it works
</blockquote></details>

The authors of `asv` doomed themselves to rediscover this big red warning about deadlocks: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.stderr.

Yes, managing subprocesses and their stdio is non-trivial under POSIX.
___
*[tests/benchmarks/makeenv.py, line 61 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7K5hO8Nx82HER99zC:-ML9RRob7hbrHZW9FCtc:b-mhnqym) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/makeenv.py#L61)):*

Is it meaningful to set an "install" timeout here during the build? Builds may take much longer than installs.
</blockquote></details>

There is no other timeout in asv, and also this is how it is done in other environments.
___
*[tests/benchmarks/makeenv.py, line 70 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7JiCe7e8nHahDJxBk:-ML9ReX-BQ3urlsgEjkE:b-iwy3y3) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/makeenv.py#L70)):*

I guess it should be `self._install_command`
</blockquote></details>

Yes, too much copy-paste. We don't use it, so this was really no-op.
___
*[tests/benchmarks/nginx.py, line 1 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7LR0sDJBaLyDpa31t:-ML9Tn3Q1EOfLGLnK40r:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/nginx.py#L1)):*

Could you add the same header as for `redis.py`? And add the descriptions of the available envvars here: `AB` and `NGINX`.
</blockquote></details>

Done.
___
*[tests/benchmarks/redis.py, line 11 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7LjGK2CuOBAXkiDy6:-ML9To_438t1GNJ76yAz:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/redis.py#L11)):*

I don't see `REDIS_SERVER_ARGS` in your PR. Please remove this and update description of `REDIS_SERVER`.
</blockquote></details>

Done.
___
*[tests/benchmarks/redis.py, line 12 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7LLdLD45KxLKtLlX9:-ML9Tpdj0zgXJFK9n6b2:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/redis.py#L12)):*

What about `REDIS_BENCHMARK`? Do you want to add it in this description?
</blockquote></details>

Done.
___
*[tests/benchmarks/redis.py, line 44 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7MFQA8yzyTi5k7t8q:-ML9VRGm3tBmZvVCX23L:b-u2atxv) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/redis.py#L44)):*

Don't you want to add `param_names = ('test')` here and use `test` instead of `t`?
</blockquote></details>

Yeah, sure.
___
*[tests/benchmarks/redis.py, line 45 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7MClJAtDtnQOKL1iB:-ML9Tsax41h8FmfBUFTf:b9z36j3) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/redis.py#L45)):*

This is not used
</blockquote></details>

It very much is: https://asv.readthedocs.io/en/stable/benchmarks.html#tracking-benchmarks
___
*[tests/benchmarks/http-root/Makefile, line 3 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0PhyD5vS_Rcarkznv:-ML9XLDg-GZz8sgDuitQ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/http-root/Makefile#L3)):*

please add a newline above this line
</blockquote></details>

Done.
___
*[tests/benchmarks/http-root/Makefile, line 5 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0Pjy78y3QfCljqon-:-ML9XLpH42Rsnn6BiCIm:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/http-root/Makefile#L5)):*

ditto
</blockquote></details>

Done.
___
*[tests/asv.conf.json, line 40 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0DZaf0tAAfAOUvsXm:-ML9Midq5sZPpDqW-o1-:b6ogdar) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/asv.conf.json#L40)):*

weird alignment of the comment
</blockquote></details>

This is from template (`asv quickstart`).
___
*[tests/asv.conf.json, line 42 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0EQ6C8qr62vJUvOV6:-ML9MqvL5ohB7Zy75sZH:beixcje) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/asv.conf.json#L42)):*

We decided in some other place to not use double spaces between sentences. Ditto for this whole file.
OTOH, was this just copied from somewhere? If so, could you comment about this at the top?
</blockquote></details>

`asv quickstart`. I don't believe this needs a comment.
___
*[tests/asv.conf.json, line 54 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0F64V0qlWR61dXeug:-ML9XN2Y707nbplbkt_r:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/asv.conf.json#L54)):*

http -> https
</blockquote></details>

Done.
___
*[tests/benchmarks/.gitignore, line 5 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7Ffu51bSMWEJOT2fP:-ML9QEBm-zW0XlMGkASe:bss4a2b) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/.gitignore#L5)):*

Why do we need `*.manifest.sgx`, `*.sig`, `*.token` lines here? We already have them in the top-level gitignore.
</blockquote></details>

Removed.
___
*[tests/benchmarks/.gitignore, line 7 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0GR7D3ZB-AohvRCY2:-ML9QB7Y0Q_O69eMpnyU:b-bqw94j) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/.gitignore#L7)):*

here and previous line: should start with `/`
</blockquote></details>

Yes, and write_1e* is also wrong.
___
*[tests/benchmarks/basic.manifest.template, line 19 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0Jif40oN1uI_RFFoH:-ML9QYav91sDa_5Binyj:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/basic.manifest.template#L19)):*

+1. Please remove it completely.
</blockquote></details>

Done.
___
*[tests/benchmarks/write_pages.c, line 21 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML0O9gW3BoasISSUFMo:-ML9XD5KAlf02cofYi6y:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/write_pages.c#L21)):*

`void *` -> `void* `
</blockquote></details>

Done.

mkow (changes requested):
> Reviewed 10 of 10 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: all files reviewed, 18 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @woju)

___
*[tests/benchmarks/__init__.py, line 2 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAhesDCChxFIe_0jUB:-MLAhesDCChxFIe_0jUC:b-pqucik) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/__init__.py#L2)):*
> ```Python
> # SPDX-License-Identifier: LGPL-3.0-or-later
> # Copyright (c) 2020 Intel Corporation,
> ```

No comma at the end
___
*[tests/benchmarks/basic.py, line 2 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAhiJ86kGXHCFxbt-K:-MLAhiJ86kGXHCFxbt-L:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/basic.py#L2)):*
> ```Python
> # SPDX-License-Identifier: LGPL-3.0-or-later
> # Copyright (c) 2020 Intel Corporation,
> ```

ditto
___
*[tests/benchmarks/nginx.py, line 2 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAiEwmEzREKUcexPVT:-MLAiEwmEzREKUcexPVU:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/nginx.py#L2)):*
> ```Python
> # SPDX-License-Identifier: LGPL-3.0-or-later
> # Copyright (c) 2020 Intel Corporation,
> ```

ditto
___
*[tests/benchmarks/nginx.py, line 10 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAiO3C44hffQtOfA8a:-MLAiO3C44hffQtOfA8b:b-dtva08) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/nginx.py#L10)):*
> ```
> three environment variables
> ```

not two?
I'd say "the following" instead, will age better :)
___
*[tests/benchmarks/redis.py, line 2 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAijeR1ctA4ZnKxRFC:-MLAijeR1ctA4ZnKxRFD:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/redis.py#L2)):*
> ```Python
> # SPDX-License-Identifier: LGPL-3.0-or-later
> # Copyright (c) 2020 Intel Corporation,
> ```

ditto
___
*[tests/benchmarks/redis.py, line 10 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAioDy-QU1uTUHymBY:-MLAioDy-QU1uTUHymBZ:bmlltlt) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/redis.py#L10)):*
> ```
> three environment variables
> ```

ditto
___
*[tests/benchmarks/write_pages.c, line 3 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAjgbt8a1rcSlGfaNk:-MLAjgbt8a1rcSlGfaNl:b-7g6akx) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/write_pages.c#L3)):*
> ```C
> /*
>  * SPDX-License-Identifier: LGPL-3.0-or-later
>  * Copyright (c) 2020 Intel Corporation,
> ```

Please make it 1-1 with other files (no comma, no empty line at the beginning, first line in a separate comment block)

dimakuv (changes requested):
> Reviewed 10 of 10 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7FB0Z5wUYyhmtzl5C:-MLCQ1pm0wcZYzWJc3f6:b2nhktp) (no related file):*

Those should be lowercase, which is a better convention, because meta-things (like Documentation, Makefile etc) should be sorted together and things like code (tests) should be separate.

https://github.com/oscarlab/graphene/issues/895#issuecomment-518721538

I wanted to do this right on the first take, because this will be used by other teams, so I don't want to change this later.
</blockquote></details>

OK.

---
___
*[tests/benchmarks/redis.py, line 45 at r4](https://reviewable.io/reviews/oscarlab/graphene/1876#-ML7MClJAtDtnQOKL1iB:-MLCRB8k27kPABpsPhLs:b2nhktp) ([raw file](https://github.com/oscarlab/graphene/blob/7ed6cf1ec39087a88274bd1174bea4d71b0554ab/tests/benchmarks/redis.py#L45)):*

It very much is: https://asv.readthedocs.io/en/stable/benchmarks.html#tracking-benchmarks
</blockquote></details>

OK.
___
*[tests/benchmarks/redis.py, line 16 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLCQvqF8xtFSgnWRYb6:-MLCQvqF8xtFSgnWRYb7:b-11ytnd) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/redis.py#L16)):*
> ```Python
> 
> .. envvar:: REDIS_BENCHMARK
>     Command executed to redis-benchmark. Defaults to ``redis-benchmark``. You can also try
> ```

`to` -> `as`

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: 12 of 17 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

___
*[tests/benchmarks/__init__.py, line 2 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAhesDCChxFIe_0jUB:-MLDrVyK-YASLO5rWtPt:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/__init__.py#L2)):*

No comma at the end
</blockquote></details>

Done.
___
*[tests/benchmarks/basic.py, line 2 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAhiJ86kGXHCFxbt-K:-MLDrZJy9V6RmvvQIYgv:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/basic.py#L2)):*

ditto
</blockquote></details>

Done.
___
*[tests/benchmarks/nginx.py, line 2 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAiEwmEzREKUcexPVT:-MLDrayuB4_5o2X5CXyt:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/nginx.py#L2)):*

ditto
</blockquote></details>

Done.
___
*[tests/benchmarks/nginx.py, line 10 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAiO3C44hffQtOfA8a:-MLDrfBP8nqIIi9b2RRp:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/nginx.py#L10)):*

> ```
> three environment variables
> ```

not two?
I'd say "the following" instead, will age better :)
</blockquote></details>

Done.
___
*[tests/benchmarks/redis.py, line 2 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAijeR1ctA4ZnKxRFC:-MLDrlcgCZ-dG7Vugl-Q:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/redis.py#L2)):*

ditto
</blockquote></details>

Done.
___
*[tests/benchmarks/redis.py, line 10 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAioDy-QU1uTUHymBY:-MLDrmGY6yTLwgruaEYB:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/redis.py#L10)):*

> ```
> three environment variables
> ```

ditto
</blockquote></details>

Done.
___
*[tests/benchmarks/redis.py, line 16 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLCQvqF8xtFSgnWRYb6:-MLDrpmlD6v8kL3Y703u:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/redis.py#L16)):*

`to` -> `as`
</blockquote></details>

Done.
___
*[tests/benchmarks/write_pages.c, line 3 at r5](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLAjgbt8a1rcSlGfaNk:-MLDwiuqEncer-5D4nop:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/716b7208e7c27fbe12384116fcc03c46a73c9cdc/tests/benchmarks/write_pages.c#L3)):*

Please make it 1-1 with other files (no comma, no empty line at the beginning, first line in a separate comment block)
</blockquote></details>

Done.

mkow (approved):
> Reviewed 5 of 5 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

dimakuv (approved):
> Reviewed 5 of 5 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLNRcvL8lFJRbQseRdN:-MLNRcvL8lFJRbQseRdO:b61hokj) (no related file):*
Jenkins fails:
```
Unable to create directory /leeroy/.pylint.d
Unable to create file /leeroy/.pylint.d/test_pal1.stats: [Errno 2] No such file or directory: '/leeroy/.pylint.d/test_pal1.stats'
MESSAGES AFFECTING THIS PR:
tests/benchmarks/makeenv.py +8:0: import-error: Unable to import 'asv.environment'
tests/benchmarks/makeenv.py +9:0: import-error: Unable to import 'asv.util'
tests/benchmarks/makeenv.py +10:0: import-error: Unable to import 'asv.console'
```

@woju  Since you don't add ASV installation in in Jenkins dockerfiles, I guess you should simply disable pylint on `tests/benchmarks` scripts?

---

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: 17 of 19 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1876#-MLNRcvL8lFJRbQseRdN:-MLNfAfxEM9d2PWpmfRK:b-ihtif6) (no related file):*

Jenkins fails:
```
Unable to create directory /leeroy/.pylint.d
Unable to create file /leeroy/.pylint.d/test_pal1.stats: [Errno 2] No such file or directory: '/leeroy/.pylint.d/test_pal1.stats'
MESSAGES AFFECTING THIS PR:
tests/benchmarks/makeenv.py +8:0: import-error: Unable to import 'asv.environment'
tests/benchmarks/makeenv.py +9:0: import-error: Unable to import 'asv.util'
tests/benchmarks/makeenv.py +10:0: import-error: Unable to import 'asv.console'
```

@woju  Since you don't add ASV installation in in Jenkins dockerfiles, I guess you should simply disable pylint on `tests/benchmarks` scripts?
</blockquote></details>

I added to dockerfiles. We'll probably need to run *something* to check that benchmarks still work, but I haven't done this, because this would add 30 min to job execution time.

---

dimakuv (approved):
> Reviewed 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> Reviewed 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1876)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #760, opened 2019-06-18
Build system plans


HW42:
Recently there has been some work on the Makefiles (like PR #732 and PR #759). Before spending too much time on details of the current Makefiles, it would be good to decide how the build system should look like in the mid-term (For example, IIRC, @mkow would like to migrate to cmake at some point).

donporter:
I don't personally have strong feelings here.  I'm just trying to get things working as needed, and happy to defer to whoever wants to drive the work.

dimakuv:
I concur with Don. I am actually fine with the current Makefile build system (especially with the recent PRs). I am more concerned with Python scripts `pal-sgx-get-token` and `pal-sgx-sign` -- their interface is ugly and they are buried so deep that most people don't recognize how to sign apps out-of-tree. I also dislike our current source tree (`lib/` dir is under Pal, Glibc is under `LibOS/shim`, regression tests are in different subdirs in LibOS and Pal).

chiache:
For the long term, I do wish to move on to a more portable building system such as cmake. It might be too much work to do before our stable release deadline. For the time being, we should explore as many options as possible and compare the pros and cons.

mkow:
I'd like to migrate to CMake, but rather somewhen in the future (next year maybe?), as we have much bigger issues now. So, for now, let's keep our Makefiles in a good condition, because this seems to be much less work and less possible problems than rewriting everything to CMake.

Also, +1 to the issues @dimakuv pointed out.

yamahata:
+1 for cmake.
Given that the migration won't happen soon, so I'm fixing my pain point in short term.

My wish list are
- out-of-source tree build
- build configuration
  "make DEBUG=1 SGX=1" isn't a way for long term.
  For multiple version support of glibc (or musl libc as optional), this would also help.
  People don't want to build all of the version.
- dependency generation
  Yeah, hw42 is resolving this. Great work!
- package script
  packaging script for .deb (or .rpm)
  e.g. "make deb" to generate .deb package.
  I'd like to tell users, "just install package" instead of "follow this install instruction."
- docker image
  we can generate docker image with something like "make docker-image".
  and we can host pre-build docker image so that casual user can download it and play graphene without hassle.

mkow:
+1 to Isaku's wishlist, especially I'd like to see out-of-tree builds.
btw. do you have any estimate how hard that migration to CMake would be and what could potentially cause troubles? (other than just generic pain of changing everything and checking whether all things still work)

[...]

PR #1879, opened 2020-10-10
Add first meson.build for installing

woju:
## Description of the changes 

This adds a possibility to actually install graphene in the system, and make packages for distributions.

This also might be a beginning of switching build system to meson. See #760 for a discussion of alternative build system, which I hope to revive.

## How to test this PR? 

Compile as usual with and without SGX (this is important that both variants are built together). Prepare some `helloworld` executable with manifest, signatures etc.

After that:
```sh
sudo apt install ninja-build

# can be replaced with `sudo apt install meson` on Debian since buster,
# but not on Ubuntu 18.04, where meson is too old
python3 -m pip install meson

# pick some suitable directory; if --prefix <path> is omitted,
# will install to /usr/local
meson setup build --prefix "$HOME"/tmp

ninja -C build # currently no-op, this is where any compilation would happen
ninja -C build install

export PATH="$HOME"/tmp/bin:"$PATH"  # or wherever you chose the prefix

cd "$HOME"/my-example-programs # with manifests and signatures and tokens already prepared
graphene ./helloworld.manifest
graphene-sgx ./helloworld.manifest.sgx
```

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+/build
+
 # No editor backup files.
 *.sw*
 *.backup
--- a/Runtime/meson.build
+++ b/Runtime/meson.build
@@ -1,0 +1,44 @@
+install_data('../Pal/src/host/Linux/libpal.so',
+    install_dir: pkglibexecdir / 'linux')
+install_subdir('../Pal/src/host/Linux/gdb_integration',
+    install_dir: pkglibexecdir / 'linux')
+
+install_data(
+    '../Pal/src/host/Linux-SGX/pal-sgx',
+    '../Pal/src/host/Linux-SGX/libpal.so',
+    install_dir: pkglibexecdir / 'linux-sgx')
+install_subdir('../Pal/src/host/Linux-SGX/gdb_integration',
+    install_dir: pkglibexecdir / 'linux-sgx')
+
+conf = configuration_data()
+conf.set_quoted('IN_GIT', '')
+
+hostpalpath_linux = prefix / pkglibexecdir / 'linux'
+conf_graphene = configuration_data()
+conf_graphene.merge_from(conf)
+conf_graphene.set('SGX', 0)
+conf_graphene.set_quoted('HOST_PAL_PATH', hostpalpath_linux)
+conf_graphene.set_quoted('LIBPAL_PATH', hostpalpath_linux / 'libpal.so')
+conf_graphene.set_quoted('PAL_CMD', hostpalpath_linux / 'libpal.so')
+
+configure_file(
+    input: 'pal_loader',
+    output: 'graphene',
+    configuration: conf_graphene,
+    install_dir: get_option('bindir'),
+)
+
+hostpalpath_linux_sgx = prefix / pkglibexecdir / 'linux-sgx'
+conf_graphene_sgx = configuration_data()
+conf_graphene_sgx.merge_from(conf)
+conf_graphene_sgx.set('SGX', 1)
+conf_graphene_sgx.set_quoted('HOST_PAL_PATH', hostpalpath_linux_sgx)
+conf_graphene_sgx.set_quoted('LIBPAL_PATH', hostpalpath_linux_sgx / 'libpal.so')
+conf_graphene_sgx.set_quoted('PAL_CMD', hostpalpath_linux_sgx / 'pal-sgx')
+
+configure_file(
+    input: 'pal_loader',
+    output: 'graphene-sgx',
+    configuration: conf_graphene_sgx,
+    install_dir: get_option('bindir'),
+)
--- a/Runtime/pal_loader
+++ b/Runtime/pal_loader
@@ -4,61 +4,79 @@
 # Copyright (C) 2019 Invisible Things Lab
 # Copyright (C) 2020 Intel Corporation
 #                    Michał Kowalczyk <mkow@invisiblethingslab.com>
+#                    Wojtek Porczyk <woju@invisiblethingslab.com>
 
-while :
-do
-	case "$1" in
-		"SGX")
-			SGX=1
-			export SGX
-			;;
-		"GDB")
-			GDB=1
-			;;
-		*)
-			break
-			;;
-	esac
-	shift
-done
 
-RUNTIME_DIR=$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")
-if [ -z "$PAL_HOST" ]; then
-    if ! command -v make >/dev/null; then
-        libpal="$RUNTIME_DIR/libpal-*.so"
-        libpal="$(echo -n "$libpal")"
-        libpal="${libpal//$RUNTIME_DIR\//}"
-        if [ "$libpal" = 'libpal-*.so' ]; then
-            echo "Unable to detect PAL_HOST. Please install the make program."
-            exit 1
+# This is how we detect whether we're installed or running from inside git repo.
+# If unchanged (i.e., if this reads IN_GIT between "at" characters), this is an
+# ordinary string and test -n "$IN_GIT" is true. But when installing, this is
+# rewritten using meson's configure_file() function to an empty string and the
+# else branch is taken. See also: ./meson.build and
+# https://mesonbuild.com/Reference-manual.html#configure_file.
+IN_GIT=@IN_GIT@
+
+if test -n "$IN_GIT"
+then
+    while :
+    do
+        case "$1" in
+            "SGX")
+                SGX=1
+                export SGX
+                ;;
+            "GDB")
+                GDB=1
+                ;;
+            *)
+                break
+                ;;
+        esac
+        shift
+    done
+
+    RUNTIME_DIR=$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")
+    if [ -z "$PAL_HOST" ]; then
+        if ! command -v make >/dev/null; then
+            libpal="$RUNTIME_DIR/libpal-*.so"
+            libpal="$(echo -n "$libpal")"
+            libpal="${libpal//$RUNTIME_DIR\//}"
+            if [ "$libpal" = 'libpal-*.so' ]; then
+                echo "Unable to detect PAL_HOST. Please install the make program."
+                exit 1
+            fi
+
+            array=("$libpal")
+            if [ ${#array[@]} -ne 1 ]; then
+                echo "Multiple libpal detected ($libpal). Please explicitly set the environment variable PAL_HOST."
+                exit 1
+            fi
+
+            PAL_HOST="${libpal%.so}"
+            PAL_HOST="${PAL_HOST#libpal-}"
+        else
+            PAL_HOST=$(make --no-print-directory --quiet -f "$RUNTIME_DIR/../Scripts/Makefile.configs" print_host 2>&1)
         fi
-
-        array=("$libpal")
-        if [ ${#array[@]} -ne 1 ]; then
-            echo "Multiple libpal detected ($libpal). Please explicitly set the environment variable PAL_HOST."
-            exit 1
-        fi
-
-        PAL_HOST="${libpal%.so}"
-        PAL_HOST="${PAL_HOST#libpal-}"
-    else
-        PAL_HOST=$(make --no-print-directory --quiet -f "$RUNTIME_DIR/../Scripts/Makefile.configs" print_host 2>&1)
     fi
+    PAL_CMD="$RUNTIME_DIR/pal-$PAL_HOST"
+    LIBPAL_PATH=$(realpath "$RUNTIME_DIR/libpal-$PAL_HOST.so")
+    HOST_PAL_PATH=$(realpath "$RUNTIME_DIR/../Pal/src/host/$PAL_HOST")
+else
+    PAL_CMD=@PAL_CMD@
+    LIBPAL_PATH=@LIBPAL_PATH@
+    HOST_PAL_PATH=@HOST_PAL_PATH@
+    SGX=@SGX@
 fi
 
 MANIFEST=
 ENVS=()
 PREFIX=()
-PAL_CMD="$RUNTIME_DIR/pal-$PAL_HOST"
-LIBPAL_PATH=$(realpath "$RUNTIME_DIR/libpal-$PAL_HOST.so")
-HOST_PAL_PATH=$(realpath "$RUNTIME_DIR/../Pal/src/host/$PAL_HOST")
 
 if [ "$GDB" == "1" ]; then
 	PREFIX=("gdb" "-q")
 	if [ -n "$INSIDE_EMACS" ]; then
 		PREFIX+=("-i=mi")
 	fi
-	if [ -v SGX ]; then
+	if [ 0"$SGX" -gt 0 ]; then
 		PREFIX+=("-x" "$HOST_PAL_PATH/gdb_integration/graphene_sgx_gdb.py")
 		ENVS+=("LD_PRELOAD=$HOST_PAL_PATH/gdb_integration/sgx_gdb.so:$LD_PRELOAD")
 	else
@@ -92,7 +110,7 @@ if [ "$MANIFEST" == "" ]; then
 fi
 
 if [ ! -f "$PAL_CMD" ]; then
-	echo "$PAL_CMD is not built"
+	echo "$PAL_CMD not found"
 	exit 1
 fi
 
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,11 @@
+project(
+    'graphene', 'c',
+    version: '1.0',
+    license: 'LGPLv3+',
+)
+
+prefix = get_option('prefix')
+pkglibexecdir = get_option('libexecdir') / meson.project_name()
+pkgdatadir = get_option('datadir') / meson.project_name()
+
+subdir('Runtime')


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

___
*[Runtime/pal_loader, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJRs1DC8zazs1u5rfAv:-MJRs1DC8zazs1u5rfAw:b-4gjuut) ([raw file](https://github.com/oscarlab/graphene/blob/c5349fe9a10a37d3d4bb359b5787d6fd1ed4e41b/Runtime/pal_loader#L9)):*
> ```
> #                    Wojtek Porczyk <woju@invisiblethingslab.com>
> 
> if test -n @IN_GIT@  # in meson, this gets zeroed out, so this will only take first branch in repo
> ```

Even with this comment, I don't understand what `IN_GIT` means... Could you give some more context?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[Runtime/pal_loader, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJRs1DC8zazs1u5rfAv:-MJSIsC6-y5VGI9M62ec:b-vwmjph) ([raw file](https://github.com/oscarlab/graphene/blob/c5349fe9a10a37d3d4bb359b5787d6fd1ed4e41b/Runtime/pal_loader#L9)):*

Even with this comment, I don't understand what `IN_GIT` means... Could you give some more context?
</blockquote></details>

This is a regular string in shell (`@` is not a special character), so `test -n @IN_GIT@` as it stands verbatim here is true (may I respectfuly remind you that `-n` stands for "nonzero", but really refers to the string length, not the integer value), so the first branch is taken. The first branch is unmodified from before this commit (just indented), so no functional change.

Meson has [configure_file](https://mesonbuild.com/Reference-manual.html#configure_file) function, which is intended for processing `.in` files and substituting things like `sed -e "s:@VARIABLE@:$VALUE:"` was done in other build systems. This is used here to substitute `@IN_GIT@` to empty value, but quoted (see `conf.set_quoted('IN_GIT', '')` line in `Runtime/meson.build`). Then in the resulting file (`graphene` or `graphene-sgx`) this reads `test -n ""`, which is false and the execution takes `else` branch.

The purpose is to have a single file which behaves both as it was before (this PR promises to not make changes and to not make meson mandatory) and also makes it possible to be installed without repository, make and gcc (which is currently invoked on every `pal_loader` invocation). For the second option I needed to hardcode the necessary variables (this is done with the same configure_file call) to absolute paths (`graphene[-sgx]` can be invoked from any directory in the system), so this have to behave differently whether this is run before or after installation.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[Runtime/pal_loader, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJRs1DC8zazs1u5rfAv:-MJSR-7U2o__2Kl_0gg-:b-4b5uz5) ([raw file](https://github.com/oscarlab/graphene/blob/c5349fe9a10a37d3d4bb359b5787d6fd1ed4e41b/Runtime/pal_loader#L9)):*

This is a regular string in shell (`@` is not a special character), so `test -n @IN_GIT@` as it stands verbatim here is true (may I respectfuly remind you that `-n` stands for "nonzero", but really refers to the string length, not the integer value), so the first branch is taken. The first branch is unmodified from before this commit (just indented), so no functional change.

Meson has [configure_file](https://mesonbuild.com/Reference-manual.html#configure_file) function, which is intended for processing `.in` files and substituting things like `sed -e "s:@VARIABLE@:$VALUE:"` was done in other build systems. This is used here to substitute `@IN_GIT@` to empty value, but quoted (see `conf.set_quoted('IN_GIT', '')` line in `Runtime/meson.build`). Then in the resulting file (`graphene` or `graphene-sgx`) this reads `test -n ""`, which is false and the execution takes `else` branch.

The purpose is to have a single file which behaves both as it was before (this PR promises to not make changes and to not make meson mandatory) and also makes it possible to be installed without repository, make and gcc (which is currently invoked on every `pal_loader` invocation). For the second option I needed to hardcode the necessary variables (this is done with the same configure_file call) to absolute paths (`graphene[-sgx]` can be invoked from any directory in the system), so this have to behave differently whether this is run before or after installation.
</blockquote></details>

Ah, ok. What I missed were two pieces of info:
1. `@IN_GIT@` is just a dummy that is replaced (by empty string) when run via Meson (`pal_loader` becomes the template for Meson, which produces the `graphene` script via sed-like replacements). The name `IN_GIT` is not very telling to me -- smth like `OLD_RUNTIME` would be more intuitive.
2. Meson requires variables to be surrounded with `@` for replacement.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[Runtime/pal_loader, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJRs1DC8zazs1u5rfAv:-MJS_qou98k2muHzM28q:bn9er8l) ([raw file](https://github.com/oscarlab/graphene/blob/c5349fe9a10a37d3d4bb359b5787d6fd1ed4e41b/Runtime/pal_loader#L9)):*

Ah, ok. What I missed were two pieces of info:
1. `@IN_GIT@` is just a dummy that is replaced (by empty string) when run via Meson (`pal_loader` becomes the template for Meson, which produces the `graphene` script via sed-like replacements). The name `IN_GIT` is not very telling to me -- smth like `OLD_RUNTIME` would be more intuitive.
2. Meson requires variables to be surrounded with `@` for replacement.
</blockquote></details>

Would you like this to be included in a comment somewhere?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[Runtime/pal_loader, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJRs1DC8zazs1u5rfAv:-MJSuWBHAnlhZSpgaTqc:ba5wqgq) ([raw file](https://github.com/oscarlab/graphene/blob/c5349fe9a10a37d3d4bb359b5787d6fd1ed4e41b/Runtime/pal_loader#L9)):*

Would you like this to be included in a comment somewhere?
</blockquote></details>

What about these:
- Add a link to the `configure_file()` Meson function (this cleared many of my questions immediately) somewhere in the docs/top-level comments
- Rename `IN_GIT` to something more telling like `OLD_RUNTIME`
- Rephrase your comment `in meson, this gets zeroed out, so this will only take first branch in repo` -> `after running through meson, IN_GIT gets zeroed out (i.e., the true branch is taken if run in git repo, false branch is taken in install)` -- smth like this

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: all files reviewed, 1 unresolved discussion (waiting on @woju)

___
*[Runtime/meson.build, line 2 at r1](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJT1YAs0Zhn5bmnHlJv:-MJT1YAs0Zhn5bmnHlJw:bvzkf7p) ([raw file](https://github.com/oscarlab/graphene/blob/c5349fe9a10a37d3d4bb359b5787d6fd1ed4e41b/Runtime/meson.build#L2)):*
> ```Meson
> install_data('../Pal/src/host/Linux/libpal.so',
>     install_dir: pkglibexecdir / 'linux')
> ```

nice. This is a starting point for distro package.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: 2 of 4 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Runtime/pal_loader, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJRs1DC8zazs1u5rfAv:-MJTCCpW2QHeOZ0Gz-a5:b-fcvai7) ([raw file](https://github.com/oscarlab/graphene/blob/c5349fe9a10a37d3d4bb359b5787d6fd1ed4e41b/Runtime/pal_loader#L9)):*

What about these:
- Add a link to the `configure_file()` Meson function (this cleared many of my questions immediately) somewhere in the docs/top-level comments
- Rename `IN_GIT` to something more telling like `OLD_RUNTIME`
- Rephrase your comment `in meson, this gets zeroed out, so this will only take first branch in repo` -> `after running through meson, IN_GIT gets zeroed out (i.e., the true branch is taken if run in git repo, false branch is taken in install)` -- smth like this
</blockquote></details>

Is this comment OK?

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJVoIH0Dzj1KA7Dldsz:-MJVoIH0Dzj1KA7Dldt-:b-bketds) (no related file):*
Minor: the last fixup commit refers to a wrong base commit ("Add benchmarks"). Correctly resolve this during squash-and-rebase.

---
___
*[Runtime/pal_loader, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJRs1DC8zazs1u5rfAv:-MJVnbgbBYQKiL779AoY:b-480tko) ([raw file](https://github.com/oscarlab/graphene/blob/c5349fe9a10a37d3d4bb359b5787d6fd1ed4e41b/Runtime/pal_loader#L9)):*

Is this comment OK?
</blockquote></details>

Yes, perfect for dummies like me :)

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJVoIH0Dzj1KA7Dldsz:-MJY3FFJ27JivPA9JKZi:b-xbk93u) (no related file):*

Minor: the last fixup commit refers to a wrong base commit ("Add benchmarks"). Correctly resolve this during squash-and-rebase.
</blockquote></details>

sorry, this was too much shell-fu

---
___
*[Runtime/pal_loader, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJRs1DC8zazs1u5rfAv:-MJY3dovDhWBgnoFPaV6:b-r4dz90) ([raw file](https://github.com/oscarlab/graphene/blob/c5349fe9a10a37d3d4bb359b5787d6fd1ed4e41b/Runtime/pal_loader#L9)):*

Yes, perfect for dummies like me :)
</blockquote></details>

You'll be not the only on wondering, how this polyglot works.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[.gitignore, line 1 at r2](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJbYDy884bfOil80s7c:-MJbYDy884bfOil80s7d:b-gtz9uz) ([raw file](https://github.com/oscarlab/graphene/blob/b939fd72cea61e86e211c68cb37a411accbd0eea/.gitignore#L1)):*
> ```ignore
> build/
> ```

should be `/build/`, otherwise it's a recursive pattern
___
*[meson.build, line 8 at r2](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJb_Ar97zRHcN3fzydb:-MJb_Ar97zRHcN3fzydc:b-50qudh) ([raw file](https://github.com/oscarlab/graphene/blob/b939fd72cea61e86e211c68cb37a411accbd0eea/meson.build#L8)):*
> ```
> pkglibexecdir
> ```

This is really hard to read without underscores, would be good to ad at least one after `pkg`.
___
*[Runtime/meson.build, line 22 at r2](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJb_kif3qFosOo7AGPO:-MJb_kif3qFosOo7AGPP:b-xtmopz) ([raw file](https://github.com/oscarlab/graphene/blob/b939fd72cea61e86e211c68cb37a411accbd0eea/Runtime/meson.build#L22)):*
> ```Meson
> conf_graphene.set_quoted('HOST_PAL_PATH', hostpalpath_linux)
> conf_graphene.set_quoted('LIBPAL_PATH', hostpalpath_linux / 'libpal.so')
> conf_graphene.set_quoted('PAL_CMD', hostpalpath_linux / 'libpal.so')
> ```

Why is this the same as `LIBPAL_PATH`?
___
*[Runtime/pal_loader, line 28 at r2](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJbadDZAbvJE5PsjZXf:-MJbadDZAbvJE5PsjZXg:bjgpewy) ([raw file](https://github.com/oscarlab/graphene/blob/b939fd72cea61e86e211c68cb37a411accbd0eea/Runtime/pal_loader#L28)):*
> ```
>                 ;;
>             "GDB")
>                 GDB=1
> ```

Why we don't support these options in the installed version?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: 3 of 4 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[.gitignore, line 1 at r2](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJbYDy884bfOil80s7c:-MJbfUYx4Yc8lbhpPiRJ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b939fd72cea61e86e211c68cb37a411accbd0eea/.gitignore#L1)):*

should be `/build/`, otherwise it's a recursive pattern
</blockquote></details>

Done.
___
*[meson.build, line 8 at r2](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJb_Ar97zRHcN3fzydb:-MJbfRlVCUT8KE3-tzvB:b-i0hnkr) ([raw file](https://github.com/oscarlab/graphene/blob/b939fd72cea61e86e211c68cb37a411accbd0eea/meson.build#L8)):*

> ```
> pkglibexecdir
> ```

This is really hard to read without underscores, would be good to ad at least one after `pkg`.
</blockquote></details>

This is a convention.
___
*[Runtime/meson.build, line 22 at r2](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJb_kif3qFosOo7AGPO:-MJbf5ewBJq9nHeZQSOi:b-eio7th) ([raw file](https://github.com/oscarlab/graphene/blob/b939fd72cea61e86e211c68cb37a411accbd0eea/Runtime/meson.build#L22)):*

Why is this the same as `LIBPAL_PATH`?
</blockquote></details>

You tell me. This is how non-sgx graphene works.
___
*[Runtime/pal_loader, line 28 at r2](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJbadDZAbvJE5PsjZXf:-MJbeS1A8AjzpLTZh5Ao:b-bon6ub) ([raw file](https://github.com/oscarlab/graphene/blob/b939fd72cea61e86e211c68cb37a411accbd0eea/Runtime/pal_loader#L28)):*

Why we don't support these options in the installed version?
</blockquote></details>

According to #1053, there are no such options.

`GDB=1 graphene-sgx whatever.manifest.sgx` still works.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[Runtime/meson.build, line 22 at r2](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJb_kif3qFosOo7AGPO:-MJc2yswE-wwNLYW9ZFZ:b-ric2ik) ([raw file](https://github.com/oscarlab/graphene/blob/b939fd72cea61e86e211c68cb37a411accbd0eea/Runtime/meson.build#L22)):*

You tell me. This is how non-sgx graphene works.
</blockquote></details>

0o
I guess only @chiache and @donporter know the answer :)
___
*[Runtime/pal_loader, line 28 at r2](https://reviewable.io/reviews/oscarlab/graphene/1879#-MJbadDZAbvJE5PsjZXf:-MJc3AGM9w8NZFEMA8_A:b-jwuorp) ([raw file](https://github.com/oscarlab/graphene/blob/b939fd72cea61e86e211c68cb37a411accbd0eea/Runtime/pal_loader#L28)):*

According to #1053, there are no such options.

`GDB=1 graphene-sgx whatever.manifest.sgx` still works.
</blockquote></details>

Ah, I missed the fact that envs still work. Then it's fine.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

yamahata (approved):
> Reviewed 1 of 4 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1879)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1883, opened 2020-10-11
[LibOS] Implement madvise + small syscall refactorings

### LINKED ISSUE #1873: [LibOS] Implement madvise
## Description of the problem

Some Linux test suites check if `madvise` syscall is implemented. I don't think implementing it has any use in supporting real-world applications (which are usually satisfied even if we return -ENOSYS from it), but adding it should be simple, as only `MADV_DONTNEED` mode is directly visible for the users and will make the tests suites happy :)

mkow:
## Description of the changes 

Changes:
* Some minor refactorings / cleanups / fixups (mostly for things which annoyed me when implementing madvise)
* madvise syscall implementation, supporting only `MADV_DONTNEED` and flags which can legally be no-ops.

See commit messages for more info.

Fixes #1873.

## How to test this PR? 

New test was added to LibOS regression, as LTP madvise tests aren't too useful for us.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)


mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: 0 of 150 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[LibOS/shim/include/shim_vma.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJOvIUg2i-t2PTJnrTF:-MJOvIUh7iomfAtUZ4pw:bbzjvh9) ([raw file](https://github.com/oscarlab/graphene/blob/f5dc5cadd65bc56508a9664d95d1826839d9147c/LibOS/shim/include/shim_vma.h#L26)):*
> ```objc
> /* TODO: split flags into internal (Graphene) and Linux; also to consider: completely remove Linux
>  * flags - we only need MAP_SHARED/MAP_PRIVATE and possibly MAP_STACK/MAP_GROWSDOWN */
> struct shim_vma {
> ```

@boryspoplawski, @dimakuv: Please take a look at this. I'm not sure about this design, but I don't have a better idea - we need to both be able to copy out the struct to external functions, and also be able to use the list directly when traversing under a lock.
___
*[Pal/src/host/Linux/db_events.c, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJP6hvaBhqevijIIOfs:-MJP6hvb3sm3N5DZai56:bhvx9ah) ([raw file](https://github.com/oscarlab/graphene/blob/3bcab01a0c0665d190b991b8823bdbbb6ed887ef/Pal/src/host/Linux/db_events.c#L5)):*
> ```C
> 
> /*
>  * db_event.c
> ```

example of a non-matching name (typo :) a lot of files are missing this "s" in this line)
___
*[Pal/src/host/Linux-SGX/elf-x86_64.h, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJP7EvF2RF759I48ZUo:-MJP7EvF2RF759I48ZUp:b-165dbd) ([raw file](https://github.com/oscarlab/graphene/blob/3bcab01a0c0665d190b991b8823bdbbb6ed887ef/Pal/src/host/Linux-SGX/elf-x86_64.h#L5)):*
> ```objc
> 
> /*
>  * dl-machine-x86_64.h
> ```

And here it completely mismatched the file...
___
*[Pal/src/host/Skeleton/db_rtld.c, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJP7Z3wCZulMm-gRDNN:-MJP7Z3xD9bkjRYMNC14:b-rnvbg) ([raw file](https://github.com/oscarlab/graphene/blob/3bcab01a0c0665d190b991b8823bdbbb6ed887ef/Pal/src/host/Skeleton/db_rtld.c#L9)):*
> ```C
>  * This file contains utilities to load ELF binaries into the memory
>  * and link them against each other.
>  * The source code in this file is imported and modified from the GNU C
> ```

plot twist: it wasn't

dimakuv (changes requested):
> Reviewed 146 of 150 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/include/shim_vma.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJOvIUg2i-t2PTJnrTF:-MJQwA7M9nMFqXvuGFEB:bsndidz) ([raw file](https://github.com/oscarlab/graphene/blob/f5dc5cadd65bc56508a9664d95d1826839d9147c/LibOS/shim/include/shim_vma.h#L26)):*

@boryspoplawski, @dimakuv: Please take a look at this. I'm not sure about this design, but I don't have a better idea - we need to both be able to copy out the struct to external functions, and also be able to use the list directly when traversing under a lock.
</blockquote></details>

By "this design" you mean defining `struct shim_vma` in this public header? I think it's fine. I am more curious what's the point of `struct shim_vma_info` now -- could we just remove it and use `shim_vma` instead everywhere, with the caveat that this is only a duplicate and not protected by lock? (in another PR)
___
*[LibOS/shim/src/sys/shim_clone.c, line 5 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQoXSV4qso57Ufxuir:-MJQoXSW8Zhpw4axFVwM:belkv59) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_clone.c#L5)):*
> ```C
> 
> /*
>  * Implementation of system call "clone" (using "clone" as "fork" is not implemented yet).
> ```

Outdated info in parenthesis (`using clone as fork ...`). Just remove it. Graphene supports fork/vfork via clone.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 374 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQrvkk8_U4BI_kJH6Q:-MJQrvkk8_U4BI_kJH6R:b-9st4y6) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L374)):*
> ```C
> 
>     if (vma->flags & (VMA_UNMAPPED | VMA_INTERNAL)) {
>         ctx->error = -EINVAL;
> ```

If we have say two "bad" VMAs, one writable file-backed mapping and one unmapped, then `ctx->error` will have the value of the last "bad" VMA, which results in `EINVAL`. But I think for understanding what went wrong in Graphene on madvise, it's better to have `ENOSYS` to take precedence over `EINVAL`. So I'd do like: `if (ctx->error != -ENOSYS) ctx->error = -EINVAL`.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 384 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQpWjrDT23MpgS9zJi:-MJQpWjrDT23MpgS9zJj:bdnzap4) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L384)):*
> ```C
> 
>     if (!(vma->prot & PROT_WRITE)) {
>         ctx->error = -ENOSYS; // Zeroing non-writable mappings is not yet implemented.
> ```

Why don't you restrict this case only for file-backed mappings (`&& vma->file`)? Why would we *not* support anonymous writable mappings?
___
*[LibOS/shim/src/sys/shim_mmap.c, line 390 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQq5L2BH0vX_Q0RsZQ:-MJQq5L2BH0vX_Q0RsZR:blxchuh) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L390)):*
> ```C
>     uintptr_t zero_start = MAX(ctx->begin, vma->begin);
>     uintptr_t zero_end = MIN(ctx->end, vma->end);
>     memset((void*)zero_start, 0, zero_end - zero_start);
> ```

Does memset work for file-backed mappings? IIRC, protected files for example don't have a "full range" mapping, but instead read from file on demand. If memset works even for files in Graphene, then no problem. But if it doesn't work, we should either ENOSYS on files altogether or do smth smarter here.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 429 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQrSQDDf1D_4fKL8Rc:-MJQrSQDDf1D_4fKL8Rd:b-em328z) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L429)):*
> ```C
>         case MADV_HWPOISON:
>         case MADV_DONTDUMP:
>         case MADV_DODUMP:
> ```

I would argue that `DONTDUMP` and `DODUMP` don't change semantics of the app and thus may be moved to "good" cases above.
___
*[LibOS/shim/test/regression/madvise.c, line 34 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQttxuAVrYrfkEpcvz:-MJQttxuAVrYrfkEpcw-:bfx504q) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L34)):*
> ```C
>     bool cleared[PAGES_CNT] = {false};
>     for (int i = 0; i < PAGES_CNT; i++) {
>         int perm_idx = ((i + 123) * 307) % PAGES_CNT; /* make the traverse non-sequential */
> ```

BTW, is this really a permutation? I don't know this math magic, could you share some link?
___
*[Pal/include/arch/x86_64/dl-machine.h, line 5 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQuXwF6slo4WZNXDvO:-MJQuXwK4xYqn2dgHWal:b-hdstai) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/Pal/include/arch/x86_64/dl-machine.h#L5)):*
> ```objc
> 
> /*
>  * This files contain architecture-specific implementation of ELF dynamic relocation functions.
> ```

`This file contains ...`
___
*[Pal/src/do-rel.h, line 5 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQv0dIBiujD51U4FaE:-MJQv0dIBiujD51U4FaF:b-7lw43g) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/Pal/src/do-rel.h#L5)):*
> ```objc
> 
> /*
>  * This files contain architecture-independent macros of ELF dynamic relocation function.
> ```

`This file contains`
___
*[Pal/src/host/Linux/db_threading.c, line 5 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQvOEh0q4ald7X_l9V:-MJQvOEh0q4ald7X_l9W:bct8lmu) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/Pal/src/host/Linux/db_threading.c#L5)):*
> ```C
> 
> /*
>  * This file contain APIs to create, exit and yield a thread.
> ```

`contains`
___
*[Pal/src/host/Skeleton/db_rtld.c, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJP7Z3wCZulMm-gRDNN:-MJQw21I9lGmBMeQHgLA:b32vhw) ([raw file](https://github.com/oscarlab/graphene/blob/3bcab01a0c0665d190b991b8823bdbbb6ed887ef/Pal/src/host/Skeleton/db_rtld.c#L9)):*

plot twist: it wasn't
</blockquote></details>

:)
___
*[Pal/src/host/Skeleton/db_threading.c, line 5 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQw5b4AzL59rmRFBtu:-MJQw5b4AzL59rmRFBtv:bct8lmu) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/Pal/src/host/Skeleton/db_threading.c#L5)):*
> ```C
> 
> /*
>  * This file contain APIs to create, exit and yield a thread.
> ```

`contains`

donporter:
Jenkins, retest Jenkins-SGX please.

boryspoplawski (changes requested):
> Reviewed 146 of 150 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, 22 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/include/shim_vma.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJOvIUg2i-t2PTJnrTF:-MJSEc0m2Hdx5fGSvNnD:b-7d6eb0) ([raw file](https://github.com/oscarlab/graphene/blob/f5dc5cadd65bc56508a9664d95d1826839d9147c/LibOS/shim/include/shim_vma.h#L26)):*

By "this design" you mean defining `struct shim_vma` in this public header? I think it's fine. I am more curious what's the point of `struct shim_vma_info` now -- could we just remove it and use `shim_vma` instead everywhere, with the caveat that this is only a duplicate and not protected by lock? (in another PR)
</blockquote></details>

I do not like this design, after all not exporting `shim_vma` was the sole purpose of defining this `struct shim_vma_info` - to hide it from outside world. I guess these are main reasons for this:
- to make this subsystem as modular as possible,
- to disallow modifications of `shim_vma` from outside of `shim_vma.c` file,
- to stop people from misusing the api, e.g. I think `traverse_vmas_in_range(..., function_that_sleeps)` has much higher chance of going through review, that if it was to be added to `shim_vma.c` - as `traverse_vmas_in_range` needs to hold a spinlock, it would be a bad idea.

My proposition: change this `traverse_vma` to `zero_memory_in_range` (or some better name).
___
*[LibOS/shim/src/shim_parser.c, line 1066 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJRhPcj-2Ay4JrZt0xl:-MJRhPckBl0hNBof8kvm:b-xvqud6) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/shim_parser.c#L1066)):*
> ```C
>             break;
>         default:
>             PRINTF("(unknown: %d)", behavior);
> ```

I would add a `break` here just for visual consistency (not blocking).
___
*[LibOS/shim/src/sys/shim_mmap.c, line 374 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQrvkk8_U4BI_kJH6Q:-MJSJ9MV2ikGGqZ0jUPV:b-h885h4) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L374)):*

If we have say two "bad" VMAs, one writable file-backed mapping and one unmapped, then `ctx->error` will have the value of the last "bad" VMA, which results in `EINVAL`. But I think for understanding what went wrong in Graphene on madvise, it's better to have `ENOSYS` to take precedence over `EINVAL`. So I'd do like: `if (ctx->error != -ENOSYS) ctx->error = -EINVAL`.
</blockquote></details>

FWIW: iteration should stop on the first error.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 390 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJSIn-t54rD9CSB1j-m:-MJSIn-t54rD9CSB1j-n:b-24lz11) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L390)):*
> ```C
>     uintptr_t zero_start = MAX(ctx->begin, vma->begin);
>     uintptr_t zero_end = MIN(ctx->end, vma->end);
>     memset((void*)zero_start, 0, zero_end - zero_start);
> ```

This seems like a bad idea: this memset could be huge and is done while holding vma spinlock.
The problem is I don't see any other way to do this ...
___
*[LibOS/shim/src/sys/shim_mmap.c, line 401 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJRy7Gl2yOk3r8OxFxR:-MJRy7Gm5SikIdTTyTKD:b-ij5ve7) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L401)):*
> ```C
> 
>     size_t len = ALIGN_UP(len_in, PAGE_SIZE);
>     if (len_in > 0 && len == 0)
> ```

Why not just check for `len_in > len`? I think it would be a bit cleaner and would show intentions behind this `if` immediately.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 415 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS-lS-3FV5V7h7Ud9f:-MJS-lS09ZdCUP2bKXDH:b-n3tvwk) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L415)):*
> ```C
>         case MADV_SEQUENTIAL:
>         case MADV_WILLNEED:
>         case MADV_FREE:
> ```

This might is not a no-op: `On a swapless system, freeing pages in a given range happens instantly, regardless of memory pressure.`
___
*[LibOS/shim/src/sys/shim_mmap.c, line 429 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQrSQDDf1D_4fKL8Rc:-MJS-3gH9ogYMcqKdHQJ:b26ecmm) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L429)):*

I would argue that `DONTDUMP` and `DODUMP` don't change semantics of the app and thus may be moved to "good" cases above.
</blockquote></details>

I think they do: we do not implement core dumps atm, but might very well do in the future.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 439 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS5OPx9wVVbx9CYbLW:-MJS5OPx9wVVbx9CYbLX:blt6d1u) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L439)):*
> ```C
>                 .error = 0,
>             };
>             bool is_continuous = traverse_vmas_in_range(start, start + len, dontneed_visitor,
> ```

This does not match the Linux behavior - it does not stop iteration on errors (as Linux does).
___
*[LibOS/shim/test/regression/madvise.c, line 21 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJSA0u50h0vJRP6uGm5:-MJSA0u50h0vJRP6uGm6:b-iau326) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L21)):*
> ```C
>     size_t page_size = getpagesize();
> 
>     void* m = mmap(NULL, PAGES_CNT * page_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS,
> ```

Addition on `void*` is not defined (well it is, but as a gcc extension), so please add appropriate casts or change to `char*`.
___
*[LibOS/shim/test/regression/madvise.c, line 23 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS5yUp8wBsOTwvWv_p:-MJS5yUp8wBsOTwvWv_q:b-ickflr) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L23)):*
> ```C
>     void* m = mmap(NULL, PAGES_CNT * page_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS,
>                    -1, 0);
>     if (!m)
> ```

`0`/`NULL` is a legal return value, you should check for `MAP_FAILED`.
___
*[LibOS/shim/test/regression/madvise.c, line 26 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS6YUp7NzQFehigsH0:-MJS6YUp7NzQFehigsH1:b-d6mn9x) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L26)):*
> ```C
>         err(1, "mmap()");
> 
>     for (int i = 0; i < PAGES_CNT; i++) {
> ```

Please use appropriate type for indexing (e.g. `size_t`).
___
*[LibOS/shim/test/regression/madvise.c, line 33 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS9uWs5Ot6ONIyfg-N:-MJS9uWs5Ot6ONIyfg-O:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L33)):*
> ```C
>     /* Clear pages one by one */
>     bool cleared[PAGES_CNT] = {false};
>     for (int i = 0; i < PAGES_CNT; i++) {
> ```

ditto
___
*[LibOS/shim/test/regression/madvise.c, line 34 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQttxuAVrYrfkEpcvz:-MJS8yuA929Ait4trM0b:bliptcr) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L34)):*

BTW, is this really a permutation? I don't know this math magic, could you share some link?
</blockquote></details>

`gcd(307, 128) == 1` hence `307` is a generator of `Z_128` (additive group of integers modulo `128`). `123` is just an arbitrary starting offset, I guess.
@mkow  `307 % 128 == 51`, so you could change that :)
___
*[LibOS/shim/test/regression/madvise.c, line 34 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS9xSF5sAcrL693u64:-MJS9xSF5sAcrL693u65:b-c3urec) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L34)):*
> ```C
>     bool cleared[PAGES_CNT] = {false};
>     for (int i = 0; i < PAGES_CNT; i++) {
>         int perm_idx = ((i + 123) * 307) % PAGES_CNT; /* make the traverse non-sequential */
> ```

ditto (indexing)
___
*[LibOS/shim/test/regression/madvise.c, line 37 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJSAfZO50X_BfkfvusQ:-MJSAfZPB1WIQwvnCxm0:b-1e74ye) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L37)):*
> ```C
>         void* curr = m + page_size * perm_idx;
> 
>         int res = madvise(curr, page_size, MADV_DONTNEED);
> ```

Maybe change 2nd argument to something smaller than `page_size` to check it works as well?
___
*[LibOS/shim/test/regression/madvise.c, line 43 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJSA_EF82mF7CamNoCZ:-MJSA_EF82mF7CamNoC_:b-c3urec) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L43)):*
> ```C
> 
>         /* Rescan the whole range and verify all magic values */
>         for (int j = 0; j < PAGES_CNT; j++) {
> ```

ditto (indexing)
___
*[Pal/src/host/Skeleton/db_rtld.c, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJP7Z3wCZulMm-gRDNN:-MJSDSsfDWu-LjdDcnl4:b-ry52oc) ([raw file](https://github.com/oscarlab/graphene/blob/3bcab01a0c0665d190b991b8823bdbbb6ed887ef/Pal/src/host/Skeleton/db_rtld.c#L9)):*

:)
</blockquote></details>

Are you sure? The (original) source file might have changed as well in the meantime, so the diff could be huge/full ...

mkow (commented):
> Reviewed 11 of 150 files at r1, 2 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: 135 of 150 files reviewed, 21 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[LibOS/shim/include/shim_vma.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJOvIUg2i-t2PTJnrTF:-MJU8mumAutkjcCxPNI4:b-gp7kbt) ([raw file](https://github.com/oscarlab/graphene/blob/f5dc5cadd65bc56508a9664d95d1826839d9147c/LibOS/shim/include/shim_vma.h#L26)):*

I do not like this design, after all not exporting `shim_vma` was the sole purpose of defining this `struct shim_vma_info` - to hide it from outside world. I guess these are main reasons for this:
- to make this subsystem as modular as possible,
- to disallow modifications of `shim_vma` from outside of `shim_vma.c` file,
- to stop people from misusing the api, e.g. I think `traverse_vmas_in_range(..., function_that_sleeps)` has much higher chance of going through review, that if it was to be added to `shim_vma.c` - as `traverse_vmas_in_range` needs to hold a spinlock, it would be a bad idea.

My proposition: change this `traverse_vma` to `zero_memory_in_range` (or some better name).
</blockquote></details>

I'd like to use this `traverse_vmas_in_range` to later implement e.g. `_vma_bkeep_change` with it and deduplicate the logic of traversing through all VMAs in some range.

I could remove the public version, but then the problem with exporting `zero_memory_in_range` is that it's not really zeroing memory, but madvise implementation which is quite weird (e.g. it should restore file mappings to the original state).

If you want more protection from external code modifying VMAs I can make the visitor parameter a pointer to const struct, but that would require either a proxy function or an ugly function cast (the internal version should stay non-const to be reusable internally).
___
*[LibOS/shim/src/shim_parser.c, line 1066 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJRhPcj-2Ay4JrZt0xl:-MJU6rBhEpnKQT1tbuV-:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/shim_parser.c#L1066)):*

I would add a `break` here just for visual consistency (not blocking).
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_clone.c, line 5 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQoXSV4qso57Ufxuir:-MJU775v6zg993NN8CnC:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_clone.c#L5)):*

Outdated info in parenthesis (`using clone as fork ...`). Just remove it. Graphene supports fork/vfork via clone.
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 374 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQrvkk8_U4BI_kJH6Q:-MJUHZAl7BN9Cb_s5I9n:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L374)):*

FWIW: iteration should stop on the first error.
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 384 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQpWjrDT23MpgS9zJi:-MJUHeTr1FaTkkkrOMzc:b-cexwk7) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L384)):*

Why don't you restrict this case only for file-backed mappings (`&& vma->file`)? Why would we *not* support anonymous writable mappings?
</blockquote></details>

Notice the `!` in front. This is for _non_-writable mappings, for which Linux madvise still works, but it would be harder to implement in Graphene.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 390 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJSIn-t54rD9CSB1j-m:-MJUI6in-PxVPmVYX0w4:b-hiqc53) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L390)):*

This seems like a bad idea: this memset could be huge and is done while holding vma spinlock.
The problem is I don't see any other way to do this ...
</blockquote></details>

Yeah, I don't think there's a better way. I hope software is not overusing madvise with this flag...
___
*[LibOS/shim/src/sys/shim_mmap.c, line 390 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQq5L2BH0vX_Q0RsZQ:-MJUIMwI1_Gez7D0uFDC:b-yl2lk0) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L390)):*

Does memset work for file-backed mappings? IIRC, protected files for example don't have a "full range" mapping, but instead read from file on demand. If memset works even for files in Graphene, then no problem. But if it doesn't work, we should either ENOSYS on files altogether or do smth smarter here.
</blockquote></details>

I think it would work, but the semantics of madvise is different - file mappings should be restored to the original, not zeroed. I corrected the bug, see now (now it's no-op for non-tainted file mappings).
___
*[LibOS/shim/src/sys/shim_mmap.c, line 401 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJRy7Gl2yOk3r8OxFxR:-MJUJOJu-eTeiD7KyKRU:b-ovwaw7) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L401)):*

Why not just check for `len_in > len`? I think it would be a bit cleaner and would show intentions behind this `if` immediately.
</blockquote></details>

Good idea.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 415 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS-lS-3FV5V7h7Ud9f:-MJUK8l81zKpCwTj2XG_:bcwsnew) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L415)):*

This might is not a no-op: `On a swapless system, freeing pages in a given range happens instantly, regardless of memory pressure.`
</blockquote></details>

Not really, it's prepended with "In Linux before version 4.12,".
___
*[LibOS/shim/src/sys/shim_mmap.c, line 429 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQrSQDDf1D_4fKL8Rc:-MJUKOj22vbS8QZaAJgN:b-q1qqo5) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L429)):*

I think they do: we do not implement core dumps atm, but might very well do in the future.
</blockquote></details>

Yup, I agree with Borys.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 439 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS5OPx9wVVbx9CYbLW:-MJUKUEzA6CTyjPLbubA:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L439)):*

This does not match the Linux behavior - it does not stop iteration on errors (as Linux does).
</blockquote></details>

Fixed.
___
*[LibOS/shim/test/regression/madvise.c, line 21 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJSA0u50h0vJRP6uGm5:-MJULcM3Eic-_8y6BqtK:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L21)):*

Addition on `void*` is not defined (well it is, but as a gcc extension), so please add appropriate casts or change to `char*`.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/madvise.c, line 23 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS5yUp8wBsOTwvWv_p:-MJULeCj8YU9Y8n67HEx:b4z3q8e) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L23)):*

`0`/`NULL` is a legal return value, you should check for `MAP_FAILED`.
</blockquote></details>

Oops, right.
___
*[LibOS/shim/test/regression/madvise.c, line 26 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS6YUp7NzQFehigsH0:-MJULuq88sV_OFwR-Dgw:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L26)):*

Please use appropriate type for indexing (e.g. `size_t`).
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/madvise.c, line 33 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS9uWs5Ot6ONIyfg-N:-MJULxRf2UyTrGAs88u5:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L33)):*

ditto
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/madvise.c, line 34 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQttxuAVrYrfkEpcvz:-MJUM7uN3Dk6P4jM8wqH:b-54glvo) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L34)):*
Borys' explanation is correct :)

> `123` is just an arbitrary starting offset, I guess.

Yup. Just to not start at the first page.

> 307 % 128 == 51, so you could change that :)

Right, I simplified it.
___
*[LibOS/shim/test/regression/madvise.c, line 34 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS9xSF5sAcrL693u64:-MJUMl1z2DuD8oL2Okwz:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L34)):*

ditto (indexing)
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/madvise.c, line 37 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJSAfZO50X_BfkfvusQ:-MJUYvHA30KHcuVsHCIQ:b-47yzv) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L37)):*

Maybe change 2nd argument to something smaller than `page_size` to check it works as well?
</blockquote></details>

Check now.
___
*[LibOS/shim/test/regression/madvise.c, line 43 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJSA_EF82mF7CamNoCZ:-MJUNz6u2V3MCr_XmTvW:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L43)):*

ditto (indexing)
</blockquote></details>

Done.
___
*[Pal/include/arch/x86_64/dl-machine.h, line 5 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQuXwF6slo4WZNXDvO:-MJU7cWf9fgoE_QtXvx3:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/Pal/include/arch/x86_64/dl-machine.h#L5)):*

`This file contains ...`
</blockquote></details>

Done.
___
*[Pal/src/do-rel.h, line 5 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQv0dIBiujD51U4FaE:-MJU7hvMAqZecit4YoC8:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/Pal/src/do-rel.h#L5)):*

`This file contains`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_threading.c, line 5 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQvOEh0q4ald7X_l9V:-MJU7ldW3uq1OO0eCMcF:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/Pal/src/host/Linux/db_threading.c#L5)):*

`contains`
</blockquote></details>

Done.
___
*[Pal/src/host/Skeleton/db_rtld.c, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJP7Z3wCZulMm-gRDNN:-MJU8BYj-AzV-CN1rBC_:b3kre) ([raw file](https://github.com/oscarlab/graphene/blob/3bcab01a0c0665d190b991b8823bdbbb6ed887ef/Pal/src/host/Skeleton/db_rtld.c#L9)):*

Are you sure? The (original) source file might have changed as well in the meantime, so the diff could be huge/full ...
</blockquote></details>

?
___
*[Pal/src/host/Skeleton/db_threading.c, line 5 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQw5b4AzL59rmRFBtu:-MJU8L2qARuWfoUNuCvb:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/Pal/src/host/Skeleton/db_threading.c#L5)):*

`contains`
</blockquote></details>

Done.

dimakuv (approved):
> Reviewed 15 of 15 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/include/shim_vma.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJOvIUg2i-t2PTJnrTF:-MJVpmZ-DP2I-ZB4LxfY:btyeoa1) ([raw file](https://github.com/oscarlab/graphene/blob/f5dc5cadd65bc56508a9664d95d1826839d9147c/LibOS/shim/include/shim_vma.h#L26)):*

I'd like to use this `traverse_vmas_in_range` to later implement e.g. `_vma_bkeep_change` with it and deduplicate the logic of traversing through all VMAs in some range.

I could remove the public version, but then the problem with exporting `zero_memory_in_range` is that it's not really zeroing memory, but madvise implementation which is quite weird (e.g. it should restore file mappings to the original state).

If you want more protection from external code modifying VMAs I can make the visitor parameter a pointer to const struct, but that would require either a proxy function or an ugly function cast (the internal version should stay non-const to be reusable internally).
</blockquote></details>

I will concur with Borys on whatever he decides. Another idea is to move the `dontneed_visitor()` inside `shim_vma.c` (not sure if it makes sense...) and revert the addition of `struct shim_vma` here.
___
*[LibOS/shim/test/regression/madvise.c, line 34 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJQttxuAVrYrfkEpcvz:-MJVsGSt52WRqa0lvfKj:b-1lurq6) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/test/regression/madvise.c#L34)):*

Borys' explanation is correct :)

> `123` is just an arbitrary starting offset, I guess.

Yup. Just to not start at the first page.

> 307 % 128 == 51, so you could change that :)

Right, I simplified it.
</blockquote></details>

Got it, thanks for the reminder about GCD.

boryspoplawski (changes requested):
> Reviewed 15 of 15 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/include/shim_vma.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJOvIUg2i-t2PTJnrTF:-MJYbz1eAHAXtelqEvin:b-o4zm7y) ([raw file](https://github.com/oscarlab/graphene/blob/f5dc5cadd65bc56508a9664d95d1826839d9147c/LibOS/shim/include/shim_vma.h#L26)):*
> I'd like to use this traverse_vmas_in_range to later implement e.g. _vma_bkeep_change with it and deduplicate the logic of traversing through all VMAs in some range.

Sure, that is a good idea, but it does not impact the rest of discussion (it could always be internal in `shim_vma.c`).

> (...) then the problem with exporting zero_memory_in_range is that it's not really zeroing memory, but madvise implementation which is quite weird (e.g. it should restore file mappings to the original state).

That was just an example name, `restore_original_mem_content` is another example. You would need to come up with some good name for it :)

>If you want more protection from external code modifying VMAs I can make the visitor parameter a pointer to const struct, but that would require either a proxy function or an ugly function cast (the internal version should stay non-const to be reusable internally).

I don't like this + this does leave `struct shim_vma` exported.

> Another idea is to move the dontneed_visitor() inside shim_vma.c (not sure if it makes sense...) and revert the addition of struct shim_vma here.

Isn't exactly what I'm proposing? ;)
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 141 at r3](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJYpF7G2aDR4hSaSpRb:-MJYpF7G2aDR4hSaSpRc:b-6pa5ky) ([raw file](https://github.com/oscarlab/graphene/blob/a87b41c003d43a6e6749072521c2b16f6a241455/LibOS/shim/src/bookkeep/shim_vma.c#L141)):*
> ```C
> 
>     while (1) {
>         if (!visitor(vma, visitor_arg))
> ```

If `visitor` returns `false` and `vma->end < end` then `is_continuous` will have a wrong value.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 415 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS-lS-3FV5V7h7Ud9f:-MJYrIUR6D_qe8gMRiDy:ba7cc5b) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L415)):*

Not really, it's prepended with "In Linux before version 4.12,".
</blockquote></details>

In my manual it is not, but it just seems to be outdated.
Otoh, isn't it semantically correct to make `MADV_FREE` a synonym to `MADV_DONTNEED`?
___
*[LibOS/shim/src/sys/shim_mmap.c, line 378 at r3](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJYqe_FBX1fs5jFBhnV:-MJYqe_FBX1fs5jFBhnW:bkkbg3z) ([raw file](https://github.com/oscarlab/graphene/blob/a87b41c003d43a6e6749072521c2b16f6a241455/LibOS/shim/src/sys/shim_mmap.c#L378)):*
> ```C
>     }
> 
>     if (vma->flags & VMA_TAINTED) {
> ```

Could you move this `if` inside the `if` below? Would be a bit cleaner IMO, as this implicitly assumes (correctly) that `vma->file` is not NULL.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 428 at r3](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJYsH5WCnApbpYRfsNG:-MJYsH5X6OlmxTsmHMEG:b-sfbh6k) ([raw file](https://github.com/oscarlab/graphene/blob/a87b41c003d43a6e6749072521c2b16f6a241455/LibOS/shim/src/sys/shim_mmap.c#L428)):*
> ```C
>         case MADV_HUGEPAGE:
>         case MADV_NOHUGEPAGE:
>             return 0; // Doing nothing is semantically correct for these modes.
> ```

Some of these flags are defined only on specific kinds of memory, e.g. `MADV_FREE` operation can be applied only to private anonymous pages.
___
*[LibOS/shim/test/regression/madvise.c, line 37 at r3](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJYu38538e5_Wf66K6t:-MJYu38538e5_Wf66K6u:b-q9i20m) ([raw file](https://github.com/oscarlab/graphene/blob/a87b41c003d43a6e6749072521c2b16f6a241455/LibOS/shim/test/regression/madvise.c#L37)):*
> ```C
>         size_t perm_idx = ((i + 123) * 51) % PAGES_CNT; /* make the traverse non-sequential */
>         char* addr = m + page_size * perm_idx;
>         size_t size = page_size - (i * 31337 % page_size); /* test aligning up of the size */
> ```

This parenthesis looks weird, either change to `((i * 31337) % page_size)` or, preferably, remove it completely.
Not blocking as it's just personal preference.
___
*[Pal/src/host/Skeleton/db_rtld.c, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJP7Z3wCZulMm-gRDNN:-MJYvau06SWwGknZLhov:b72qcr) ([raw file](https://github.com/oscarlab/graphene/blob/3bcab01a0c0665d190b991b8823bdbbb6ed887ef/Pal/src/host/Skeleton/db_rtld.c#L9)):*

?
</blockquote></details>

Are you sure that this file was not imported from glibc?
Update: ok, now I've actually read whole file :)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: 149 of 150 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/include/shim_vma.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJOvIUg2i-t2PTJnrTF:-MJZ49UP5WQKaKAhpZqw:bkwu1as) ([raw file](https://github.com/oscarlab/graphene/blob/f5dc5cadd65bc56508a9664d95d1826839d9147c/LibOS/shim/include/shim_vma.h#L26)):*
> Isn't exactly what I'm proposing? ;)

Nah, you proposed a generic name which would make this function look like a universal utility, when actually the implementation will be madvise-specific. `restore_original_mem_content` is a nice name, but it's still not exactly what's inside - the very specific madvise handling logic (I'm mostly talking about all the different error codes we need to emulate - this logic definitely belongs to the madvise implementation, not shim bookkeeping, that's the problem with this solution).
___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 141 at r3](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJYpF7G2aDR4hSaSpRb:-MJZ-qg031-2L6PLkVGD:bs65rue) ([raw file](https://github.com/oscarlab/graphene/blob/a87b41c003d43a6e6749072521c2b16f6a241455/LibOS/shim/src/bookkeep/shim_vma.c#L141)):*

If `visitor` returns `false` and `vma->end < end` then `is_continuous` will have a wrong value.
</blockquote></details>

Yup, but I think this matches the documentation: "Returns whether the traversed range was continuously covered by VMAs" (_traversed_ range). I'm not sure if it's worth changing the semantics to always iterate the rest of the array, just to retrieve this one value.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 415 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS-lS-3FV5V7h7Ud9f:-MJZ1sVN5vOTLbho13ZN:b82jfoo) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L415)):*

In my manual it is not, but it just seems to be outdated.
Otoh, isn't it semantically correct to make `MADV_FREE` a synonym to `MADV_DONTNEED`?
</blockquote></details>

Maybe it is, but doing nothing also is (AFAIK). Why would we do the zeroing when it's not needed?
___
*[LibOS/shim/src/sys/shim_mmap.c, line 378 at r3](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJYqe_FBX1fs5jFBhnV:-MJZ1iNM8hhIe3c97ERn:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/a87b41c003d43a6e6749072521c2b16f6a241455/LibOS/shim/src/sys/shim_mmap.c#L378)):*

Could you move this `if` inside the `if` below? Would be a bit cleaner IMO, as this implicitly assumes (correctly) that `vma->file` is not NULL.
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 428 at r3](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJYsH5WCnApbpYRfsNG:-MJZ23QlELZJm28FBkBo:bnl085k) ([raw file](https://github.com/oscarlab/graphene/blob/a87b41c003d43a6e6749072521c2b16f6a241455/LibOS/shim/src/sys/shim_mmap.c#L428)):*

Some of these flags are defined only on specific kinds of memory, e.g. `MADV_FREE` operation can be applied only to private anonymous pages.
</blockquote></details>

Yeah, I know, but I'm not sure if we want to implement all the error handling for all of them, there are a lot of cases which will make the implementation significantly more complex. Also, I don't think not returning an error in this case is a problem in any realistic scenario.
___
*[LibOS/shim/test/regression/madvise.c, line 37 at r3](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJYu38538e5_Wf66K6t:-MJZ2VL11XCSY7bIiT5q:b-b9mjyb) ([raw file](https://github.com/oscarlab/graphene/blob/a87b41c003d43a6e6749072521c2b16f6a241455/LibOS/shim/test/regression/madvise.c#L37)):*

This parenthesis looks weird, either change to `((i * 31337) % page_size)` or, preferably, remove it completely.
Not blocking as it's just personal preference.
</blockquote></details>

I prefer it this way, to explicitly group the expression into 2 chunks: the size and the random decrease.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/include/shim_vma.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJOvIUg2i-t2PTJnrTF:-MJZNCW47b9SMfsMx1fg:b2xs4jc) ([raw file](https://github.com/oscarlab/graphene/blob/f5dc5cadd65bc56508a9664d95d1826839d9147c/LibOS/shim/include/shim_vma.h#L26)):*

> Isn't exactly what I'm proposing? ;)

Nah, you proposed a generic name which would make this function look like a universal utility, when actually the implementation will be madvise-specific. `restore_original_mem_content` is a nice name, but it's still not exactly what's inside - the very specific madvise handling logic (I'm mostly talking about all the different error codes we need to emulate - this logic definitely belongs to the madvise implementation, not shim bookkeeping, that's the problem with this solution).
</blockquote></details>

As I said before, this was just an example name. `shim_vma.c` already contains mmap/munmap/mprotect specific handling, I see no reason to threat madvise specially.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 415 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS-lS-3FV5V7h7Ud9f:-MJZABYyEi4r1V6c_t8-:b-3h447b) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L415)):*

Maybe it is, but doing nothing also is (AFAIK). Why would we do the zeroing when it's not needed?
</blockquote></details>

Because 1) it would also do error checking, b) it is what the app asked for after all.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 428 at r3](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJYsH5WCnApbpYRfsNG:-MJZATTvCOfqSGj196F0:bovvc2u) ([raw file](https://github.com/oscarlab/graphene/blob/a87b41c003d43a6e6749072521c2b16f6a241455/LibOS/shim/src/sys/shim_mmap.c#L428)):*

Yeah, I know, but I'm not sure if we want to implement all the error handling for all of them, there are a lot of cases which will make the implementation significantly more complex. Also, I don't think not returning an error in this case is a problem in any realistic scenario.
</blockquote></details>

If not returning an error on erroneous condition is not a problem, then returning `-ENOSYS` also shouldn't be.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/include/shim_vma.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJOvIUg2i-t2PTJnrTF:-MJ_kZFb8R0fin7UGn-N:boa2o7f) ([raw file](https://github.com/oscarlab/graphene/blob/f5dc5cadd65bc56508a9664d95d1826839d9147c/LibOS/shim/include/shim_vma.h#L26)):*

As I said before, this was just an example name. `shim_vma.c` already contains mmap/munmap/mprotect specific handling, I see no reason to threat madvise specially.
</blockquote></details>

I agree with Borys -- it looks like we all are fine with putting madvise-specific `dontneed_visitor()` in `shim_vma.c`. Maybe with an additional level of indirection in error codes (`dontneed_visitor()` throws some Graphene-internal error codes, and `shim_do_madvise()` replaces them with actual Linux error codes).

mkow (commented):
> Reviewed 3 of 3 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/include/shim_vma.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJOvIUg2i-t2PTJnrTF:-MJaQp6bB4I8bNc1bJ28:bwdfbtv) ([raw file](https://github.com/oscarlab/graphene/blob/f5dc5cadd65bc56508a9664d95d1826839d9147c/LibOS/shim/include/shim_vma.h#L26)):*

I agree with Borys -- it looks like we all are fine with putting madvise-specific `dontneed_visitor()` in `shim_vma.c`. Maybe with an additional level of indirection in error codes (`dontneed_visitor()` throws some Graphene-internal error codes, and `shim_do_madvise()` replaces them with actual Linux error codes).
</blockquote></details>

Ok, done, but without the error codes abstraction - I don't think this call will be ever used for anything else than madvise.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 415 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS-lS-3FV5V7h7Ud9f:-MJaOtun-Fj438McwxrX:b5soapn) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L415)):*
> b) it is what the app asked for after all.

Nah, it asked to "maybe zero the memory if you want", with the intention of communicating to the OS that it may now reclaim the pages. But we can skip this zeroing, at least according to `man`. Especially when it makes sense for performance.

Ad errors: I think for this case pretending that we succeeded would be better, I wouldn't be surprised if apps checked the return code (just as a good practice), but didn't depend on the exact semantics of the call. Anyways, I can either add error checking or return -ENOSYS, I guess you wanted the former?
___
*[LibOS/shim/src/sys/shim_mmap.c, line 428 at r3](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJYsH5WCnApbpYRfsNG:-MJaPz9r7cW-MIzOl1Z_:bks0tw4) ([raw file](https://github.com/oscarlab/graphene/blob/a87b41c003d43a6e6749072521c2b16f6a241455/LibOS/shim/src/sys/shim_mmap.c#L428)):*

If not returning an error on erroneous condition is not a problem, then returning `-ENOSYS` also shouldn't be.
</blockquote></details>

That's not true, see the thread above.

dimakuv (approved):
> Reviewed 2 of 3 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 155 at r6](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJayGZ76sw6VCzH8gn6:-MJayGZ76sw6VCzH8gn7:bosazey) ([raw file](https://github.com/oscarlab/graphene/blob/62c7f400944ada4788ff9ff88879412c2f246141/LibOS/shim/src/bookkeep/shim_vma.c#L155)):*
> ```C
>  */
> // TODO: Probably other VMA functions could make use of this helper.
> static bool __traverse_vmas_in_range(uintptr_t begin, uintptr_t end, traverse_visitor visitor,
> ```

Names beginning with double underscore are reserved, use just one.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 415 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS-lS-3FV5V7h7Ud9f:-MJb-5JM772dssOmLIOY:b46q5ew) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L415)):*

> b) it is what the app asked for after all.

Nah, it asked to "maybe zero the memory if you want", with the intention of communicating to the OS that it may now reclaim the pages. But we can skip this zeroing, at least according to `man`. Especially when it makes sense for performance.

Ad errors: I think for this case pretending that we succeeded would be better, I wouldn't be surprised if apps checked the return code (just as a good practice), but didn't depend on the exact semantics of the call. Anyways, I can either add error checking or return -ENOSYS, I guess you wanted the former?
</blockquote></details>

You can leave it as it is for now, I just don't like it :)
Btw, we won't be able to implement most of the options (e.g. `MADV_DONTNEED` on files) with current approach (holding the lock) ...

yamahata (approved):
> Reviewed 135 of 150 files at r1, 12 of 15 files at r3, 2 of 3 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/src/bookkeep/shim_vma.c, line 155 at r6](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJayGZ76sw6VCzH8gn6:-MJbz4jbDeh8KG-Al5UW:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/62c7f400944ada4788ff9ff88879412c2f246141/LibOS/shim/src/bookkeep/shim_vma.c#L155)):*

Names beginning with double underscore are reserved, use just one.
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 415 at r2](https://reviewable.io/reviews/oscarlab/graphene/1883#-MJS-lS-3FV5V7h7Ud9f:-MJbz9lx9saxs4STfhfw:b-dej5j) ([raw file](https://github.com/oscarlab/graphene/blob/19f70b9184f3d24dd70e3da70d8e4b53f9f542e6/LibOS/shim/src/sys/shim_mmap.c#L415)):*

You can leave it as it is for now, I just don't like it :)
Btw, we won't be able to implement most of the options (e.g. `MADV_DONTNEED` on files) with current approach (holding the lock) ...
</blockquote></details>

Yup, I just hope we won't ever need this weird semantics.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

boryspoplawski (approved):
> Reviewed 1 of 3 files at r5, 1 of 1 files at r7, 17 of 17 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

yamahata (approved):
> Reviewed 1 of 1 files at r7, 17 of 17 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv (approved):
> Reviewed 1 of 1 files at r7, 14 of 17 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1883)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1895, opened 2020-10-16
Loader rework pre-cleanup

mkow:
## Description of the changes 

This PR contains a bunch of simple clean-ups which were required by the more complex clean-ups (coming soon), which are needed by the loader rework (coming less soon).

## How to test this PR? 

CI + check compilation warnings on GCC before and now (see commit message about the `-Wno-gnu-alignof-expression` fix).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)


yamahata (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)* status: 0 of 51 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[LibOS/shim/src/Makefile, line 24 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJilFPb5L85cIZ_pxrG:-MJilFPb5L85cIZ_pxrH:b-29gsah) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/LibOS/shim/src/Makefile#L24)):*
> ```Makefile
> 	# Some of the code uses alignof on expressions, which is a GNU extension.
> 	# Silence Clang - it complains but does support it.
> 	CFLAGS += -Wno-gnu-alignof-expression
> ```

minor: $(cc-option,-Wno-gnu-alignof-expression)

boryspoplawski (changes requested):
> Reviewed 10 of 11 files at r1, 41 of 41 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[Pal/lib/string/utils.c, line 27 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJiofd1Ap-ww3q8360W:-MJiofd1Ap-ww3q8360X:bsp06gm) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/lib/string/utils.c#L27)):*
> ```C
> char* alloc_concat3(const char* a, size_t a_len, const char* b, size_t b_len,
>                     const char* c, size_t c_len) {
>     a_len = (a_len != (size_t)-1) ? a_len : (a ? strlen(a) : 0);
> ```

Non-blocking: I think this cast is not needed (`!=` will promote -1 to `size_t` implicitly).
___
*[Pal/lib/string/utils.c, line 35 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJipKa0Adjqsao6rtJ4:-MJipKa0Adjqsao6rtJ5:b-d3pro3) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/lib/string/utils.c#L35)):*
> ```C
>         return NULL;
> 
>     if (a_len)
> ```

Non-blocking: this ifs are not needed (`memcpy` with `0` size is ok).
___
*[Pal/src/db_main.c, line 127 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJiqc9K0khm1xOkJMFo:-MJiqc9K0khm1xOkJMFp:bnmofac) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/src/db_main.c#L127)):*
> ```C
>         } else {
>             *ptr = alloc_concat(str, len, "=", 1);
>         }
> ```

Please add error checking here (`*ptr` might be `NULL`).
___
*[Pal/src/db_process.c, line 22 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJiqxtP6UOoBUiZxfeF:-MJiqxtQDKEeljZeE95S:b-7r9k75) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/src/db_process.c#L22)):*
> ```C
>     ENTER_PAL_CALL(DkProcessCreate);
> 
>     assert(uri);
> ```

This is PAL API, we should check for errors here (not just assert).
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 704 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJit3CL6rFIesDTCugg:-MJit3CL6rFIesDTCugh:b-gje21j) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/src/host/Linux-SGX/enclave_framework.c#L704)):*
> ```C
> 
> static int init_trusted_file(const char* key, const char* uri) {
>     char* cskey;
> ```

Please move this definition to the place of initialization.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 738 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJitkd57s-P2xaRNXZq:-MJitkd57s-P2xaRNXZr:bux14yv) ([raw file](https://github.com/oscarlab/graphene/blob/bd5d56b1a0def07c59e7a31510e939ff35f8be7b/Pal/src/host/Linux-SGX/enclave_framework.c#L738)):*
> ```C
>     char* cfgbuf = NULL;
>     ssize_t cfgsize;
>     int nuris, ret;
> ```

I will miss `nuris` ...
___
*[Pal/src/host/Linux-SGX/sgx_process.c, line 65 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJivCZlCeYwxXt_rapM:-MJivCZlCeYwxXt_rapN:b-8mf2ij) ([raw file](https://github.com/oscarlab/graphene/blob/bd5d56b1a0def07c59e7a31510e939ff35f8be7b/Pal/src/host/Linux-SGX/sgx_process.c#L65)):*
> ```C
>     int fds[2] = {-1, -1};
> 
>     if (!uri || !strstartswith_static(uri, URI_PREFIX_FILE))
> ```

Why remove only half of the error checking? If we assume this will be always called with correct arguments ...

mkow (commented):
> Reviewed 1 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)* status: 47 of 51 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/src/Makefile, line 24 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJilFPb5L85cIZ_pxrG:-MJimIW35npyS60FKGtD:b6skyp9) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/LibOS/shim/src/Makefile#L24)):*

minor: $(cc-option,-Wno-gnu-alignof-expression)
</blockquote></details>

Good point, I forgot we have this function!
___
*[Pal/lib/string/utils.c, line 27 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJiofd1Ap-ww3q8360W:-MJivwUG-lEwqJFqZnO-:b-hrdd63) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/lib/string/utils.c#L27)):*

Non-blocking: I think this cast is not needed (`!=` will promote -1 to `size_t` implicitly).
</blockquote></details>

Yup, I know, but I wanted to make this more explicit. I don't have strong feelings about this, I can change it if you have stronger preference for removing ;)
___
*[Pal/lib/string/utils.c, line 35 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJipKa0Adjqsao6rtJ4:-MJiwYeU7zHxyCAKC0SF:b-tisqsg) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/lib/string/utils.c#L35)):*

Non-blocking: this ifs are not needed (`memcpy` with `0` size is ok).
</blockquote></details>

I also thought about this, but the problem is that the standard is unclear if `src` can be `NULL` given `size=0` (we actually call it like this above). It only says "The behavior is undefined if either dest or src is an invalid or null pointer", so I'm afraid that some implementations or sanitizers may fire asserts on this.
___
*[Pal/src/db_main.c, line 127 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJiqc9K0khm1xOkJMFo:-MJiyQ4RCR4XbN01WXEj:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/src/db_main.c#L127)):*

Please add error checking here (`*ptr` might be `NULL`).
</blockquote></details>

Done.
___
*[Pal/src/db_process.c, line 22 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJiqxtP6UOoBUiZxfeF:-MJiyiRn74BV5TQumuu9:b-xl5awe) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/src/db_process.c#L22)):*

This is PAL API, we should check for errors here (not just assert).
</blockquote></details>

I don't think it makes sense and other places usually don't check for this (see e.g. DkStreamOpen). That something is an API doesn't mean we need to check for NULLs in release builds, the assumption is that all pointer arguments can't be NULL unless specified otherwise in the docs/headers (at least in case of internal APIs, which are called by us).
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 704 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJit3CL6rFIesDTCugg:-MJj-8aaCIStLjz1RMuA:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/src/host/Linux-SGX/enclave_framework.c#L704)):*

Please move this definition to the place of initialization.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 738 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJitkd57s-P2xaRNXZq:-MJj-9wA2RfYYSLnjkGT:bo5olor) ([raw file](https://github.com/oscarlab/graphene/blob/bd5d56b1a0def07c59e7a31510e939ff35f8be7b/Pal/src/host/Linux-SGX/enclave_framework.c#L738)):*

I will miss `nuris` ...
</blockquote></details>

me too! what a great variable name it was...
___
*[Pal/src/host/Linux-SGX/sgx_process.c, line 65 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJivCZlCeYwxXt_rapM:-MJj-JKr9SHyDaDDVey8:b-rg9a1x) ([raw file](https://github.com/oscarlab/graphene/blob/bd5d56b1a0def07c59e7a31510e939ff35f8be7b/Pal/src/host/Linux-SGX/sgx_process.c#L65)):*

Why remove only half of the error checking? If we assume this will be always called with correct arguments ...
</blockquote></details>

Because being-NULL is Graphene-controlled, but the contents or `uri` come from the app itself. Validating all the arguments in all functions doesn't make sense, we have to assume that the called abides the interface (and for tricky cases we can add asserts).

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)* status: 47 of 51 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

boryspoplawski (changes requested):
> Reviewed 1 of 11 files at r1, 1 of 41 files at r2, 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/lib/string/utils.c, line 27 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJiofd1Ap-ww3q8360W:-MJj2b6D9aJJkU3kDy7B:bhx1dz1) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/lib/string/utils.c#L27)):*

Yup, I know, but I wanted to make this more explicit. I don't have strong feelings about this, I can change it if you have stronger preference for removing ;)
</blockquote></details>

No, don't worry, just looked weird that you cast here, but not above in function call ;)
___
*[Pal/lib/string/utils.c, line 35 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJipKa0Adjqsao6rtJ4:-MJj5Ty8E6P3m6EpUw6E:b-j8ec3w) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/lib/string/utils.c#L35)):*

I also thought about this, but the problem is that the standard is unclear if `src` can be `NULL` given `size=0` (we actually call it like this above). It only says "The behavior is undefined if either dest or src is an invalid or null pointer", so I'm afraid that some implementations or sanitizers may fire asserts on this.
</blockquote></details>

I see that this quote comes from cppreference.com, but actual standard is silent about this (gives no such requirement).
___
*[Pal/src/db_process.c, line 22 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJiqxtP6UOoBUiZxfeF:-MJj6TWQ321pj4M7RcDU:bnc6al3) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/src/db_process.c#L22)):*

I don't think it makes sense and other places usually don't check for this (see e.g. DkStreamOpen). That something is an API doesn't mean we need to check for NULLs in release builds, the assumption is that all pointer arguments can't be NULL unless specified otherwise in the docs/headers (at least in case of internal APIs, which are called by us).
</blockquote></details>

They usually do check this (see e.g. DkStreamRead), but some indeed do not. If you think they shouldn't (which I even might agree with) then create another PR removing it from all calls, but here I would suggesting adding it (at least for consistency with other functions).
This is a very minor issue, so unblocking.
___
*[Pal/src/host/Linux-SGX/sgx_process.c, line 65 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJivCZlCeYwxXt_rapM:-MJj7p6REPGlRDtWCPjo:b1hwrwp) ([raw file](https://github.com/oscarlab/graphene/blob/bd5d56b1a0def07c59e7a31510e939ff35f8be7b/Pal/src/host/Linux-SGX/sgx_process.c#L65)):*

Because being-NULL is Graphene-controlled, but the contents or `uri` come from the app itself. Validating all the arguments in all functions doesn't make sense, we have to assume that the called abides the interface (and for tricky cases we can add asserts).
</blockquote></details>

No, the `URI_PREFIX_FILE` (which is `file:`) does come from Graphene not the app.

dimakuv:

dimakuv (changes requested):
> Reviewed 10 of 11 files at r1, 37 of 41 files at r2, 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[Pal/include/lib/api.h, line 167 at r3](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJk1w7S5EWf75o-3pZd:-MJk1w7S5EWf75o-3pZe:b-ef4p25) ([raw file](https://github.com/oscarlab/graphene/blob/bd5d56b1a0def07c59e7a31510e939ff35f8be7b/Pal/include/lib/api.h#L167)):*
> ```objc
>     (static_strlen(str) + 1 > (max)                                   \
>      ? NULL                                                           \
>      : memcpy(var, str, static_strlen(str) + 1) + static_strlen(str))
> ```

What about removing this macro and replacing with normal `strcpy()`?
___
*[Pal/src/host/Linux-SGX/sgx_process.c, line 65 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJivCZlCeYwxXt_rapM:-MJk3sUj2B8k5KZ6KZTc:b-8sogei) ([raw file](https://github.com/oscarlab/graphene/blob/bd5d56b1a0def07c59e7a31510e939ff35f8be7b/Pal/src/host/Linux-SGX/sgx_process.c#L65)):*

No, the `URI_PREFIX_FILE` (which is `file:`) does come from Graphene not the app.
</blockquote></details>

I think Borys is right. Moreover, it is already checked once in `_DkProcessCreate()`.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)* status: 50 of 51 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[Pal/include/lib/api.h, line 167 at r3](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJk1w7S5EWf75o-3pZd:-MJlKVvo8lx2UOyuXnWT:b-s216xv) ([raw file](https://github.com/oscarlab/graphene/blob/bd5d56b1a0def07c59e7a31510e939ff35f8be7b/Pal/include/lib/api.h#L167)):*

What about removing this macro and replacing with normal `strcpy()`?
</blockquote></details>

The problem with this is that's actually strncpy, not strcpy. Also, despite the name, it doesn't match semantics of neither strcpy nor strncpy (it returns a shifted pointer and silently skips copying anything if destination is too small). So, I left this for later, as it looked much harder to change.
___
*[Pal/src/db_process.c, line 22 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJiqxtP6UOoBUiZxfeF:-MJlOp2gAqvx0yeWTDWr:b20303w) ([raw file](https://github.com/oscarlab/graphene/blob/6cc681ced2cfae681f98fa1441c67f7ee51a253b/Pal/src/db_process.c#L22)):*

They usually do check this (see e.g. DkStreamRead), but some indeed do not. If you think they shouldn't (which I even might agree with) then create another PR removing it from all calls, but here I would suggesting adding it (at least for consistency with other functions).
This is a very minor issue, so unblocking.
</blockquote></details>

But it's already inconsistent, so I'd prefer being consistent with the version which makes more sense. Creating another PR to clean this up is a good idea, will do this.
___
*[Pal/src/host/Linux-SGX/sgx_process.c, line 65 at r2](https://reviewable.io/reviews/oscarlab/graphene/1895#-MJivCZlCeYwxXt_rapM:-MJlPt7RA-0sr3grdw-D:b-8ssw8) ([raw file](https://github.com/oscarlab/graphene/blob/bd5d56b1a0def07c59e7a31510e939ff35f8be7b/Pal/src/host/Linux-SGX/sgx_process.c#L65)):*

I think Borys is right. Moreover, it is already checked once in `_DkProcessCreate()`.
</blockquote></details>

Ok, right, then we can safely remove this check.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)* status: :shipit: complete! all files reviewed, all discussions resolved

yamahata (approved):
> Reviewed 10 of 11 files at r1, 35 of 41 files at r2, 4 of 4 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1895)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1901, opened 2020-10-18
Loader rework part 2: remove support for loader.exec and sgx.sigfile 

mkow:
## Description of the changes 

This is the second PR in the loader rework series. Previous one was #1895.

It mostly removes and simplifies the logic of resolving the manifests. In some cases it makes life harder (see the changes in Examples), but only temporarily - it will be followed up with a change that will make all this complexity unnecessary (there will be one central manifest per Graphene run and optional per-binary "patches").

See commit messages for a slightly more granular description of the changes.

Caveat: This PR introduces the following regressions:
- Configuring multiprocess workloads may be a bit harder temporarily - the manifest needs now to reside in the same directory as the corresponding executable and have a matching name. This can be worked around for system binaries mounted from the host by symlinking them and putting manifest next to the symlink (and using the symlink to spawn processes).
- gcc and R examples are broken now and were disabled in CI, because their host OS dependencies are to hard to fix up. I already spend a few days trying to fix up all the examples after this change and just gave up with these two. All the symlink fixups are only temporal and won't be required soon, and these examples could be re-enabled then.
- I'm not sure how to fix up GSC. @vahldiek could you take a look?

I'd prefer to not spend even more time on temporary workarounds and instead spend this time on finishing the whole rework as soon as possible. Alternative would be to just submit one huge PR later, but that way the whole process would take too much time for everyone and I'd die rebasing the changes multiple times :)

## How to test this PR? 

CI + manual testing for examples not covered by CI.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)


dimakuv (changes requested):
> Reviewed 7 of 7 files at r1, 11 of 11 files at r2, 96 of 96 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-s9GV03rRbz0g7x07:-MK-s9GV03rRbz0g7x08:b-irs114) (no related file):*
For GSC, you need to add the `-exec` argument to pal-sgx-sign here: https://github.com/oscarlab/graphene/blob/36edd357fac526467025bd6676cfd452a326191a/Tools/gsc/sign_manifests.py#L12-L18

Also, this whole `binary_path` replacement can be removed (together with the helper func `get_binary_path()`):
https://github.com/oscarlab/graphene/blob/36edd357fac526467025bd6676cfd452a326191a/Tools/gsc/finalize_manifests.py#L106-L113

And then there must be symlinks for all the test manifests here: https://github.com/oscarlab/graphene/tree/master/Tools/gsc/test. Definitely need @vahldiek here.

---
___
*[Documentation/tutorials/pytorch/index.rst, line 299 at r3](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-hMM-BC6SuRC8wwmr:-MK-hMM-BC6SuRC8wwms:b-fdxzy2) ([raw file](https://github.com/oscarlab/graphene/blob/cbe5c5e2b8240c83913381eb6cdd5e6854a4112e/Documentation/tutorials/pytorch/index.rst#L299)):*
> ```reStructuredText
> 
> #. Signs the manifest and generates the SGX signature file containing SIGSTRUCT
>    (``pytorch.sig``).
> ```

This should be `python3.sig`?
___
*[Documentation/tutorials/pytorch/index.rst, line 301 at r3](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-hRBq6kg90x1KAznV:-MK-hRBq6kg90x1KAznW:b-fdxzy2) ([raw file](https://github.com/oscarlab/graphene/blob/cbe5c5e2b8240c83913381eb6cdd5e6854a4112e/Documentation/tutorials/pytorch/index.rst#L301)):*
> ```reStructuredText
>    (``pytorch.sig``).
> 
> #. Creates a dummy EINITTOKEN token file ``pytorch.token`` (this file is used
> ```

This should be `python3.sig`?
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 725 at r3](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-oSRQANnYwXMkmtk6:-MK-oSRQANnYwXMkmtk7:boheogx) ([raw file](https://github.com/oscarlab/graphene/blob/cbe5c5e2b8240c83913381eb6cdd5e6854a4112e/Pal/src/host/Linux-SGX/sgx_main.c#L725)):*
> ```C
> static int load_enclave(struct pal_enclave* enclave, int manifest_fd, char* manifest_path,
>                         char* exec_path, char* args, size_t args_size, char* env, size_t env_size,
>                         bool exec_uri_inferred, bool need_gsgx) {
> ```

Could you rename `exec_uri_inferred` -> `exec_path_from_manifest_path` or something like this? Definitely remove `uri` in the name.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 798 at r3](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-ofqUEoKt_dK1vga4:-MK-ofqUEoKt_dK1vga5:bg6pun1) ([raw file](https://github.com/oscarlab/graphene/blob/cbe5c5e2b8240c83913381eb6cdd5e6854a4112e/Pal/src/host/Linux-SGX/sgx_main.c#L798)):*
> ```C
>             // from the manifest file name, but it doesn't exist, and let
>             // the enclave go a bit further.  Go ahead and warn the user,
>             // though.
> ```

I doubt this "advanced usage" has any real meaning. Maybe a very specific LibOS implementation (that doesn't load apps but does runs very specific code)?.. Anyway, even though I want to remove this, it is for another PR.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 929 at r3](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-pP3mEc9KVdSHlzZm:-MK-pP3n4nOJe8sm6KHN:b-pgik2v) ([raw file](https://github.com/oscarlab/graphene/blob/cbe5c5e2b8240c83913381eb6cdd5e6854a4112e/Pal/src/host/Linux-SGX/sgx_main.c#L929)):*
> ```C
>     int ret = 0;
>     bool exec_uri_inferred = false; // Handle the case where the exec uri is inferred from
>                                     // the manifest name somewhat differently
> ```

`somewhat differently`, lol :) When we change this name to smth like `exec_path_from_manifest_path`, the comment won't be needed
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 985 at r3](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-quwg2sR1ou5jNfNI:-MK-quwg2sR1ou5jNfNJ:bhwqcvn) ([raw file](https://github.com/oscarlab/graphene/blob/cbe5c5e2b8240c83913381eb6cdd5e6854a4112e/Pal/src/host/Linux-SGX/sgx_main.c#L985)):*
> ```C
>             ret = -EINVAL;
>             goto out;
>         }
> ```

Could you move this logic of "read for four bytes from the presumed executable file" after the manifest-path resolution below? I.e., closer to memcmp.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 1032 at r3](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-qWALCKYyGIneiHNZ:-MK-qWALCKYyGIneiHN_:b-cu7exm) ([raw file](https://github.com/oscarlab/graphene/blob/cbe5c5e2b8240c83913381eb6cdd5e6854a4112e/Pal/src/host/Linux-SGX/sgx_main.c#L1032)):*
> ```C
>             goto out;
>         }
>         SGX_DBG(DBG_E, "exec_path: %s\n", exec_path); //tmp!
> ```

What is this `tmp!`?
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 1038 at r3](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-qYoI47XNPDZwxlbk:-MK-qYoI47XNPDZwxlbl:b-cu7exm) ([raw file](https://github.com/oscarlab/graphene/blob/cbe5c5e2b8240c83913381eb6cdd5e6854a4112e/Pal/src/host/Linux-SGX/sgx_main.c#L1038)):*
> ```C
>             goto out;
>         }
>         SGX_DBG(DBG_E, "manifest_path: %s\n", manifest_path); //tmp!
> ```

What is this `tmp!`?

vahldiek (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-s9GV03rRbz0g7x07:-MK13q2W8e2obct3-Fbs:b216f6i) (no related file):*

For GSC, you need to add the `-exec` argument to pal-sgx-sign here: https://github.com/oscarlab/graphene/blob/36edd357fac526467025bd6676cfd452a326191a/Tools/gsc/sign_manifests.py#L12-L18

Also, this whole `binary_path` replacement can be removed (together with the helper func `get_binary_path()`):
https://github.com/oscarlab/graphene/blob/36edd357fac526467025bd6676cfd452a326191a/Tools/gsc/finalize_manifests.py#L106-L113

And then there must be symlinks for all the test manifests here: https://github.com/oscarlab/graphene/tree/master/Tools/gsc/test. Definitely need @vahldiek here.
</blockquote></details>

@dimakuv explanation is quite good, there are three tasks:
1) Change template 
https://github.com/oscarlab/graphene/blob/fbb2fcada7de66c960d0178ab6cd6209ef5ca88d/Tools/gsc/templates/manifest.template#L4
Instead of setting exec, I think it needs to use the argv_0 overwrite. I'm not sure if there is a different way to specify this that would be easier.

2) Generate a symlink
https://github.com/oscarlab/graphene/blob/fbb2fcada7de66c960d0178ab6cd6209ef5ca88d/Tools/gsc/finalize_manifests.py#L113
At this point all information is available to create the symlink (target location = `binary_path` and binary name = `executable`).

3) Adding `-exec` to the signing process. 
https://github.com/oscarlab/graphene/blob/fbb2fcada7de66c960d0178ab6cd6209ef5ca88d/Tools/gsc/sign_manifests.py#L12
Therefore you have to use the following snippet to find the binary name given a GSC manifest filename. (The snippet below is from `finalize_manifest.py`)
``` 
executable = manifest[:manifest.rfind('.manifest')] if (
            manifest.rfind('.manifest') != -1) else manifest
```

Unfortunately, I have something planned for Tuesday. I can provide more complete code by Wednesday end of day. 

---

vahldiek:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-s9GV03rRbz0g7x07:-MK15xiZ2unUMnxfFws0:bxqbz84) (no related file):*
> And then there must be symlinks for all the test manifests here: https://github.com/oscarlab/graphene/tree/master/Tools/gsc/test. 

I'm not sure what this is about. I was hoping that GSC can generate the symlinks.

---

vahldiek (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-s9GV03rRbz0g7x07:-MKCKu3l25WC47mBMJ5-:bar5z92) (no related file):*
<details><summary><i>Previously, vahldiek (Anjo Vahldiek-Oberwagner) wrote…</i></summary><blockquote>

> And then there must be symlinks for all the test manifests here: https://github.com/oscarlab/graphene/tree/master/Tools/gsc/test. 

I'm not sure what this is about. I was hoping that GSC can generate the symlinks.
</blockquote></details>

If you haven't had time to fix GSC, I've pushed a tested fix to my personal fork (https://github.com/vahldiek/graphene/tree/mkow/loader-rework). This branch starts from this PR and adds a single commit. Turns out that the first step is not required only 2 and 3. 

---

mkow:
Jenkins, retest Jenkins please (`apps.LTP.fork11` timed out, known issue)

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

vahldiek (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-s9GV03rRbz0g7x07:-MKEQKmo-p5bL0tSftIf:bgrr5nv) (no related file):*
> btw. when looking at GSC code, this whole logic with finding the matching executable with which seems to be a bad design. For example, if I understood it correctly, it's impossible to use two different binaries with the same names, but in different folders.
This is correct and I don't like it myself. This was the attempt to reduce the amount of cmd line arguments (since binary name is taken from manifest filename). And I think under the current implementation, the loader wouldn't allow multiple binaries with the same filename anyways, since they have to be in the same folder or does the loader still allow execing into manifests in different folders?

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-s9GV03rRbz0g7x07:-MKEv5jf1FqH7FJU60J_:ba19v5k) (no related file):*
> And I think under the current implementation, the loader wouldn't allow multiple binaries with the same filename anyways, since they have to be in the same folder or does the loader still allow execing into manifests in different folders?

By "current implementation" you mean this PR or current master?

---

vahldiek (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-s9GV03rRbz0g7x07:-MKEQIx_6EYdLjnH6Z3_:b45b87f) (no related file):*

> And I think under the current implementation, the loader wouldn't allow multiple binaries with the same filename anyways, since they have to be in the same folder or does the loader still allow execing into manifests in different folders?

By "current implementation" you mean this PR or current master?
</blockquote></details>

This PR in particular. But even on master GSC doesn't allow multiple applications with the same executable filename. This stems from the fact that `executable.manifest` is the expect input. 

Anyway, I think this is something we can only change with more inputs to GSC, maybe that is something we want to look at like `gsc build image [executable_path_in_docker_image manifest_file]*`

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1901#-MK-s9GV03rRbz0g7x07:-MKG5VPG0cZVMWWVmsDo:bli4mf4) (no related file):*
<details><summary><i>Previously, vahldiek (Anjo Vahldiek-Oberwagner) wrote…</i></summary><blockquote>

This PR in particular. But even on master GSC doesn't allow multiple applications with the same executable filename. This stems from the fact that `executable.manifest` is the expect input. 

Anyway, I think this is something we can only change with more inputs to GSC, maybe that is something we want to look at like `gsc build image [executable_path_in_docker_image manifest_file]*`
</blockquote></details>

Then why wouldn't it allow multiple execs? This PR makes Graphene look for manifest by just attaching ".manifest" to the executable full path, so you always have 1-1 mapping. Your GSC implementation strips the path to only basename, which may cause collisions.

Anyways, this is something we need to probably redesign in GSC. But let's wait with this until I finish my loader rework.

---

yamahata (approved):
> Reviewed 5 of 7 files at r1, 8 of 11 files at r2, 94 of 96 files at r3, 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

yamahata (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/db_main.c, line 325 at r4](https://reviewable.io/reviews/oscarlab/graphene/1901#-MKGEtYaCA-7boStVEYT:-MKGEtYaCA-7boStVEYU:b-klfhmm) ([raw file](https://github.com/oscarlab/graphene/blob/36edd357fac526467025bd6676cfd452a326191a/Pal/src/db_main.c#L325)):*
> ```C
>     /* if there is no executable, try to find one in the manifest */
>     if (!exec_handle && g_pal_state.root_config) {
>         ret = get_config(g_pal_state.root_config, "loader.exec", uri_buf, URI_MAX);
> ```

Can we emit some warning if the manifest still has loader.exec or sgx.sigfile?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: 5 of 111 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @yamahata)

___
*[Pal/src/db_main.c, line 325 at r4](https://reviewable.io/reviews/oscarlab/graphene/1901#-MKGEtYaCA-7boStVEYT:-MKHJgvt2IJOGkBIbhpr:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/36edd357fac526467025bd6676cfd452a326191a/Pal/src/db_main.c#L325)):*

Can we emit some warning if the manifest still has loader.exec or sgx.sigfile?
</blockquote></details>

Done.

yamahata (approved):
> Reviewed 8 of 11 files at r6, 98 of 98 files at r7.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1901)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1902, opened 2020-10-19
[LibOS] Simplify process copying (clone) code

boryspoplawski:
## Description of the changes 
Check commit messages for details. Reviewing commit by commit is suggested :) 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)


boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: 0 of 51 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[LibOS/shim/src/sys/shim_clone.c, line 269 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MJyulmA8wAZRiGKOlS4:-MJyulmA8wAZRiGKOlS5:b-iyrmm9) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L269)):*
> ```C
> 
>         /* TODO: broken
>         ret = prepare_ipc_leader();
> ```

Old code called it in `fork` but never in `clone`, even if it was used to create a new process. From what I can tell, this function is supposed to find the IPC leader process. Unfortunately there's some bug in it and uncommenting it causes some ipc messages disappearing, e.g. message about signal being sent in `pipe_ocloexec` regression test.
Any unfortunate soul that knows IPC better than me willing to debug it?
Note that all tests work with this being commented and that this function might not have been tested well previously - it was used only in `fork` and glibc uses `clone` to implement `fork`.

dimakuv (changes requested):
> Reviewed 46 of 46 files at r1, 6 of 6 files at r2, 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[LibOS/shim/src/shim_context-x86_64.c, line 30 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-5SGwCs6TnsOOHVRN:-MK-5SGxCus3Z4kJeoVc:b-7ddpp2) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/shim_context-x86_64.c#L30)):*
> ```C
> 
>     /* In case you are wondering like I once did: this function is used only to restore context in
>      * the child thread/process after clone/fork so it can safely set `rax` to 0. */
> ```

I am confused by this comment. Why "only" (what was the other case that you thought it was used for)? Why "so it can safely set rax to 0" (how else would you restore the context passed to child in `shim_context` struct)? Please expand a bit.

Ok, I think what you meant is that "restore_context" sounds like "restore app context after syscall invocation" (which is actually done as part of `syscallas()` asm function). Maybe `restore_child_context_after_clone` or simply `restore_child_context` is a better name, and then this comment may be deleted.
___
*[LibOS/shim/src/bookkeep/shim_thread.c, line 604 at r2](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-4NDlBzNmPWlPvS1V:-MK-4NDm0ziQcFF493tr:byd2i2h) ([raw file](https://github.com/oscarlab/graphene/blob/f3958ba923af156c3230951c5dec0b42d56f0716/LibOS/shim/src/bookkeep/shim_thread.c#L604)):*
> ```C
>             new_tcb->tp        = NULL;
>             new_tcb->debug_buf = NULL;
>             new_tcb->vma_cache = NULL;
> ```

Is this an unrelated bug fix? I'm fine with this change in this commit, just want to be sure it's not something that crept in from another PR.
___
*[LibOS/shim/src/sys/shim_clone.c, line 343 at r2](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-JUYN9Htz7zDfM9jj:-MK-JUYN9Htz7zDfM9jk:b-1578mw) ([raw file](https://github.com/oscarlab/graphene/blob/f3958ba923af156c3230951c5dec0b42d56f0716/LibOS/shim/src/sys/shim_clone.c#L343)):*
> ```C
>             thread->stack_top = (char*)vma_info.addr + vma_info.length;
>             thread->stack_red = thread->stack = vma_info.addr;
>             parent_stack = (void*)shim_context_get_sp(&self->shim_tcb->context);
> ```

This `parent_stack` variable seems to have saved the parent-thread's context and later restored it. Looks like it was a useless action, and you just removed it? Because you create a new thread struct + TCB anyway and don't modify the parent-thread's context.
___
*[LibOS/shim/src/sys/shim_clone.c, line 143 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-8hx1CBo4mh8MJ3N7:-MK-8hx23Rq0IYjKRp7F:bhbup2b) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L143)):*
> ```C
>         put_handle(thread->exec);
>         thread->exec = NULL;
>     }
> ```

Do you know what this unset-thread-exec means? I see that `get_new_thread()` copies the `exec` field of the parent thread (parent process). I also see that `shim_do_exec()` replaces the `exec` field with the new process's name. But I still don't understand the logic of unsetting the `exec` field here -- shouldn't the new child process (and its main thread which we migrate here) contain the `exec` field, which is the same as the parent's main thread?
___
*[LibOS/shim/src/sys/shim_clone.c, line 222 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-CGzSCr_sDm2SSh-J:-MK-CGzSCr_sDm2SSh-K:bwjy15) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L222)):*
> ```C
>          * here would break too many applications ... */
>         // TODO: either implement these flags (shouldn't be hard) or return an error
>         flags |= CLONE_FS | CLONE_SYSVSEM;
> ```

BTW, I don't believe there are sane apps that do `CLONE_VM` without `CLONE_FS`/`CLONE_SYSVSEM`.
___
*[LibOS/shim/src/sys/shim_clone.c, line 269 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MJyulmA8wAZRiGKOlS4:-MK-EFnD86Wi33xgtV1P:bokno22) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L269)):*

Old code called it in `fork` but never in `clone`, even if it was used to create a new process. From what I can tell, this function is supposed to find the IPC leader process. Unfortunately there's some bug in it and uncommenting it causes some ipc messages disappearing, e.g. message about signal being sent in `pipe_ocloexec` regression test.
Any unfortunate soul that knows IPC better than me willing to debug it?
Note that all tests work with this being commented and that this function might not have been tested well previously - it was used only in `fork` and glibc uses `clone` to implement `fork`.
</blockquote></details>

I created an issue for this: https://github.com/oscarlab/graphene/issues/1903. I'll try to debug and fix it.
___
*[LibOS/shim/src/sys/shim_clone.c, line 288 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-Ieo466l6Zb3cGV17:-MK-Ieo466l6Zb3cGV18:b-idrzzz) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L288)):*
> ```C
>         memcpy(shim_tcb, self->shim_tcb, sizeof(*shim_tcb));
>         __shim_tcb_init(shim_tcb);
>         shim_tcb->tp = NULL;
> ```

Can you mention that this will be correctly updated during `thread` restore from the checkpoint?
___
*[LibOS/shim/src/sys/shim_clone.c, line 313 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-Lwr-4VrPbmOsq8VX:-MK-Lwr-4VrPbmOsq8VY:bpvrbfn) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L313)):*
> ```C
>         thread->in_vm    = false;
>         add_thread(thread);
>         set_as_child(self, thread);
> ```

These four lines should be moved right before `put_thread(thread); enable_preempt(NULL); return tid`. Logically they belong there. It will also untangle the logic of refcounts of `thread`.

Currently `add_thread()` and `set_as_child()` both increase the refcount of `thread`. If the creation process of the thread fails, it seems that we have a memory leak (`thread` won't be freed).
___
*[LibOS/shim/src/sys/shim_clone.c, line 321 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-KHpB6gAP0jIb7hev:-MK-KHpB6gAP0jIb7hew:bd290ob) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L321)):*
> ```C
>         thread->handle_map = NULL;
>         thread->shim_tcb = NULL;
>         unlock(&thread->lock);
> ```

Why do we need lock/unlock for the newly allocated thread? Is it a safety precaution in case `create_process_and_send_checkpoint()` does some async stuff with the thread? I don't think we'll ever have async code like this. This lock/unlock looks like it can be safely removed.

Also, if you would need this lock, don't you have a data race on `shim_tcb` then? You just free it in the line below, and `thread` will point to some garbage.
___
*[LibOS/shim/src/sys/shim_clone.c, line 324 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-Kat85kfKcEkwuZN4:-MK-Kat85kfKcEkwuZN5:b-9te1g5) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L324)):*
> ```C
> 
>         /* We do not need the child process' tcb - free it. */
>         free(shim_tcb);
> ```

Was there a memory leak before?
___
*[LibOS/shim/src/sys/shim_clone.c, line 349 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-McFgD2ijfTmekQhz:-MK-McFgD2ijfTmekQi-:b-mfd9ca) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L349)):*
> ```C
>     if (!(flags & CLONE_FILES)) {
>         /* If CLONE_FILES is not given, the new thread should receive its own copy of the
>          * descriptors table. */
> ```

...and Graphene probably breaks :) But logically this change is correct though it's a weird combination of flags.
___
*[LibOS/shim/src/sys/shim_clone.c, line 383 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-N7L29xYPngcZBDw6:-MK-N7L29xYPngcZBDw7:bfvr8vz) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L383)):*
> ```C
>     new_args.thread  = thread;
>     new_args.parent  = self;
>     new_args.stack   = user_stack_addr ?: (void*)shim_context_get_sp(&self->shim_tcb->context);
> ```

Was this a bug before? I guess we never noticed this because POSIX requires the parent thread to create the new stack and pass it in `user_stack_addr`.
___
*[LibOS/shim/src/sys/shim_fork.c, line 14 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-NtiC6DtsWpvzx7A9:-MK-NtiC6DtsWpvzx7AA:b8lmby5) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_fork.c#L14)):*
> ```C
> 
> int shim_do_fork(void) {
>     return shim_do_clone(SIGCHLD, NULL, NULL, NULL, NULL);
> ```

Hurray! We are one step closer to sanity :)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[LibOS/shim/include/shim_lock.h, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-LvJJ87zckUd_w9M7:-MK-LvJKD5LiOL_Tokiv:b-slukee) ([raw file](https://github.com/oscarlab/graphene/blob/03a0996737b57c72054ef0380252c91d4fed23a3/LibOS/shim/include/shim_lock.h#L1)):*
> ```objc
> #ifndef _SHIM_LOCK_H_
> ```

please add license + copyright (I guess the same as in shim_internal.h?)
___
*[LibOS/shim/src/sys/shim_clone.c, line 269 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MJyulmA8wAZRiGKOlS4:-MK-MQ763zoTEW2g0b_3:b-xybzfz) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L269)):*

Old code called it in `fork` but never in `clone`, even if it was used to create a new process. From what I can tell, this function is supposed to find the IPC leader process. Unfortunately there's some bug in it and uncommenting it causes some ipc messages disappearing, e.g. message about signal being sent in `pipe_ocloexec` regression test.
Any unfortunate soul that knows IPC better than me willing to debug it?
Note that all tests work with this being commented and that this function might not have been tested well previously - it was used only in `fork` and glibc uses `clone` to implement `fork`.
</blockquote></details>

Seems we've found an unfortunate: https://github.com/oscarlab/graphene/issues/1903
___
*[LibOS/shim/src/sys/shim_clone.c, line 357 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-NltH51LESjoZdTK7:-MK-NltH51LESjoZdTK8:b-11rk94) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L357)):*
> ```C
>     }
> 
>     /* Currently CLONE_SIGHAND is always set here, since CLONE_VM implies CLONE_THREAD (which implies
> ```

please fix wrapping

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[LibOS/shim/include/shim_lock.h, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-LvJJ87zckUd_w9M7:-MK-TfkRBNOW7Jp-RpRK:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/03a0996737b57c72054ef0380252c91d4fed23a3/LibOS/shim/include/shim_lock.h#L1)):*

please add license + copyright (I guess the same as in shim_internal.h?)
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_context-x86_64.c, line 30 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-5SGwCs6TnsOOHVRN:-MK-UWo62BQriU94EmzX:b3ycsqw) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/shim_context-x86_64.c#L30)):*

I am confused by this comment. Why "only" (what was the other case that you thought it was used for)? Why "so it can safely set rax to 0" (how else would you restore the context passed to child in `shim_context` struct)? Please expand a bit.

Ok, I think what you meant is that "restore_context" sounds like "restore app context after syscall invocation" (which is actually done as part of `syscallas()` asm function). Maybe `restore_child_context_after_clone` or simply `restore_child_context` is a better name, and then this comment may be deleted.
</blockquote></details>

Well, `restore_context` seems to me like a self-explanatory name: take this `context` and restore it in the current thread. It could be used in other places like signal handling or syscalls ...
Also I think it should be fork/clone code that does this "set `rax` to 0", not this function. For now I'll just rename the function.
___
*[LibOS/shim/src/bookkeep/shim_thread.c, line 604 at r2](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-4NDlBzNmPWlPvS1V:-MK-XUlVBOngHlpBcVkz:b-vvbup9) ([raw file](https://github.com/oscarlab/graphene/blob/f3958ba923af156c3230951c5dec0b42d56f0716/LibOS/shim/src/bookkeep/shim_thread.c#L604)):*

Is this an unrelated bug fix? I'm fine with this change in this commit, just want to be sure it's not something that crept in from another PR.
</blockquote></details>

Yes, just found it while reviewing tcb usages. I think this was (accidentally) working correctly before, as this field was reinitialized in checkpointing.
___
*[LibOS/shim/src/sys/shim_clone.c, line 343 at r2](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-JUYN9Htz7zDfM9jj:-MK-d76p1QsERRjECAhy:b8tug4k) ([raw file](https://github.com/oscarlab/graphene/blob/f3958ba923af156c3230951c5dec0b42d56f0716/LibOS/shim/src/sys/shim_clone.c#L343)):*

This `parent_stack` variable seems to have saved the parent-thread's context and later restored it. Looks like it was a useless action, and you just removed it? Because you create a new thread struct + TCB anyway and don't modify the parent-thread's context.
</blockquote></details>

Actually I should have not removed it - `shim_tcb.context.regs` is a pointer and we do shallow copy above, so this was not useless. Restored `parent_stack` saving.
I was mislead by the fact that fs_base is stored inside the context itself. We probably need to look why `regs` are not stored directly ...
___
*[LibOS/shim/src/sys/shim_clone.c, line 143 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-8hx1CBo4mh8MJ3N7:-MK-YI3zAGQsZbEQXqSF:b1x74vt) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L143)):*

Do you know what this unset-thread-exec means? I see that `get_new_thread()` copies the `exec` field of the parent thread (parent process). I also see that `shim_do_exec()` replaces the `exec` field with the new process's name. But I still don't understand the logic of unsetting the `exec` field here -- shouldn't the new child process (and its main thread which we migrate here) contain the `exec` field, which is the same as the parent's main thread?
</blockquote></details>

AFAIK this function is run in the parent process, so this basically cleans `exec` field of a structure describing a child process in the parent. I'll be removing this in future PR(s), so don't wrap your head around it :)
___
*[LibOS/shim/src/sys/shim_clone.c, line 222 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-CGzSCr_sDm2SSh-J:-MK-ZL7d7wEKtPVSXt9r:b-yrqh24) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L222)):*

BTW, I don't believe there are sane apps that do `CLONE_VM` without `CLONE_FS`/`CLONE_SYSVSEM`.
</blockquote></details>

Neither do I, but there was a comment here saying just that ...
___
*[LibOS/shim/src/sys/shim_clone.c, line 288 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-Ieo466l6Zb3cGV17:-MK-_Jn33ZaKHskxSwOB:b91upt1) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L288)):*

Can you mention that this will be correctly updated during `thread` restore from the checkpoint?
</blockquote></details>

I'm not sure what to exactly write here - this field won't be explicitly updated anywhere, but `set_cur_thread` (done among others in checkpoint restoration) will set it.
Or should I just write the above? :)
___
*[LibOS/shim/src/sys/shim_clone.c, line 313 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-Lwr-4VrPbmOsq8VX:-MK-e92ZAcYYjUkE0aml:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L313)):*

These four lines should be moved right before `put_thread(thread); enable_preempt(NULL); return tid`. Logically they belong there. It will also untangle the logic of refcounts of `thread`.

Currently `add_thread()` and `set_as_child()` both increase the refcount of `thread`. If the creation process of the thread fails, it seems that we have a memory leak (`thread` won't be freed).
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_clone.c, line 321 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-KHpB6gAP0jIb7hev:-MK-eH5eB5T8XU_Gvdpu:bkj6wyk) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L321)):*
I didn't add these locks, they were here before. As far as I can tell you are right and they can be safely removed.
> and `thread` will point to some garbage.

`thread->shim_tcb` is cleaned under that lock, so this case should not happen (but also this `thread` should not yet be visible to others)

Also note that soon doing a fork will not be tracked using **threads** :)
___
*[LibOS/shim/src/sys/shim_clone.c, line 324 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-Kat85kfKcEkwuZN4:-MK-euD9-9_nRkE9WEhH:b-32mwhc) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L324)):*

Was there a memory leak before?
</blockquote></details>

Before `shim_tcb` was allocated on stack. Actually I reverted it back - we do not need to allocate this struct on heap.
___
*[LibOS/shim/src/sys/shim_clone.c, line 349 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-McFgD2ijfTmekQhz:-MK-rdJUB0gcAxwOeNnZ:b-p2miu) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L349)):*

...and Graphene probably breaks :) But logically this change is correct though it's a weird combination of flags.
</blockquote></details>

Yeah is a weird combination, but hopefully Graphene can handle it :)
___
*[LibOS/shim/src/sys/shim_clone.c, line 357 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-NltH51LESjoZdTK7:-MK-r2fZAfsD1CLhCz4y:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L357)):*

please fix wrapping
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_clone.c, line 383 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-N7L29xYPngcZBDw6:-MK-s2G13bhyv3d9AHhd:bnvao81) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L383)):*

Was this a bug before? I guess we never noticed this because POSIX requires the parent thread to create the new stack and pass it in `user_stack_addr`.
</blockquote></details>

Yeah, but it's a super corner case that should not happen in reality: `CLONE_VM` + `NULL` as `child_stack` (which means duplicate parent's stack) is not super useful and very hard to handle correctly in user application (two processes using same stack at the same time).

dimakuv (approved):
> Reviewed 1 of 1 files at r4, 4 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/src/shim_context-x86_64.c, line 30 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-5SGwCs6TnsOOHVRN:-MK-vayf6eXRbsmiXVjt:b-leyoxt) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/shim_context-x86_64.c#L30)):*

Well, `restore_context` seems to me like a self-explanatory name: take this `context` and restore it in the current thread. It could be used in other places like signal handling or syscalls ...
Also I think it should be fork/clone code that does this "set `rax` to 0", not this function. For now I'll just rename the function.
</blockquote></details>

`self-explanatory name` in an oxymoron in Graphene :) But I agree that the fork/clone code already calls a helper "fixup regs on syscall return", so it would make sense to also call a new helper "set rax to 0 for child".
___
*[LibOS/shim/src/sys/shim_clone.c, line 343 at r2](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-JUYN9Htz7zDfM9jj:-MK-zUk-DGgCKUQ7FfGi:b8bmrav) ([raw file](https://github.com/oscarlab/graphene/blob/f3958ba923af156c3230951c5dec0b42d56f0716/LibOS/shim/src/sys/shim_clone.c#L343)):*

Actually I should have not removed it - `shim_tcb.context.regs` is a pointer and we do shallow copy above, so this was not useless. Restored `parent_stack` saving.
I was mislead by the fact that fs_base is stored inside the context itself. We probably need to look why `regs` are not stored directly ...
</blockquote></details>

I looked at Graphene code, and I don't see a need to have `shim_tcb.context.regs` to be a separate object in memory. Looks like the whole context can be tucked inside the TCB object. Work for some other PR...
___
*[LibOS/shim/src/sys/shim_clone.c, line 288 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-Ieo466l6Zb3cGV17:-MK-wr06CAFGAy9_Qee9:b-t39rhd) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L288)):*

I'm not sure what to exactly write here - this field won't be explicitly updated anywhere, but `set_cur_thread` (done among others in checkpoint restoration) will set it.
Or should I just write the above? :)
</blockquote></details>

Yes, that's what I meant. But Ok, it doesn't deserve a long comment. Ignore this.

yamahata (changes requested):
only minor fix. Overall it looks good to me.
(except ns leader stuff.)

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902#-:-MK12PyQApvOtnI1xXmu:b-t4o2ir)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/src/sys/shim_clone.c, line 343 at r2](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-JUYN9Htz7zDfM9jj:-MK12bfg3zadtrRIz5Ks:b-4m8h1i) ([raw file](https://github.com/oscarlab/graphene/blob/f3958ba923af156c3230951c5dec0b42d56f0716/LibOS/shim/src/sys/shim_clone.c#L343)):*

I looked at Graphene code, and I don't see a need to have `shim_tcb.context.regs` to be a separate object in memory. Looks like the whole context can be tucked inside the TCB object. Work for some other PR...
</blockquote></details>

It was left over from the old implementation.
Yes, now it's easy to put regs into context directly.
___
*[LibOS/shim/src/sys/shim_clone.c, line 185 at r5](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK0wBWoD2HElcIUNT2T:-MK0wBWoD2HElcIUNT2U:b-1uusvz) ([raw file](https://github.com/oscarlab/graphene/blob/d9492bba6fae96cd75724a9023c99e333a1eb5ec/LibOS/shim/src/sys/shim_clone.c#L185)):*
> ```
> flags &= ~CLONE_DETACHED;
> ```

CLONE_DETACHED flags needs to be in supported_flags or this line should be before the above check.

yamahata (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/src/sys/shim_clone.c, line 343 at r2](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-JUYN9Htz7zDfM9jj:-MK16K3I8tcZPfiroGLZ:b4s04hf) ([raw file](https://github.com/oscarlab/graphene/blob/f3958ba923af156c3230951c5dec0b42d56f0716/LibOS/shim/src/sys/shim_clone.c#L343)):*

It was left over from the old implementation.
Yes, now it's easy to put regs into context directly.
</blockquote></details>

I remember it now. With the current implementation, shim can be called nested with signal handler. That's the reason why the region for saved registers is allocated on stack.
If it's fixed(probably with #1218 ) to guarantee that shim can not be called nested, context.regs can be allocated in context.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: 51 of 52 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @yamahata)

___
*[LibOS/shim/src/sys/shim_clone.c, line 185 at r5](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK0wBWoD2HElcIUNT2T:-MK28DyQ4Y_94ZQV3wzs:b-pqg1gj) ([raw file](https://github.com/oscarlab/graphene/blob/d9492bba6fae96cd75724a9023c99e333a1eb5ec/LibOS/shim/src/sys/shim_clone.c#L185)):*

> ```
> flags &= ~CLONE_DETACHED;
> ```

CLONE_DETACHED flags needs to be in supported_flags or this line should be before the above check.
</blockquote></details>

It is in `supported_flags` (3rd line)

mkow (changes requested):
> Reviewed 46 of 46 files at r1, 6 of 6 files at r2, 4 of 4 files at r3, 1 of 1 files at r4, 4 of 4 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @yamahata)

___
*[LibOS/shim/src/sys/shim_clone.c, line 120 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK0EYqT6zY_gmh0P3cg:-MK0EYqT6zY_gmh0P3ch:b-5x2tvl) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L120)):*
> ```C
> 
> static BEGIN_MIGRATION_DEF(fork, struct shim_thread* thread,
>         struct shim_process_ipc_info* process_ipc_info) {
> ```

please align to `(`
___
*[LibOS/shim/src/sys/shim_clone.c, line 149 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK2-devA9hYSY_JBgJp:-MK2-dew5LhHE7sE1MmB:btoaw3t) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L149)):*
> ```C
> 
> int shim_do_clone(int flags, void* user_stack_addr, int* parent_tidptr, int* child_tidptr,
>                   void* tls) {
> ```

While we're here, please fix types to match the kernel definition:
`long clone(unsigned long clone_flags, unsigned long newsp, int* parent_tidptr, int* child_tidptr, unsigned long  tls)`
___
*[LibOS/shim/src/sys/shim_clone.c, line 196 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK24nhI-WuLlc68y7DM:-MK24nhJ-Sy9jJqR1EUo:b-ervucx) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L196)):*
> ```
> they do not happen in the wild, so they are explicitly disallowed for now
> ```
This is not the reason for not allowing them, the reason is that they are hard to emulate. This is the reason why it shouldn't be a big deal that we don't support them.
___
*[LibOS/shim/src/sys/shim_clone.c, line 198 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK25AUi-OnhSGT709A3:-MK25AUi-OnhSGT709A4:ban7dp0) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L198)):*
> ```C
>      * the Linux allows for such combinations, they do not happen in the wild, so they are
>      * explicitly disallowed for now. */
>     if (flags & CLONE_VM)
> ```

I'd feel safer with curly braces around this `if` :)
___
*[LibOS/shim/src/sys/shim_clone.c, line 263 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK266pX6b4NQ-zV5Xlu:-MK266pX6b4NQ-zV5Xlv:b-d6tjct) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L263)):*
> ```C
> 
>         if ((flags & CSIGNAL) != SIGCHLD) {
>             debug("Currently only SIGCHLD is supported as child-death signal.\n");
> ```

please add more context (at least that this error comes from `clone()`)
___
*[LibOS/shim/src/sys/shim_clone.c, line 285 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK27Xcl03K2xqLxxrir:-MK27Xcl03K2xqLxxris:b-r0bopm) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L285)):*
> ```C
>         }
>         /* Preemption is disabled and we are copying our own tcb, which should be ok to do,
>         * even without any locks. */
> ```

please fix comment alignment (one more space)
___
*[LibOS/shim/src/sys/shim_clone.c, line 295 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK20kTu1mjXMxulK0jP:-MK20kTu1mjXMxulK0jQ:b-875o85) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L295)):*
> ```C
>         }
> 
>         if (user_stack_addr) {
> ```

Do we actually support `user_stack_addr == 0`? Citing `man`: "child_stack argument may be zero, in which case copy-on-write semantics ensure that the child gets separate copies of stack pages when either process modifies the stack. In this case, for correct operation, the CLONE_VM option should not be specified."
___
*[LibOS/shim/src/sys/shim_fork.c, line 14 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-NtiC6DtsWpvzx7A9:-MK29LMr7VCr5lfcgJDq:b-ra0xjj) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_fork.c#L14)):*

Hurray! We are one step closer to sanity :)
</blockquote></details>

:partying_face:

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @yamahata)

___
*[LibOS/shim/src/sys/shim_clone.c, line 120 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK0EYqT6zY_gmh0P3cg:-MK2CUON6z5dT2eLG_TB:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L120)):*

please align to `(`
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_clone.c, line 149 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK2-devA9hYSY_JBgJp:-MK2Fi4mEyAqex_c_QTn:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L149)):*

While we're here, please fix types to match the kernel definition:
`long clone(unsigned long clone_flags, unsigned long newsp, int* parent_tidptr, int* child_tidptr, unsigned long  tls)`
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_clone.c, line 196 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK24nhI-WuLlc68y7DM:-MK2Foaf7KtAMdxpoPXs:b-oqspdq) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L196)):*

> ```
> they do not happen in the wild, so they are explicitly disallowed for now
> ```
This is not the reason for not allowing them, the reason is that they are hard to emulate. This is the reason why it shouldn't be a big deal that we don't support them.
</blockquote></details>

I do not agree: I think emulating `CLONE_VM` without any of these 2 flags should be doable and not so hard. The reason we do not support it is that there has been no request to do so - no app tested under Graphene actually used it.
___
*[LibOS/shim/src/sys/shim_clone.c, line 198 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK25AUi-OnhSGT709A3:-MK2GTNKE_X0DbMZl_1R:b4o6zn5) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L198)):*

I'd feel safer with curly braces around this `if` :)
</blockquote></details>

This is what I prefer too :)
___
*[LibOS/shim/src/sys/shim_clone.c, line 263 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK266pX6b4NQ-zV5Xlu:-MK2Gf5n3ounIEdlkK4U:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L263)):*

please add more context (at least that this error comes from `clone()`)
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_clone.c, line 285 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK27Xcl03K2xqLxxrir:-MK2Gmc7BE3AscyLcWU3:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L285)):*

please fix comment alignment (one more space)
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_clone.c, line 295 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK20kTu1mjXMxulK0jP:-MK2H-MABgNHZ8-8CLUo:bid6nes) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L295)):*

Do we actually support `user_stack_addr == 0`? Citing `man`: "child_stack argument may be zero, in which case copy-on-write semantics ensure that the child gets separate copies of stack pages when either process modifies the stack. In this case, for correct operation, the CLONE_VM option should not be specified."
</blockquote></details>

Why not? I believe the Linux kernel supports it and it's easy for as to follow. `(If the child shares the parent's memory because of the use of the CLONE_VM flag, then no copy-on-write duplication occurs and chaos is likely to result.)` If user asks for chaos, the chaos they shall get.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @yamahata)

___
*[LibOS/shim/src/sys/shim_clone.c, line 149 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK2-devA9hYSY_JBgJp:-MK2K0KF2EdB6YBU4M8N:b-wes8wa) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L149)):*

Done.
</blockquote></details>

What about the ret val?
___
*[LibOS/shim/src/sys/shim_clone.c, line 196 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK24nhI-WuLlc68y7DM:-MK2KHX0A_DnAMT14mDk:b-b128dk) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L196)):*

I do not agree: I think emulating `CLONE_VM` without any of these 2 flags should be doable and not so hard. The reason we do not support it is that there has been no request to do so - no app tested under Graphene actually used it.
</blockquote></details>

Ok, whatever :)
___
*[LibOS/shim/src/sys/shim_clone.c, line 295 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK20kTu1mjXMxulK0jP:-MK2KXxiESeMEgjLNCfr:b9hf0hq) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L295)):*

Why not? I believe the Linux kernel supports it and it's easy for as to follow. `(If the child shares the parent's memory because of the use of the CLONE_VM flag, then no copy-on-write duplication occurs and chaos is likely to result.)` If user asks for chaos, the chaos they shall get.
</blockquote></details>

I asked whether we support it, not whether we should ;) The code neither seemed to handle this case nor error-out, but sounds like you've already thought about it, so I'm unblocking.

boryspoplawski:
Jenkins, retest Jenkins-18.04 please (fork11 failed, known issue)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: 50 of 54 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @yamahata)

___
*[LibOS/shim/src/sys/shim_clone.c, line 149 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK2-devA9hYSY_JBgJp:-MK2NtuJ3Z0qVcS3lQwo:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L149)):*

What about the ret val?
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_clone.c, line 295 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK20kTu1mjXMxulK0jP:-MK2O0TM5Sj_fgdJGkd2:b782yw4) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L295)):*

I asked whether we support it, not whether we should ;) The code neither seemed to handle this case nor error-out, but sounds like you've already thought about it, so I'm unblocking.
</blockquote></details>

Yeah, we do support it :)

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @yamahata)

yamahata (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/sys/shim_clone.c, line 185 at r5](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK0wBWoD2HElcIUNT2T:-MK2SWds1lj0mk7nPKWF:b-fyg999) ([raw file](https://github.com/oscarlab/graphene/blob/d9492bba6fae96cd75724a9023c99e333a1eb5ec/LibOS/shim/src/sys/shim_clone.c#L185)):*

It is in `supported_flags` (3rd line)
</blockquote></details>

Oops. My bad.

dimakuv (dismissed):
> Reviewed 1 of 1 files at r6, 4 of 4 files at r7, 4 of 4 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (dismissed):
> Reviewed 6 of 6 files at r10, 8 of 8 files at r11.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> Reviewed 6 of 6 files at r10, 8 of 8 files at r11.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: :shipit: complete! all files reviewed, all discussions resolved

yamahata (dismissed):
> Reviewed 39 of 46 files at r1, 1 of 1 files at r4, 6 of 6 files at r10, 8 of 8 files at r11.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: :shipit: complete! all files reviewed, all discussions resolved

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: 53 of 54 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @yamahata)

___
*[LibOS/shim/src/sys/shim_clone.c, line 313 at r3](https://reviewable.io/reviews/oscarlab/graphene/1902#-MK-Lwr-4VrPbmOsq8VX:-MK5wRXh5GNdvesOVoJ2:b-9e5gka) ([raw file](https://github.com/oscarlab/graphene/blob/e20d6ec725ef11200a9923180332dd5c8d14da4b/LibOS/shim/src/sys/shim_clone.c#L313)):*

Done.
</blockquote></details>

This was wrong thing to do in the end. Right after `create_process_and_send_checkpoint` the child process is alive and can send us ipc messages. It can even exit quickly and send us info about exit code, but if it's not on `thread_list`, Graphene will ignore such message and we will never learn about child exiting.

mkow (approved):
> Reviewed 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

yamahata (approved):
> Reviewed 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1902)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1930, opened 2020-10-29
[Pal/lib] Download and patch TOML-parser library tomlc99

dimakuv:
## Description of the changes 

This library will be used to parse Graphene manifest files written in TOML syntax. We patch the library slightly to remove unsupported `toml_parse_file()` and `toml_rtod()`, as well as `errno()` handling.

## How to test this PR? 

All tests must pass. I added a dummy `toml_parse()` to check that Graphene downloads, builds and uses TOML correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1930)


yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1930)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1930)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Pal/lib/Makefile, line 166 at r1](https://reviewable.io/reviews/oscarlab/graphene/1930#-MKr0p65DMvfmZuetURh:-MKr0p6616vJCUWvvZK-:bk3mtgv) ([raw file](https://github.com/oscarlab/graphene/blob/ba8f5b14d264b667e441d3bc104962e1e4c148c1/Pal/lib/Makefile#L166)):*
> ```Makefile
> 	../../Scripts/download --output $@ --url $(TOML_H_URI) --sha256 $(TOML_H_CHECKSUM)
> 
> toml.c: ../include/lib/toml.h
> ```

why this dependency?
___
*[Pal/src/Makefile, line 123 at r1](https://reviewable.io/reviews/oscarlab/graphene/1930#-MKr2Ywk9s1ING3PsB51:-MKr2Ywk9s1ING3PsB52:bkbtlcd) ([raw file](https://github.com/oscarlab/graphene/blob/ba8f5b14d264b667e441d3bc104962e1e4c148c1/Pal/src/Makefile#L123)):*
> ```Makefile
> 	$(call cmd,ar_a_o)
> 
> # db_main needs graphene-lib because it uses its TOML parser
> ```

Hmm, aren't we missing this for all other PAL objects which use PAL lib? How are they linked to the lib?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1930)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[Pal/lib/Makefile, line 166 at r1](https://reviewable.io/reviews/oscarlab/graphene/1930#-MKr0p65DMvfmZuetURh:-MKt1Io25hp7aNPp0aby:b962t3) ([raw file](https://github.com/oscarlab/graphene/blob/ba8f5b14d264b667e441d3bc104962e1e4c148c1/Pal/lib/Makefile#L166)):*

why this dependency?
</blockquote></details>

Not sure what you suggest. We need to download both files, this dependency is one way to do it (and I thought it was clean). Do you want to download both files under the same target?
___
*[Pal/src/Makefile, line 123 at r1](https://reviewable.io/reviews/oscarlab/graphene/1930#-MKr2Ywk9s1ING3PsB51:-MKt-3Ud3jy1MANltdi0:b2hmt42) ([raw file](https://github.com/oscarlab/graphene/blob/ba8f5b14d264b667e441d3bc104962e1e4c148c1/Pal/src/Makefile#L123)):*

Hmm, aren't we missing this for all other PAL objects which use PAL lib? How are they linked to the lib?
</blockquote></details>

The problem is not in linking but in compiling. I am downloading `toml.h` from the internet, which is a new thing in Graphene source code (except `uthash.h` but that case is special). And then the compiler needs to actually find this `toml.h` file when compiling e.g. `db_main.c` (because it contains `#include toml.h`).

So I need a dependency: "please download toml.h before compiling db_main.c". However, these are located in two different Makefiles, and there is no way to trigger a make action on `toml.h` from here. So I'm falling back to "please build the whole graphene-lib library (which includes downloading toml.h) before compiling db_main.c".

I'm not sure how to make it more explicit without rearranging Makefiles. I see several options to fix this:
1. Leave as is. A bit hacky but works and generally meaningful.
2. Add a dependency on `graphene-lib` library to all compilations. A bit of an overkill, but also generally meaningful ("first build the helper library, then everything else").
3. Add `toml.h` header in Graphene repo and only download `toml.c` (this assumes that `toml.h` doesn't change much and that we are able to keep them in sync when moving to newer revisions of TOML).

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1930)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @yamahata)

___
*[Pal/lib/Makefile, line 166 at r1](https://reviewable.io/reviews/oscarlab/graphene/1930#-MKr0p65DMvfmZuetURh:-MKwRbfO9FtG4mbUfpTr:bl40490) ([raw file](https://github.com/oscarlab/graphene/blob/ba8f5b14d264b667e441d3bc104962e1e4c148c1/Pal/lib/Makefile#L166)):*

Not sure what you suggest. We need to download both files, this dependency is one way to do it (and I thought it was clean). Do you want to download both files under the same target?
</blockquote></details>

I mean, why downloading toml.c depends on downloading toml.h? Can't we just remove this dependency?
___
*[Pal/src/Makefile, line 123 at r1](https://reviewable.io/reviews/oscarlab/graphene/1930#-MKr2Ywk9s1ING3PsB51:-MKwS4g00lohbIrnk53v:bwxgd2v) ([raw file](https://github.com/oscarlab/graphene/blob/ba8f5b14d264b667e441d3bc104962e1e4c148c1/Pal/src/Makefile#L123)):*

The problem is not in linking but in compiling. I am downloading `toml.h` from the internet, which is a new thing in Graphene source code (except `uthash.h` but that case is special). And then the compiler needs to actually find this `toml.h` file when compiling e.g. `db_main.c` (because it contains `#include toml.h`).

So I need a dependency: "please download toml.h before compiling db_main.c". However, these are located in two different Makefiles, and there is no way to trigger a make action on `toml.h` from here. So I'm falling back to "please build the whole graphene-lib library (which includes downloading toml.h) before compiling db_main.c".

I'm not sure how to make it more explicit without rearranging Makefiles. I see several options to fix this:
1. Leave as is. A bit hacky but works and generally meaningful.
2. Add a dependency on `graphene-lib` library to all compilations. A bit of an overkill, but also generally meaningful ("first build the helper library, then everything else").
3. Add `toml.h` header in Graphene repo and only download `toml.c` (this assumes that `toml.h` doesn't change much and that we are able to keep them in sync when moving to newer revisions of TOML).
</blockquote></details>

ad 3.: I don't understand what makes toml.h special (different from the headers in the repo) and how moving it to the repo would help. If db_main.c uses it, then it will have a dependency on it (because of the header dependency resolving we have), and this dependency will trigger downloading.
If we don't have this dependency automatically for whatever reason, then it means that our makefiles are broken and are not noticing changes in the headers.

CC: @yamahata

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1930)* status: 5 of 6 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @yamahata)

___
*[Pal/lib/Makefile, line 166 at r1](https://reviewable.io/reviews/oscarlab/graphene/1930#-MKr0p65DMvfmZuetURh:-ML7tpT22vlt1fgbY-TT:b-vtao8x) ([raw file](https://github.com/oscarlab/graphene/blob/ba8f5b14d264b667e441d3bc104962e1e4c148c1/Pal/lib/Makefile#L166)):*

I mean, why downloading toml.c depends on downloading toml.h? Can't we just remove this dependency?
</blockquote></details>

If I understand GNU Make and GCC correctly, there is no "smartness" in these tools. Our Makefiles have dependency tracking (`-MD -MP`) but this only works for already-existing header files. However, `toml.h` is an autogenerated file. This would mean that Make would need to look at `toml.o` target, parse `toml.c` to extract header files, find them in include dirs one by one, and then when Make detects `toml.h`, somehow recurse into itself (because there is no such file on the hard disk) and find the target `toml.h`. This is impossible because Make is not recursive. (https://stackoverflow.com/questions/5229561/gnu-make-generating-automatic-dependencies-with-generated-header-files)

So since `toml.h` is autogenerated, we need an explicit dependency. Maybe @yamahata could correct me and suggest a better approach.
___
*[Pal/src/Makefile, line 123 at r1](https://reviewable.io/reviews/oscarlab/graphene/1930#-MKr2Ywk9s1ING3PsB51:-ML7v3mg0a9e0o88RZhB:bvdk11l) ([raw file](https://github.com/oscarlab/graphene/blob/ba8f5b14d264b667e441d3bc104962e1e4c148c1/Pal/src/Makefile#L123)):*

ad 3.: I don't understand what makes toml.h special (different from the headers in the repo) and how moving it to the repo would help. If db_main.c uses it, then it will have a dependency on it (because of the header dependency resolving we have), and this dependency will trigger downloading.
If we don't have this dependency automatically for whatever reason, then it means that our makefiles are broken and are not noticing changes in the headers.

CC: @yamahata
</blockquote></details>

See my other comment. I don't think the statement `and this dependency will trigger downloading` is correct (make doesn't recourse in itself).

But I would like to go with my option 2 and add `graphene-lib` as an order dependency on all `.o` files.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1930)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @yamahata)

___
*[Pal/lib/Makefile, line 166 at r1](https://reviewable.io/reviews/oscarlab/graphene/1930#-MKr0p65DMvfmZuetURh:-MLALfvH05dSYUn4dE2R:bj0xu9w) ([raw file](https://github.com/oscarlab/graphene/blob/ba8f5b14d264b667e441d3bc104962e1e4c148c1/Pal/lib/Makefile#L166)):*

If I understand GNU Make and GCC correctly, there is no "smartness" in these tools. Our Makefiles have dependency tracking (`-MD -MP`) but this only works for already-existing header files. However, `toml.h` is an autogenerated file. This would mean that Make would need to look at `toml.o` target, parse `toml.c` to extract header files, find them in include dirs one by one, and then when Make detects `toml.h`, somehow recurse into itself (because there is no such file on the hard disk) and find the target `toml.h`. This is impossible because Make is not recursive. (https://stackoverflow.com/questions/5229561/gnu-make-generating-automatic-dependencies-with-generated-header-files)

So since `toml.h` is autogenerated, we need an explicit dependency. Maybe @yamahata could correct me and suggest a better approach.
</blockquote></details>

Ok, seems our dependency resolution works differently than I thought.
___
*[Pal/src/Makefile, line 123 at r1](https://reviewable.io/reviews/oscarlab/graphene/1930#-MKr2Ywk9s1ING3PsB51:-MLAM4hTElVLTh6IGyT5:b8h8bwn) ([raw file](https://github.com/oscarlab/graphene/blob/ba8f5b14d264b667e441d3bc104962e1e4c148c1/Pal/src/Makefile#L123)):*

See my other comment. I don't think the statement `and this dependency will trigger downloading` is correct (make doesn't recourse in itself).

But I would like to go with my option 2 and add `graphene-lib` as an order dependency on all `.o` files.
</blockquote></details>

Ok, sounds good, but why using an order-only dependency?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1930)* status: 5 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @yamahata)

___
*[Pal/src/Makefile, line 123 at r1](https://reviewable.io/reviews/oscarlab/graphene/1930#-MKr2Ywk9s1ING3PsB51:-MLCchVZ4t1YScF2gFM7:b-88i64m) ([raw file](https://github.com/oscarlab/graphene/blob/ba8f5b14d264b667e441d3bc104962e1e4c148c1/Pal/src/Makefile#L123)):*

Ok, sounds good, but why using an order-only dependency?
</blockquote></details>

Done. I wasn't sure if there is a need to rebuild Pal source files when the helper `graphene_lib` implementation changes (because at compile time we only care about headers generated/used by`graphene_lib` and not the actual implementation). But I don't see any downsides to rebuild Pal source files when `graphene_lib` changes, so I changed order-only dependency into normal one.
___
*[Pal/src/Makefile, line 107 at r3](https://reviewable.io/reviews/oscarlab/graphene/1930#-MLCdpYH9f2Ucum3GhSE:-MLCdpYI-iS99VhhKUY1:b42bpg5) ([raw file](https://github.com/oscarlab/graphene/blob/f25e1a53e4f76c4a6ee161176c89ad721b5fb76a/Pal/src/Makefile#L107)):*
> ```Makefile
> ifneq ($(pal_lib),)
> $(pal_lib): $(addprefix $(OBJ_DIR)/,$(objs)) \
> 	    $(host_lib) $(graphene_lib) $(pal_lib_deps)
> ```

FYI: When we generate the final PAL shared library, we specify `graphene_lib` as a regular dependency. See my other comment.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1930)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1930)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1930)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1938, opened 2020-11-04
Miscelaneous inline asm fixes

boryspoplawski:
## Description of the changes 

Mostly adds missing "memory" clobber which caused some nasty bugs - gcc
optimized those inline asms and assumed values returned by them never
change.

Don't ask me how long it took me to debug this ...

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)

### DIFF
--- a/LibOS/shim/include/arch/x86_64/shim_internal-arch.h
+++ b/LibOS/shim/include/arch/x86_64/shim_internal-arch.h
@@ -10,10 +10,13 @@
         __stack_top &= ~0xf;                                    \
         __stack_top -= 8;                                       \
         __asm__ volatile (                                      \
-            "movq %0, %%rbp\n"                                  \
             "movq %0, %%rsp\n"                                  \
-            "jmpq *%1\n"                                        \
-            ::"r"(__stack_top), "r"(func), "D"(arg): "memory"); \
+            "xorq %%rbp, %%rbp\n"                               \
+            "jmpq *%%rcx\n"                                     \
+            :                                                   \
+            : "r"(__stack_top), "c"(func), "D"(arg)             \
+            : "memory");                                        \
+        __builtin_unreachable();                                \
     } while (0)
 
 #define CALL_ELF_ENTRY(ENTRY, ARGP)      \
--- a/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h
+++ b/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h
@@ -135,34 +135,11 @@ static inline void shim_regs_set_syscallnr(struct shim_regs* sr, uint64_t sc_num
                       sizeof(ret) == 1,                                 \
                       "SHIM_TCB_GET can be used only for "              \
                       "8, 4, 2, or 1-byte(s) members");                 \
-        switch (sizeof(ret)) {                                          \
-        case 8:                                                         \
-            __asm__("movq %%gs:%c1, %0\n"                               \
-                    : "=r"(ret)                                         \
-                    : "i" (offsetof(PAL_TCB, libos_tcb) +               \
-                           offsetof(shim_tcb_t, member)));              \
-            break;                                                      \
-        case 4:                                                         \
-            __asm__("movl %%gs:%c1, %0\n"                               \
-                    : "=r"(ret)                                         \
-                    : "i" (offsetof(PAL_TCB, libos_tcb) +               \
-                           offsetof(shim_tcb_t, member)));              \
-            break;                                                      \
-        case 2:                                                         \
-            __asm__("movw %%gs:%c1, %0\n"                               \
-                    : "=r"(ret)                                         \
-                    : "i" (offsetof(PAL_TCB, libos_tcb) +               \
-                           offsetof(shim_tcb_t, member)));              \
-            break;                                                      \
-        case 1:                                                         \
-            __asm__("movb %%gs:%c1, %0\n"                               \
-                    : "=r"(ret)                                         \
-                    : "i" (offsetof(PAL_TCB, libos_tcb) +               \
-                           offsetof(shim_tcb_t, member)));              \
-            break;                                                      \
-        default:                                                        \
-            __abort();                                                  \
-        }                                                               \
+        __asm__("mov %%gs:%c1, %0\n"                                    \
+                : "=r"(ret)                                             \
+                : "i" (offsetof(PAL_TCB, libos_tcb) +                   \
+                       offsetof(shim_tcb_t, member))                    \
+                : "memory");                                            \
         ret;                                                            \
     })
 
@@ -180,25 +157,29 @@ static inline void shim_regs_set_syscallnr(struct shim_regs* sr, uint64_t sc_num
             __asm__("movq %0, %%gs:%c1\n"                               \
                     :: "ir"(value),                                     \
                      "i"(offsetof(PAL_TCB, libos_tcb) +                 \
-                         offsetof(shim_tcb_t, member)));                \
+                         offsetof(shim_tcb_t, member))                  \
+                    : "memory");                                        \
             break;                                                      \
         case 4:                                                         \
             __asm__("movl %0, %%gs:%c1\n"                               \
                     :: "ir"(value),                                     \
                      "i"(offsetof(PAL_TCB, libos_tcb) +                 \
-                         offsetof(shim_tcb_t, member)));                \
+                         offsetof(shim_tcb_t, member))                  \
+                    : "memory");                                        \
             break;                                                      \
         case 2:                                                         \
             __asm__("movw %0, %%gs:%c1\n"                               \
                     :: "ir"(value),                                     \
                      "i"(offsetof(PAL_TCB, libos_tcb) +                 \
-                         offsetof(shim_tcb_t, member)));                \
+                         offsetof(shim_tcb_t, member))                  \
+                    : "memory");                                        \
             break;                                                      \
         case 1:                                                         \
             __asm__("movb %0, %%gs:%c1\n"                               \
                     :: "ir"(value),                                     \
                      "i"(offsetof(PAL_TCB, libos_tcb) +                 \
-                         offsetof(shim_tcb_t, member)));                \
+                         offsetof(shim_tcb_t, member))                  \
+                    : "memory");                                        \
             break;                                                      \
         default:                                                        \
             __abort();                                                  \
--- a/LibOS/shim/src/bookkeep/shim_signal.c
+++ b/LibOS/shim/src/bookkeep/shim_signal.c
@@ -451,7 +451,7 @@ bool test_user_memory(void* addr, size_t size, bool write) {
     tcb->test_range.start     = addr;
     tcb->test_range.end       = addr + size - 1;
     /* enforce compiler to store tcb->test_range into memory */
-    __asm__ volatile("" ::: "memory");
+    COMPILER_BARRIER();
 
     /* Try to read or write into one byte inside each page */
     void* tmp = addr;
@@ -466,7 +466,7 @@ bool test_user_memory(void* addr, size_t size, bool write) {
 
 ret_fault:
     /* enforce compiler to load tcb->test_range.has_fault below */
-    __asm__ volatile("" : "=m"(tcb->test_range.has_fault));
+    COMPILER_BARRIER();
 
     /* If any read or write into the target region causes an exception,
      * the control flow will immediately jump to here. */
@@ -517,7 +517,7 @@ bool test_user_string(const char* addr) {
     tcb->test_range.has_fault = false;
     tcb->test_range.cont_addr = &&ret_fault;
     /* enforce compiler to store tcb->test_range into memory */
-    __asm__ volatile("" ::: "memory");
+    COMPILER_BARRIER();
 
     do {
         /* Add the memory region to the watch list. This is not racy because
@@ -538,7 +538,7 @@ bool test_user_string(const char* addr) {
 
 ret_fault:
     /* enforce compiler to load tcb->test_range.has_fault below */
-    __asm__ volatile("" : "=m"(tcb->test_range.has_fault));
+    COMPILER_BARRIER();
 
     /* If any read or write into the target region causes an exception,
      * the control flow will immediately jump to here. */
--- a/LibOS/shim/src/bookkeep/shim_thread.c
+++ b/LibOS/shim/src/bookkeep/shim_thread.c
@@ -474,7 +474,7 @@ void cleanup_thread(IDTYPE caller, void* arg) {
 
     /* wait on clear_child_tid_pal; this signals that PAL layer exited child thread */
     while (__atomic_load_n(&thread->clear_child_tid_pal, __ATOMIC_RELAXED) != 0)
-        cpu_pause();
+        CPU_RELAX();
 
     /* notify parent if any */
     release_clear_child_tid(thread->clear_child_tid);
--- a/LibOS/shim/src/ipc/shim_ipc_helper.c
+++ b/LibOS/shim/src/ipc/shim_ipc_helper.c
@@ -851,6 +851,7 @@ static void shim_ipc_helper_prepare(void* arg) {
     self->stack_top = stack + IPC_HELPER_STACK_SIZE;
     self->stack     = stack;
     __SWITCH_STACK(self->stack_top, shim_ipc_helper, NULL);
+    /* UNREACHABLE */
 }
 
 /* this should be called with the ipc_helper_lock held */
--- a/LibOS/shim/src/sys/shim_exec.c
+++ b/LibOS/shim/src/sys/shim_exec.c
@@ -152,7 +152,7 @@ static int shim_do_execve_rtld(struct shim_handle* hdl, const char** argv, const
         .new_auxv = new_auxv
     };
     __SWITCH_STACK(new_argp, &__shim_do_execve_rtld, &arg);
-    return 0;
+    /* UNREACHABLE */
 }
 
 static BEGIN_MIGRATION_DEF(execve, struct shim_thread* thread,
--- a/Pal/include/arch/x86_64/cpu.h
+++ b/Pal/include/arch/x86_64/cpu.h
@@ -5,10 +5,6 @@
 
 #include <stdint.h>
 
-static inline void cpu_pause(void) {
-    __asm__ volatile("pause");
-}
-
 enum PAL_CPUID_WORD {
     PAL_CPUID_WORD_EAX = 0,
     PAL_CPUID_WORD_EBX = 1,
@@ -35,7 +31,7 @@ static inline uint64_t get_tsc(void) {
     return lo | ((uint64_t)hi << 32);
 }
 
-#define CPU_RELAX() __asm__ __volatile__("rep; nop" ::: "memory")
+#define CPU_RELAX() __asm__ volatile("pause")
 
 /* some non-Intel clones support out of order store; WMB() ceases to be a nop for these */
 #define MB()  __asm__ __volatile__("mfence" ::: "memory")
--- a/Pal/include/arch/x86_64/pal-arch.h
+++ b/Pal/include/arch/x86_64/pal-arch.h
@@ -36,7 +36,7 @@ typedef struct pal_tcb {
 
 static inline PAL_TCB* pal_get_tcb(void) {
     PAL_TCB* tcb;
-    __asm__("movq %%gs:%c1,%q0" : "=r"(tcb) : "i"(offsetof(struct pal_tcb, self)));
+    __asm__("movq %%gs:%c1, %0" : "=r"(tcb) : "i"(offsetof(struct pal_tcb, self)) : "memory");
     return tcb;
 }
 
--- a/Pal/include/arch/x86_64/pal_internal-arch.h
+++ b/Pal/include/arch/x86_64/pal_internal-arch.h
@@ -1,0 +1,25 @@
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
+/* Copyright (C) 2020 Intel Corporation
+ *                    Borys Popławski <borysp@invisiblethingslab.com>
+ */
+
+#ifndef PAL_INTERNAL_ARCH_H_
+#define PAL_INTERNAL_ARCH_H_
+
+#include "assert.h"
+
+#define ARCH_PROBE_STACK(size, page_size)           \
+    assert(size / page_size > 0);                   \
+    __asm__ volatile(                               \
+        "mov %%rsp, %%rdx\n"                        \
+        "1:\n"                                      \
+        "sub %0, %%rsp\n"                           \
+        "orq $0, (%%rsp)\n"                         \
+        "loop 1b\n"                                 \
+        "mov %%rdx, %%rsp\n"                        \
+        :                                           \
+        : "ri"(page_size), "c"(size / page_size)    \
+        : "memory", "cc", "rdx"                     \
+    )
+
+#endif // PAL_INTERNAL_ARCH_H_
--- a/Pal/include/lib/spinlock.h
+++ b/Pal/include/lib/spinlock.h
@@ -104,7 +104,7 @@ static inline void spinlock_lock(spinlock_t* lock) {
     do {
         /* This check imposes no inter-thread ordering, thus does not slow other threads. */
         while (__atomic_load_n(&lock->lock, __ATOMIC_RELAXED) != SPINLOCK_UNLOCKED)
-            cpu_pause();
+            CPU_RELAX();
         /* Seen lock as free, check if it still is, this time with acquire semantics (but only
          * if we really take it). */
         val = SPINLOCK_UNLOCKED;
@@ -136,7 +136,7 @@ static inline int spinlock_lock_timeout(spinlock_t* lock, unsigned long iteratio
                 return 1;
             }
             iterations--;
-            cpu_pause();
+            CPU_RELAX();
         }
         /* Seen lock as free, check if it still is, this time with acquire semantics (but only
          * if we really take it). */
--- a/Pal/lib/string/memcpy.c
+++ b/Pal/lib/string/memcpy.c
@@ -15,7 +15,7 @@ void* memcpy(void* restrict dest, const void* restrict src, size_t count) {
      * memcpy() is heavily used in Linux-SGX PAL to copy data in/out of SGX enclave. Experiments
      * with Redis 5.0 show perf improvement of using "rep movsb" at 3-5% for 4KB payloads over
      * previous implementation taken from Glibc 2.23. */
-    __asm__ volatile("rep movsb" : "+&D"(d), "+&c"(count) : "S"(src) : "cc", "memory");
+    __asm__ volatile("rep movsb" : "+D"(d), "+c"(count), "+S"(src) :: "cc", "memory");
 #else
     const char* s = src;
     while (count--)
--- a/Pal/lib/string/memset.c
+++ b/Pal/lib/string/memset.c
@@ -15,7 +15,7 @@ void* memset(void* dest, int ch, size_t count) {
      * operations for software in common situations like memory copy and set operations"
      * Intel 64 and IA-32 Architectures Optimization Reference Manual
      */
-    __asm__ volatile("rep stosb" : "+&D"(d), "+&c"(count) : "a"((uint8_t)ch) : "cc", "memory");
+    __asm__ volatile("rep stosb" : "+D"(d), "+c"(count) : "a"((uint8_t)ch) : "cc", "memory");
 #else
     while (count--)
         *d++ = ch;
--- a/Pal/regression/Memory.c
+++ b/Pal/regression/Memory.c
@@ -1,3 +1,5 @@
+/* XXX: What on earth is this supposed to be, an attempt to fit most UBs in one file? */
+
 #include "api.h"
 #include "pal.h"
 #include "pal_debug.h"
--- a/Pal/src/db_rtld.c
+++ b/Pal/src/db_rtld.c
@@ -956,6 +956,8 @@ void DkDebugDetachBinary(PAL_PTR start_addr) {
 #ifdef __x86_64__
 void* stack_before_call __attribute_unused = NULL;
 
+/* TODO: Why on earth do we call loaded libraries entry points?!?
+ * I won't bother fixing this asm, it needs to be purged. */
 #define CALL_ENTRY(l, cookies)                                                       \
     ({                                                                               \
         long ret;                                                                    \
--- a/Pal/src/host/Linux-SGX/db_pipes.c
+++ b/Pal/src/host/Linux-SGX/db_pipes.c
@@ -386,7 +386,7 @@ static int64_t pipe_read(PAL_HANDLE handle, uint64_t offset, uint64_t len, void*
     } else {
         /* normal pipe, use a secure session (should be already initialized) */
         while (!__atomic_load_n(&handle->pipe.handshake_done, __ATOMIC_ACQUIRE))
-            cpu_pause();
+            CPU_RELAX();
 
         if (!handle->pipe.ssl_ctx)
             return -PAL_ERROR_NOTCONNECTION;
@@ -426,7 +426,7 @@ static int64_t pipe_write(PAL_HANDLE handle, uint64_t offset, uint64_t len, cons
     } else {
         /* normal pipe, use a secure session (should be already initialized) */
         while (!__atomic_load_n(&handle->pipe.handshake_done, __ATOMIC_ACQUIRE))
-            cpu_pause();
+            CPU_RELAX();
 
         if (!handle->pipe.ssl_ctx)
             return -PAL_ERROR_NOTCONNECTION;
@@ -455,7 +455,7 @@ static int pipe_close(PAL_HANDLE handle) {
         }
     } else if (handle->pipe.fd != PAL_IDX_POISON) {
         while (!__atomic_load_n(&handle->pipe.handshake_done, __ATOMIC_ACQUIRE))
-            cpu_pause();
+            CPU_RELAX();
 
         if (handle->pipe.ssl_ctx) {
             _DkStreamSecureFree((LIB_SSL_CONTEXT*)handle->pipe.ssl_ctx);
--- a/Pal/src/host/Linux-SGX/sgx_enclave.c
+++ b/Pal/src/host/Linux-SGX/sgx_enclave.c
@@ -730,7 +730,7 @@ static int rpc_thread_loop(void* arg) {
                 (void)INLINE_SYSCALL(nanosleep, 2, &tv, /*rem=*/NULL);
             } else {
                 spin_attempts++;
-                cpu_pause();
+                CPU_RELAX();
             }
             continue;
         }
--- a/Pal/src/host/Linux-SGX/sgx_exception.c
+++ b/Pal/src/host/Linux-SGX/sgx_exception.c
@@ -33,18 +33,14 @@
 
 #if defined(__x86_64__)
 /* in x86_64 kernels, sigaction is required to have a user-defined restorer */
-#define DEFINE_RESTORE_RT(syscall) DEFINE_RESTORE_RT2(syscall)
-#define DEFINE_RESTORE_RT2(syscall)          \
-    __asm__(                                 \
-        "    nop\n"                          \
-        ".align 16\n"                        \
-        ".LSTART_restore_rt:\n"              \
-        "    .type __restore_rt,@function\n" \
-        "__restore_rt:\n"                    \
-        "    movq $" #syscall                \
-        ", %rax\n"                           \
-        "    syscall\n");
-DEFINE_RESTORE_RT(__NR_rt_sigreturn)
+__asm__(
+".align 16\n"
+".LSTART_restore_rt:\n"
+".type __restore_rt,@function\n"
+"__restore_rt:\n"
+"movq $" XSTRINGIFY(__NR_rt_sigreturn) ", %rax\n"
+"syscall\n"
+);
 
 /* workaround for an old GAS (2.27) bug that incorrectly omits relocations when referencing this
  * symbol */
--- a/Pal/src/host/Linux-SGX/sgx_main.c
+++ b/Pal/src/host/Linux-SGX/sgx_main.c
@@ -10,6 +10,7 @@
  *   #define ElfW(type)       _ElfW(Elf, __ELF_NATIVE_CLASS, type)"
  * errors.
  */
+#include "pal_internal-arch.h"
 #include "pal_linux.h"
 #include "pal_linux_error.h"
 #include "pal_rtld.h"
@@ -892,21 +893,12 @@ static int load_enclave(struct pal_enclave* enclave, int manifest_fd, char* mani
     return 0;
 }
 
-/* Grow stack of main thread to THREAD_STACK_SIZE by allocating a large dummy array and probing
- * each stack page (Linux dynamically grows the stack of the main thread but gets confused with
+/* Grow the stack of the main thread to THREAD_STACK_SIZE by probing each stack page above current
+ * stack pointer (Linux dynamically grows the stack of the main thread but gets confused with
  * huge-jump stack accesses coming from within the enclave). Note that other, non-main threads
  * are created manually via clone(.., THREAD_STACK_SIZE, ..) and thus do not need this hack. */
-static void __attribute__((noinline)) force_linux_to_grow_stack(void) {
-    char dummy[THREAD_STACK_SIZE];
-    for (uint64_t i = 0; i < sizeof(dummy); i += PRESET_PAGESIZE) {
-        /* touch each page on the stack just to make it is not optimized away */
-        __asm__ volatile(
-            "movq %0, %%rbx\r\n"
-            "movq (%%rbx), %%rbx\r\n"
-            :
-            : "r"(&dummy[i])
-            : "%rbx");
-    }
+static void force_linux_to_grow_stack(void) {
+    ARCH_PROBE_STACK(THREAD_STACK_SIZE, PRESET_PAGESIZE);
 }
 
 int main(int argc, char* argv[], char* envp[]) {
--- a/Pal/src/host/Linux-SGX/sgx_thread.c
+++ b/Pal/src/host/Linux-SGX/sgx_thread.c
@@ -222,22 +222,21 @@ noreturn void thread_exit(int status) {
 
     /* free the thread stack (via munmap) and exit; note that exit() needs a "status" arg
      * but it could be allocated on a stack, so we must put it in register and do asm */
-    __asm__ volatile("cmpq $0, %%rdi \n\t"     /* check if tcb->stack != NULL */
-                     "je 1f \n\t"
-                     "syscall \n\t"            /* all args are already prepared, call munmap */
-                     "1: \n\t"
-                     "movq %%rdx, %%rax \n\t"  /* prepare for exit: rax = __NR_exit */
-                     "movq %%rbx, %%rdi \n\t"  /* prepare for exit: rdi = status    */
-                     "syscall \n\t"            /* all args are prepared, call exit  */
-                     : /* no output regs since we don't return from exit */
-                     : "a"(__NR_munmap), "D"(tcb->stack), "S"(THREAD_STACK_SIZE + ALT_STACK_SIZE),
-                       "d"(__NR_exit), "b"(status)
-                     : "cc", "rcx", "r11", "memory"  /* syscall instr clobbers cc, rcx, and r11 */
+    __asm__ volatile("cmpq $0, %%rdi \n"        /* check if tcb->stack != NULL */
+                     "je 1f \n"
+                     "syscall \n"               /* all args are already prepared, call munmap */
+                     "1: \n"
+                     "mov %[nr_exit], %%rax \n"
+                     "mov %[exit_code], %%edi \n"
+                     "syscall \n"               /* all args are prepared, call exit  */
+                     "ud2 \n"
+                     "jmp 1b \n"
+                     :
+                     : "a" (__NR_munmap), "D" (tcb->stack), "S" (THREAD_STACK_SIZE + ALT_STACK_SIZE),
+                       [nr_exit] "i" (__NR_exit), [exit_code] "r" (status)
+                     : "memory", "rcx", "r11"
     );
-
-    while (true) {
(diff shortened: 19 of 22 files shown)


dimakuv (changes requested):
> Reviewed 21 of 21 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHne5k3JDzop_mmNy9:-MLHne5l9ayWD9FSZ-o_:b-t7uzur) (no related file):*
Jenkins SGX (non-debug) pipelines failed with:
```
sgx_main.c: In function 'force_linux_to_grow_stack':
sgx_main.c:902:44: error: comparison of unsigned expression >= 0 is always true [-Werror=type-limits]
     for (uint64_t i = sizeof(dummy) - 1; i >= 0; i -= PRESET_PAGESIZE) {
                                            ^
sgx_main.c:901:10: error: variable 'tmp' set but not used [-Werror=unused-but-set-variable]
     char tmp;
          ^
sgx_main.c:904:13: error: 'dummy[i]' is used uninitialized in this function [-Werror=uninitialized]
         tmp = *(volatile char*)&dummy[i];
             ^
```

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHvp2w15d8T0j4KcHb:-MLHvp2xAZkji12v2Ktz:bkv90fl) (no related file):*
@yamahata Could you review this PR? You wrote some of this inline asm.

---
___
*[LibOS/shim/include/arch/x86_64/shim_internal-arch.h, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHofcU9qHTU7z6197p:-MLHofcU9qHTU7z6197q:bbfwkbk) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/LibOS/shim/include/arch/x86_64/shim_internal-arch.h#L15)):*
> ```objc
>             "movq %0, %%rsp\n"                                  \
>             "xorq %%rbp, %%rbp\n"                               \
>             "jmpq *%%rcx\n"                                     \
> ```

Was there a bug here? The old version seems legit...
___
*[LibOS/shim/include/arch/x86_64/shim_tcb-arch.h, line 142 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHrfCF1DyIZOSBJKY6:-MLHrfCG9fBB_S7bYVxG:b2b2lo3) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h#L142)):*
> ```objc
>                 : "i" (offsetof(PAL_TCB, libos_tcb) +                   \
>                        offsetof(shim_tcb_t, member))                    \
>                 : "memory");                                            \
> ```

Why did you remove the switch statement? I understand that the GCC assembler may be smart enough to deduce the correct mov instruction from type of `ret`, but this looks like a dangerous assumption.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 902 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHuwVw843rwRrY8XHx:-MLHuwVw843rwRrY8XHy:b6fod15) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux-SGX/sgx_main.c#L902)):*
> ```C
>     char dummy[THREAD_STACK_SIZE];
>     char tmp;
>     for (uint64_t i = sizeof(dummy) - 1; i >= 0; i -= PRESET_PAGESIZE) {
> ```

Why did you revert the direction? Is it just to have a better correspondence to the "grow the stack" description? I see no functional difference here.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 903 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHuYUAApHgDDI9Q3XZ:-MLHuYUAApHgDDI9Q3X_:b-nwi2wd) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux-SGX/sgx_main.c#L903)):*
> ```C
>     char tmp;
>     for (uint64_t i = sizeof(dummy) - 1; i >= 0; i -= PRESET_PAGESIZE) {
>         /* touch each page on the stack just to make it is not optimized away */
> ```

`make` -> `make sure`

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHvx3C7fPzrEoSCtlS:-MLHvx3C7fPzrEoSCtlT:b-dvkgfa) (no related file):*
> Don't ask me how long it took me to debug this ...

I actually wanted to ask :) Do you know which of these inline asms led to your bug? And in which app and how it manifested?

---

mkow (changes requested):
> Reviewed 21 of 21 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/include/arch/x86_64/shim_tcb-arch.h, line 142 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHrfCF1DyIZOSBJKY6:-MLIFAYhC4bkPnekD4VD:bvylp9f) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h#L142)):*

Why did you remove the switch statement? I understand that the GCC assembler may be smart enough to deduce the correct mov instruction from type of `ret`, but this looks like a dangerous assumption.
</blockquote></details>

I tested it and it seems to work will all the sizes. I think "r" always selects register with a proper size. But I don't see this behavior documented, GCC docs are terrible...
___
*[Pal/regression/Memory.c, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIH2tME0uGDYu9Q0nj:-MLIH2tME0uGDYu9Q0nk:b7ctzjl) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/regression/Memory.c#L1)):*
> ```C
> /* XXX: What on earth is this supposed to be, an attempt to fit most UBs in one file? */
> ```

+1, always had the same question when looking at this file :)
___
*[Pal/src/host/Linux/db_exception.c, line 33 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIHkyNETtzQcNkAJzh:-MLIHkyNETtzQcNkAJzi:b-pouvfp) ([raw file](https://github.com/oscarlab/graphene/blob/ec270d07616f30a1bb5bdc83e193ff6d33c8f016/Pal/src/host/Linux/db_exception.c#L33)):*
> ```C
> #define DEFINE_RESTORE_RT2(syscall)          \
>     __asm__(                                 \
>         "    nop\n"                          \
> ```

Any idea why it started with a `nop`?
___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLIIipz0BDIBN_x-Vom:b-1jiggo) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*
> ```C
>         "syscall \n"            /* rdi arg is already prepared, call exit */
>         "2: \n"
>         "hlt \n"
> ```

isn't `hlt` privileged?
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 39 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIw34BGQfP2Mbt8Xx:-MLIIw34BGQfP2Mbt8Xy:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/ec270d07616f30a1bb5bdc83e193ff6d33c8f016/Pal/src/host/Linux-SGX/sgx_exception.c#L39)):*
> ```C
> #define DEFINE_RESTORE_RT2(syscall)          \
>     __asm__(                                 \
>         "    nop\n"                          \
> ```

ditto
___
*[Pal/src/host/Linux-SGX/sgx_thread.c, line 233 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIJK-u219mK1N3Uy7w:-MLIJK-v8M9NWGm2uTGm:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux-SGX/sgx_thread.c#L233)):*
> ```C
>                      "syscall \n"           /* all args are prepared, call exit  */
>                      "2: \n"
>                      "hlt \n"
> ```

ditto

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: 20 of 22 files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHne5k3JDzop_mmNy9:-MLIYQjWANEfNOrmta_m:b82i390) (no related file):*

Jenkins SGX (non-debug) pipelines failed with:
```
sgx_main.c: In function 'force_linux_to_grow_stack':
sgx_main.c:902:44: error: comparison of unsigned expression >= 0 is always true [-Werror=type-limits]
     for (uint64_t i = sizeof(dummy) - 1; i >= 0; i -= PRESET_PAGESIZE) {
                                            ^
sgx_main.c:901:10: error: variable 'tmp' set but not used [-Werror=unused-but-set-variable]
     char tmp;
          ^
sgx_main.c:904:13: error: 'dummy[i]' is used uninitialized in this function [-Werror=uninitialized]
         tmp = *(volatile char*)&dummy[i];
             ^
```
</blockquote></details>

Sorry, it was late. Rewritten completely.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHvx3C7fPzrEoSCtlS:-MLIjK5YEn4sxgXwHSZC:bwaxpu) (no related file):*

> Don't ask me how long it took me to debug this ...

I actually wanted to ask :) Do you know which of these inline asms led to your bug? And in which app and how it manifested?
</blockquote></details>

It was the lack of "memory" clobber in `shim_tcb-arch.h`. Spotted in LTP's abort01 (probably others too, but I debugged this one):
1) it uses `wait` to, well, wait for the child,
2) `wait` uses `thread_sleep` to sleep if the child did not yet exit,
3) `thread_sleep` is `static inline` function in some `.h` file, so it gets inlined inside `wait`,
4) `thread_sleep` uses `PAL_ERRNO` to obtain errno code in case something fails, `PAL_ERRNO` extracts errno from TCB (and does some translation, not important),
5) this extraction missed the "memory" clobber, so gcc on non-debug builds (i.e. the one with optimizations turned on, `-O2` I think) optimized this errno extraction and moved it to the begining of `wait` (because `thread_sleep` was inlined inside it) - after all this inline asm did only depend on some constant offset, so it could even be moved across compiler barries,
6) so if `thread_sleep` failed, it reported some random errno (obtained at the beginning of the function).

How it manifested?
Sometimes `wait` would fail with `ENOENT` - it happened when `SIGCHLD` was handled while doing `wait`, so it was supposed to report `EINTR`, but due to the bug it reported `ENOENT`.

---
___
*[LibOS/shim/include/arch/x86_64/shim_internal-arch.h, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHofcU9qHTU7z6197p:-MLIS43LET2saipVxYuj:b7cxa28) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/LibOS/shim/include/arch/x86_64/shim_internal-arch.h#L15)):*

Was there a bug here? The old version seems legit...
</blockquote></details>

gcc could pick `rbp` as `%1` (I did not hit this bug, but it looks possible).
Also arbitrarily setting `rbp` to `rsp` makes no sense, better zero it out to aid debuggers.
___
*[LibOS/shim/include/arch/x86_64/shim_tcb-arch.h, line 142 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHrfCF1DyIZOSBJKY6:-MLISOkU8k0ipvXyMwaI:b-sdavdp) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h#L142)):*

I tested it and it seems to work will all the sizes. I think "r" always selects register with a proper size. But I don't see this behavior documented, GCC docs are terrible...
</blockquote></details>

Yeah, that was my reasoning: since we always use a register ("r") then we can use `mov` without size suffix.
Should I revert the old version (switch)? Indeed gcc doesn't tell us anything about this case ...
___
*[Pal/src/host/Linux/db_exception.c, line 33 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIHkyNETtzQcNkAJzh:-MLISmSpEtuRjulp9RMk:bcyiw11) ([raw file](https://github.com/oscarlab/graphene/blob/ec270d07616f30a1bb5bdc83e193ff6d33c8f016/Pal/src/host/Linux/db_exception.c#L33)):*

Any idea why it started with a `nop`?
</blockquote></details>

No and hopefully there are no cursed hacks around it anywhere ...
___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLIStpl8ScwEQqlGMQy:b-sn5h7m) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*

isn't `hlt` privileged?
</blockquote></details>

Yeah and it will crash the process and if somehow it does not, we have a safety loop around it. Generally speaking exit syscall should succeed, this is just a safety net.
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 39 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIw34BGQfP2Mbt8Xx:-MLIZ0pt0zRQkqjExgxX:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/ec270d07616f30a1bb5bdc83e193ff6d33c8f016/Pal/src/host/Linux-SGX/sgx_exception.c#L39)):*

ditto
</blockquote></details>

ditto
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 902 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHuwVw843rwRrY8XHx:-MLIYARx-sdDx9a_8_XG:b-wo62mz) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux-SGX/sgx_main.c#L902)):*

Why did you revert the direction? Is it just to have a better correspondence to the "grow the stack" description? I see no functional difference here.
</blockquote></details>

This function made no sense (pre and post changes). Rewritten.
___
*[Pal/src/host/Linux-SGX/sgx_thread.c, line 233 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIJK-u219mK1N3Uy7w:-MLIZ3OR2k72yjpGE9CV:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux-SGX/sgx_thread.c#L233)):*

ditto
</blockquote></details>

ditto

yamahata (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: 20 of 22 files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIufCJ9DG5sd3S0Nkf:-MLIufCK2EamRaIqWDRr:bjsai64) (no related file):*
can you please share the combination of compiler version and optimization options to hit the issue?

---
___
*[LibOS/shim/include/arch/x86_64/shim_internal-arch.h, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHofcU9qHTU7z6197p:-MLIvZH-6GEw3K96XJRm:bs1ob6m) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/LibOS/shim/include/arch/x86_64/shim_internal-arch.h#L15)):*

gcc could pick `rbp` as `%1` (I did not hit this bug, but it looks possible).
Also arbitrarily setting `rbp` to `rsp` makes no sense, better zero it out to aid debuggers.
</blockquote></details>

it’s correct to clear rbp to indicate the end of stack to gdb. my bad.
___
*[LibOS/shim/include/arch/x86_64/shim_tcb-arch.h, line 142 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHrfCF1DyIZOSBJKY6:-MLIweJr4eUbg9evnNT9:bqflain) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/LibOS/shim/include/arch/x86_64/shim_tcb-arch.h#L142)):*

I tested it and it seems to work will all the sizes. I think "r" always selects register with a proper size. But I don't see this behavior documented, GCC docs are terrible...
</blockquote></details>

I think gas selects actual op code. As long as llvm is happy with it, I’m fine with this.
___
*[Pal/src/db_rtld.c, line 960 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIyiE68bB_rkcaP1bV:-MLIyiE83irCpPtWNdUa:bvxivcu) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/db_rtld.c#L960)):*
> ```C
> 
> /* TODO: Why on earth do we call loaded libraries entry points?!?
>  * I won't bother fixing this asm, it needs to be purged. */
> ```

this was a hack for multiple liboses  to load, I guess.
it is needed to call constructor instead (which isn’t needed at the moment).
___
*[Pal/src/host/Linux/db_exception.c, line 33 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIHkyNETtzQcNkAJzh:-MLJ-ckE1MrdbOgNyp_Y:b-cbmtb7) ([raw file](https://github.com/oscarlab/graphene/blob/ec270d07616f30a1bb5bdc83e193ff6d33c8f016/Pal/src/host/Linux/db_exception.c#L33)):*

No and hopefully there are no cursed hacks around it anywhere ...
</blockquote></details>

https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/x86_64/sigaction.c.html#74
for gdb. the code was copied from glibc.
___
*[Pal/src/host/Linux-SGX/sgx_thread.c, line 240 at r2](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLJ-zBnDXPsgnqEVAd2:-MLJ-zBnDXPsgnqEVAd3:b-5is2wg) ([raw file](https://github.com/oscarlab/graphene/blob/8df2a96c9c204c75bb8d677fc4f45ffcd38d0c35/Pal/src/host/Linux-SGX/sgx_thread.c#L240)):*
> ```C
>                      : "memory"
>     );
>     __builtin_unreachable();
> ```

nice clean up

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: 20 of 22 files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLHvp2w15d8T0j4KcHb:-MLJ4mvE6aLMpXWIhbTL:b-896fix) (no related file):*

@yamahata Could you review this PR? You wrote some of this inline asm.
</blockquote></details>

Done.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIufCJ9DG5sd3S0Nkf:-MLJ30L45uj6-PSNFhvy:bqk7lxb) (no related file):*

can you please share the combination of compiler version and optimization options to hit the issue?
</blockquote></details>

Ubuntu 16.04:
```
$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609
```
Options: just
`make -j8 WERROR=1 GLIBC_VERSION=2.27`
from Graphene, which I guess is just `-O2`?
Note that this is very subtle issue, so gcc might not always generate code hitting it.

---
___
*[Pal/src/db_rtld.c, line 960 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIyiE68bB_rkcaP1bV:-MLJ4IMzDhQqzeJw4hm8:b-xjgwxy) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/db_rtld.c#L960)):*

this was a hack for multiple liboses  to load, I guess.
it is needed to call constructor instead (which isn’t needed at the moment).
</blockquote></details>

Yeah, if that's required we should call global constructors, not entry points ...
___
*[Pal/src/host/Linux/db_exception.c, line 33 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIHkyNETtzQcNkAJzh:-MLJ4SFj2mF7GwGoba_g:b-b1t6xy) ([raw file](https://github.com/oscarlab/graphene/blob/ec270d07616f30a1bb5bdc83e193ff6d33c8f016/Pal/src/host/Linux/db_exception.c#L33)):*

https://code.woboq.org/userspace/glibc/sysdeps/unix/sysv/linux/x86_64/sigaction.c.html#74
for gdb. the code was copied from glibc.
</blockquote></details>

Still no idea why we would need a `nop` there ...
Anyway, gdb should have no problem with it.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @yamahata)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @yamahata)

___
*[Pal/include/arch/x86_64/pal_internal-arch.h, line 6 at r2](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLJEWT567_OcKjQYhnF:-MLJEWT567_OcKjQYhnG:b-q8cqri) ([raw file](https://github.com/oscarlab/graphene/blob/8df2a96c9c204c75bb8d677fc4f45ffcd38d0c35/Pal/include/arch/x86_64/pal_internal-arch.h#L6)):*
> ```objc
>  */
> 
> #ifndef _PAL_INTERNAL_ARCH_H
> ```

This is a reserved name in C, please fix it up when doing the final rebase.
___
*[Pal/include/arch/x86_64/pal_internal-arch.h, line 17 at r2](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLJG4WmBBxEzddtTHAm:-MLJG4WnEMKREKlcxzxT:bj5c5es) ([raw file](https://github.com/oscarlab/graphene/blob/8df2a96c9c204c75bb8d677fc4f45ffcd38d0c35/Pal/include/arch/x86_64/pal_internal-arch.h#L17)):*
> ```objc
>         "1:\n"                                      \
>         "sub %0, %%rsp\n"                           \
>         "orq $0, (%%rsp)\n"                         \
> ```

This gives me flashbacks from reading https://marcan.st/2017/12/debugging-an-evil-go-runtime-bug/. We have plans for supporting Go, so this may actually cause problems, I think. I'm not sure how to fix this though.
___
*[Pal/include/arch/x86_64/pal_internal-arch.h, line 25 at r2](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLJFg5Z1YbKPTcdkf4U:-MLJFg5_-ma7B-1hqtLS:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/8df2a96c9c204c75bb8d677fc4f45ffcd38d0c35/Pal/include/arch/x86_64/pal_internal-arch.h#L25)):*
> ```objc
>     )
> 
> #endif // _PAL_INTERNAL_ARCH_H
> ```

ditto
___
*[Pal/src/db_rtld.c, line 960 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIyiE68bB_rkcaP1bV:-MLJGXifEHryZZwXZipV:b-mivr5f) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/db_rtld.c#L960)):*
> this was a hack for multiple liboses to load, I guess.

but this doesn't work, it's noreturn (despite the fact someone called it in a loop...)
___
*[Pal/src/host/Linux/db_exception.c, line 33 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIHkyNETtzQcNkAJzh:-MLJH6YfE7gg_qGlmqZh:b-bcm145) ([raw file](https://github.com/oscarlab/graphene/blob/ec270d07616f30a1bb5bdc83e193ff6d33c8f016/Pal/src/host/Linux/db_exception.c#L33)):*

Still no idea why we would need a `nop` there ...
Anyway, gdb should have no problem with it.
</blockquote></details>

Yes, I totally don't get that comment. This isn't a call target, this `nop` is just a dangling byte.
___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLJHEu39SUc01Mokqa0:bq1wfgs) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*

Yeah and it will crash the process and if somehow it does not, we have a safety loop around it. Generally speaking exit syscall should succeed, this is just a safety net.
</blockquote></details>

But this looks like you tried to implement an infinite loop which sleeps the CPU (like kernels do). If you want to crash the process here then just issue `int3` or something similar, to be explicit.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: 21 of 22 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv, @mkow, and @yamahata)

___
*[Pal/include/arch/x86_64/pal_internal-arch.h, line 6 at r2](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLJEWT567_OcKjQYhnF:-MLKF_sr0wn26u5phYc2:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/8df2a96c9c204c75bb8d677fc4f45ffcd38d0c35/Pal/include/arch/x86_64/pal_internal-arch.h#L6)):*

This is a reserved name in C, please fix it up when doing the final rebase.
</blockquote></details>

Done.
___
*[Pal/include/arch/x86_64/pal_internal-arch.h, line 17 at r2](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLJG4WmBBxEzddtTHAm:-MLKENdjDKJ-suLdU8xv:bpp41ea) ([raw file](https://github.com/oscarlab/graphene/blob/8df2a96c9c204c75bb8d677fc4f45ffcd38d0c35/Pal/include/arch/x86_64/pal_internal-arch.h#L17)):*

This gives me flashbacks from reading https://marcan.st/2017/12/debugging-an-evil-go-runtime-bug/. We have plans for supporting Go, so this may actually cause problems, I think. I'm not sure how to fix this though.
</blockquote></details>

This only happens on the main thread in untrusted code, so this has nothing to do with small go stacks.
___
*[Pal/include/arch/x86_64/pal_internal-arch.h, line 25 at r2](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLJFg5Z1YbKPTcdkf4U:-MLKFbL6DaouMvVdCMMK:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/8df2a96c9c204c75bb8d677fc4f45ffcd38d0c35/Pal/include/arch/x86_64/pal_internal-arch.h#L25)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/db_rtld.c, line 960 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIyiE68bB_rkcaP1bV:-MLKEXZsCvOdsiYsB2gJ:bd85ebn) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/db_rtld.c#L960)):*

> this was a hack for multiple liboses to load, I guess.

but this doesn't work, it's noreturn (despite the fact someone called it in a loop...)
</blockquote></details>

No, if it was only to load our known subset of libraries, which we created, so we could have made their entry point return, it might have worked. That doesn't change the fact that this needs to be removed :)
___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLKEtPiB069mkwgjaWC:b-ewtxx2) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*

But this looks like you tried to implement an infinite loop which sleeps the CPU (like kernels do). If you want to crash the process here then just issue `int3` or something similar, to be explicit.
</blockquote></details>

I would very much like such loop, but this is not possible in ring 3, this is the least we can do to show our intentions. `int3` looks like a debug break, doesn't show what we want to achieve here: just stop the cpu, if the exit syscall failed, there's nothing more we can do.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @mkow, and @yamahata)

___
*[Pal/include/arch/x86_64/pal_internal-arch.h, line 17 at r2](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLJG4WmBBxEzddtTHAm:-MLKXhqj9_5m95LvqKcO:b-ce6wp8) ([raw file](https://github.com/oscarlab/graphene/blob/8df2a96c9c204c75bb8d677fc4f45ffcd38d0c35/Pal/include/arch/x86_64/pal_internal-arch.h#L17)):*

This only happens on the main thread in untrusted code, so this has nothing to do with small go stacks.
</blockquote></details>

Oh, right, then it should be fine.
___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLKZ73-7GxGpefaqt18:bfflos5) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*

I would very much like such loop, but this is not possible in ring 3, this is the least we can do to show our intentions. `int3` looks like a debug break, doesn't show what we want to achieve here: just stop the cpu, if the exit syscall failed, there's nothing more we can do.
</blockquote></details>

`int3` is nice because it actually stops the debugger, if present, and if not it crashes the process (same as `hlt`, but explicitly, and doesn't pretend that it's an inf busy loop).

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow and @yamahata)

___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLKaniUEUR5tDmN-JR_:bo0rugi) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*

`int3` is nice because it actually stops the debugger, if present, and if not it crashes the process (same as `hlt`, but explicitly, and doesn't pretend that it's an inf busy loop).
</blockquote></details>

1) `hlt` does not make cpu busy (exactly the opposite),
2) the idea was to look like loop halting the cpu, because we would very much like to do that (but we cannot),
3) if use app handles `SIGTRAP` then `int3` will just run the handler - not what we want.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @mkow, and @yamahata)

___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLKcgJ9BidwUgBwA3dS:b-sc8v06) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*
> hlt does not make cpu busy (exactly the opposite),

In ring0 yes, but I'd still call this busy loop.

> the idea was to look like loop halting the cpu, because we would very much like to do that (but we cannot),

For me this totally looks like an infinitely spinning loop, but it isn't and the code is misleading (it's only such a loop in ring0).

> if use app handles SIGTRAP then int3 will just run the handler - not what we want.

Same for the exception generated by `hlt`? I don't get this argument.

My main problem with this code is that it's _very_ misleading.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow and @yamahata)

___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLKjTjC2v4B3UVV_P03:biocb3g) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*
> For me this totally looks like an infinitely spinning loop, but it isn't and the code is misleading (it's only such a loop in ring0).

It should be a inf loop halting the cpu, but we cannot do that, so we just crash. So this code exactly shows the intention, I wouldn't call that misleading.
> Same for the exception generated by hlt? I don't get this argument.

The difference is that handling `int3` will resume the code execution after it and `hlt` before (effectively redoing segfault). That would make a difference if we removed the loop (which I guess is part of your suggestion).
>My main problem with this code is that it's very misleading.

Changing it to `int3` will make it ever more misleading, it will look like someone tried to do some weird debugger trap.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @mkow, and @yamahata)

___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLKnW0x18IrpYbo3Hgw:b-3wmr7v) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*
> It should be a inf loop halting the cpu, but we cannot do that, so we just crash. So this code exactly shows the intention, I wouldn't call that misleading.

Uh, but it's not doing this and never will, at least when implemented using `hlt`. We show the intention to the reader and then actually do something else.

> The difference is that handling `int3` will resume the code execution after it and `hlt` before (effectively redoing segfault). That would make a difference if we removed the loop (which I guess is part of your suggestion).

With the loop as you noticed it doesn't make a difference. The loop makes this code loop like `hlt` actually worked like in ring0.

> Changing it to int3 will make it ever more misleading, it will look like someone tried to do some weird debugger trap.

Which is a good thing? Because it will actually be a debugger trap (both look like a debugger trap and will be a debugger trap).

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow and @yamahata)

___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLKqOnH6rPSyO8ckaIe:bnqll7) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*
> Uh, but it's not doing this and never will, at least when implemented using hlt. We show the intention to the reader and then actually do something else.

We do exactly what the user sees.
> With the loop as you noticed it doesn't make a difference. The loop makes this code loop like hlt actually worked like in ring0.

The loop is there just as a sanity check, because it doesn't cost us anything to put.
> Which is a good thing? Because it will actually be a debugger trap (both look like a debugger trap and will be a debugger trap).

But we do not want a debugger trap!

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @mkow, and @yamahata)

___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLKsbqTBqDPZLpf0C9Q:bfax5wr) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*
> We do exactly what the user sees.

No, the reader sees "wait until next interrupt" instruction, but we actually throw an exception here because the context doesn't allow for this instruction.

> The loop is there just as a sanity check, because it doesn't cost us anything to put.

And it makes `hlt` confusing and `int3` working equally well, but more straightforward.

> But we do not want a debugger trap!

Why not? Even if, then why not just generate an exception explicitly instead of doing some shenanigans?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow and @yamahata)

___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLKv7L47z_oUzXjO35E:b-gej66l) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*
> No, the reader sees "wait until next interrupt" instruction, but we actually throw an exception here because the context doesn't allow for this instruction.

Since the use is aware we are in ring3, they can expect an exception.
> And it makes hlt confusing and int3 working equally well, but more straightforward.

I guess that's the whole point here: to me `int3` is confusing here, more than `hlt`. We could go for `ud2` as a middle ground, I guess.
> Why not? Even if, then why not just generate an exception explicitly instead of doing some shenanigans?

Doing a debugger exception not in a context where we expect to be under a debugger sounds like shenanigans to me.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @mkow, and @yamahata)

___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLKxGJ062ECJeSZEMVn:b1xwcuv) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*
> Since the use is aware we are in ring3, they can expect an exception.

You can use the same argument to defend any misleading code and just say that the reader knows everything and will notice all the traps.

> We could go for ud2 as a middle ground, I guess.

I'm fine with `ud2`.

> Doing a debugger exception not in a context where we expect to be under a debugger sounds like shenanigans to me.

How does it differ from any other exception? It's not even handled differently, any exception will stop the debugger here if it's present and if it's not it will kill the process. My point is: if you want to throw an exception then throw it explicitly, not write something which on first reading looks like an infinite loop which just quietly spins.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: 20 of 22 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @yamahata)

___
*[Pal/src/host/Linux/db_threading.c, line 262 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIIipz0BDIBN_x-Vol:-MLLB11y1DJJySSgUYw5:b-zbf9q5) ([raw file](https://github.com/oscarlab/graphene/blob/5c1e003c0ec77e94a30a05d12db777b3f2bfe88a/Pal/src/host/Linux/db_threading.c#L262)):*

> Since the use is aware we are in ring3, they can expect an exception.

You can use the same argument to defend any misleading code and just say that the reader knows everything and will notice all the traps.

> We could go for ud2 as a middle ground, I guess.

I'm fine with `ud2`.

> Doing a debugger exception not in a context where we expect to be under a debugger sounds like shenanigans to me.

How does it differ from any other exception? It's not even handled differently, any exception will stop the debugger here if it's present and if it's not it will kill the process. My point is: if you want to throw an exception then throw it explicitly, not write something which on first reading looks like an infinite loop which just quietly spins.
</blockquote></details>

I've changed it to `ud2` and moved loop start to before syscall. Cleaned up the asm a bit too.

yamahata (approved):
> Reviewed 18 of 21 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux/db_exception.c, line 33 at r1](https://reviewable.io/reviews/oscarlab/graphene/1938#-MLIHkyNETtzQcNkAJzh:-MLLVWn2DZODVr4ErR0-:bpmv2ym) ([raw file](https://github.com/oscarlab/graphene/blob/ec270d07616f30a1bb5bdc83e193ff6d33c8f016/Pal/src/host/Linux/db_exception.c#L33)):*

Yes, I totally don't get that comment. This isn't a call target, this `nop` is just a dangling byte.
</blockquote></details>

TBH I don't get the original comment. I'm guessing disassembler in gdb?

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

dimakuv (dismissed):
> Reviewed 1 of 1 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

dimakuv:
Jenkins, retest Jenkins please (`select04` LTP test failed, unrelated)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1938)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #1947, opened 2020-11-06
Allow redirecting debug logs to a file


pwmarcz:
I'm working with Graphene tests, and find myself switching `loader.debug_type = inline` and `loader.debug_type = none`. The default debug messages go to stdout, which makes them really hard to separate from normal program output.

I see that the default implementation opens `/dev/tty` as debug stream. There is also `loader.debug_type = file`, but it's broken: writing to a file in PAL (using `DkStreamWrite`) requires providing an absolute offset, and we pass 0, so new data overwrites the old.

## Possible solutions

* **Output logs to stderr**. Make `inline` output to `/dev/stderr` instead, or introduce `loader.debug_type = stderr`. That would already improve the situation, as it's easier to separate the logs. However, while I think it's worth doing (are there any arguments for *not* using stderr by deault? It seems to be a standard for logging), the output will be mixed with stderr from application, so it would be good to have an option to write to a file.

* **Keep track of the file offset in LibOS.** Unfortunately this requires synchronizing the offset across multiple threads, and also multiple processes. One way would be to introduce a separate thread in one of the processes, and use IPC to transfer messages to that process.

  I don't like this solution because it's complicated and pretty brittle: enabling debugging can cause things to go wrong in a harder-to-debug way; log messages might be delayed, and so on.

  If we wanted to avoid synchronizing between processes, we could write to multiple files. But I don't think that's a good solution; on the contrary, it makes it harder to debug communication between processes.

* **Allow `DkStreamWrite` to append to a file somehow.** I undestand this is not desirable to allow as it makes PAL less deterministic (although I'm not fully familiar with the argument). However, there is a `PAL_ACCESS_APPEND` option - is it supposed to do anything? There are also other resource types than "file" (e.g. device, pipe, or socket) that obviously allow such mode of output.

* **Introduce a special PAL API for logging**. In a simplest version, that would be something like `DkDebugLog(PAL_NUM count, PAL_PTR buffer)`. LibOS would just call this function, and PAL would be set up to write to stdout, stderr, or an open file (probably just by keeping a file descriptor internally and calling `write` on it).

  I like this one because I think "output a message" seems like a natural API for logging, much more so than "write to a file at an absolute offset". I think the API could also be extended with more information in the future (such as log type or level).

CC @mkow @donporter

mkow:
I like the option with a new PAL API.
[...]

PR #1965, opened 2020-11-13
[Pal,LibOS] Fix loader.debug_type = file

pwmarcz:
See https://github.com/oscarlab/graphene/issues/1947.

## Description of the changes 

This makes `loader.debug_type = "file"` usable again by introducing a new `DkDebugLog` call.

## How to test this PR? 

Run a program with:

```toml
loader.debug_type = "file"
loader.debug_file = "graphene.log"
```

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -28,13 +28,16 @@ Debug type
 
 ::
 
-    loader.debug_type = "[none|inline]"
+    loader.debug_type = "[none|inline|file]"
     (Default: "none")
+
+    loader.debug_file = "[PATH]"
 
 This specifies the debug option while running the library OS. If the debug type
 is ``none``, no debug output will be printed to standard output. If the debug
 type is ``inline``, a dmesg-like debug output will be printed inline with
-standard output.
+standard output. If the debug type is ``file``, debug output will be written to
+the file specified in ``loader.debug_file``.
 
 Preloaded libraries
 ^^^^^^^^^^^^^^^^^^^
--- a/Documentation/pal/host-abi.rst
+++ b/Documentation/pal/host-abi.rst
@@ -370,6 +370,9 @@ cryptographically-strong random bits, flush portions of instruction caches,
 increment and decrement the reference counts on objects shared between threads,
 and to obtain an attestation report and quote.
 
+.. doxygenfunction:: DkDebugLog
+   :project: pal
+
 .. doxygenfunction:: DkSystemTimeQuery
    :project: pal
 
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -36,7 +36,7 @@ struct debug_buf {
 #include "pal_debug.h"
 #include "pal_error.h"
 
-extern PAL_HANDLE debug_handle;
+extern bool g_debug_log_enabled;
 
 #include <stdarg.h>
 
@@ -47,7 +47,7 @@ void debug_vprintf(const char* fmt, va_list ap) __attribute__((format(printf, 1,
 
 #define debug(fmt, ...)                       \
     do {                                      \
-        if (debug_handle)                     \
+        if (g_debug_log_enabled)              \
             debug_printf(fmt, ##__VA_ARGS__); \
     } while (0)
 
@@ -120,11 +120,11 @@ static inline int64_t get_cur_preempt(void) {
     r func(PROTO_ARGS_##n(args));
 
 #define PARSE_SYSCALL1(name, ...) \
-    if (debug_handle)             \
+    if (g_debug_log_enabled)      \
         parse_syscall_before(__NR_##name, #name, ##__VA_ARGS__);
 
 #define PARSE_SYSCALL2(name, ...) \
-    if (debug_handle)             \
+    if (g_debug_log_enabled)      \
         parse_syscall_after(__NR_##name, #name, ##__VA_ARGS__);
 
 void parse_syscall_before(int sysno, const char* name, int nr, ...);
--- a/LibOS/shim/include/shim_thread.h
+++ b/LibOS/shim/include/shim_thread.h
@@ -137,7 +137,7 @@ void debug_setprefix(shim_tcb_t* tcb);
 
 static inline void debug_setbuf(shim_tcb_t* tcb,
                                 struct debug_buf* debug_buf) {
-    if (!debug_handle)
+    if (!g_debug_log_enabled)
         return;
 
     tcb->debug_buf = debug_buf ? debug_buf : malloc(sizeof(struct debug_buf));
--- a/LibOS/shim/include/shim_utils.h
+++ b/LibOS/shim/include/shim_utils.h
@@ -13,8 +13,6 @@
 
 struct shim_handle;
 
-void sysparser_printf(const char* fmt, ...);
-
 /* quick hash function based on Robert Jenkins' hash algorithm */
 static inline uint64_t hash64(uint64_t key) {
     key = (~key) + (key << 21);
--- a/LibOS/shim/src/shim_init.c
+++ b/LibOS/shim/src/shim_init.c
@@ -421,7 +421,7 @@ fail:
 extern PAL_HANDLE thread_start_event;
 
 noreturn void* shim_init(int argc, void* args) {
-    debug_handle = PAL_CB(debug_stream);
+    g_debug_log_enabled = PAL_CB(enable_debug_log);
     g_process_ipc_info.vmid = (IDTYPE)PAL_CB(process_id);
 
     /* create the initial TCB, shim can not be run without a tcb */
--- a/LibOS/shim/src/shim_parser.c
+++ b/LibOS/shim/src/shim_parser.c
@@ -482,10 +482,6 @@ static inline int is_pointer_or_long(const char* type) {
     do {                 \
         debug_putch(ch); \
     } while (0)
-#define VPRINTF(fmt, ap)        \
-    do {                        \
-        debug_vprintf(fmt, ap); \
-    } while (0)
 
 struct flag_table {
     const char *name;
@@ -556,15 +552,8 @@ static inline void skip_syscall_args(va_list* ap) {
         va_arg(*ap, int);
 }
 
-void sysparser_printf(const char* fmt, ...) {
-    va_list ap;
-    va_start(ap, fmt);
-    VPRINTF(fmt, ap);
-    va_end(ap);
-}
-
 void parse_syscall_before(int sysno, const char* name, int nr, ...) {
-    if (!debug_handle)
+    if (!g_debug_log_enabled)
         return;
 
     struct parser_table* parser = &syscall_parser_table[sysno];
@@ -600,7 +589,7 @@ dotdotdot:
 }
 
 void parse_syscall_after(int sysno, const char* name, int nr, ...) {
-    if (!debug_handle)
+    if (!g_debug_log_enabled)
         return;
 
     struct parser_table* parser = &syscall_parser_table[sysno];
--- a/LibOS/shim/src/sys/shim_exit.c
+++ b/LibOS/shim/src/sys/shim_exit.c
@@ -214,8 +214,7 @@ noreturn int shim_do_exit_group(int error_code) {
 
     error_code &= 0xFF;
 
-    if (debug_handle)
-        sysparser_printf("---- shim_exit_group (returning %d)\n", error_code);
+    debug("---- shim_exit_group (returning %d)\n", error_code);
 
     process_exit(error_code, 0);
 }
@@ -225,8 +224,7 @@ noreturn int shim_do_exit(int error_code) {
 
     error_code &= 0xFF;
 
-    if (debug_handle)
-        sysparser_printf("---- shim_exit (returning %d)\n", error_code);
+    debug("---- shim_exit (returning %d)\n", error_code);
 
     thread_exit(error_code, 0);
 }
--- a/LibOS/shim/src/utils/printf.c
+++ b/LibOS/shim/src/utils/printf.c
@@ -1,6 +1,7 @@
 /* SPDX-License-Identifier: LGPL-3.0-or-later */
 /* Copyright (C) 2014 Stony Brook University */
 
+#include <assert.h>
 #include <stdarg.h>
 #include <stdint.h>
 
@@ -10,13 +11,25 @@
 #include "shim_internal.h"
 #include "shim_ipc.h"
 
-PAL_HANDLE debug_handle = NULL;
+bool g_debug_log_enabled = false;
 
-static inline int debug_fputs(const char* buf, int len) {
-    if (DkStreamWrite(debug_handle, 0, len, (void*)buf, NULL) == (PAL_NUM)len)
-        return 0;
-    else
-        return -1;
+static inline int debug_fputs(const char* buf, size_t size) {
+    size_t bytes = 0;
+
+    while (bytes < size) {
+        PAL_NUM x = DkDebugLog((void*)(buf + bytes), size - bytes);
+        if (x == PAL_STREAM_ERROR) {
+            int err = PAL_ERRNO();
+            if (err == EINTR || err == EAGAIN || err == EWOULDBLOCK) {
+                continue;
+            }
+            return -err;
+        }
+
+        bytes += x;
+    }
+
+    return 0;
 }
 
 static int debug_fputch(void* f, int ch, void* b) {
@@ -98,7 +111,7 @@ void debug_printf(const char* fmt, ...) {
 }
 
 void debug_setprefix(shim_tcb_t* tcb) {
-    if (!debug_handle)
+    if (!g_debug_log_enabled)
         return;
 
     struct debug_buf* buf = tcb->debug_buf;
--- a/LibOS/shim/test/regression/.gitignore
+++ b/LibOS/shim/test/regression/.gitignore
@@ -16,6 +16,8 @@
 /bootstrap_static
 /cpuid
 /debug
+/debug_log_file
+/debug_log_inline
 /debug_regs-x86_64
 /dev
 /env_from_file
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -91,6 +91,8 @@ manifests = \
 	manifest \
 	argv_from_file.manifest \
 	attestation.manifest \
+	debug_log_file.manifest \
+	debug_log_inline.manifest \
 	env_from_file.manifest \
 	env_from_host.manifest \
 	eventfd.manifest \
@@ -117,6 +119,8 @@ exec_target = \
 	$(c_executables) \
 	$(cxx_executables) \
 	argv_from_file.manifest \
+	debug_log_file.manifest \
+	debug_log_inline.manifest \
 	env_from_file.manifest \
 	env_from_host.manifest \
 	file_check_policy_allow_all_but_log.manifest \
@@ -126,6 +130,8 @@ exec_target = \
 
 target = \
 	argv_from_file \
+	debug_log_file \
+	debug_log_inline \
 	env_from_host \
 	env_from_file \
 	file_check_policy_allow_all_but_log \
@@ -177,6 +183,9 @@ LDLIBS-fp_multithread += -lm
 # correspondence).
 
 argv_from_file env_from_host env_from_file: bootstrap
+	$(call cmd,ln_sfr)
+
+debug_log_file debug_log_inline: bootstrap
 	$(call cmd,ln_sfr)
 
 file_check_policy_allow_all_but_log file_check_policy_strict: file_check_policy
--- a/LibOS/shim/test/regression/debug_log_file.manifest.template
+++ b/LibOS/shim/test/regression/debug_log_file.manifest.template
@@ -1,0 +1,15 @@
+loader.argv0_override = "bootstrap"
+
+loader.preload = "file:../../src/libsysdb.so"
+loader.env.LD_LIBRARY_PATH = "/lib"
+loader.debug_type = "file"
+loader.debug_file = "tmp/debug_log_file.log"
+
+fs.mount.lib.type = "chroot"
+fs.mount.lib.path = "/lib"
+fs.mount.lib.uri = "file:../../../../Runtime"
+
+sgx.trusted_files.ld = "file:../../../../Runtime/ld-linux-x86-64.so.2"
+sgx.trusted_files.libc = "file:../../../../Runtime/libc.so.6"
+
+sgx.static_address = 1
--- a/LibOS/shim/test/regression/debug_log_inline.manifest.template
+++ b/LibOS/shim/test/regression/debug_log_inline.manifest.template
@@ -1,0 +1,14 @@
+loader.argv0_override = "bootstrap"
+
+loader.preload = "file:../../src/libsysdb.so"
+loader.env.LD_LIBRARY_PATH = "/lib"
+loader.debug_type = "inline"
+
+fs.mount.lib.type = "chroot"
+fs.mount.lib.path = "/lib"
+fs.mount.lib.uri = "file:../../../../Runtime"
+
+sgx.trusted_files.ld = "file:../../../../Runtime/ld-linux-x86-64.so.2"
+sgx.trusted_files.libc = "file:../../../../Runtime/libc.so.6"
+
+sgx.static_address = 1
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -47,8 +47,7 @@ class TC_01_Bootstrap(RegressionTestCase):
         with open('argv_test_input', 'wb') as f:
             f.write(result.stdout)
         try:
-            manifest = self.get_manifest('argv_from_file')
-            stdout, _ = self.run_binary([manifest, 'WRONG', 'ARGUMENTS'])
+            stdout, _ = self.run_binary(['argv_from_file', 'WRONG', 'ARGUMENTS'])
             self.assertIn('# of arguments: %d\n' % len(args), stdout)
             for i, arg in enumerate(args):
                 self.assertIn('argv[%d] = %s\n' % (i, arg), stdout)
@@ -62,8 +61,7 @@ class TC_01_Bootstrap(RegressionTestCase):
             'some weir:d\nvar_name': ' even we\nirder\tvalue',
         }
         manifest_envs = {'LD_LIBRARY_PATH': '/lib'}
-        manifest = self.get_manifest('env_from_host')
-        stdout, _ = self.run_binary([manifest], env=host_envs)
+        stdout, _ = self.run_binary(['env_from_host'], env=host_envs)
         self.assertIn('# of envs: %d\n' % (len(host_envs) + len(manifest_envs)), stdout)
         for _, (key, val) in enumerate({**host_envs, **manifest_envs}.items()):
             # We don't enforce any specific order of envs, so we skip checking the index.
@@ -78,8 +76,7 @@ class TC_01_Bootstrap(RegressionTestCase):
         with open('env_test_input', 'wb') as f:
             f.write(result.stdout)
         try:
-            manifest = self.get_manifest('env_from_file')
-            stdout, _ = self.run_binary([manifest], env=host_envs)
+            stdout, _ = self.run_binary(['env_from_file'], env=host_envs)
             self.assertIn('# of envs: %d\n' % (len(envs) + len(manifest_envs)), stdout)
             for _, arg in enumerate(envs + manifest_envs):
                 # We don't enforce any specific order of envs, so we skip checking the index.
@@ -199,8 +196,7 @@ class TC_01_Bootstrap(RegressionTestCase):
     @unittest.skipUnless(HAS_SGX, 'This test relies on SGX-specific manifest options.')
     def test_501_init_fail2(self):
         try:
-            manifest = self.get_manifest('init_fail2')
-            self.run_binary([manifest], timeout=60)
+            self.run_binary(['init_fail2'], timeout=60)
             self.fail('expected to return nonzero (and != 42)')
         except subprocess.CalledProcessError as e:
             self.assertNotEqual(e.returncode, 42, 'expected returncode != 42')
@@ -213,8 +209,7 @@ class TC_01_Bootstrap(RegressionTestCase):
 
     @unittest.skipUnless(HAS_SGX, 'This test is only meaningful on SGX PAL')
     def test_601_multi_pthread_exitless(self):
-        manifest = self.get_manifest('multi_pthread_exitless')
-        stdout, _ = self.run_binary([manifest], timeout=60)
+        stdout, _ = self.run_binary(['multi_pthread_exitless'], timeout=60)
 
         # Multiple thread creation
         self.assertIn('128 Threads Created', stdout)
@@ -231,8 +226,7 @@ class TC_01_Bootstrap(RegressionTestCase):
         self.assertIn('FE_TOWARDZERO parent: 42.5 = 42.0, -42.5 = -42.0', stdout)
 
     def test_700_debug_log_inline(self):
-        manifest = self.get_manifest('debug_log_inline')
-        stdout, _ = self.run_binary([manifest])
+        stdout, _ = self.run_binary(['debug_log_inline'])
         self._verify_debug_log(stdout)
 
     def test_701_debug_log_file(self):
@@ -240,8 +234,7 @@ class TC_01_Bootstrap(RegressionTestCase):
         if os.path.exists(log_path):
             os.remove(log_path)
 
-        manifest = self.get_manifest('debug_log_file')
-        self.run_binary([manifest])
+        self.run_binary(['debug_log_file'])
 
         with open(log_path) as log_file:
             log = log_file.read()
@@ -270,27 +263,25 @@ class TC_02_OpenMP(RegressionTestCase):
     'only relevant to SGX.')
 class TC_03_FileCheckPolicy(RegressionTestCase):
     def test_000_strict_success(self):
-        manifest = self.get_manifest('file_check_policy_strict')
-        stdout, _ = self.run_binary([manifest, 'trusted_testfile'])
+        stdout, _ = self.run_binary(['file_check_policy_strict', 'trusted_testfile'])
 
         self.assertIn('file_check_policy succeeded', stdout)
 
     def test_001_strict_fail(self):
-        manifest = self.get_manifest('file_check_policy_strict')
         with self.expect_returncode(2):
-            self.run_binary([manifest, 'unknown_testfile'])
+            self.run_binary(['file_check_policy_strict', 'unknown_testfile'])
 
     def test_002_allow_all_but_log_success(self):
-        manifest = self.get_manifest('file_check_policy_allow_all_but_log')
-        stdout, stderr = self.run_binary([manifest, 'unknown_testfile'])
+        stdout, stderr = self.run_binary(['file_check_policy_allow_all_but_log',
+                                          'unknown_testfile'])
 
         self.assertIn('Allowing access to an unknown file due to file_check_policy settings: '
                       'file:unknown_testfile', stderr)
         self.assertIn('file_check_policy succeeded', stdout)
 
     def test_003_allow_all_but_log_fail(self):
-        manifest = self.get_manifest('file_check_policy_allow_all_but_log')
-        stdout, stderr = self.run_binary([manifest, 'trusted_testfile'])
+        stdout, stderr = self.run_binary(['file_check_policy_allow_all_but_log',
+                                          'trusted_testfile'])
 
         self.assertNotIn('Allowing access to an unknown file due to file_check_policy settings: '
                          'file:trusted_testfile', stderr)
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -230,6 +230,30 @@ class TC_01_Bootstrap(RegressionTestCase):
         self.assertIn('FE_TOWARDZERO  child: 42.5 = 42.0, -42.5 = -42.0', stdout)
         self.assertIn('FE_TOWARDZERO parent: 42.5 = 42.0, -42.5 = -42.0', stdout)
 
+    def test_700_debug_log_inline(self):
+        manifest = self.get_manifest('debug_log_inline')
+        stdout, _ = self.run_binary([manifest])
+        self._verify_debug_log(stdout)
+
+    def test_701_debug_log_file(self):
+        log_path = 'tmp/debug_log_file.log'
+        if os.path.exists(log_path):
+            os.remove(log_path)
+
+        manifest = self.get_manifest('debug_log_file')
+        self.run_binary([manifest])
+
+        with open(log_path) as log_file:
+            log = log_file.read()
+
+        self._verify_debug_log(log)
+
+    def _verify_debug_log(self, log: str):
+        self.assertIn('Host:', log)
+        self.assertIn('Shim process initialized', log)
+        self.assertIn('--- shim_exit_group', log)
+
+
 @unittest.skipUnless(HAS_SGX,
     'This test is only meaningful on SGX PAL because only SGX catches raw '
     'syscalls and redirects to Graphene\'s LibOS. If we will add seccomp to '
--- a/Pal/include/pal/pal.h
+++ b/Pal/include/pal/pal.h
@@ -134,7 +134,7 @@ typedef struct PAL_CONTROL_ {
     PAL_STR executable;         /*!< executable name */
     PAL_HANDLE parent_process;  /*!< handle of parent process */
     PAL_HANDLE first_thread;    /*!< handle of first thread */
-    PAL_HANDLE debug_stream;    /*!< debug stream */
+    PAL_BOL enable_debug_log;   /*!< enable debug log calls */
 
     /*
      * Memory layout
@@ -361,6 +361,9 @@ PAL_NUM DkStreamRead(PAL_HANDLE handle, PAL_NUM offset, PAL_NUM count, PAL_PTR b
  *
  * If the handle is a file, `offset` must be specified at each call of DkStreamWrite. `dest` can be
  * used to specify the remote socket address if the handle is a UDP socket.
+ *
+ * \return number of bytes written if succeeded, PAL_STREAM_ERROR on failure (in which case
+ *  PAL_ERRNO() is set)
  */
 PAL_NUM DkStreamWrite(PAL_HANDLE handle, PAL_NUM offset, PAL_NUM count, PAL_PTR buffer,
                       PAL_STR dest);
@@ -677,6 +680,17 @@ void DkObjectClose(PAL_HANDLE objectHandle);
 /*
  * MISC
  */
+
+/*!
+ * \brief Output a message to the debug stream.
+ *
+ * Works only if the debug stream has been initialized, which can be checked by looking at
+ * `g_pal_control.enable_debug_log`.
+ *
+ * \return number of bytes written if succeeded, PAL_STREAM_ERROR on failure (in which case
+ *  PAL_ERRNO() is set)
+ */
+PAL_NUM DkDebugLog(PAL_PTR buffer, PAL_NUM size);
 
 /*!
  * \brief Get the current time
--- a/Pal/regression/Bootstrap.c
+++ b/Pal/regression/Bootstrap.c
@@ -38,7 +38,7 @@ int main(int argc, char** argv, char** envp) {
 
     /* test debug stream */
     char* msg = "Written to Debug Stream\n";
-    DkStreamWrite(pal_control.debug_stream, 0, strlen(msg), msg, NULL);
+    DkDebugLog(msg, strlen(msg));
 
     /* Allocation Alignment */
     pal_printf("Allocation Alignment: %ld\n", pal_control.alloc_align);
--- a/Pal/regression/test_pal.py
+++ b/Pal/regression/test_pal.py
@@ -190,13 +190,13 @@ class TC_01_Bootstrap(RegressionTestCase):
 
     def test_104_manifest_as_executable_name(self):
         manifest = self.get_manifest('Bootstrap2')
-        _, stderr = self.run_binary([manifest])
+        _, stderr = self.run_binary(['Bootstrap2'])
         self.assertIn('User Program Started', stderr)
         self.assertIn('Loaded Manifest: file:' + manifest, stderr)
 
     def test_105_manifest_as_argument(self):
         manifest = self.get_manifest('Bootstrap4')
-        _, stderr = self.run_binary([manifest])
+        _, stderr = self.run_binary(['Bootstrap4'])
         self.assertIn('Loaded Manifest: file:' + manifest, stderr)
         self.assertIn('Loaded Executable: file:Bootstrap', stderr)
 
@@ -210,12 +210,12 @@ class TC_01_Bootstrap(RegressionTestCase):
     @unittest.skipUnless(HAS_SGX, 'this test requires SGX')
     def test_120_8gb_enclave(self):
         manifest = self.get_manifest('Bootstrap6')
-        _, stderr = self.run_binary([manifest], timeout=360)
+        _, stderr = self.run_binary(['Bootstrap6'], timeout=360)
         self.assertIn('Loaded Manifest: file:' + manifest, stderr)
         self.assertIn('Executable Range OK', stderr)
 
     def test_130_large_number_of_items_in_manifest(self):
-        _, stderr = self.run_binary([self.get_manifest('Bootstrap7')])
+        _, stderr = self.run_binary(['Bootstrap7'])
         self.assertIn('key1000=na', stderr)
         self.assertIn('key1=na', stderr)
 
@@ -566,8 +566,7 @@ class TC_20_SingleProcess(RegressionTestCase):
 
     @unittest.skipUnless(HAS_SGX, 'This test is only meaningful on SGX PAL')
     def test_511_thread2_exitless(self):
-        manifest = self.get_manifest('Thread2_exitless')
-        _, stderr = self.run_binary([manifest], timeout=60)
+        _, stderr = self.run_binary(['Thread2_exitless'], timeout=60)
 
         # Thread Cleanup: Exit by return.
         self.assertIn('Thread 2 ok.', stderr)
--- a/Pal/src/db_main.c
+++ b/Pal/src/db_main.c
@@ -170,19 +170,21 @@ static void set_debug_type(void) {
     if (!debug_type)
         return;
 
-    PAL_HANDLE handle = NULL;
+    bool enable_debug_log = false;
 
     if (!strcmp(debug_type, "inline")) {
-        ret = _DkStreamOpen(&handle, URI_PREFIX_DEV "tty", PAL_ACCESS_WRONLY, 0, 0, 0);
+        // TODO: use /dev/stderr instead?
+        ret = _DkInitDebugStream("/dev/stdout");
+        enable_debug_log = true;
     } else if (!strcmp(debug_type, "file")) {
         char* debug_file = NULL;
         ret = toml_string_in(g_pal_state.manifest_root, "loader.debug_file", &debug_file);
         if (ret < 0 || !debug_file)
             INIT_FAIL_MANIFEST(PAL_ERROR_DENIED, "Cannot find/parse \'loader.debug_file\'");
 
-        ret = _DkStreamOpen(&handle, debug_file, PAL_ACCESS_WRONLY, PAL_SHARE_OWNER_W,
-                            PAL_CREATE_TRY, 0);
+        ret = _DkInitDebugStream(debug_file);
         free(debug_file);
+        enable_debug_log = true;
     } else if (!strcmp(debug_type, "none")) {
         ret = 0;
     } else {
@@ -195,7 +197,7 @@ static void set_debug_type(void) {
     if (ret < 0)
         INIT_FAIL(-ret, "Cannot open debug stream");
 
-    g_pal_control.debug_stream = handle;
+    g_pal_control.enable_debug_log = enable_debug_log;
 }
 
 /* Loads a file containing a concatenation of C-strings. The resulting array of pointers is
--- a/Pal/src/db_streams.c
+++ b/Pal/src/db_streams.c
@@ -751,3 +751,15 @@ const char* _DkStreamRealpath(PAL_HANDLE hdl) {
 
     return ops->getrealpath(hdl);
 }
+
+PAL_NUM DkDebugLog(PAL_PTR buffer, PAL_NUM size) {
+    ENTER_PAL_CALL(DkDebugLog);
+
+    ssize_t ret = _DkDebugLog(buffer, size);
+    if (ret < 0) {
+        _DkRaiseFailure(-ret);
+        ret = PAL_STREAM_ERROR;
+    }
+
+    LEAVE_PAL_CALL_RETURN(ret);
+}
--- a/Pal/src/host/Linux-SGX/db_streams.c
+++ b/Pal/src/host/Linux-SGX/db_streams.c
@@ -33,6 +33,8 @@ typedef __kernel_pid_t pid_t;
 #define DUMMYPAYLOAD     "dummypayload"
 #define DUMMYPAYLOADSIZE (sizeof(DUMMYPAYLOAD))
 
+static int g_debug_fd = -1;
+
 struct hdl_header {
     uint8_t fds;       /* bitmask of host file descriptors corresponding to PAL handle */
     size_t  data_size; /* total size of serialized PAL handle */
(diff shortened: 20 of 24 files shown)


mkow (changes requested):
> Reviewed 18 of 18 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1IjH1AB-k7plltiRY:-MM1IjH1AB-k7plltiRZ:bdygloh) (no related file):*
The last fixup commit ("fixup! [LibOS,Pal,Examples,GSC,Docs] Move manifest parsing to TOML") targets a wrong one (from another PR).

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1JwgA3dFHECd9sWrg:-MM1JwgA3dFHECd9sWrh:b-gbv1dh) (no related file):*
Could you add a test for it, so that we won't break this again? You can probably reuse the `Bootstrap` binary from LibOS regression, as other tests often do.

---
___
*[Pal/src/host/Linux/db_streams.c, line 31 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1ca9bDXG8kjUESJKu:-MM1ca9bDXG8kjUESJKv:bnswi30) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/host/Linux/db_streams.c#L31)):*
> ```C
> #include <sys/socket.h>
> 
> static int g_debug_fd;
> ```

Could you initialize it to -1? (for the `if` in `_DkDebugLog`)
___
*[Pal/src/host/Linux/db_streams.c, line 381 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1_sfb8atqeWyeHXXx:-MM1_sfd3Wjr8NBsBw1E:bmnb1ps) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/host/Linux/db_streams.c#L381)):*
> ```C
> }
> 
> int _DkDebugLog(const void *buf, int size) {
> ```

`void *buf` -> `void* buf`
`int size` -> `size_t size`
___
*[Pal/src/host/Linux-SGX/db_streams.c, line 36 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1dJ_3EkZFQaA91E35:-MM1dJ_3EkZFQaA91E36:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/host/Linux-SGX/db_streams.c#L36)):*
> ```C
> #define DUMMYPAYLOADSIZE (sizeof(DUMMYPAYLOAD))
> 
> static int g_debug_fd;
> ```

ditto
___
*[Pal/src/host/Linux-SGX/db_streams.c, line 415 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1dKPjDUimunVgKoPZ:-MM1dKPjDUimunVgKoP_:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/host/Linux-SGX/db_streams.c#L415)):*
> ```C
> }
> 
> int _DkDebugLog(const void* buf, int size) {
> ```

ditto
___
*[Pal/src/host/Skeleton/db_streams.c, line 41 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1dc1J00qNvUW-zlnz:-MM1dc1J00qNvUW-zlo-:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/host/Skeleton/db_streams.c#L41)):*
> ```C
> }
> 
> int _DkDebugLog(const void *buf, int size) {
> ```

ditto

dimakuv (changes requested):
> Reviewed 18 of 18 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1JwgA3dFHECd9sWrg:-MMFJVij0QCl75p12Qtb:b-arytjs) (no related file):*

Could you add a test for it, so that we won't break this again? You can probably reuse the `Bootstrap` binary from LibOS regression, as other tests often do.
</blockquote></details>

+1, need a test for it.

---
___
*[LibOS/shim/src/utils/printf.c, line 16 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMFG1l6CU695Tf8tC1r:-MMFG1l773_RvjahATq6:b73tdpv) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/LibOS/shim/src/utils/printf.c#L16)):*
> ```C
> 
> static inline int debug_fputs(const char* buf, int len) {
>     if (DkDebugLog((void*)buf, len) == (PAL_NUM)len)
> ```

Maybe better write in a while loop until everything is written out? This will cover cases of partial writes due to interrupts. You can look at how Graphene calls `DkStreamWrite()` for inspiration.
___
*[Pal/include/pal/pal.h, line 685 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMFJHr81goNOHKcywUP:-MMFJHr81goNOHKcywUQ:b-rb4w96) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/include/pal/pal.h#L685)):*
> ```objc
>  *
>  * Works only if the debug stream has been initialized, which can be checked by looking at
>  * `g_pal_control.enable_debug_log`.
> ```

Please add a short description of what the function returns (see `DkStreamWrite()` for inspiration).
___
*[Pal/src/db_streams.c, line 760 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMFIDj65s2xS0iPE0Qc:-MMFIDj65s2xS0iPE0Qd:bh5wghl) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/db_streams.c#L760)):*
> ```C
>     int ret = _DkDebugLog(buffer, size);
> 
>     LEAVE_PAL_CALL_RETURN(ret);
> ```

This is wrong semantics. In the stream PAL API, we return `PAL_STREAM_ERROR` on failures, and always set the `PAL_ERRNO` via `_DkRiaseFailure(-ret)`. Please change the return value here exactly as in `DkStreamWrite()`.
___
*[Pal/src/pal_internal.h, line 341 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMFHbdh5-6h83XDFHLB:-MMFHbdh5-6h83XDFHLC:b-u6erzh) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/pal_internal.h#L341)):*
> ```objc
> 
> int _DkInitDebugStream(const char* path);
> int _DkDebugLog(const void* buf, int size);
> ```

`int size` -> `size_t size`

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: 11 of 23 files reviewed, 11 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1IjH1AB-k7plltiRY:-MMFXmp63aJiKNWUzUTb:biq08i4) (no related file):*

The last fixup commit ("fixup! [LibOS,Pal,Examples,GSC,Docs] Move manifest parsing to TOML") targets a wrong one (from another PR).
</blockquote></details>

Ouch, right. I squashed these initial commits (but kept the new changes as fixup).

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1JwgA3dFHECd9sWrg:-MMFeWFTCCNLG8Wi7cO9:bkz4b3f) (no related file):*

+1, need a test for it.
</blockquote></details>

Done. I added two tests: `debug_log_inline` and `debug_log_file`.

---
___
*[LibOS/shim/src/utils/printf.c, line 16 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMFG1l6CU695Tf8tC1r:-MMFVLka-FkGmETgTvbZ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/LibOS/shim/src/utils/printf.c#L16)):*

Maybe better write in a while loop until everything is written out? This will cover cases of partial writes due to interrupts. You can look at how Graphene calls `DkStreamWrite()` for inspiration.
</blockquote></details>

Done.
___
*[Pal/include/pal/pal.h, line 685 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMFJHr81goNOHKcywUP:-MMFW2lx5Ze8pZGPIrAQ:b14bpnm) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/include/pal/pal.h#L685)):*

Please add a short description of what the function returns (see `DkStreamWrite()` for inspiration).
</blockquote></details>

Done. This was also not specified for `DkStreamWrite`, so I added it to the doc comment there as well.
___
*[Pal/src/db_streams.c, line 760 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMFIDj65s2xS0iPE0Qc:-MMFXCWM3kMQOzCqxSTC:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/db_streams.c#L760)):*

This is wrong semantics. In the stream PAL API, we return `PAL_STREAM_ERROR` on failures, and always set the `PAL_ERRNO` via `_DkRiaseFailure(-ret)`. Please change the return value here exactly as in `DkStreamWrite()`.
</blockquote></details>

Done.
___
*[Pal/src/pal_internal.h, line 341 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMFHbdh5-6h83XDFHLB:-MMFY6rc2htNjvDpPi-e:b-mrt6yr) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/pal_internal.h#L341)):*

`int size` -> `size_t size`
</blockquote></details>

Done. I changed the code to use `size_t` and `ssize_t` consistently.
___
*[Pal/src/host/Linux/db_streams.c, line 31 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1ca9bDXG8kjUESJKu:-MMFYIg0CuZe6ucndKFx:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/host/Linux/db_streams.c#L31)):*

Could you initialize it to -1? (for the `if` in `_DkDebugLog`)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_streams.c, line 381 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1_sfb8atqeWyeHXXx:-MMFYYYH30vlHp2Wiyvn:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/host/Linux/db_streams.c#L381)):*

`void *buf` -> `void* buf`
`int size` -> `size_t size`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_streams.c, line 36 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1dJ_3EkZFQaA91E35:-MMFYd4Q80Z_e0nSLXTx:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/host/Linux-SGX/db_streams.c#L36)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_streams.c, line 415 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1dKPjDUimunVgKoPZ:-MMFYe1W6l2fXZDA7FZ4:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/host/Linux-SGX/db_streams.c#L415)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Skeleton/db_streams.c, line 41 at r1](https://reviewable.io/reviews/oscarlab/graphene/1965#-MM1dc1J00qNvUW-zlnz:-MMFYf4W3L4_cpXttRCT:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/97318ace65b9ab15b529d55f5985f2c31a9e4aae/Pal/src/host/Skeleton/db_streams.c#L41)):*

ditto
</blockquote></details>

Done.

dimakuv (changes requested):
> Reviewed 12 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[LibOS/shim/test/regression/test_libos.py, line 269 at r2](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMFgN7V7yOe8LB1_Fwi:-MMFgN7W2dtJ4HH--i61:b-r2nngl) ([raw file](https://github.com/oscarlab/graphene/blob/b26da9ea26443f139423aa2d918f521e219e9c38/LibOS/shim/test/regression/test_libos.py#L269)):*
> ```Python
>         self.assertIn('Host: Linux', log)
>         self.assertIn('Shim process initialized', log)
>         self.assertIn('Shim process initialized', log)
> ```

Why do you have the same check twice?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: 22 of 23 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/test/regression/test_libos.py, line 269 at r2](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMFgN7V7yOe8LB1_Fwi:-MMFx8Pb41E31YWfjMBQ:b7n7dfn) ([raw file](https://github.com/oscarlab/graphene/blob/b26da9ea26443f139423aa2d918f521e219e9c38/LibOS/shim/test/regression/test_libos.py#L269)):*

Why do you have the same check twice?
</blockquote></details>

I was too quick :) Fixed now.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (changes requested):
> Reviewed 11 of 12 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/test/regression/test_libos.py, line 249 at r3](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMGkfE_77xu6fPoJml9:-MMGkfEaDsLK7AwREAPw:bqu9m7t) ([raw file](https://github.com/oscarlab/graphene/blob/2755beaeaa7083a20dfd10ea002b0c82c4260a80/LibOS/shim/test/regression/test_libos.py#L249)):*
> ```Python
> 
>     def test_700_debug_log_inline(self):
>         manifest = self.get_manifest('debug_log_inline')
> ```

I think this isn't needed anymore, you can just run `debug_log_inline` directly and it should work (ditto below)
___
*[LibOS/shim/test/regression/test_libos.py, line 266 at r3](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMGl7Nf1e-cVo8g9k_d:-MMGl7Nf1e-cVo8g9k_e:b-5geo6n) ([raw file](https://github.com/oscarlab/graphene/blob/2755beaeaa7083a20dfd10ea002b0c82c4260a80/LibOS/shim/test/regression/test_libos.py#L266)):*
> ```Python
>         self._verify_debug_log(log)
> 
>     def _verify_debug_log(self, log: str):
> ```

Won't these tests fail when testing without `DEBUG=1`? Could you verify in such builds that the log doesn't exist/doesn't contain those lines?
___
*[LibOS/shim/test/regression/test_libos.py, line 267 at r3](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMGl2GU8d12y-pEa7nN:-MMGl2GVANpzHnYz02-l:bnwn4r7) ([raw file](https://github.com/oscarlab/graphene/blob/2755beaeaa7083a20dfd10ea002b0c82c4260a80/LibOS/shim/test/regression/test_libos.py#L267)):*
> ```Python
> 
>     def _verify_debug_log(self, log: str):
>         self.assertIn('Host: Linux', log)
> ```

This will fail on non-Linux backends.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: 22 of 24 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/test/regression/test_libos.py, line 249 at r3](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMGkfE_77xu6fPoJml9:-MMGrNDFBk2-_nqeocUF:b-bzgauu) ([raw file](https://github.com/oscarlab/graphene/blob/2755beaeaa7083a20dfd10ea002b0c82c4260a80/LibOS/shim/test/regression/test_libos.py#L249)):*

I think this isn't needed anymore, you can just run `debug_log_inline` directly and it should work (ditto below)
</blockquote></details>

Okay, but since I did this by copying other tests, I'd like to fix them too. I added a separate commit for that.
___
*[LibOS/shim/test/regression/test_libos.py, line 266 at r3](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMGl7Nf1e-cVo8g9k_d:-MMGpfyb6RhpjYshpSP-:b-w6e2cb) ([raw file](https://github.com/oscarlab/graphene/blob/2755beaeaa7083a20dfd10ea002b0c82c4260a80/LibOS/shim/test/regression/test_libos.py#L266)):*

Won't these tests fail when testing without `DEBUG=1`? Could you verify in such builds that the log doesn't exist/doesn't contain those lines?
</blockquote></details>

`DEBUG=1` doesn't control log visibility in LibOS, and these messages do show up.

(it does in SGX PAL... it's kind of a mess, but for now I would rather keep the test this way).
___
*[LibOS/shim/test/regression/test_libos.py, line 267 at r3](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMGl2GU8d12y-pEa7nN:-MMGrBkoB1RF5HBietlU:bs11tqm) ([raw file](https://github.com/oscarlab/graphene/blob/2755beaeaa7083a20dfd10ea002b0c82c4260a80/LibOS/shim/test/regression/test_libos.py#L267)):*

This will fail on non-Linux backends.
</blockquote></details>

Changed to look for `'Host:'`.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

yamahata (approved):
> Reviewed 11 of 18 files at r1, 11 of 12 files at r2, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (approved):
> Reviewed 4 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

yamahata (dismissed):
> Reviewed 4 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

boryspoplawski (changes requested):
> Reviewed 10 of 18 files at r1, 9 of 12 files at r2, 1 of 2 files at r4, 4 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_internal.h, line 39 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIstXtBKFObCXxGobx:-MMIstXtBKFObCXxGoby:bejbvae) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/LibOS/shim/include/shim_internal.h#L39)):*
> ```objc
> #include "pal_error.h"
> 
> extern bool g_enable_debug_log;
> ```

Imo `g_debug_log_enabled` would be better
___
*[LibOS/shim/src/utils/printf.c, line 20 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMItuHl5hP0j1kq5qVY:-MMItuHm1XhR8jw6L5Jr:b-ljgg9l) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/LibOS/shim/src/utils/printf.c#L20)):*
> ```C
> 
>     while (bytes < size) {
>         PAL_NUM x = DkDebugLog((void*)(buf + bytes), size - bytes);
> ```

This cast is superfluous.
___
*[Pal/include/pal/pal.h, line 365 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIwowR-WbO_fYdiZQi:-MMIwowSBfzfwqUZbnbq:bsdmnkv) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/include/pal/pal.h#L365)):*
> ```objc
>  * used to specify the remote socket address if the handle is a UDP socket.
>  *
>  * \return Returns number of bytes written if succeeded, and PAL_STREAM_ERROR on failure (in which
> ```

Imo you should skip `Returns` and `,` before `and`.
___
*[Pal/include/pal/pal.h, line 690 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIwzKp5F6LN4zwUTjg:-MMIwzKp5F6LN4zwUTjh:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/include/pal/pal.h#L690)):*
> ```objc
>  * `g_pal_control.enable_debug_log`.
>  *
>  * \return Returns number of bytes written if succeeded, and PAL_STREAM_ERROR on failure (in which
> ```

ditto
___
*[Pal/src/host/Linux/db_streams.c, line 369 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyyjd-sNapDahEqM6:-MMIyyjd-sNapDahEqM7:boumj68) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L369)):*
> ```C
> }
> 
> int _DkInitDebugStream(const char* path) {
> ```

Please add a check that `g_debug_fd` is not yet initialized and fail if it is (or just close the old fd).
___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMIyeQl5zCGzr5Eww2-:b-wkpgcp) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*
> ```C
> 
> int _DkInitDebugStream(const char* path) {
>     int ret = INLINE_SYSCALL(open, 3, path, O_WRONLY | O_APPEND | O_CREAT, 0600);
> ```

Please change this to `S_*` constants instead of hardcoded number.
___
*[Pal/src/host/Linux-SGX/db_streams.c, line 403 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIzWnRCi4dAljPesOR:-MMIzWnRCi4dAljPesOS:b-5dfur0) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux-SGX/db_streams.c#L403)):*
> ```C
> }
> 
> int _DkInitDebugStream(const char* path) {
> ```

ditto (initialization check)
___
*[Pal/src/host/Linux-SGX/db_streams.c, line 404 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIzZ1J252y_qZiJZjI:-MMIzZ1J252y_qZiJZjJ:b-2bf1k1) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux-SGX/db_streams.c#L404)):*
> ```C
> 
> int _DkInitDebugStream(const char* path) {
>     int ret = ocall_open(path, O_WRONLY | O_APPEND | O_CREAT, 0600);
> ```

ditto (hardcoded number)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski and @pwmarcz)

___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMJ-22r7yQunvb3hUpy:buqrnp3) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*

Please change this to `S_*` constants instead of hardcoded number.
</blockquote></details>

Please don't. Those `S_*` constants are unreadable mess, even Linux community is against them. See https://lwn.net/Articles/696231/.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @pwmarcz)

___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMJ1H0fA73vv8_u07h8:b-iy3zm5) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*

Please don't. Those `S_*` constants are unreadable mess, even Linux community is against them. See https://lwn.net/Articles/696231/.
</blockquote></details>

I disagree. While `S_*` constant are a mess (I don't like them neither), they are still better than pure octal numbers. I've already seen missing leading `0` (in Graphene too, I believe). While `PERM_rw_______` would look much nicer, we don't have such macros currently.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski and @pwmarcz)

___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMJ6UR09ajghtgqoe95:bl2bsm5) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*
> they are still better than pure octal numbers.

I totally disagree, everyone can read octal permissions on Linux, but no one can read this `S_*` mess. The difference in code readability here is really huge.

> I've already seen missing leading `0` (in Graphene too, I believe).

That's easy to check in the review and this is the only possible bug when using octals. With `S_*` you have a bunch of totally random letters glued together, how is this less error prone?

> While `PERM_rw_______` would look much nicer, we don't have such macros currently.

I agree that these are better. `PERM_*` > octals > `S_*`.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @pwmarcz)

___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMJ9tw96oEHZ0co7BPv:bnf7r0d) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*
> I totally disagree, everyone can read octal permissions on Linux, but no one can read this S_* mess. The difference in code readability here is really huge.

And I totally disagree whit that. Give me `0142` and I have no idea what permission is that. Otoh `S_IXUSR | S_IRGRP | S_IWOTH` is ugly, but self contained, I can easily read what the permissions will be.

> That's easy to check in the review and this is the only possible bug when using octals. With S_* you have a bunch of totally random letters glued together, how is this less error prone?

Easy, yet somehow sometimes slips through. I disagree they are totally random, `S_IWOTH`: `S_I` - this is the problematic part, wth it stands for I do not know, but it's constant, so you can just ignore it, `W` - write, `OTH` - others. If you write `S_JWOTH` it won't even compile, where as if you write `0776`, who know's if that's a bug or intended.

>I agree that these are better. PERM_* > octals > S_*.

`PERM_*` >>> `S_I*` > octals

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 8 unresolved discussions (waiting on @pwmarcz)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 8 unresolved discussions (waiting on @boryspoplawski and @pwmarcz)

___
*[LibOS/shim/include/shim_internal.h, line 39 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIstXtBKFObCXxGobx:-MMKi_b--LV2kfKVlKYI:b3e7tmb) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/LibOS/shim/include/shim_internal.h#L39)):*

Imo `g_debug_log_enabled` would be better
</blockquote></details>

It's called this way because it comes from `pal_control.enable_debug_log` field, and that structure as `disable_aslr` already. Do you think it's worth it to change it only here (so that we have something like `g_debug_log_enabled = pal_control.enable_debug_log`)?
___
*[LibOS/shim/src/utils/printf.c, line 20 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMItuHl5hP0j1kq5qVY:-MMKjMvr88IVDNI7knXt:b2177g5) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/LibOS/shim/src/utils/printf.c#L20)):*

This cast is superfluous.
</blockquote></details>

It's not, because we are dropping `const`:
```
utils/printf.c: In function 'debug_fputs':
utils/printf.c:20:32: error: passing argument 1 of 'DkDebugLog' discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
         PAL_NUM x = DkDebugLog((buf + bytes), size - bytes);
                                ^
In file included from utils/printf.c:9:0:
../../../Pal/include/pal/pal.h:693:9: note: expected 'PAL_PTR {aka void *}' but argument is of type 'const char *'
 PAL_NUM DkDebugLog(PAL_PTR buffer, PAL_NUM size);
         ^~~~~~~~~~
```
___
*[Pal/include/pal/pal.h, line 365 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIwowR-WbO_fYdiZQi:-MMKkMsZEyHRoKI5c12S:bm3yqt3) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/include/pal/pal.h#L365)):*

Imo you should skip `Returns` and `,` before `and`.
</blockquote></details>

OK, done.
___
*[Pal/include/pal/pal.h, line 690 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIwzKp5F6LN4zwUTjg:-MMKkQco8x1HdDfuR-df:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/include/pal/pal.h#L690)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_streams.c, line 369 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyyjd-sNapDahEqM6:-MMKluxt1PTaKGcm0I59:bcpnu0h) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L369)):*

Please add a check that `g_debug_fd` is not yet initialized and fail if it is (or just close the old fd).
</blockquote></details>

I added a close.
___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMKmMGFAnfGqaDnQbPO:b-tp3cn9) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*

> I totally disagree, everyone can read octal permissions on Linux, but no one can read this S_* mess. The difference in code readability here is really huge.

And I totally disagree whit that. Give me `0142` and I have no idea what permission is that. Otoh `S_IXUSR | S_IRGRP | S_IWOTH` is ugly, but self contained, I can easily read what the permissions will be.

> That's easy to check in the review and this is the only possible bug when using octals. With S_* you have a bunch of totally random letters glued together, how is this less error prone?

Easy, yet somehow sometimes slips through. I disagree they are totally random, `S_IWOTH`: `S_I` - this is the problematic part, wth it stands for I do not know, but it's constant, so you can just ignore it, `W` - write, `OTH` - others. If you write `S_JWOTH` it won't even compile, where as if you write `0776`, who know's if that's a bug or intended.

>I agree that these are better. PERM_* > octals > S_*.

`PERM_*` >>> `S_I*` > octals
</blockquote></details>

So what do we do? I'm happy to change to `S_I*` but please agree on something.

The current code has a pretty even split between octal permissions and `S_*`. Also, the [coding style](https://graphene.readthedocs.io/en/latest/devel/coding-style.html) document currently has an example with `0600` :)

`PERM_*` looks nice but from what I can see the linked patch did not go forward? If we do introduce it, I think it should be in a separate PR that also converts existing code.
___
*[Pal/src/host/Linux-SGX/db_streams.c, line 403 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIzWnRCi4dAljPesOR:-MMKlycI0x3goFWKW340:b-16uuh6) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux-SGX/db_streams.c#L403)):*

ditto (initialization check)
</blockquote></details>

I added a close, same as above.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 8 unresolved discussions (waiting on @boryspoplawski, @dimakuv, @pwmarcz, and @yamahata)

___
*[LibOS/shim/src/utils/printf.c, line 20 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMItuHl5hP0j1kq5qVY:-MMLZTjS9DAKRoaVjFns:boi4vi2) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/LibOS/shim/src/utils/printf.c#L20)):*

It's not, because we are dropping `const`:
```
utils/printf.c: In function 'debug_fputs':
utils/printf.c:20:32: error: passing argument 1 of 'DkDebugLog' discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
         PAL_NUM x = DkDebugLog((buf + bytes), size - bytes);
                                ^
In file included from utils/printf.c:9:0:
../../../Pal/include/pal/pal.h:693:9: note: expected 'PAL_PTR {aka void *}' but argument is of type 'const char *'
 PAL_NUM DkDebugLog(PAL_PTR buffer, PAL_NUM size);
         ^~~~~~~~~~
```
</blockquote></details>

Can you add `const` to the API then? Those `PAL_*` types will go away anyways, they don't make any sense.
___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMLZdEq1pkT9U5TFZxM:b-wdnlzn) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*

So what do we do? I'm happy to change to `S_I*` but please agree on something.

The current code has a pretty even split between octal permissions and `S_*`. Also, the [coding style](https://graphene.readthedocs.io/en/latest/devel/coding-style.html) document currently has an example with `0600` :)

`PERM_*` looks nice but from what I can see the linked patch did not go forward? If we do introduce it, I think it should be in a separate PR that also converts existing code.
</blockquote></details>

@dimakuv @yamahata What's your preference here?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 8 unresolved discussions (waiting on @boryspoplawski, @dimakuv, @mkow, @pwmarcz, and @yamahata)

___
*[LibOS/shim/include/shim_internal.h, line 39 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIstXtBKFObCXxGobx:-MMLYtbv2wSy6yRquTX8:bvifuf3) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/LibOS/shim/include/shim_internal.h#L39)):*

It's called this way because it comes from `pal_control.enable_debug_log` field, and that structure as `disable_aslr` already. Do you think it's worth it to change it only here (so that we have something like `g_debug_log_enabled = pal_control.enable_debug_log`)?
</blockquote></details>

Yeah, that would be fine
___
*[LibOS/shim/src/utils/printf.c, line 20 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMItuHl5hP0j1kq5qVY:-MMLZwhYDqpvsVWXN3FH:ba3dk63) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/LibOS/shim/src/utils/printf.c#L20)):*

Can you add `const` to the API then? Those `PAL_*` types will go away anyways, they don't make any sense.
</blockquote></details>

Then it's a question why Pal API does not have const ...
But I guess it's an existing problem, so unblocking
___
*[Pal/include/pal/pal.h, line 365 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIwowR-WbO_fYdiZQi:-MML_CYZ3TPn1ZKBlXiU:b-4zwpsz) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/include/pal/pal.h#L365)):*

OK, done.
</blockquote></details>

Did you forgot to push your changes? :)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 8 unresolved discussions (waiting on @boryspoplawski, @dimakuv, @pwmarcz, and @yamahata)

___
*[LibOS/shim/src/utils/printf.c, line 20 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMItuHl5hP0j1kq5qVY:-MMLa3s99cwB0WbsjZEi:b-2biob4) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/LibOS/shim/src/utils/printf.c#L20)):*
> Then it's a question why Pal API does not have const ...

I don't know any reason, all those typedefs doesn't make any sense. We should be just using proper C types there.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 8 unresolved discussions (waiting on @boryspoplawski, @pwmarcz, and @yamahata)

___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMLaBx_1rG4xDEmFW8f:bblphbt) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*

@dimakuv @yamahata What's your preference here?
</blockquote></details>

I used `S_*` in my code. So slight preference for macros, but I don't care much.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: 16 of 24 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, and @yamahata)

___
*[LibOS/shim/include/shim_internal.h, line 39 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIstXtBKFObCXxGobx:-MMLnrRSEtW0oaDe5R3f:bm3yqt3) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/LibOS/shim/include/shim_internal.h#L39)):*

Yeah, that would be fine
</blockquote></details>

OK, done.
___
*[LibOS/shim/src/utils/printf.c, line 20 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMItuHl5hP0j1kq5qVY:-MMLnvlO2n3urhFGxUc6:b-4nzn9p) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/LibOS/shim/src/utils/printf.c#L20)):*
Okay... I assumed these are all `PAL_PTR` instead of proper pointer types for a good reason :)

>Can you add const to the API then? Those PAL_* types will go away anyways, they don't make any sense.

I would rather not. That would mean either using `const PAL_PTR`, `const void*` or `const char*`. None these appear in`pal.h` which uses `PAL_*` types exclusively (there is some usage of `const PAL_PTR` in two internal functions, but even these do not have `const` in headers, just implementation).

If there is no good reason for using `PAL_PTR`, `PAL_NUM` etc., then of course, let's change that in bulk (for example, file-by-file?), but adding one function to `pal.h` that is different from everything else feels like a bad way to do it.
___
*[Pal/include/pal/pal.h, line 365 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIwowR-WbO_fYdiZQi:-MMLn1AMENQtUKSdxVI3:b-qvsz5x) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/include/pal/pal.h#L365)):*

Did you forgot to push your changes? :)
</blockquote></details>

Yeah... I think I did one rebase too many or something. Anyway, I pushed the new commits now.
___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMLrB-l8_KsbxH4n6Iw:b-7sws6u) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*

I used `S_*` in my code. So slight preference for macros, but I don't care much.
</blockquote></details>

Then, in the interest of moving this forward, let's go with `S_*` here? I updated the code.
___
*[Pal/src/host/Linux-SGX/db_streams.c, line 404 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIzZ1J252y_qZiJZjI:-MMLr_TD6jeH-XngxLJM:b-av1uo0) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux-SGX/db_streams.c#L404)):*

ditto (hardcoded number)
</blockquote></details>

Changed to `S_*`.

boryspoplawski (changes requested):
> Reviewed 8 of 8 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[LibOS/shim/src/utils/printf.c, line 20 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMItuHl5hP0j1kq5qVY:-MMLtg8Q-WlyZ8Brb279:btblzbm) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/LibOS/shim/src/utils/printf.c#L20)):*

Okay... I assumed these are all `PAL_PTR` instead of proper pointer types for a good reason :)

>Can you add const to the API then? Those PAL_* types will go away anyways, they don't make any sense.

I would rather not. That would mean either using `const PAL_PTR`, `const void*` or `const char*`. None these appear in`pal.h` which uses `PAL_*` types exclusively (there is some usage of `const PAL_PTR` in two internal functions, but even these do not have `const` in headers, just implementation).

If there is no good reason for using `PAL_PTR`, `PAL_NUM` etc., then of course, let's change that in bulk (for example, file-by-file?), but adding one function to `pal.h` that is different from everything else feels like a bad way to do it.
</blockquote></details>

I agree with Paweł, this is not the place for such change.
___
*[Pal/src/host/Linux/db_streams.c, line 371 at r6](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMLu5oHEJqimG4XcLqb:-MMLu5oI8m21PEutkaTz:b-itxax9) ([raw file](https://github.com/oscarlab/graphene/blob/5a361bee8b87ed2cf461b0d97788e4e8098a272b/Pal/src/host/Linux/db_streams.c#L371)):*
> ```C
> int _DkInitDebugStream(const char* path) {
>     if (g_debug_fd >= 0) {
>         INLINE_SYSCALL(close, 1, g_debug_fd);
> ```

Please add error checking (I know, it's `close`, but still)
___
*[Pal/src/host/Linux-SGX/db_streams.c, line 405 at r6](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMLvHhBEXi4wZL5c4GF:-MMLvHhCCYBZp2E9bwRQ:b-dw3cqm) ([raw file](https://github.com/oscarlab/graphene/blob/5a361bee8b87ed2cf461b0d97788e4e8098a272b/Pal/src/host/Linux-SGX/db_streams.c#L405)):*
> ```C
> int _DkInitDebugStream(const char* path) {
>     if (g_debug_fd >= 0) {
>         ocall_close(g_debug_fd);
> ```

Please add error checking

mkow (approved):
> Reviewed 8 of 8 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/src/utils/printf.c, line 20 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMItuHl5hP0j1kq5qVY:-MMLv0fbEzjUqWbLSkIU:b-i0hchr) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/LibOS/shim/src/utils/printf.c#L20)):*

Okay... I assumed these are all `PAL_PTR` instead of proper pointer types for a good reason :)

>Can you add const to the API then? Those PAL_* types will go away anyways, they don't make any sense.

I would rather not. That would mean either using `const PAL_PTR`, `const void*` or `const char*`. None these appear in`pal.h` which uses `PAL_*` types exclusively (there is some usage of `const PAL_PTR` in two internal functions, but even these do not have `const` in headers, just implementation).

If there is no good reason for using `PAL_PTR`, `PAL_NUM` etc., then of course, let's change that in bulk (for example, file-by-file?), but adding one function to `pal.h` that is different from everything else feels like a bad way to do it.
</blockquote></details>

Yeah, but dropping const sounds even worse to me ;) Anyways, I'm unblocking this discussion.
___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMLvUIY92Kb4LEswng-:bk0iihs) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*

Then, in the interest of moving this forward, let's go with `S_*` here? I updated the code.
</blockquote></details>

Ugly :P
@pwmarcz Would you have time to create a quick PR afterwards and introduce `PERM_*` macros?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: 22 of 24 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMM-exE6HVRaSVitwlg:b-yikse8) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*
Some cleanup in headers was necessary to get `#include <linux/stat.h>` to work.

>@pwmarcz Would you have time to create a quick PR afterwards and introduce PERM_* macros?

OK, I'll either create a PR or an issue.
___
*[Pal/src/host/Linux/db_streams.c, line 371 at r6](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMLu5oHEJqimG4XcLqb:-MMM3bn1CHe2MeSBs3B3:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/5a361bee8b87ed2cf461b0d97788e4e8098a272b/Pal/src/host/Linux/db_streams.c#L371)):*

Please add error checking (I know, it's `close`, but still)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_streams.c, line 405 at r6](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMLvHhBEXi4wZL5c4GF:-MMM3e668_g6sM6Fo8K-:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/5a361bee8b87ed2cf461b0d97788e4e8098a272b/Pal/src/host/Linux-SGX/db_streams.c#L405)):*

Please add error checking
</blockquote></details>

Done.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

boryspoplawski (dismissed):
> Reviewed 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMM9L-45xZAdSJuWUQs:b-1k0d9q) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*

>Some cleanup in headers was necessary to get #include <linux/stat.h> to work.

Aaand it doesn't compile under Ubuntu 16. This is becoming too big of a time waste here, so we agreed with @boryspoplawski to admit defeat and keep this as octal for now.
</blockquote></details>

Yeah, Graphene includes are a big mess right now ...

mkow (dismissed):
> Reviewed 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux/db_streams.c, line 370 at r5](https://reviewable.io/reviews/oscarlab/graphene/1965#-MMIyeQl5zCGzr5Eww1z:-MMMAy0h2eIgi1ODCnoH:b-vfgm4a) ([raw file](https://github.com/oscarlab/graphene/blob/369acc2e1267653eba8e7573e251431a2c961e77/Pal/src/host/Linux/db_streams.c#L370)):*

Yeah, Graphene includes are a big mess right now ...
</blockquote></details>

yay! :partying_face: :)

dimakuv (dismissed):
> Reviewed 2 of 4 files at r5, 6 of 8 files at r6, 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (approved):
> Reviewed 3 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

boryspoplawski (dismissed):
> Reviewed 3 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

yamahata (dismissed):
> Reviewed 6 of 8 files at r6, 2 of 2 files at r8, 3 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1965)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1970, opened 2020-11-16
Add /sys pseudo-filesystem support

### LINKED ISSUE #1783: [LibOS, PAL] Add support for pseudo sysfs to get platform topology information
## Description of the problem
Many libraries such as tbb, hwloc use sysfs-devices-system to get the underlying platform topology. But currently this isn't emulated in LibOS and PAL. So propose to support the following as pseudo files as a start. Based on feedback I can extend/trim these as needed.

/sys/devices/system/node
/sys/devices/system/cpu/cpu#/topology/core_siblings
/sys/devices/system/cpu/cpu#/topology/thread_siblings
/sys/devices/system/cpu/offline
/sys/devices/system/cpu/online
/sys/devices/system/cpu/possible

## Expected results
Avoid crashes with latest openVINO 2020.4 with tbb/hwloc libraries and also address the performance gap due to missing topology info.

## Actual results
With openVINO 2020.4 benchmark observed the following issues due to these missing topology info.
1. openVINO 2020.4 uses topology info to identify thread siblings and tries not to schedule on the sibling cores. Since the feature is missing, observe ~12% degrade in benchmark latency. 
2. openVINO 2020.4 uses tbb libraries which in turn use hwloc to get numa information. hwloc uses sysfs and since this is not supported, openVINO ends up getting the number of numa nodes as zero resulting in a divide-by-zero exception.

vijaydhanraj:
## Description of the changes 
This PR adds support for /sys pseudo-filesystem which in turn helps to enable open-source libraries such as "Hardware Locality" to relay underlying hardware topology-related information to applications. This is critical for ML benchmarks which seem to use topology-related info for better scheduling of threads.

 Fixes #1783 

## How to test this PR? 
Please use the newly added regression test "sysfs_common.c" to test the PR. Another option would be to run openVINO benchmark which exercies lot flows introduced in this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)

### DIFF
--- a/Examples/openvino/Makefile
+++ b/Examples/openvino/Makefile
@@ -60,8 +60,8 @@ $(OPENVINO_DIR)/README.md:
 
 $(OPENVINO_DIR)/inference-engine/bin/intel64/$(OPENVINO_BUILD)/object_detection_sample_ssd: $(OPENVINO_DIR)/README.md
 	cd $(OPENVINO_DIR)/inference-engine && mkdir -p build
-	cd $(OPENVINO_DIR)/inference-engine/build && cmake -DTHREADING=OMP -DCMAKE_BUILD_TYPE=$(OPENVINO_BUILD) ../..
-	$(MAKE) -C $(OPENVINO_DIR)/inference-engine/build
+	cd $(OPENVINO_DIR)/inference-engine/build && cmake -DCMAKE_BUILD_TYPE=$(OPENVINO_BUILD) ../..
+	$(MAKE) -C $(OPENVINO_DIR)/inference-engine/build -j8
 
 object_detection_sample_ssd.manifest: object_detection_sample_ssd.manifest.template
 	sed -e 's|$$(GRAPHENE_DIR)|'"$(GRAPHENE_DIR)"'|g' \
--- a/Examples/openvino/object_detection_sample_ssd.manifest.template
+++ b/Examples/openvino/object_detection_sample_ssd.manifest.template
@@ -13,7 +13,7 @@ loader.debug_type = "$(GRAPHENE_DEBUG)"
 loader.insecure__use_cmdline_argv = 1
 
 # Environment variables
-loader.env.LD_LIBRARY_PATH = "/lib:$(ARCH_LIBDIR):/usr/$(ARCH_LIBDIR)"
+loader.env.LD_LIBRARY_PATH = /lib:$(ARCH_LIBDIR):/usr/$(ARCH_LIBDIR):$(OPENVINO_DIR)/inference-engine/temp/tbb/lib
 
 # Mounted FSes. The following "chroot" FSes mount a part of the host FS into the
 # guest. Other parts of the host FS will not be available in the guest.
@@ -80,12 +80,14 @@ sgx.trusted_files.libnssnis = "file:$(ARCH_LIBDIR)/libnss_nis.so.2"
 sgx.trusted_files.libnsl = "file:$(ARCH_LIBDIR)/libnsl.so.1"
 
 # OpenVINO C++ dependencies
-sgx.trusted_files.libcpp = "file:/usr/$(ARCH_LIBDIR)/libstdc++.so.6"
-sgx.trusted_files.libgcc = "file:$(ARCH_LIBDIR)/libgcc_s.so.1"
-
-# OpenVINO TBB dependencies
-sgx.trusted_files.libtbbmalloc = "file:/usr/$(ARCH_LIBDIR)/libtbbmalloc.so.2"
-sgx.trusted_files.libtbb = "file:/usr/$(ARCH_LIBDIR)/libtbb.so.2"
+sgx.trusted_files.libcpp = file:/usr/$(ARCH_LIBDIR)/libstdc++.so.6
+sgx.trusted_files.libgcc = file:$(ARCH_LIBDIR)/libgcc_s.so.1
+sgx.trusted_files.libnuma    = file:/usr/$(ARCH_LIBDIR)/libnuma.so.1
+sgx.trusted_files.libxml2    = file:/usr/$(ARCH_LIBDIR)/libxml2.so.2
+sgx.trusted_files.libicuuc   = file:/usr/$(ARCH_LIBDIR)/libicuuc.so.60
+sgx.trusted_files.libz       = file:$(ARCH_LIBDIR)/libz.so.1
+sgx.trusted_files.liblzma    = file:$(ARCH_LIBDIR)/liblzma.so.5
+sgx.trusted_files.libicudata = file:/usr/$(ARCH_LIBDIR)/libicudata.so.60
 
 # OpenVINO main libraries
 sgx.trusted_files.libformatreader = "file:$(OPENVINO_DIR)/bin/intel64/$(OPENVINO_BUILD)/lib/libformat_reader.so"
@@ -108,8 +110,12 @@ sgx.trusted_files.libimgcodecs = "file:$(OPENVINO_DIR)/inference-engine/temp/ope
 sgx.trusted_files.libimgproc = "file:$(OPENVINO_DIR)/inference-engine/temp/opencv_4.3.0_ubuntu18/opencv/lib/libopencv_imgproc.so.4.3"
 sgx.trusted_files.libopencvcore = "file:$(OPENVINO_DIR)/inference-engine/temp/opencv_4.3.0_ubuntu18/opencv/lib/libopencv_core.so.4.3"
 
-# OpenVINO OMP libraries
-sgx.trusted_files.libiomp5 = "file:$(OPENVINO_DIR)/inference-engine/temp/omp/lib/libiomp5.so"
+#OpenVINO TBB dependencies
+sgx.trusted_files.libtbbmalloc = file:$(OPENVINO_DIR)/inference-engine/temp/tbb/lib/libtbbmalloc.so.2
+sgx.trusted_files.libtbb       = file:$(OPENVINO_DIR)/inference-engine/temp/tbb/lib/libtbb.so.2
+sgx.trusted_files.libtbbbind   = file:$(OPENVINO_DIR)/inference-engine/temp/tbb/lib/libtbbbind.so.2
+sgx.trusted_files.libhwloc     = file:$(OPENVINO_DIR)/inference-engine/temp/tbb/lib/libhwloc.so.5
+$(LIBTBB_DEBUG)
 
 # OpenVINO SSD300 model
 sgx.trusted_files.pluginsxml = "file:$(OPENVINO_DIR)/bin/intel64/$(OPENVINO_BUILD)/lib/plugins.xml"
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -483,11 +483,16 @@ extern struct shim_mount eventfd_builtin_fs;
 #define FILE_RW_MODE 0666
 #define FILE_R_MODE  0444
 
+#define SYSFS_FILESZ    256
+
 extern struct shim_fs_ops dev_fs_ops;
 extern struct shim_d_ops dev_d_ops;
 
 extern struct shim_fs_ops proc_fs_ops;
 extern struct shim_d_ops proc_d_ops;
+
+extern struct shim_fs_ops sys_fs_ops;
+extern struct shim_d_ops sys_d_ops;
 
 struct pseudo_name_ops {
     int (*match_name)(const char* name);
@@ -552,5 +557,16 @@ ssize_t str_read(struct shim_handle* hdl, void* buf, size_t count);
 ssize_t str_write(struct shim_handle* hdl, const void* buf, size_t count);
 off_t str_seek(struct shim_handle* hdl, off_t offset, int whence);
 int str_flush(struct shim_handle* hdl);
+
+/* /sys fs related common APIs*/
+const char* extract_filename (const char* pathname);
+int extract_num_from_path(const char* path);
+int sys_info_mode(const char* name, mode_t* mode);
+int sys_info_stat(const char* name, struct stat* buf);
+int sys_dir_open(struct shim_handle* hdl, const char* name, int flags);
+int sys_dir_mode(const char* name, mode_t* mode);
+int sys_dir_stat(const char* name, struct stat* buf);
+int sys_match_resource_num(const char* pathname);
+int sys_list_resource_num(const char* pathname, struct shim_dirent** buf, int len);
 
 #endif /* _SHIM_FS_H_ */
--- a/LibOS/shim/src/Makefile
+++ b/LibOS/shim/src/Makefile
@@ -76,6 +76,10 @@ objs = \
 	fs/proc/thread.o \
 	fs/socket/fs.o \
 	fs/str/fs.o \
+	fs/sys/fs.o \
+	fs/sys/cpu_info.o \
+	fs/sys/cache_info.o \
+	fs/sys/node_info.o \
 	ipc/shim_ipc.o \
 	ipc/shim_ipc_child.o \
 	ipc/shim_ipc_helper.o \
--- a/LibOS/shim/src/fs/proc/info.c
+++ b/LibOS/shim/src/fs/proc/info.c
@@ -143,7 +143,7 @@ static int proc_cpuinfo_open(struct shim_handle* hdl, const char* name, int flag
         len += ret;                                                  \
     } while (0)
 
-    for (size_t n = 0; n < pal_control.cpu_info.online_logical_cores; n++) {
+    for (size_t n = 0; n < pal_control.cpu_info.num_online_logical_cores; n++) {
         /* Below strings must match exactly the strings retrieved from /proc/cpuinfo
          * (see Linux's arch/x86/kernel/cpu/proc.c) */
         ADD_INFO("processor\t: %lu\n", n);
--- a/LibOS/shim/src/fs/proc/thread.c
+++ b/LibOS/shim/src/fs/proc/thread.c
@@ -647,6 +647,7 @@ static int proc_thread_dir_stat(const char* name, struct stat* buf) {
     buf->st_gid = thread->gid;
     unlock(&thread->lock);
     buf->st_size = 4096;
+    buf->st_nlink = 3;       //FIXME: need atleast 3 `/proc/self/task`
 
     put_thread(thread);
     return 0;
@@ -726,12 +727,25 @@ const struct pseudo_fs_ops fs_thread = {
     .stat = &proc_thread_dir_stat,
 };
 
+static const struct pseudo_dir dir_task = {
+    .size = 1,
+    .ent =
+        {
+            {
+                .name_ops = &nm_thread,
+                .fs_ops   = &fs_thread,
+                .type     = LINUX_DT_DIR,
+            },
+        },
+};
+
 const struct pseudo_dir dir_thread = {
-    .size = 5,
+    .size = 6,
     .ent  = {
         {.name = "cwd",  .fs_ops = &fs_thread_link, .type = LINUX_DT_LNK},
         {.name = "exe",  .fs_ops = &fs_thread_link, .type = LINUX_DT_LNK},
         {.name = "root", .fs_ops = &fs_thread_link, .type = LINUX_DT_LNK},
-        {.name = "fd",   .fs_ops = &fs_thread_fd,   .dir = &dir_fd},
+        {.name = "fd",   .fs_ops = &fs_thread_fd,   .dir  = &dir_fd},
         {.name = "maps", .fs_ops = &fs_thread_maps, .type = LINUX_DT_REG},
+        {.name = "task", .fs_ops = &fs_thread,      .dir  = &dir_task },
     }};
--- a/LibOS/shim/src/fs/shim_fs.c
+++ b/LibOS/shim/src/fs/shim_fs.c
@@ -40,6 +40,11 @@ struct shim_fs mountable_fs[] = {
         .fs_ops = &dev_fs_ops,
         .d_ops  = &dev_d_ops,
     },
+    {
+        .name   = "sys",
+        .fs_ops = &sys_fs_ops,
+        .d_ops  = &sys_d_ops,
+    },
 };
 
 struct shim_mount* builtin_fs[] = {
@@ -150,6 +155,13 @@ static int __mount_sys(struct shim_dentry* root) {
 
     if ((ret = mount_fs("chroot", URI_PREFIX_DEV "tty", "/dev/tty", dev_dent, NULL, 0)) < 0) {
         debug("Mounting terminal device failed (%d)\n", ret);
+        return ret;
+    }
+
+    debug("mounting as sys filesystem: /sys\n");
+
+    if ((ret = mount_fs("sys", NULL, "/sys", root, NULL, 0)) < 0) {
+        debug("mounting sys filesystem failed (%d)\n", ret);
         return ret;
     }
 
--- a/LibOS/shim/src/fs/sys/cache_info.c
+++ b/LibOS/shim/src/fs/sys/cache_info.c
@@ -1,0 +1,142 @@
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
+/* Copyright (C) 2020 Intel Corporation */
+
+/*!
+ * \file
+ *
+ * This file contains the implementation of `/sys/devices/system/cpu/cpuX/cache`
+ */
+
+#include "shim_fs.h"
+
+static int cache_info_open (struct shim_handle* hdl, const char* name, int flags) {
+    __UNUSED(hdl);
+    __UNUSED(flags);
+
+    int len;
+    char* str = malloc(SYSFS_FILESZ);
+    if (!str)
+        return -ENOMEM;
+
+    const char* filename = extract_filename(name);
+    int cpunum = extract_num_from_path(name);
+    if (cpunum < 0 )
+        return -ENOENT;
+
+    int idx = extract_num_from_path(strstr(name, "index"));
+    if (idx < 0 )
+        return -ENOENT;
+
+    if (!strcmp(filename, "shared_cpu_map")) {
+        const char* cachemap;
+        cachemap = pal_control.topo_info.core_topology[cpunum].cache[idx].shared_cpu_map;
+        len =  strlen(cachemap) + 1;
+        memcpy(str, cachemap, len);
+
+    } else if (!strcmp(filename, "level")) {
+        const char* level = pal_control.topo_info.core_topology[cpunum].cache[idx].level;
+        len =  strlen(level) + 1;
+        memcpy(str, level, len);
+
+    } else if (!strcmp(filename, "type")) {
+        const char* type = pal_control.topo_info.core_topology[cpunum].cache[idx].type;
+        len =  strlen(type) + 1;
+        memcpy(str, type, len);
+
+    } else if (!strcmp(filename, "size")) {
+        const char* size = pal_control.topo_info.core_topology[cpunum].cache[idx].size;
+        len =  strlen(size) + 1;
+        memcpy(str, size, len);
+
+    } else if (!strcmp(filename, "coherency_line_size")) {
+        const char* coherency;
+        coherency = pal_control.topo_info.core_topology[cpunum].cache[idx].coherency_line_size;
+        len =  strlen(coherency) + 1;
+        memcpy(str, coherency, len);
+
+    } else if (!strcmp(filename, "number_of_sets")) {
+        const char* sets  = pal_control.topo_info.core_topology[cpunum].cache[idx].number_of_sets;
+        len =  strlen(sets) + 1;
+        memcpy(str, sets, len);
+
+    } else if (!strcmp(filename, "physical_line_partition")) {
+        const char* partition;
+        partition = pal_control.topo_info.core_topology[cpunum].cache[idx].physical_line_partition;
+        len =  strlen(partition) + 1;
+        memcpy(str, partition, len);
+
+    } else {
+        debug("Unsupported Filepath %s\n", name);
+        return -ENOENT;
+    }
+
+    struct shim_str_data* data = malloc(sizeof(struct shim_str_data));
+    if (!data) {
+        free(str);
+        return -ENOMEM;
+    }
+
+    memset(data, 0, sizeof(struct shim_str_data));
+    data->str          = str;
+    data->len          = len;
+    hdl->type          = TYPE_STR;
+    hdl->flags         = flags & ~O_RDONLY;
+    hdl->acc_mode      = MAY_READ;
+    hdl->info.str.data = data;
+
+    return 0;
+}
+
+static const struct pseudo_fs_ops cache_idx_info = {
+    .mode = &sys_info_mode,
+    .stat = &sys_info_stat,
+    .open = &cache_info_open,
+};
+
+static const struct pseudo_dir cache_idx_dir = {
+    .size = 7,
+    .ent  = {
+              { .name   = "shared_cpu_map",
+                .fs_ops = &cache_idx_info,
+                .type   = LINUX_DT_REG },
+              { .name   = "level",
+                .fs_ops = &cache_idx_info,
+                .type   = LINUX_DT_REG },
+              { .name   = "type",
+                .fs_ops = &cache_idx_info,
+                .type   = LINUX_DT_REG },
+              { .name   = "size",
+                .fs_ops = &cache_idx_info,
+                .type   = LINUX_DT_REG },
+              { .name   = "coherency_line_size",
+                .fs_ops = &cache_idx_info,
+                .type   = LINUX_DT_REG },
+              { .name   = "number_of_sets",
+                .fs_ops = &cache_idx_info,
+                .type   = LINUX_DT_REG },
+              { .name   = "physical_line_partition",
+                .fs_ops = &cache_idx_info,
+                .type   = LINUX_DT_REG },
+            }
+};
+
+static const struct pseudo_fs_ops cache_info = {
+    .mode = &sys_dir_mode,
+    .stat = &sys_dir_stat,
+    .open = &sys_dir_open,
+};
+
+static const struct pseudo_name_ops cache_ops = {
+    .match_name = &sys_match_resource_num,
+    .list_name  = &sys_list_resource_num,
+};
+
+const struct pseudo_dir cpunum_cache_dir = {
+    .size = 1,
+    .ent  = {
+              { .name_ops = &cache_ops,
+                .fs_ops   = &cache_info,
+                .dir      = &cache_idx_dir,
+                .type     = LINUX_DT_DIR },
+            }
+};
--- a/LibOS/shim/src/fs/sys/cpu_info.c
+++ b/LibOS/shim/src/fs/sys/cpu_info.c
@@ -1,0 +1,168 @@
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
+/* Copyright (C) 2020 Intel Corporation */
+
+/*!
+ * \file
+ *
+ * This file contains the implementation of `/sys/devices/system/cpu` and its sub-directories.
+ */
+
+#include "shim_fs.h"
+extern const struct pseudo_dir cpunum_cache_dir;
+
+static int cpu_info_open(struct shim_handle* hdl, const char* name, int flags) {
+    __UNUSED(hdl);
+    __UNUSED(flags);
+
+    int len;
+    char* str = malloc(SYSFS_FILESZ);
+    if (!str)
+        return -ENOMEM;
+
+    const char* filename = extract_filename(name);
+    if (!strcmp(filename, "online")) {
+        const char* online_info;
+        if (strstr(name, "cpu/cpu")) {
+            int cpunum = extract_num_from_path(name);
+            if (cpunum < 0 )
+                return -ENOENT;
+            online_info = pal_control.topo_info.core_topology[cpunum].is_logical_core_online;
+        }
+        else
+            online_info = pal_control.topo_info.online_logical_cores;
+
+        len = strlen(online_info) + 1;
+        memcpy(str, online_info, len);
+
+    } else if (!strcmp(filename, "possible")) {
+        const char* possible_info = pal_control.topo_info.possible_logical_cores;
+        len = strlen(possible_info) + 1;
+        memcpy(str, possible_info, len);
+
+    } else if (!strcmp(filename, "core_id")) {
+        int cpunum = extract_num_from_path(name);
+        if (cpunum < 0 )
+            return -ENOENT;
+
+        const char* core_id;
+        core_id = pal_control.topo_info.core_topology[cpunum].core_id;
+        len = strlen(core_id) + 1;
+        memcpy(str, core_id, len);
+
+    } else if (!strcmp(filename, "physical_package_id")) {
+        int cpunum = extract_num_from_path(name);
+        if (cpunum < 0 )
+            return -ENOENT;
+        /* we have already collected this info as part of /proc/cpuinfo. So reuse it */
+        snprintf(str, sizeof(str), "%d", pal_control.cpu_info.cpu_socket[cpunum]);
+        len = strlen(str) + 1;
+        str[len] = '\0';
+
+    } else if (!strcmp(filename, "core_siblings")) {
+        int cpunum = extract_num_from_path(name);
+        if (cpunum < 0 )
+            return -ENOENT;
+
+        const char* coresiblings_info;
+        coresiblings_info = pal_control.topo_info.core_topology[cpunum].core_siblings;
+        len = strlen(coresiblings_info) + 1;
+        memcpy(str, coresiblings_info, len);
+
+    } else if (!strcmp(filename, "thread_siblings")) {
+        int cpunum = extract_num_from_path(name);
+        if (cpunum < 0 )
+            return -ENOENT;
+
+        const char* threadsiblings_info;
+        threadsiblings_info = pal_control.topo_info.core_topology[cpunum].thread_siblings;
+        len = strlen(threadsiblings_info) + 1;
+        memcpy(str, threadsiblings_info, len);
+
+    } else {
+        debug("Unsupported Filepath %s\n", name);
+        return -ENOENT;
+    }
+
+    struct shim_str_data* data = malloc(sizeof(struct shim_str_data));
+    if (!data) {
+        free(str);
+        return -ENOMEM;
+    }
+
+    memset(data, 0, sizeof(struct shim_str_data));
+    data->str          = str;
+    data->len          = len;
+    hdl->type          = TYPE_STR;
+    hdl->flags         = flags & ~O_RDONLY;
+    hdl->acc_mode      = MAY_READ;
+    hdl->info.str.data = data;
+
+    return 0;
+}
+
+static const struct pseudo_fs_ops cpu_info = {
+    .mode = &sys_info_mode,
+    .stat = &sys_info_stat,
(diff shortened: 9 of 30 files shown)


dimakuv (changes requested):
> Reviewed 4 of 10 files at r1, 4 of 4 files at r2, 12 of 12 files at r3, 8 of 8 files at r4, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @vijaydhanraj)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFb--u3gQ8PyA9iLKa:-MMFb--v0rBXm-UJjORX:bdl4vmw) (no related file):*
Thanks for the PR!

Major comments:

1. Please revert your commit r1. The name change to `num_online_logical_cores` is not particularly interesting, and I don't think it is worth the change (`CPU_INFO` and `TOPO_INFO` are clearly different, and you have different types, so I don't see a problem).

2. Please move commit r2 (where you add the test) to the very end. Otherwise we have a git history where this commit reports test failures (if someone wants to e.g. git-bisect).

3. Merge commits r3 and r4 together -- they don't work without each other. (Again, if someone wants to e.g. git-bisect, Graphene won't be even buildable in-between these commits.)

4. It looks like the new functions in `db_main.c` and `sgx_main.c` are 100% duplicate code. Please move them and their definitions and required macros/structs into `Pal/src/host/Linux-common` and `Pal/include/host/Linux-common`. Create a new file there for these functions.

After all these changes, you'll have no choice but to rebase  this PR. Alternatively, I would recommend to close this PR, do the changes I described, and submit a new PR (otherwise reviewers will be confused).

---
___
*[Examples/openvino/Makefile, line 64 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFcPRvBoJ_9yu9kZlE:-MMFcPRvBoJ_9yu9kZlF:b-10s7xg) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/Examples/openvino/Makefile#L64)):*
> ```Makefile
> 	cd $(OPENVINO_DIR)/inference-engine && mkdir -p build
> 	cd $(OPENVINO_DIR)/inference-engine/build && cmake -DCMAKE_BUILD_TYPE=$(OPENVINO_BUILD) ../..
> 	$(MAKE) -C $(OPENVINO_DIR)/inference-engine/build -j8
> ```

`-j8 `not needed. The `$(MAKE)` already propagates the command-line options the user put.
___
*[Examples/openvino/object_detection_sample_ssd.manifest.template, line 16 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFc_Er5KSRZILHlIUu:-MMFc_Er5KSRZILHlIUv:bsmph1k) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/Examples/openvino/object_detection_sample_ssd.manifest.template#L16)):*
> ```
> 
> # Environment variables
> loader.env.LD_LIBRARY_PATH = /lib:$(ARCH_LIBDIR):/usr/$(ARCH_LIBDIR):$(OPENVINO_DIR)/inference-engine/temp/tbb/lib
> ```

How can this work? We moved to TOML now, so all strings must be in double-quotes.
___
*[LibOS/shim/include/shim_fs.h, line 562 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFcjGK5x0waXJWdwRh:-MMFcjGK5x0waXJWdwRi:bf7b438) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/include/shim_fs.h#L562)):*
> ```objc
> 
> /* /sys fs related common APIs*/
> const char* extract_filename (const char* pathname);
> ```

In many places in this PR, you have a space after the function name. Please remove everywhere.
___
*[LibOS/shim/src/fs/proc/thread.c, line 650 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFctdq5QY4060ZqJ9s:-MMFctdq5QY4060ZqJ9t:b87p8fc) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/proc/thread.c#L650)):*
> ```C
>     unlock(&thread->lock);
>     buf->st_size = 4096;
>     buf->st_nlink = 3;       //FIXME: need atleast 3 `/proc/self/task`
> ```

Why is this?
___
*[LibOS/shim/src/fs/proc/thread.c, line 750 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFcwe1-HjKtZrt4Tmm:-MMFcwe1-HjKtZrt4Tmn:b-w3hftr) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/proc/thread.c#L750)):*
> ```C
>         {.name = "fd",   .fs_ops = &fs_thread_fd,   .dir  = &dir_fd},
>         {.name = "maps", .fs_ops = &fs_thread_maps, .type = LINUX_DT_REG},
>         {.name = "task", .fs_ops = &fs_thread,      .dir  = &dir_task },
> ```

I'm confused. You're adding a new `/proc/[..]/task` subdir, but you don't add its implementation? Did you forget to git-add some files?
___
*[LibOS/shim/src/fs/sys/cache_info.c, line 5 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFd7CZ-h5XkC_Mh2i6:-MMFd7CZ-h5XkC_Mh2i7:bv2qmo3) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cache_info.c#L5)):*
> ```C
> 
> /*!
>  * \file
> ```

Please remove this line.
___
*[LibOS/shim/src/fs/sys/cpu_info.c, line 5 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFdC5hBqxoZqtZjzbf:-MMFdC5hBqxoZqtZjzbg:bv2qmo3) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cpu_info.c#L5)):*
> ```C
> 
> /*!
>  * \file
> ```

Please remove this line.
___
*[LibOS/shim/src/fs/sys/cpu_info.c, line 137 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFdDJ3ADA25uM0SIC1:-MMFdDJ3ADA25uM0SIC2:b-s1ngo5) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cpu_info.c#L137)):*
> ```C
>     .ent  = {
>               { .name   = "online",
> 		        .fs_ops = &cpu_info,
> ```

Please replace all tabs with spaces in your PR.
___
*[LibOS/shim/src/fs/sys/fs.c, line 5 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFdFlPAgvg7X3D5LZG:-MMFdFlPAgvg7X3D5LZH:bv2qmo3) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/fs.c#L5)):*
> ```C
> 
> /*!
>  * \file
> ```

Please remove this line.
___
*[LibOS/shim/src/fs/sys/fs.c, line 15 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFdRlO7uc9D3CUudMR:-MMFdRlO7uc9D3CUudMS:b6f47pr) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/fs.c#L15)):*
> ```C
> 
> /* Extract name of the file from the given path*/
> const char* extract_filename (const char* pathname) {
> ```

Please check if you can replace some of these helper functions with what we already have in Graphene: https://github.com/oscarlab/graphene/blob/master/Pal/include/lib/api.h#L244-L245
___
*[LibOS/shim/src/fs/sys/node_info.c, line 5 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFd_wY2vTT8laJShV1:-MMFd_wY2vTT8laJShV2:bv2qmo3) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/node_info.c#L5)):*
> ```C
> 
> /*!
>  * \file
> ```

Please remove this line.
___
*[Pal/include/arch/x86_64/pal-arch.h, line 30 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFdgzr8_6iIKToBZSx:-MMFdgzr8_6iIKToBZSy:b-6dakyx) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/Pal/include/arch/x86_64/pal-arch.h#L30)):*
> ```objc
> #define PAL_SYSFS_INT_FILESZ      16
> #define PAL_SYSFS_BUF_FILESZ      64
> #define PAL_SYSFS_MAP_FILESZ      256
> ```

Could you give some rationale and rule-of-thumb memory requirements for these constants and the general topology descriptions? I remember that on 160-core machines, it would require MBs of memory?
___
*[Pal/include/arch/x86_64/pal-arch.h, line 237 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFe0jI6burDIQzaHgY:-MMFe0jI6burDIQzaHgZ:be0vunn) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/Pal/include/arch/x86_64/pal-arch.h#L237)):*
> ```objc
> typedef struct PAL_NUMA_HUGEPAGE_INFO_ {
>     char nr_hugepages[PAL_SYSFS_INT_FILESZ];
> } PAL_NUMA_HUGEPAGE_INFO;
> ```

Is there really a point in showing hugepages to the user app? Can't we just always return `0` or something?
___
*[Pal/src/host/Linux/db_main.c, line 341 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFeD18AAgjzdx3ceoi:-MMFeD18AAgjzdx3ceoj:b-uchu07) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/Pal/src/host/Linux/db_main.c#L341)):*
> ```C
> }
> 
> int get_num_cache_level(const char* path) {
> ```

All this stuff must be deduplicated (you have exactly the same in `sgx_main.c`). Please see my top-level comment on `Linux-common/`.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @vijaydhanraj)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFb--u3gQ8PyA9iLKa:-MMHAJb6EumefNql1AZV:b-gmmadt) (no related file):*

Thanks for the PR!

Major comments:

1. Please revert your commit r1. The name change to `num_online_logical_cores` is not particularly interesting, and I don't think it is worth the change (`CPU_INFO` and `TOPO_INFO` are clearly different, and you have different types, so I don't see a problem).

2. Please move commit r2 (where you add the test) to the very end. Otherwise we have a git history where this commit reports test failures (if someone wants to e.g. git-bisect).

3. Merge commits r3 and r4 together -- they don't work without each other. (Again, if someone wants to e.g. git-bisect, Graphene won't be even buildable in-between these commits.)

4. It looks like the new functions in `db_main.c` and `sgx_main.c` are 100% duplicate code. Please move them and their definitions and required macros/structs into `Pal/src/host/Linux-common` and `Pal/include/host/Linux-common`. Create a new file there for these functions.

After all these changes, you'll have no choice but to rebase  this PR. Alternatively, I would recommend to close this PR, do the changes I described, and submit a new PR (otherwise reviewers will be confused).
</blockquote></details>

I split it so that it could be easy to review but you are right if we do git-bisect it might break. Thanks for #4, didn't know we could have a common functionality for both Linux and Liux-SGX PAL.

Will do the above changes and create a new PR.

---
___
*[Examples/openvino/object_detection_sample_ssd.manifest.template, line 16 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFc_Er5KSRZILHlIUu:-MMHBfJQ3nAcvglAS6eF:b-dlr8mh) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/Examples/openvino/object_detection_sample_ssd.manifest.template#L16)):*

How can this work? We moved to TOML now, so all strings must be in double-quotes.
</blockquote></details>

The version I was working with didn't have TOML changes and did a git rebase before pushing. It might be due to this. I will add double-quotes.
___
*[LibOS/shim/src/fs/proc/thread.c, line 650 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFctdq5QY4060ZqJ9s:-MMHCoMRDQVQ3SAIrn0R:bapgr32) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/proc/thread.c#L650)):*

Why is this?
</blockquote></details>

Libs like "hwloc" uses /proc/../task this to estimate number of tids. So created a dummy of 3 for `.`, `..` and one `tid`
___
*[LibOS/shim/src/fs/proc/thread.c, line 750 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFcwe1-HjKtZrt4Tmm:-MMHF1w4Dsm-vjxTLmWh:b5sa2w6) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/proc/thread.c#L750)):*

I'm confused. You're adding a new `/proc/[..]/task` subdir, but you don't add its implementation? Did you forget to git-add some files?
</blockquote></details>

I don't think so. Idea is to read the `/proc/../task` directory and its contents to have an estimate of the number of tids. What else is needed here?
___
*[Pal/include/arch/x86_64/pal-arch.h, line 30 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFdgzr8_6iIKToBZSx:-MMHHAWX9HJHYjpo0934:b-crm65i) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/Pal/include/arch/x86_64/pal-arch.h#L30)):*

Could you give some rationale and rule-of-thumb memory requirements for these constants and the general topology descriptions? I remember that on 160-core machines, it would require MBs of memory?
</blockquote></details>

so the idea was to represent plain integers (only numeric values), buffers which can be numeric with text for example `1024576 KB` and  cpu_maps like `00000000,ffffffff,00000000,ffffffff`. Based on this, came up with `PAL_SYSFS_INT_FILESZ`, `PAL_SYSFS_BUF_FILESZ` and `PAL_SYSFS_MAP_FILESZ` respectively.

So each CPU could take ~3KB and if we extend to 160 machines it would be around 480KB. In PoC version I did set large values for these and it did result in ~2MB.
___
*[Pal/include/arch/x86_64/pal-arch.h, line 237 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFe0jI6burDIQzaHgY:-MMHLT8f0_MA82dxWn3E:b924daq) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/Pal/include/arch/x86_64/pal-arch.h#L237)):*

Is there really a point in showing hugepages to the user app? Can't we just always return `0` or something?
</blockquote></details>

yes let us remove this and I will return 0. Since this might require write permissions from the user which we don't allow.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 15 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @vijaydhanraj)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFb--u3gQ8PyA9iLKa:-MMIUHkI4bKvH4K-scFE:bflkpcy) (no related file):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

I split it so that it could be easy to review but you are right if we do git-bisect it might break. Thanks for #4, didn't know we could have a common functionality for both Linux and Liux-SGX PAL.

Will do the above changes and create a new PR.
</blockquote></details>

@dimakuv regarding #4, although the functions defined in `db_main.c` and `sgx_main.c` are similar but they do return types. The functions in `sgx_main.c` return Unix error code on failure whereas the functions in `db_main.c` return PAL error code. I think for this reason only we have functions like `get_hw_resource` defined in both files. Not sure how to tackle this.

---

yamahata (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @vijaydhanraj)

___
*[LibOS/shim/src/fs/proc/thread.c, line 730 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMJWc8QBObkDrHgIm3l:-MMJWc8QBObkDrHgIm3m:b-k16hkq) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/proc/thread.c#L730)):*
> ```C
> };
> 
> static const struct pseudo_dir dir_task = {
> ```

Is this the change to procfs?
If so, please split the change from /sysfs change into independent changeset.
___
*[LibOS/shim/src/fs/sys/cache_info.c, line 30 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMJWuQ7CE1AmHwFppar:-MMJWuQ7CE1AmHwFppas:baocu5o) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cache_info.c#L30)):*
> ```C
>         return -ENOENT;
> 
>     if (!strcmp(filename, "shared_cpu_map")) {
> ```

Can we add one more member to struct pseudo_ent to carry private data?
Then we can drop those strcmp pattern. and consolidate same logics.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @vijaydhanraj)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFb--u3gQ8PyA9iLKa:-MMKMOlB8sQ_fJDT1VpP:b-s9vyrr) (no related file):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

@dimakuv regarding #4, although the functions defined in `db_main.c` and `sgx_main.c` are similar but they do return types. The functions in `sgx_main.c` return Unix error code on failure whereas the functions in `db_main.c` return PAL error code. I think for this reason only we have functions like `get_hw_resource` defined in both files. Not sure how to tackle this.
</blockquote></details>

Regarding 4. Just use Unix error codes in the shared code, and then do `unix_to_pal_error(ret)` in the callers in `db_main.c`. We should also move `get_hw_resource()` in the shared code then, with the same trick.

---
___
*[LibOS/shim/src/fs/proc/thread.c, line 650 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFctdq5QY4060ZqJ9s:-MMKMgQ-53Dt9qDoQveR:b-mserwb) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/proc/thread.c#L650)):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

Libs like "hwloc" uses /proc/../task this to estimate number of tids. So created a dummy of 3 for `.`, `..` and one `tid`
</blockquote></details>

Please add this to your comment
___
*[LibOS/shim/src/fs/proc/thread.c, line 750 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFcwe1-HjKtZrt4Tmm:-MMKMuoj3ACEazhzcKEv:b-pedlc6) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/proc/thread.c#L750)):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

I don't think so. Idea is to read the `/proc/../task` directory and its contents to have an estimate of the number of tids. What else is needed here?
</blockquote></details>

I am just surprised why we already had all the logic to implement `/proc/../task` in Graphene but didn't add this. Looks like a separate change in logic, maybe you submit this particular part as a separate PR? See also Isaku's comment above.
___
*[Pal/include/arch/x86_64/pal-arch.h, line 30 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFdgzr8_6iIKToBZSx:-MMKNBqCDNUZIIr6FLZI:b-l7oamp) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/Pal/include/arch/x86_64/pal-arch.h#L30)):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

so the idea was to represent plain integers (only numeric values), buffers which can be numeric with text for example `1024576 KB` and  cpu_maps like `00000000,ffffffff,00000000,ffffffff`. Based on this, came up with `PAL_SYSFS_INT_FILESZ`, `PAL_SYSFS_BUF_FILESZ` and `PAL_SYSFS_MAP_FILESZ` respectively.

So each CPU could take ~3KB and if we extend to 160 machines it would be around 480KB. In PoC version I did set large values for these and it did result in ~2MB.
</blockquote></details>

We need a comment like what you just wrote in this file (before these new macros).

2MB sounds good.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @vijaydhanraj, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFb--u3gQ8PyA9iLKa:-MMLZ6xF3mTbXTWB9l5c:bihsu1p) (no related file):*

Regarding 4. Just use Unix error codes in the shared code, and then do `unix_to_pal_error(ret)` in the callers in `db_main.c`. We should also move `get_hw_resource()` in the shared code then, with the same trick.
</blockquote></details>

sounds good, will do.

---
___
*[LibOS/shim/src/fs/proc/thread.c, line 730 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMJWc8QBObkDrHgIm3l:-MMLYxCV2ou00BWsyeTh:b-xgnzkd) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/proc/thread.c#L730)):*

Is this the change to procfs?
If so, please split the change from /sysfs change into independent changeset.
</blockquote></details>

Sure will do
___
*[LibOS/shim/src/fs/sys/cache_info.c, line 30 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMJWuQ7CE1AmHwFppar:-MMLZDId9hCETrdLl5-7:bd2pgny) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cache_info.c#L30)):*

Can we add one more member to struct pseudo_ent to carry private data?
Then we can drop those strcmp pattern. and consolidate same logics.

</blockquote></details>

@yamahata can you please clarify this? Trying to understand what is that we want to have as private data. Do you mean some sort of encoding for each file so that we can map this encoding with the right filename?

yamahata (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @vijaydhanraj, and @yamahata)

___
*[LibOS/shim/src/fs/sys/cache_info.c, line 30 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMJWuQ7CE1AmHwFppar:-MMMQAQR1Aqi9ysbI-mj:b-rj8lab) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cache_info.c#L30)):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

@yamahata can you please clarify this? Trying to understand what is that we want to have as private data. Do you mean some sort of encoding for each file so that we can map this encoding with the right filename?
</blockquote></details>

The purpose is to remove those ugly repeated strcmp().
- add void *data member to struct pseudo_ent.
- change pseudo_fs_ops::open(hdl, name, flags, void* data): add data argument
   (or pseudo_ent itself can be passed)
- cache_info_open uses pseudo_ent::data to elimiate strcmp()

more concreately

static void copy_shared_cpu_map(...) {
   // copy cachemap.
}
pseudo_ent {
    .data = &copy_shared_cpu_map,
}

and then
cache_info_open() {
  ...
  ((cast to function)data)(...);
  ...
}


as bonus those are common to other sysfs. add data member to each sysfs files.
You're repeating the same pattern four times. Let's consolidate them.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @vijaydhanraj, and @yamahata)

___
*[LibOS/shim/src/fs/sys/cache_info.c, line 30 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMJWuQ7CE1AmHwFppar:-MMOARwzEbPuahpgxLd7:brtwwqv) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cache_info.c#L30)):*

The purpose is to remove those ugly repeated strcmp().
- add void *data member to struct pseudo_ent.
- change pseudo_fs_ops::open(hdl, name, flags, void* data): add data argument
   (or pseudo_ent itself can be passed)
- cache_info_open uses pseudo_ent::data to elimiate strcmp()

more concreately

static void copy_shared_cpu_map(...) {
   // copy cachemap.
}
pseudo_ent {
    .data = &copy_shared_cpu_map,
}

and then
cache_info_open() {
  ...
  ((cast to function)data)(...);
  ...
}


as bonus those are common to other sysfs. add data member to each sysfs files.
You're repeating the same pattern four times. Let's consolidate them.

</blockquote></details>

if I understand correctly, we would end up having a copy function for each of the file right? There is another approach to avoid the strcmp as well. Basically, define independent `pseudo_fs_ops` for each of the file and have its own open function. For example, 
```
static const struct pseudo_fs_ops cache_shared_cpu_map = {
    .mode = &sys_info_mode,
    .stat = &sys_info_stat,
    .open = &shared_cpu_map_open,
};

static shared_cpu_map_open(...) {
     //copy cachemap
}
```
But I wanted to avoid this and consolidate all the copies into a single function which will be defeated with your suggestion.

Also, `pseduo_ent` struct is used in many other places where `data` member will not be used and we would end up increasing the size of the struct.
 I think we can live with `strcmp` function and avoid overcomplicating the implementation. But if you strongly feel otherwise, please let me know I can change it.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @vijaydhanraj, and @yamahata)

___
*[Examples/openvino/Makefile, line 64 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFcPRvBoJ_9yu9kZlE:-MMOfh4v2hzp11vMo8QI:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/Examples/openvino/Makefile#L64)):*

`-j8 `not needed. The `$(MAKE)` already propagates the command-line options the user put.
</blockquote></details>

Done.
___
*[LibOS/shim/include/shim_fs.h, line 562 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFcjGK5x0waXJWdwRh:-MMOfjZG2-gXX1S-2PCF:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/include/shim_fs.h#L562)):*

In many places in this PR, you have a space after the function name. Please remove everywhere.
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/proc/thread.c, line 650 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFctdq5QY4060ZqJ9s:-MMOfkvEAdKNTd_OcUMs:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/proc/thread.c#L650)):*

Please add this to your comment
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/proc/thread.c, line 750 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFcwe1-HjKtZrt4Tmm:-MMOfmbOANuhMMKcXSF2:b-2pai6i) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/proc/thread.c#L750)):*

I am just surprised why we already had all the logic to implement `/proc/../task` in Graphene but didn't add this. Looks like a separate change in logic, maybe you submit this particular part as a separate PR? See also Isaku's comment above.
</blockquote></details>

Separated this as a different patchset but not into a separate PR as without this the openVINO benchmark_app will fail.
___
*[LibOS/shim/src/fs/sys/cache_info.c, line 5 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFd7CZ-h5XkC_Mh2i6:-MMOg3cJBmHF8O8BgqIJ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cache_info.c#L5)):*

Please remove this line.
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/sys/cpu_info.c, line 5 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFdC5hBqxoZqtZjzbf:-MMOg5XMAXm399iU_cnK:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cpu_info.c#L5)):*

Please remove this line.
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/sys/cpu_info.c, line 137 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFdDJ3ADA25uM0SIC1:-MMOg6ou61sOsOZkbhXS:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cpu_info.c#L137)):*

Please replace all tabs with spaces in your PR.
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/sys/fs.c, line 5 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFdFlPAgvg7X3D5LZG:-MMOg7uV6-4zKQAzt2l_:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/fs.c#L5)):*

Please remove this line.
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/sys/fs.c, line 15 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFdRlO7uc9D3CUudMR:-MMOg8r1BLm2q3WqYHWh:b-y8vvjl) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/fs.c#L15)):*

Please check if you can replace some of these helper functions with what we already have in Graphene: https://github.com/oscarlab/graphene/blob/master/Pal/include/lib/api.h#L244-L245
</blockquote></details>

yes used `get_base_name` and removed `extract_filename` function.
___
*[LibOS/shim/src/fs/sys/node_info.c, line 5 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFd_wY2vTT8laJShV1:-MMOgNYL7tOMQcSEHGxl:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/node_info.c#L5)):*

Please remove this line.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_main.c, line 341 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFeD18AAgjzdx3ceoi:-MMOgPyQAFqw5JcEshVr:bae36za) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/Pal/src/host/Linux/db_main.c#L341)):*

All this stuff must be deduplicated (you have exactly the same in `sgx_main.c`). Please see my top-level comment on `Linux-common/`.
</blockquote></details>

Created a new file called `topo_info` and moved all the common functionality to it.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @vijaydhanraj, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMFb--u3gQ8PyA9iLKa:-MMOggTq9Fi3c6iEZKB-:b-opaz95) (no related file):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

sounds good, will do.
</blockquote></details>

Created a new PR, https://github.com/oscarlab/graphene/pull/1975 which takes into account all of these suggestions and added #Closes link to this PR.

---

yamahata (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @vijaydhanraj, and @yamahata)

___
*[LibOS/shim/src/fs/sys/cache_info.c, line 30 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMJWuQ7CE1AmHwFppar:-MMR4GrAAOOgzuhb6qwu:byv83lo) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cache_info.c#L30)):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

if I understand correctly, we would end up having a copy function for each of the file right? There is another approach to avoid the strcmp as well. Basically, define independent `pseudo_fs_ops` for each of the file and have its own open function. For example, 
```
static const struct pseudo_fs_ops cache_shared_cpu_map = {
    .mode = &sys_info_mode,
    .stat = &sys_info_stat,
    .open = &shared_cpu_map_open,
};

static shared_cpu_map_open(...) {
     //copy cachemap
}
```
But I wanted to avoid this and consolidate all the copies into a single function which will be defeated with your suggestion.

Also, `pseduo_ent` struct is used in many other places where `data` member will not be used and we would end up increasing the size of the struct.
 I think we can live with `strcmp` function and avoid overcomplicating the implementation. But if you strongly feel otherwise, please let me know I can change it.
</blockquote></details>

How about this?
Does this make sense to you?

cache_info_open(..., copy_fn) {
   ...
   copy_fn()
   ...
}

static copy_cpu_map() {
   ...
}

static shared_cpu_map_open() {
   cache_info_open(coppy_cpu_map)
}

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @vijaydhanraj, and @yamahata)

___
*[LibOS/shim/src/fs/sys/cache_info.c, line 30 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMJWuQ7CE1AmHwFppar:-MMR_0WMEStufiQ6NlKk:bscfopw) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cache_info.c#L30)):*

How about this?
Does this make sense to you?

cache_info_open(..., copy_fn) {
   ...
   copy_fn()
   ...
}

static copy_cpu_map() {
   ...
}

static shared_cpu_map_open() {
   cache_info_open(coppy_cpu_map)
}




</blockquote></details>

Now we don't add a data member to `pseudo_ent` that is good but we still end up creating a copy function for each filename. For instance, just under cache, we have 7 filenames so 7 new functions for just doing the same copy operation (btw even in the current implementation I am calling memcpy for each filename, I will fix this and make it common). Another question is who will call `shared_cpu_map_open`? Are you saying we should assign it to `open` function in `pseudo_fs_ops` struct?

yamahata (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @vijaydhanraj, and @yamahata)

___
*[LibOS/shim/src/fs/sys/cache_info.c, line 30 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMJWuQ7CE1AmHwFppar:-MMWWagQAT2_1OkbJKi5:b-ddhfs5) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cache_info.c#L30)):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

Now we don't add a data member to `pseudo_ent` that is good but we still end up creating a copy function for each filename. For instance, just under cache, we have 7 filenames so 7 new functions for just doing the same copy operation (btw even in the current implementation I am calling memcpy for each filename, I will fix this and make it common). Another question is who will call `shared_cpu_map_open`? Are you saying we should assign it to `open` function in `pseudo_fs_ops` struct? 
</blockquote></details>

Let me write down more details.

sysfs_open_common(..., (*copy_fn)(...)) {
   ...
   copy_fn(...); /* call function pointer instead of bunch of strcmp */
   ..l
}

static copy_shared_cpu_map() {
   // copy cpu map
}

static shared_cpu_map_open() {
  // pass the function pointer to sysfs_open_common()
  sysfs_open_common(..., &copy_shared_cpu_map);
}
...

pseudo_fs_ops shared_cpu_map_ops = {
  .open = &shared_cpu_map_open,
}
...
.end = {
  .name = "shared_cpu_map",
  .fs_ops = &shared_cpu_map_ops
}

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1970)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @vijaydhanraj, and @yamahata)

___
*[LibOS/shim/src/fs/sys/cache_info.c, line 30 at r5](https://reviewable.io/reviews/oscarlab/graphene/1970#-MMJWuQ7CE1AmHwFppar:-MMWd2Np6iAQ7qcSvbYA:bxx2ej2) ([raw file](https://github.com/oscarlab/graphene/blob/56fd34b277bd25e6332a83e7ed423c26b768369c/LibOS/shim/src/fs/sys/cache_info.c#L30)):*

Let me write down more details.

sysfs_open_common(..., (*copy_fn)(...)) {
   ...
   copy_fn(...); /* call function pointer instead of bunch of strcmp */
   ..l
}

static copy_shared_cpu_map() {
   // copy cpu map
}

static shared_cpu_map_open() {
  // pass the function pointer to sysfs_open_common()
  sysfs_open_common(..., &copy_shared_cpu_map);
}
...

pseudo_fs_ops shared_cpu_map_ops = {
  .open = &shared_cpu_map_open,
}
...
.end = {
  .name = "shared_cpu_map",
  .fs_ops = &shared_cpu_map_ops
}
</blockquote></details>

Here `shared_cpu_map_open` calls a common function `sysfs_open_common` which will call a function pointer to do specific copy of cpumap. So for each file we will end up creating,
1. `pseudo_fs_ops` struct which will have opens for each file.
2. Create an open function for each file which will do nothing but call a common open.
3. Create a copy function for each file which will be called by a common open.

As I had suggested earlier (repasting as not sure how to link :( ), we could simply remove the common open function and combine 2 and 3 as one function for each file which will directly do the copy. This would be cleaner in my opinion to avoid `strcmp`. 
```
static const struct pseudo_fs_ops cache_shared_cpu_map = {
    .mode = &sys_info_mode,
    .stat = &sys_info_stat,
    .open = &shared_cpu_map_open,
};

static shared_cpu_map_open(...) {
     //copy cachemap
}
```

But I didn't want to do this, as we will need to create a `pseudo_fs_ops` struct and an open function for each file like below. If we use `strcmp` we can avoid both creating individual `pseudo_fs_ops` for each file and can have a common open.
```
pseudo_fs_ops shared_cpu_map_ops = {
    .mode = &sys_info_mode,
    .stat = &sys_info_stat,
    **.open = &shared_cpu_map_open,**
}

pseudo_fs_ops level_ops = {
    .mode = &sys_info_mode,
    .stat = &sys_info_stat,
   **.open = &level_open,**
}

pseudo_fs_ops type = {
    .mode = &sys_info_mode,
    .stat = &sys_info_stat,
   **.open = &type_open,**
}
...
.
.

```

PR #1979, opened 2020-11-18
Introduce log levels, convert inner PAL to the logging system

pwmarcz:
I still want to verify this manually so I'm keeping it as draft until I do.

## Description of the changes

* Introduce log levels to PAL. The client (i.e. LibOS or Pal) is responsible for calling or not calling `DkDebugLog`, but the log level determined on startup.
* The logging level is "error" if `debug_type` is not specified (so that the messages previously output from `SGX_DBG` are still being output), and "info" / "trace" if it's specified, depending on the `DEBUG` variable.
  * Of course, later we should have an explicit `loader.debug_level` variable, but for now, I wanted to mimic the current behaviour.
* Convert `SGX_DBG()` to PAL logging system. Unfortunately this doesn't include the "outer" PAL as the logging system is initialized from inside. I don't have a good idea yet how to handle it.

## How to test this PR?

Run workloads with various combinations of `Linux / Linux-SGX`, `DEBUG`, `loader.debug_type`.

## Possible next steps

* Output to stderr, not stdout! (Didn't want to do it here because it will probably involve updating tests)
* Do something about the rest of `SGX_DBG` (not sure how exactly, yet)
* Make log level directly controllable via manifest (instead of `debug_type = inline`, there should be `debug_level = info/error/...`) - this also looks like a bigger change, and a breaking one...
* Make the log prefix (currently, process and thread numbers in LibOS) the same in LibOS and PAL

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1979)

### DIFF
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -36,7 +36,7 @@ struct debug_buf {
 #include "pal_debug.h"
 #include "pal_error.h"
 
-extern bool g_debug_log_enabled;
+extern unsigned int g_debug_log_level;
 
 #include <stdarg.h>
 
@@ -45,11 +45,16 @@ void debug_puts(const char* str);
 void debug_putch(int ch);
 void debug_vprintf(const char* fmt, va_list ap) __attribute__((format(printf, 1, 0)));
 
-#define debug(fmt, ...)                       \
-    do {                                      \
-        if (g_debug_log_enabled)              \
-            debug_printf(fmt, ##__VA_ARGS__); \
-    } while (0)
+#define _debug(level, fmt...)                          \
+    do {                                               \
+        if ((level) <= g_debug_log_level)              \
+            debug_printf(fmt);                         \
+    }  while(0)
+
+#define debug_error(fmt...)    _debug(PAL_LOG_ERROR, fmt)
+#define debug_info(fmt...)     _debug(PAL_LOG_INFO, fmt)
+#define debug_trace(fmt...)    _debug(PAL_LOG_TRACE, fmt)
+#define debug(fmt...)          _debug(PAL_LOG_INFO, fmt)
 
 #if 0
 #define DEBUG_BREAK_ON_FAILURE() DEBUG_BREAK()
@@ -120,11 +125,11 @@ static inline int64_t get_cur_preempt(void) {
     r func(PROTO_ARGS_##n(args));
 
 #define PARSE_SYSCALL1(name, ...) \
-    if (g_debug_log_enabled)      \
+    if (g_debug_log_level >= PAL_LOG_TRACE)      \
         parse_syscall_before(__NR_##name, #name, ##__VA_ARGS__);
 
 #define PARSE_SYSCALL2(name, ...) \
-    if (g_debug_log_enabled)      \
+    if (g_debug_log_level >= PAL_LOG_TRACE)      \
         parse_syscall_after(__NR_##name, #name, ##__VA_ARGS__);
 
 void parse_syscall_before(int sysno, const char* name, int nr, ...);
--- a/LibOS/shim/include/shim_thread.h
+++ b/LibOS/shim/include/shim_thread.h
@@ -137,7 +137,7 @@ void debug_setprefix(shim_tcb_t* tcb);
 
 static inline void debug_setbuf(shim_tcb_t* tcb,
                                 struct debug_buf* debug_buf) {
-    if (!g_debug_log_enabled)
+    if (g_debug_log_level == PAL_LOG_NONE)
         return;
 
     tcb->debug_buf = debug_buf ? debug_buf : malloc(sizeof(struct debug_buf));
--- a/LibOS/shim/src/shim_init.c
+++ b/LibOS/shim/src/shim_init.c
@@ -421,7 +421,9 @@ fail:
 extern PAL_HANDLE thread_start_event;
 
 noreturn void* shim_init(int argc, void* args) {
-    g_debug_log_enabled = PAL_CB(enable_debug_log);
+    /* For now, assume all debug() calls are at PAL_LOG_INFO level. */
+    g_debug_log_enabled = PAL_CB(debug_log_level) >= PAL_LOG_INFO;
+
     g_process_ipc_info.vmid = (IDTYPE)PAL_CB(process_id);
 
     /* create the initial TCB, shim can not be run without a tcb */
--- a/LibOS/shim/src/shim_init.c
+++ b/LibOS/shim/src/shim_init.c
@@ -421,8 +421,7 @@ fail:
 extern PAL_HANDLE thread_start_event;
 
 noreturn void* shim_init(int argc, void* args) {
-    /* For now, assume all debug() calls are at PAL_LOG_INFO level. */
-    g_debug_log_enabled = PAL_CB(debug_log_level) >= PAL_LOG_INFO;
+    g_debug_log_level = PAL_CB(debug_log_level);
 
     g_process_ipc_info.vmid = (IDTYPE)PAL_CB(process_id);
 
--- a/LibOS/shim/src/shim_parser.c
+++ b/LibOS/shim/src/shim_parser.c
@@ -553,7 +553,7 @@ static inline void skip_syscall_args(va_list* ap) {
 }
 
 void parse_syscall_before(int sysno, const char* name, int nr, ...) {
-    if (!g_debug_log_enabled)
+    if (g_debug_log_level < PAL_LOG_TRACE)
         return;
 
     struct parser_table* parser = &syscall_parser_table[sysno];
@@ -589,7 +589,7 @@ dotdotdot:
 }
 
 void parse_syscall_after(int sysno, const char* name, int nr, ...) {
-    if (!g_debug_log_enabled)
+    if (g_debug_log_level < PAL_LOG_TRACE)
         return;
 
     struct parser_table* parser = &syscall_parser_table[sysno];
--- a/LibOS/shim/src/utils/printf.c
+++ b/LibOS/shim/src/utils/printf.c
@@ -11,7 +11,7 @@
 #include "shim_internal.h"
 #include "shim_ipc.h"
 
-bool g_debug_log_enabled = false;
+unsigned int g_debug_log_level = PAL_LOG_NONE;
 
 static inline int debug_fputs(const char* buf, size_t size) {
     size_t bytes = 0;
@@ -111,7 +111,7 @@ void debug_printf(const char* fmt, ...) {
 }
 
 void debug_setprefix(shim_tcb_t* tcb) {
-    if (!g_debug_log_enabled)
+    if (g_debug_log_level == PAL_LOG_NONE)
         return;
 
     struct debug_buf* buf = tcb->debug_buf;
--- a/Pal/include/pal/pal.h
+++ b/Pal/include/pal/pal.h
@@ -87,6 +87,15 @@ typedef union pal_handle {
 
 #include "pal-arch.h"
 
+/*! Debug log level */
+enum {
+    PAL_LOG_NONE   = 0,
+    PAL_LOG_ERROR  = 1,
+    PAL_LOG_INFO   = 2,
+    PAL_LOG_TRACE  = 3,
+    PAL_LOG_ALL    = 4,
+};
+
 /********** PAL TYPE DEFINITIONS **********/
 enum {
     pal_type_file,
@@ -134,7 +143,7 @@ typedef struct PAL_CONTROL_ {
     PAL_STR executable;         /*!< executable name */
     PAL_HANDLE parent_process;  /*!< handle of parent process */
     PAL_HANDLE first_thread;    /*!< handle of first thread */
-    PAL_BOL enable_debug_log;   /*!< enable debug log calls */
+    PAL_NUM debug_log_level;    /*!< debug log level */
 
     /*
      * Memory layout
@@ -684,8 +693,8 @@ void DkObjectClose(PAL_HANDLE objectHandle);
 /*!
  * \brief Output a message to the debug stream.
  *
- * Works only if the debug stream has been initialized, which can be checked by looking at
- * `g_pal_control.enable_debug_log`.
+ * Works only if the debug stream has been initialized, which can be checked by verifying if
+ * `g_pal_control.debug_log_level` is greater than `PAL_LOG_NONE`.
  *
  * \return number of bytes written if succeeded, PAL_STREAM_ERROR on failure (in which case
  *  PAL_ERRNO() is set)
--- a/Pal/src/db_main.c
+++ b/Pal/src/db_main.c
@@ -156,6 +156,20 @@ static int insert_envs_from_manifest(const char*** envpp) {
     return 0;
 }
 
+/*
+ * Log levels. For now, determine them from loader.debug_type and DEBUG flag.
+ *
+ * TODO: Introduce explicit loader.debug_level option.
+ */
+#ifdef DEBUG
+#define LOG_LEVEL_DEFAULT PAL_LOG_TRACE
+#else
+#define LOG_LEVEL_DEFAULT PAL_LOG_INFO
+#endif
+
+#define LOG_LEVEL_DISABLED PAL_LOG_ERROR
+
+
 static void set_debug_type(void) {
     int ret = 0;
 
@@ -167,15 +181,14 @@ static void set_debug_type(void) {
     if (ret < 0)
         INIT_FAIL_MANIFEST(PAL_ERROR_DENIED, "Cannot parse \'loader.debug_type\'");
 
-    if (!debug_type)
-        return;
+    PAL_NUM debug_log_level = false;
 
-    bool enable_debug_log = false;
-
-    if (!strcmp(debug_type, "inline")) {
-        // TODO: use /dev/stderr instead?
+    if (!debug_type || !strcmp(debug_type, "none")) {
         ret = _DkInitDebugStream("/dev/stdout");
-        enable_debug_log = true;
+        debug_log_level = LOG_LEVEL_DISABLED;
+    } else if (!strcmp(debug_type, "inline")) {
+        ret = _DkInitDebugStream("/dev/stdout");
+        debug_log_level = LOG_LEVEL_DEFAULT;
     } else if (!strcmp(debug_type, "file")) {
         char* debug_file = NULL;
         ret = toml_string_in(g_pal_state.manifest_root, "loader.debug_file", &debug_file);
@@ -184,20 +197,19 @@ static void set_debug_type(void) {
 
         ret = _DkInitDebugStream(debug_file);
         free(debug_file);
-        enable_debug_log = true;
-    } else if (!strcmp(debug_type, "none")) {
-        ret = 0;
+        debug_log_level = LOG_LEVEL_DEFAULT;
     } else {
         INIT_FAIL_MANIFEST(PAL_ERROR_INVAL, "Unknown \'loader.debug_type\' "
                            "(allowed: `inline`, `file`, `none`)");
     }
 
-    free(debug_type);
+    if (debug_type)
+        free(debug_type);
 
     if (ret < 0)
         INIT_FAIL(-ret, "Cannot open debug stream");
 
-    g_pal_control.enable_debug_log = enable_debug_log;
+    g_pal_control.debug_log_level = debug_log_level;
 }
 
 /* Loads a file containing a concatenation of C-strings. The resulting array of pointers is
--- a/Pal/src/host/Linux-SGX/db_exception.c
+++ b/Pal/src/host/Linux-SGX/db_exception.c
@@ -113,14 +113,14 @@ static void save_pal_context(PAL_CONTEXT* ctx, sgx_cpu_context_t* uc,
 static void emulate_rdtsc_and_print_warning(sgx_cpu_context_t* uc) {
     static int first = 0;
     if (__atomic_exchange_n(&first, 1, __ATOMIC_RELAXED) == 0) {
-        SGX_DBG(DBG_E, "WARNING: all RDTSC/RDTSCP instructions are emulated (imprecisely) via "
+        debug_error("WARNING: all RDTSC/RDTSCP instructions are emulated (imprecisely) via "
                        "gettime() syscall.\n");
     }
 
     uint64_t usec;
     int res = _DkSystemTimeQuery(&usec);
     if (res < 0) {
-        SGX_DBG(DBG_E, "_DkSystemTimeQuery() failed in unrecoverable context, exiting.\n");
+        debug_error("_DkSystemTimeQuery() failed in unrecoverable context, exiting.\n");
         _DkProcessExit(1);
     }
     /* FIXME: Ideally, we would like to scale microseconds back to RDTSC clock cycles */
@@ -157,14 +157,14 @@ static bool handle_ud(sgx_cpu_context_t* uc) {
     } else if (instr[0] == 0xf3 && (instr[1] & ~1) == 0x48 && instr[2] == 0x0f &&
                instr[3] == 0xae && instr[4] >> 6 == 0b11 && ((instr[4] >> 3) & 0b111) < 4) {
         /* A disabled {RD,WR}{FS,GS}BASE instruction generated a #UD */
-        SGX_DBG(DBG_E, "The {RD,WR}{FS,GS}BASE instruction is not currently enabled. "
+        debug_error("The {RD,WR}{FS,GS}BASE instruction is not currently enabled. "
                        "Please reload Graphene SGX kernel module.\n");
         return false;
     } else if (instr[0] == 0x0f && instr[1] == 0x05) {
         /* syscall: LibOS may know how to handle this */
         return false;
     }
-    SGX_DBG(DBG_E, "Unknown or illegal instruction at RIP 0x%016lx\n", uc->rip);
+    debug_error("Unknown or illegal instruction at RIP 0x%016lx\n", uc->rip);
     return false;
 }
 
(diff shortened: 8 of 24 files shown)


yamahata (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1979)* status: 0 of 24 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz)

___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 331 at r1](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMR66jR9OL-b455b6-p:-MMR66jS7aBEQiHBUn2V:b-akb6fx) ([raw file](https://github.com/oscarlab/graphene/blob/e5988071baeeb48cac19b502a1a1b72a91b43529/Pal/src/host/Linux-SGX/pal_linux.h#L331)):*
> ```
>   if ((class) & DBG_LEVEL)
> ```

So are we going to drop the filtering by component?

So far DBG_LEVEL was NOT level, but filtering flag of component.
Looking at the change, E corresponds  to error, I corresponds to info, D corresponds to debug.
What about S, P, M?

Anyway please delete unused DBG_ flag. At least DBG_P isn't used now.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1979)* status: 0 of 24 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @yamahata)

___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 331 at r1](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMR66jR9OL-b455b6-p:-MMREV5J1Ih-Rz3XaW3Y:b9l4kor) ([raw file](https://github.com/oscarlab/graphene/blob/e5988071baeeb48cac19b502a1a1b72a91b43529/Pal/src/host/Linux-SGX/pal_linux.h#L331)):*
>So are we going to drop the filtering by component?

That was suggested in [the issue](https://github.com/oscarlab/graphene/issues/1947#issuecomment-728182489), I think it's a good idea to simplify the system and replace compontents with levels. But I'm happy to change if somebody needs it.

>What about S, P, M?

M probably corresponds to memory, no idea about the others. But again, maybe we don't need this distinction and levels are enough?

>Anyway please delete unused DBG_ flag. At least DBG_P isn't used now.

Okay, I will.

dimakuv (commented):
> Reviewed 24 of 24 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1979)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz and @yamahata)

___
*[LibOS/shim/include/shim_internal.h, line 56 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUkijg933rfnDngfWt:-MMUkijg933rfnDngfWu:braautf) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/LibOS/shim/include/shim_internal.h#L56)):*
> ```objc
> #define debug_error(fmt...)    _debug(PAL_LOG_ERROR, fmt)
> #define debug_info(fmt...)     _debug(PAL_LOG_INFO, fmt)
> #define debug_trace(fmt...)    _debug(PAL_LOG_TRACE, fmt)
> ```

I think we want something in-between INFO and TRACE.
- TRACE implies tracing absolutely all syscalls.
- INFO is not about syscalls but more about interesting stuff happening in Graphene.

We need something like MINIMALTRACE that doesn't trace "noisy" syscalls like `gettimeofday()` and `schedyield()`. Not sure about a good name for it. This can be outsourced to a follow-up PR if you want.
___
*[LibOS/shim/include/shim_internal.h, line 128 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUlJ_E7IMJ6zKptCFW:-MMUlJ_E7IMJ6zKptCFX:bgs8mpi) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/LibOS/shim/include/shim_internal.h#L128)):*
> ```objc
> 
> #define PARSE_SYSCALL1(name, ...) \
>     if (g_debug_log_level >= PAL_LOG_TRACE)      \
> ```

Do we even need this check given that we have the same check inside `parse_syscall_before()`?
___
*[LibOS/shim/include/shim_internal.h, line 132 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUlMua96mjkCge0A9K:-MMUlMua96mjkCge0A9L:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/LibOS/shim/include/shim_internal.h#L132)):*
> ```objc
> 
> #define PARSE_SYSCALL2(name, ...) \
>     if (g_debug_log_level >= PAL_LOG_TRACE)      \
> ```

ditto
___
*[Pal/src/host/Linux-SGX/db_exception.c, line 116 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUmFoi7gNAvPlxVmq3:-MMUmFoi7gNAvPlxVmq4:biw9nf6) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/Pal/src/host/Linux-SGX/db_exception.c#L116)):*
> ```C
>     static int first = 0;
>     if (__atomic_exchange_n(&first, 1, __ATOMIC_RELAXED) == 0) {
>         debug_error("WARNING: all RDTSC/RDTSCP instructions are emulated (imprecisely) via "
> ```

We can change this to `debug_info`.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 431 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUmesDB8m1nrJo-4zF:-MMUmesDB8m1nrJo-4zG:biinjf) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/Pal/src/host/Linux-SGX/db_files.c#L431)):*
> ```C
> 
>     if (!(prot & PAL_PROT_WRITECOPY) && (prot & PAL_PROT_WRITE)) {
>         debug_error(
> ```

`debug_info`
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 199 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUmpp279lEUoGQEAbO:-MMUmpp279lEUoGQEAbP:biinjf) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/Pal/src/host/Linux-SGX/db_main-x86_64.c#L199)):*
> ```C
>     ci->cpu_bogomips = get_bogomips();
>     if (ci->cpu_bogomips == 0.0) {
>         debug_error(
> ```

`debug_info`
___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 140 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUnqli3qSOC5daUuKf:-MMUnqli3qSOC5daUuKg:b-1gr8qr) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/Pal/src/host/Linux-SGX/enclave_pages.c#L140)):*
> ```C
>     while (check_vma_above && addr + size > check_vma_above->bottom) {
>         if (check_vma_above->is_pal_internal != is_pal_internal) {
>             debug_trace("VMA %p-%p (internal=%d) overlaps with %p-%p (internal=%d)\n", addr,
> ```

Let's remove this completely, too much noise, and this subsystem works fine
___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 152 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUnxVJ3mHwavkrM_Oq:-MMUnxVJ3mHwavkrM_Or:b-1gr8qr) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/Pal/src/host/Linux-SGX/enclave_pages.c#L152)):*
> ```C
>     while (check_vma_below && addr < check_vma_below->top) {
>         if (check_vma_below->is_pal_internal != is_pal_internal) {
>             debug_trace("VMA %p-%p (internal=%d) overlaps with %p-%p (internal=%d)\n", addr,
> ```

Let's remove this completely, too much noise, and this subsystem works fine
___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 179 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUnyEF6-rG-3FXx7fM:-MMUnyEF6-rG-3FXx7fN:b-1gr8qr) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/Pal/src/host/Linux-SGX/enclave_pages.c#L179)):*
> ```C
>            vma_above->is_pal_internal == vma->is_pal_internal) {
>         /* newly created VMA grows into above VMA; expand newly created VMA and free above-VMA */
>         debug_trace("Merge %p-%p and %p-%p\n", vma->bottom, vma->top, vma_above->bottom,
> ```

Let's remove this completely, too much noise, and this subsystem works fine
___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 196 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUnyoYBPuodfWXcF6P:-MMUnyoYBPuodfWXcF6Q:b-1gr8qr) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/Pal/src/host/Linux-SGX/enclave_pages.c#L196)):*
> ```C
>            vma_below->is_pal_internal == vma->is_pal_internal) {
>         /* newly created VMA grows into below VMA; expand newly create VMA and free below-VMA */
>         debug_trace("Merge %p-%p and %p-%p\n", vma->bottom, vma->top, vma_below->bottom,
> ```

Let's remove this completely, too much noise, and this subsystem works fine
___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 212 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUnzVZ9oFiiYXkIKYU:-MMUnzVZ9oFiiYXkIKYV:b-1gr8qr) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/Pal/src/host/Linux-SGX/enclave_pages.c#L212)):*
> ```C
>     INIT_LIST_HEAD(vma, list);
>     LISTP_ADD_AFTER(vma, vma_above, &g_heap_vma_list, list);
>     debug_trace("Created vma %p-%p\n", vma->bottom, vma->top);
> ```

Let's remove this completely, too much noise, and this subsystem works fine
___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 242 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUo-HYDDBAJ3eGeBC4:-MMUo-HYDDBAJ3eGeBC5:b-1gr8qr) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/Pal/src/host/Linux-SGX/enclave_pages.c#L242)):*
> ```C
>     assert(access_ok(addr, size));
> 
>     debug_trace("Allocating %lu bytes in enclave memory at %p (%s)\n", size, addr,
> ```

Let's remove this completely, too much noise, and this subsystem works fine
___
*[Pal/src/host/Linux-SGX/enclave_pages.c, line 305 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUo-y-5bzsTN2KV-AT:-MMUo-y-5bzsTN2KV-AU:b-1gr8qr) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/Pal/src/host/Linux-SGX/enclave_pages.c#L305)):*
> ```C
>     }
> 
>     debug_trace("Freeing %lu bytes in enclave memory at %p\n", size, addr);
> ```

Let's remove this completely, too much noise, and this subsystem works fine
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 561 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUoJYJC2TNEOWQz0wV:-MMUoJYJC2TNEOWQz0wW:b-1gr8qr) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/Pal/src/host/Linux-SGX/enclave_pf.c#L561)):*
> ```C
>     pf_status_t pfs;
> 
>     debug_trace("flush_pf_maps: pf %p, buf %p, remove %d\n", pf, buffer, remove);
> ```

Let's remove this completely, too much noise, and this subsystem works fine
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 577 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUoKrxEFMJ_VgC0tp7:-MMUoKrxEFMJ_VgC0tp8:b-1gr8qr) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/Pal/src/host/Linux-SGX/enclave_pf.c#L577)):*
> ```C
>         assert(PF_SUCCESS(pfs));
> 
>         debug_trace("flush_pf_maps: pf %p, buf %p, map size %lu, offset %lu\n", map_pf,
> ```

Let's remove this completely, too much noise, and this subsystem works fine
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 331 at r1](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMR66jR9OL-b455b6-p:-MMUoZcN7PnPEJIK7mlz:b2sc67m) ([raw file](https://github.com/oscarlab/graphene/blob/e5988071baeeb48cac19b502a1a1b72a91b43529/Pal/src/host/Linux-SGX/pal_linux.h#L331)):*

>So are we going to drop the filtering by component?

That was suggested in [the issue](https://github.com/oscarlab/graphene/issues/1947#issuecomment-728182489), I think it's a good idea to simplify the system and replace compontents with levels. But I'm happy to change if somebody needs it.

>What about S, P, M?

M probably corresponds to memory, no idea about the others. But again, maybe we don't need this distinction and levels are enough?

>Anyway please delete unused DBG_ flag. At least DBG_P isn't used now.

Okay, I will.
</blockquote></details>

I suggest dropping "filter by component". Why can't we remove `SGX_DBG()` completely again?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1979)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @pwmarcz, and @yamahata)

___
*[LibOS/shim/include/shim_internal.h, line 56 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUkijg933rfnDngfWt:-MMVJKRkEGtIyGO6kE7F:b-2doyqa) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/LibOS/shim/include/shim_internal.h#L56)):*

I think we want something in-between INFO and TRACE.
- TRACE implies tracing absolutely all syscalls.
- INFO is not about syscalls but more about interesting stuff happening in Graphene.

We need something like MINIMALTRACE that doesn't trace "noisy" syscalls like `gettimeofday()` and `schedyield()`. Not sure about a good name for it. This can be outsourced to a follow-up PR if you want.
</blockquote></details>

The level between INFO and TRACE is often called DEBUG, but it's kind of awkward because the whole feature is called "debug" or "debug log".

How about we rename `debug` to `log`? E.g. `log_info()`, `log_debug()`, `log_trace()`.
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 331 at r1](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMR66jR9OL-b455b6-p:-MMVJp3fD2MOLevHv-fJ:bg9kxs8) ([raw file](https://github.com/oscarlab/graphene/blob/e5988071baeeb48cac19b502a1a1b72a91b43529/Pal/src/host/Linux-SGX/pal_linux.h#L331)):*

I suggest dropping "filter by component". Why can't we remove `SGX_DBG()` completely again?
</blockquote></details>

Two problems:

1. Right now the logging system is inside the enclave (`_DkDebugLog`, `g_debug_log_fd`, `ocall_write`),
2. Many of the remaining `SGX_DBG()` calls happen during enclave initialization, before we initialize the logging system. That's  because we initialize the logging system inside `pal_main()`, so that the same code can be used for all hosts.

I guess a better solution would be to initialize the logging system earlier, in `sgx_main`. Then we could pass the debug log FD to the enclave. And the inner PAL would decide what to log (i.e. log level) by looking at the manifest.

That looks like a tricky change to get right, without too much duplication, etc. I would like to get this PR done first (and probably shape the API better in the process), and also update manifest format to include log level. That's why I propose to keep the rest of `SGX_DBG()` as-is for the moment.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1979)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @pwmarcz, and @yamahata)

___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 331 at r1](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMR66jR9OL-b455b6-p:-MMVUAPCDoNlCxmfmrbV:bjqtrb0) ([raw file](https://github.com/oscarlab/graphene/blob/e5988071baeeb48cac19b502a1a1b72a91b43529/Pal/src/host/Linux-SGX/pal_linux.h#L331)):*
>I guess a better solution would be to initialize the logging system earlier, in sgx_main. Then we could pass the debug log FD to the enclave.

Alternatively, we don't pass anything inside, but have initialization code (based on manifest) both in `sgx_main` (outside) and `pal_main` (inside the enclave). Effectively we would have two logging systems with the same API. It doesn't sound too ugly, but I'm not sure how you feel about parsing the manifest so early?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1979)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz and @yamahata)

___
*[LibOS/shim/include/shim_internal.h, line 56 at r2](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMUkijg933rfnDngfWt:-MMZygffDWbWmzAqPqi0:b-fp05e) ([raw file](https://github.com/oscarlab/graphene/blob/756a2e41ce794699ad5d9486410e7bfe8b87066f/LibOS/shim/include/shim_internal.h#L56)):*

The level between INFO and TRACE is often called DEBUG, but it's kind of awkward because the whole feature is called "debug" or "debug log".

How about we rename `debug` to `log`? E.g. `log_info()`, `log_debug()`, `log_trace()`.
</blockquote></details>

Both `DEBUG` and `log_*()` proposals sound good to me.
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 331 at r1](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMR66jR9OL-b455b6-p:-MMZzFeRBbhU-vp0ZlZy:b-9u5n13) ([raw file](https://github.com/oscarlab/graphene/blob/e5988071baeeb48cac19b502a1a1b72a91b43529/Pal/src/host/Linux-SGX/pal_linux.h#L331)):*

>I guess a better solution would be to initialize the logging system earlier, in sgx_main. Then we could pass the debug log FD to the enclave.

Alternatively, we don't pass anything inside, but have initialization code (based on manifest) both in `sgx_main` (outside) and `pal_main` (inside the enclave). Effectively we would have two logging systems with the same API. It doesn't sound too ugly, but I'm not sure how you feel about parsing the manifest so early?
</blockquote></details>

Do we actually want to rely on the manifest during bootstrapping? I mean, we print some messages even before loading and parsing the manifest. Shouldn't we just do normal `printf/write`? And we can simply have the logic of "print all errors, but print warnings/info only if command-line has --verbose-bootstrap" or smth like this.

I personally don't like the idea of relying on the manifest option during bootstrap.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1979)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @pwmarcz, and @yamahata)

___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 331 at r1](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMR66jR9OL-b455b6-p:-MM_J6ImDDpoegoQq1H9:b-cf1cob) ([raw file](https://github.com/oscarlab/graphene/blob/e5988071baeeb48cac19b502a1a1b72a91b43529/Pal/src/host/Linux-SGX/pal_linux.h#L331)):*

Do we actually want to rely on the manifest during bootstrapping? I mean, we print some messages even before loading and parsing the manifest. Shouldn't we just do normal `printf/write`? And we can simply have the logic of "print all errors, but print warnings/info only if command-line has --verbose-bootstrap" or smth like this.

I personally don't like the idea of relying on the manifest option during bootstrap.
</blockquote></details>

Yes, that's what I was worried about. So bootstrapping messages can be special. Still, there are also outer PAL messages *after* bootstrapping, e.g. messages from inside OCALLs. Should they also be logged to console/stderr, or should they follow the configuration in manifest?

Some possibilities I see:

1. Introduce a separate API for outer PAL, analogous to the inner one: `sgx_log_info()`, `sgx_log_warning()` etc. It would not be configured using manifest (for example: output is always stderr, the level is hardcoded depending on `DEBUG` variable).

    This is the simplest solution but at the cost of having messages from OCALLs go somewhere else.

2. Make the outer logging system use the same names as the inner one (`log_info()`, `log_warning()` etc.). Initialize it to hardcoded values (stderr, level) as above, but as the inner PAL initializes, it also sets up the outer one, so that further messages end up in a log file.

    That's a bit more complicated, but more user-friendly.

3. Have separate logging systems for bootstrapping and other outer-PAL messages, so that it's clear which messages go to console, and which ones possibly go to file.

    Honestly, that sounds pretty bad. Not only we have two systems to maintain (three, if you count the inner one), you need to also be careful not to call the wrong one before it's initialized.

What do you think?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1979)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz and @yamahata)

___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 331 at r1](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMR66jR9OL-b455b6-p:-MM_Ymdh69dMqI6hB1k1:b-b479pj) ([raw file](https://github.com/oscarlab/graphene/blob/e5988071baeeb48cac19b502a1a1b72a91b43529/Pal/src/host/Linux-SGX/pal_linux.h#L331)):*

Yes, that's what I was worried about. So bootstrapping messages can be special. Still, there are also outer PAL messages *after* bootstrapping, e.g. messages from inside OCALLs. Should they also be logged to console/stderr, or should they follow the configuration in manifest?

Some possibilities I see:

1. Introduce a separate API for outer PAL, analogous to the inner one: `sgx_log_info()`, `sgx_log_warning()` etc. It would not be configured using manifest (for example: output is always stderr, the level is hardcoded depending on `DEBUG` variable).

    This is the simplest solution but at the cost of having messages from OCALLs go somewhere else.

2. Make the outer logging system use the same names as the inner one (`log_info()`, `log_warning()` etc.). Initialize it to hardcoded values (stderr, level) as above, but as the inner PAL initializes, it also sets up the outer one, so that further messages end up in a log file.

    That's a bit more complicated, but more user-friendly.

3. Have separate logging systems for bootstrapping and other outer-PAL messages, so that it's clear which messages go to console, and which ones possibly go to file.

    Honestly, that sounds pretty bad. Not only we have two systems to maintain (three, if you count the inner one), you need to also be careful not to call the wrong one before it's initialized.

What do you think?
</blockquote></details>

I prefer option 2.

Also, do we really have debug messages during untrusted execution of OCALLs? If we do, we should probably just remove these. OCALLs must be literally "just execute the syscall and return the result" (or generally do one simple operation), so what is there to output?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1979)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @pwmarcz, and @yamahata)

___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 331 at r1](https://reviewable.io/reviews/oscarlab/graphene/1979#-MMR66jR9OL-b455b6-p:-MMaF2r3-xZH4QkwFeWG:bdhkc73) ([raw file](https://github.com/oscarlab/graphene/blob/e5988071baeeb48cac19b502a1a1b72a91b43529/Pal/src/host/Linux-SGX/pal_linux.h#L331)):*

I prefer option 2.

Also, do we really have debug messages during untrusted execution of OCALLs? If we do, we should probably just remove these. OCALLs must be literally "just execute the syscall and return the result" (or generally do one simple operation), so what is there to output?
</blockquote></details>

Looking at it more closely, yes, there's just a few messages directly in OCALLs, but I see also other places that in outer PAL that log messages: `rpc_thread_loop`, error messages on exit, starting a new thread. It would be nice to have these as part of the full log.

Still, I think that can be done in the next steps, right now I would rather leave this part (outer `SGX_DBG` calls) as is.

PR #1997, opened 2020-11-25
[Pal/Linux-SGX] Get rid of sgx driver submodule

woju:
## Description of the changes 

- `link-intel-driver.py` is imported from oscarlab/graphene-sgx-driver and the submodule is removed.
- Interactive `input()` is removed, because it is unwieldy in build system.
- There are some changes in includes and include dirs, to include only the new file.

## How to test this PR? 

Jenkins, as usual.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)

### DIFF
--- a/.ci/lib/stage-build-sgx.jenkinsfile
+++ b/.ci/lib/stage-build-sgx.jenkinsfile
@@ -18,19 +18,12 @@ stage('build') {
         # test the build with the DCAP driver v1.6 and clean up afterwards
 
         cd "$WORKSPACE"
-
-        ISGX_DRIVER_PATH=/opt/intel/SGXDataCenterAttestationPrimitives/driver/linux \
-            make ${MAKEOPTS} -C Pal/src/host/Linux-SGX/sgx-driver
-        make ${MAKEOPTS}
-
+        make ${MAKEOPTS} ISGX_DRIVER_PATH=/opt/intel/SGXDataCenterAttestationPrimitives/driver/linux
         make ${MAKEOPTS} clean
-        make ${MAKEOPTS} -C Pal/src/host/Linux-SGX/sgx-driver distclean
     '''
 
     sh '''
-        ISGX_DRIVER_PATH=/opt/intel/linux-sgx-driver \
-            make ${MAKEOPTS} -C Pal/src/host/Linux-SGX/sgx-driver
-        make ${MAKEOPTS}
+        make ${MAKEOPTS} ISGX_DRIVER_PATH=/opt/intel/linux-sgx-driver
     '''
 
     try {
--- a/.ci/run-pylint
+++ b/.ci/run-pylint
@@ -18,7 +18,6 @@ find . -name \*.py \
     -and -not -path ./LibOS/shim/test/ltp/build/\* \
     -and -not -path ./LibOS/shim/test/ltp/install/\* \
     -and -not -path ./Examples/pytorch/\* \
-    -and -not -path ./Pal/src/host/Linux-SGX/sgx-driver/\* \
 | sed 's/./\\&/g' \
 | xargs "${PYLINT}" "$@" \
     Pal/src/host/Linux-SGX/signer/pal-sgx-get-token \
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "Pal/src/host/Linux-SGX/sgx-driver"]
-	path = Pal/src/host/Linux-SGX/sgx-driver
-	url = https://github.com/oscarlab/graphene-sgx-driver.git
 [submodule "LibOS/shim/test/ltp/src"]
 	path = LibOS/shim/test/ltp/src
 	url = https://github.com/linux-test-project/ltp.git
--- a/Documentation/building.rst
+++ b/Documentation/building.rst
@@ -72,23 +72,24 @@ Run the following commands on Ubuntu to install SGX-related dependencies::
     sudo apt install -y python3-pip
     sudo /usr/bin/pip3 install protobuf
 
-2. Install the Linux kernel patched with FSGSBASE
-"""""""""""""""""""""""""""""""""""""""""""""""""
+2a. Install the Linux kernel patched with FSGSBASE
+""""""""""""""""""""""""""""""""""""""""""""""""""
 
 FSGSBASE is a feature in recent processors which allows direct access to the FS
 and GS segment base addresses. For more information about FSGSBASE and its
 benefits, see `this discussion <https://lwn.net/Articles/821719>`__.
+FSGSBASE patchset was merged in 5.9. For older kernels it is available as
+`separate patches <https://github.com/oscarlab/graphene-sgx-driver/tree/master/fsgsbase_patches>`__.
 
-Work is being done to include FSGSBASE enabling in the upstream Linux kernel.
-Currently, the FSGSBASE enabling code is out-of-tree, requiring some patches to
-the kernel.
+The following instructions to patch and compile a Linux kernel with FSGSBASE
+support below are written around Ubuntu 18.04 LTS (Bionic Beaver) with a Linux
+5.4 LTS stable kernel but can be adapted for other distros as necessary. These
+instructions ensure that the resulting kernel has FSGSBASE support and up to
+date security mitigations.
 
-Enabling FSGSBASE support requires building and installing a custom kernel with
-backported patches. The instructions to patch and compile a Linux kernel with
-FSGSBASE support below are written around Ubuntu 18.04 LTS (Bionic Beaver) with
-a Linux 5.4 LTS stable kernel but can be adapted for other distros as necessary.
-These instructions ensure that the resulting kernel has FSGSBASE support and up
-to date security mitigations.
+#. Clone the repository with patches::
+
+       git clone https://github.com/oscarlab/graphene-sgx-driver
 
 #. Setup a build environment for kernel development following `the instructions
    in the Ubuntu wiki <https://wiki.ubuntu.com/KernelTeam/GitKernelBuild>`__.
@@ -100,7 +101,7 @@ to date security mitigations.
 
 #. Apply the provided FSGSBASE patches to the kernel source tree::
 
-       git am <graphene-dir>/Pal/src/host/Linux-SGX/sgx-driver/fsgsbase_patches/*.patch
+       git am <graphene-sgx-driver>/fsgsbase_patches/*.patch
 
    The conversation regarding this patchset can be found in the kernel mailing
    list archives `here
@@ -125,6 +126,27 @@ SDK/PSW, and Graphene itself (see next steps). Note that older versions of
 these software packages may not work with recent Linux kernels like 5.4. We
 recommend to use commit ``b7ccf6f`` of the Intel SGX Linux Driver for Intel SGX
 DCAP and commit ``0e71c22`` of the Intel SGX SDK/PSW.
+
+2b. Install the Graphene FSGSBASE driver (not for production)
+"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+If you followed step 2a and installed the patched Linux kernel, skip this step.
+Otherwise, you will need a Graphene-specific Linux driver that enables the
+FSGSBASE feature available in recent processors.
+
+.. warning::
+
+   This module is a |~| quick-and-dirty hack with dangerous security hole
+   (allows unauthorized local privilege escalation). "Do not use for production"
+   is not a |~| joke. We use it only for testing on very old kernels where the
+   patchset does not apply cleanly.
+
+To install the Graphene FSGSBASE driver, run the following commands::
+
+   git clone https://github.com/oscarlab/graphene-sgx-driver
+   cd graphene-sgx-driver
+   make
+   sudo insmod gsgx.ko
 
 3. Generate signing keys
 """"""""""""""""""""""""
@@ -157,22 +179,6 @@ Alternatively, if you want to use the DCAP versions of the SDK and driver,
 download and install it from:
 
 - https://github.com/intel/SGXDataCenterAttestationPrimitives
-
-5. Install the Graphene SGX driver (not for production)
-"""""""""""""""""""""""""""""""""""""""""""""""""""""""
-
-If you followed step 1 and installed the patched Linux kernel, skip this step.
-Otherwise, you will need a Graphene-specific Linux driver that enables the
-FSGSBASE feature available in recent processors.
-
-To install the Graphene SGX driver, run the following commands::
-
-   git submodule update --init -- Pal/src/host/Linux-SGX/sgx-driver
-   cd Pal/src/host/Linux-SGX/sgx-driver
-   make
-   # the console will prompt you for the path to the Intel SGX driver code
-   # (simply press ENTER if you use the in-kernel Intel SGX driver)
-   sudo insmod gsgx.ko
 
 Building
 ^^^^^^^^
--- a/Documentation/cloud-deployment.rst
+++ b/Documentation/cloud-deployment.rst
@@ -40,7 +40,6 @@ Building
 
        git clone https://github.com/oscarlab/graphene.git
        cd graphene
-       git submodule update --init -- Pal/src/host/Linux-SGX/sgx-driver/
 
 #. Prepare the signing keys::
 
--- a/Documentation/quickstart.rst
+++ b/Documentation/quickstart.rst
@@ -55,7 +55,6 @@ second command should list the process status of :command:`aesm_service`.
 
       git clone https://github.com/oscarlab/graphene.git
       cd graphene
-      git submodule update --init -- Pal/src/host/Linux-SGX/sgx-driver/
       export GRAPHENE_DIR=$PWD
 
 #. Prepare a signing key::
--- a/LICENSE.addendum.txt
+++ b/LICENSE.addendum.txt
@@ -1,9 +1,5 @@
 Graphene itself is licensed under the LGPL-3.0-or-later.
 
-Graphene on an SGX host requires a kernel driver (in
-Pal/src/host/Linux-SGX/sgx-driver), which is separately licensed under
-the GPL.
-
 Graphene also includes the following third party sources (and licenses):
 
 mbedtls crypto libraries - Apache 2.0
--- a/Pal/Makefile
+++ b/Pal/Makefile
@@ -33,7 +33,6 @@ format:
 	                               -path ./src/host/Linux-SGX/tools/common/cJSON.h -prune -o \
 	                               -path ./src/host/Linux-SGX/tools/common/cJSON-*/cJSON.c -prune -o \
 	                               -path ./src/host/Linux-SGX/tools/common/cJSON-*/cJSON.h -prune -o \
-	                               -path ./src/host/Linux-SGX/sgx-driver -prune -o \
 	                               \( -name '*.h' -o -name '*.c' \) -print)
 
 .PHONY: distclean
--- a/Pal/src/host/Linux-SGX/.gitignore
+++ b/Pal/src/host/Linux-SGX/.gitignore
@@ -1,6 +1,7 @@
 /asm-offsets.h
 /generated-offsets.s
 /generated_offsets.py
+/gsgx.h
 /pal-sgx
 /quote/aesm.pb-c.c
 /quote/aesm.pb-c.h
--- a/Pal/src/host/Linux-SGX/Makefile
+++ b/Pal/src/host/Linux-SGX/Makefile
@@ -14,8 +14,7 @@ CFLAGS += \
 	-I../../../include/lib \
 	-I../../../lib/crypto/mbedtls/include \
 	-I../../../lib/crypto/mbedtls/crypto/include \
-	-Iprotected-files \
-	-Isgx-driver
+	-Iprotected-files
 
 # Some of the code uses alignof on expressions, which is a GNU extension.
 # Silence Clang - it complains but does support it.
@@ -89,7 +88,7 @@ urts-asm-objs = sgx_entry.o
 graphene_lib = .lib/graphene-lib.a
 
 .PHONY: all
-all: sgx-driver/sgx.h $(host_files) tools
+all: gsgx.h $(host_files) tools
 
 libpal-Linux-SGX.a: $(enclave-objs) $(enclave-asm-objs)
 	$(call cmd,ar_a_o)
@@ -138,8 +137,8 @@ gdb_integration/sgx_gdb.so: gdb_integration/sgx_gdb.c
 
 enclave_entry.o sgx_entry.o: asm-offsets.h
 
-sgx-driver/sgx.h:
-	$(MAKE) -C sgx-driver $(notdir $@)
+gsgx.h: gsgx.h.in
+	./link-intel-driver.py < $< > $@
 
 ifeq ($(filter %clean,$(MAKECMDGOALS)),)
 include $(wildcard *.d) $(wildcard gdb_integration/*.d)
@@ -154,6 +153,7 @@ tools:
 CLEAN_FILES += $(notdir $(pal_static) $(pal_lib) $(pal_loader))
 CLEAN_FILES += gdb_integration/sgx_gdb.so
 CLEAN_FILES += quote/aesm.pb-c.c quote/aesm.pb-c.h quote/aesm.pb-c.d quote/aesm.pb-c.o
+CLEAN_FILES += gsgx.h
 
 .PHONY: clean_
 clean_:
--- a/Pal/src/host/Linux-SGX/Makefile
+++ b/Pal/src/host/Linux-SGX/Makefile
@@ -163,12 +163,10 @@ clean_:
 
 .PHONY: clean
 clean: clean_
-	$(MAKE) -C sgx-driver $@
 	$(MAKE) -C tools $@
 
 .PHONY: distclean
 distclean: clean_
-	$(MAKE) -C sgx-driver $@
 	$(MAKE) -C tools $@
 
 .PHONY: test
--- a/Pal/src/host/Linux-SGX/db_misc.c
+++ b/Pal/src/host/Linux-SGX/db_misc.c
@@ -18,12 +18,7 @@
 #include "pal_linux.h"
 #include "pal_linux_defs.h"
 #include "pal_security.h"
-/* sgx.h is required to define SGX_DCAP,
- * and doesn't have a definition for __packed */
-#ifndef __packed
-#define __packed __attribute__((packed))
-#endif
-#include "sgx.h"
+#include "gsgx.h"
 #include "sgx_api.h"
 #include "sgx_attest.h"
 #include "toml.h"
--- a/Pal/src/host/Linux-SGX/generated-offsets.c
+++ b/Pal/src/host/Linux-SGX/generated-offsets.c
@@ -175,7 +175,4 @@ __attribute__((__used__)) static void dummy(void) {
 #ifdef SGX_DCAP
     DEFINE(SGX_DCAP, SGX_DCAP);
 #endif
-#ifdef SGX_DCAP_16_OR_LATER
-    DEFINE(SGX_DCAP_16_OR_LATER, SGX_DCAP_16_OR_LATER);
-#endif
 }
--- a/Pal/src/host/Linux-SGX/generated-offsets.c
+++ b/Pal/src/host/Linux-SGX/generated-offsets.c
@@ -11,12 +11,7 @@
 #include "sgx_arch.h"
 #include "sgx_tls.h"
 
-/* sgx.h header from the Intel SGX driver assumes that `__packed` macro was defined */
-#ifndef __packed
-#define __packed __attribute__((packed))
-#endif
-#include "sgx.h"
-#undef __packed
+#include "gsgx.h"
 
 __attribute__((__used__)) static void dummy(void) {
     /* defines in sgx_arch.h */
--- a/Pal/src/host/Linux-SGX/gsgx.h
+++ b/Pal/src/host/Linux-SGX/gsgx.h
@@ -1,0 +1,52 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+/* (C) Copyright 2020 Intel Corporation
+ *                    Dmitrii Kuvaiskii <dmitrii.kuvaiskii@intel.com>
+ */
+
+#ifndef __ARCH_GSGX_H__
+#define __ARCH_GSGX_H__
+
+#ifndef __packed
+#define __packed __attribute__((packed))
+#endif
+
+#include <linux/stddef.h>
+#include <linux/types.h>
+
+#include "sgx.h"
+
+#define GSGX_FILE  "/dev/gsgx"
+#define GSGX_MINOR MISC_DYNAMIC_MINOR
+
+/* Graphene needs the below subset of SGX instructions' return values */
+#ifndef SGX_INVALID_SIG_STRUCT
+#define SGX_INVALID_SIG_STRUCT  1
+#endif
+
+#ifndef SGX_INVALID_ATTRIBUTE
+#define SGX_INVALID_ATTRIBUTE   2
+#endif
+
+#ifndef SGX_INVALID_MEASUREMENT
+#define SGX_INVALID_MEASUREMENT 4
+#endif
+
+#ifndef SGX_INVALID_SIGNATURE
+#define SGX_INVALID_SIGNATURE   8
+#endif
+
+#ifndef SGX_INVALID_EINITTOKEN
+#define SGX_INVALID_EINITTOKEN  16
+#endif
+
+#ifndef SGX_INVALID_CPUSVN
+#define SGX_INVALID_CPUSVN      32
+#endif
+
+/* SGX_INVALID_LICENSE was renamed to SGX_INVALID_EINITTOKEN in SGX driver 2.1:
+ *   https://github.com/intel/linux-sgx-driver/commit/a7997dafe184d7d527683d8d46c4066db205758d */
+#ifndef SGX_INVALID_LICENSE
+#define SGX_INVALID_LICENSE     SGX_INVALID_EINITTOKEN
+#endif
+
+#endif /* __ARCH_GSGX_H__ */
--- a/Pal/src/host/Linux-SGX/gsgx.h
+++ b/Pal/src/host/Linux-SGX/gsgx.h
@@ -1,4 +1,4 @@
-/* SPDX-License-Identifier: GPL-2.0-only */
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
 /* (C) Copyright 2020 Intel Corporation
  *                    Dmitrii Kuvaiskii <dmitrii.kuvaiskii@intel.com>
  */
--- a/Pal/src/host/Linux-SGX/gsgx.h.in
+++ b/Pal/src/host/Linux-SGX/gsgx.h.in
@@ -13,10 +13,9 @@
 #include <linux/stddef.h>
 #include <linux/types.h>
 
-#include "sgx.h"
+#include "@DRIVER_SGX_H@"
 
 #define GSGX_FILE  "/dev/gsgx"
-#define GSGX_MINOR MISC_DYNAMIC_MINOR
 
 /* Graphene needs the below subset of SGX instructions' return values */
 #ifndef SGX_INVALID_SIG_STRUCT
@@ -47,6 +46,14 @@
  *   https://github.com/intel/linux-sgx-driver/commit/a7997dafe184d7d527683d8d46c4066db205758d */
 #ifndef SGX_INVALID_LICENSE
 #define SGX_INVALID_LICENSE     SGX_INVALID_EINITTOKEN
+#endif
+
+#define ISGX_FILE "@ISGX_FILE@"
+
+@DEFINE_DCAP@
+
+#ifdef SGX_DCAP
+# define SGX_DCAP_16_OR_LATER 1
 #endif
 
 #endif /* __ARCH_GSGX_H__ */
--- a/Pal/src/host/Linux-SGX/gsgx.h.in
+++ b/Pal/src/host/Linux-SGX/gsgx.h.in
@@ -52,8 +52,4 @@
 
 @DEFINE_DCAP@
 
-#ifdef SGX_DCAP
-# define SGX_DCAP_16_OR_LATER 1
-#endif
-
 #endif /* __ARCH_GSGX_H__ */
--- a/Pal/src/host/Linux-SGX/link-intel-driver.py
+++ b/Pal/src/host/Linux-SGX/link-intel-driver.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+
+import sys, os, shutil
+
+DRIVER_VERSIONS = {
+        'sgx_user.h':                 '/dev/isgx',
+        'include/uapi/asm/sgx_oot.h': '/dev/sgx/enclave',
+        'include/uapi/asm/sgx.h':     '/dev/sgx/enclave',
+        'sgx_in_kernel.h':            '/dev/sgx/enclave',
+}
+
+def find_intel_sgx_driver():
+    """
+    Graphene only needs one header from the Intel SGX Driver:
+      - sgx_user.h for non-DCAP, older version of the driver
+        (https://github.com/intel/linux-sgx-driver)
+      - include/uapi/asm/sgx_oot.h for DCAP 1.6+ version of the driver
+        (https://github.com/intel/SGXDataCenterAttestationPrimitives)
+      - include/uapi/asm/sgx.h for in-kernel 20+ version of the driver
+        (https://lore.kernel.org/linux-sgx/20190417103938.7762-1-jarkko.sakkinen@linux.intel.com/)
+      - default sgx_in_kernel.h for in-kernel 32+ version of the driver
+        (https://lore.kernel.org/linux-sgx/20200716135303.276442-1-jarkko.sakkinen@linux.intel.com)
+
+    This function returns the required header from the SGX driver.
+    """
+    isgx_driver_path = os.getenv("ISGX_DRIVER_PATH")
+    if not isgx_driver_path:
+        msg = 'Enter the Intel SGX driver dir with C headers (or press ENTER for in-kernel driver): '
+        isgx_driver_path = os.path.expanduser(input(msg))
+
+    if not isgx_driver_path or isgx_driver_path.strip() == '':
+        # user did not specify any driver path, use default in-kernel driver's C header
+        isgx_driver_path = os.path.dirname(os.path.abspath(__file__))
+
+    for header_path, dev_path in DRIVER_VERSIONS.items():
+        abs_header_path = os.path.abspath(os.path.join(isgx_driver_path, header_path))
+        if os.path.exists(abs_header_path):
+            return abs_header_path, dev_path
+
+    raise Exception("Could not find the header from the Intel SGX Driver")
+
+
+def main():
+    """ Find and copy header/device paths from Intel SGX Driver"""
+    header_path, dev_path = find_intel_sgx_driver()
+
+    this_header_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'sgx.h')
+    shutil.copyfile(header_path, this_header_path)
+
+    with open(this_header_path, 'a') as f:
+        f.write('\n\n#ifndef ISGX_FILE\n#define ISGX_FILE "%s"\n#endif\n' % dev_path)
+        if dev_path == '/dev/sgx' or dev_path == '/dev/sgx/enclave':
+            f.write('\n\n#ifndef SGX_DCAP\n#define SGX_DCAP 1\n#endif\n')
+        if dev_path == '/dev/sgx/enclave':
+            f.write('\n\n#ifndef SGX_DCAP_16_OR_LATER\n#define SGX_DCAP_16_OR_LATER 1\n#endif\n')
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/Pal/src/host/Linux-SGX/link-intel-driver.py
+++ b/Pal/src/host/Linux-SGX/link-intel-driver.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 
-import sys, os, shutil
+import os
+import string
+import sys
 
 DRIVER_VERSIONS = {
         'sgx_user.h':                 '/dev/isgx',
@@ -9,8 +11,8 @@ DRIVER_VERSIONS = {
         'sgx_in_kernel.h':            '/dev/sgx/enclave',
 }
 
-def find_intel_sgx_driver():
-    """
+def find_intel_sgx_driver(isgx_driver_path):
+    '''
     Graphene only needs one header from the Intel SGX Driver:
       - sgx_user.h for non-DCAP, older version of the driver
         (https://github.com/intel/linux-sgx-driver)
@@ -22,38 +24,54 @@ def find_intel_sgx_driver():
         (https://lore.kernel.org/linux-sgx/20200716135303.276442-1-jarkko.sakkinen@linux.intel.com)
 
     This function returns the required header from the SGX driver.
-    """
-    isgx_driver_path = os.getenv("ISGX_DRIVER_PATH")
-    if not isgx_driver_path:
-        msg = 'Enter the Intel SGX driver dir with C headers (or press ENTER for in-kernel driver): '
-        isgx_driver_path = os.path.expanduser(input(msg))
-
-    if not isgx_driver_path or isgx_driver_path.strip() == '':
-        # user did not specify any driver path, use default in-kernel driver's C header
-        isgx_driver_path = os.path.dirname(os.path.abspath(__file__))
-
+    '''
     for header_path, dev_path in DRIVER_VERSIONS.items():
         abs_header_path = os.path.abspath(os.path.join(isgx_driver_path, header_path))
         if os.path.exists(abs_header_path):
             return abs_header_path, dev_path
 
-    raise Exception("Could not find the header from the Intel SGX Driver")
+    print('Could not find the header from the Intel SGX Driver (ISGX_DRIVER_PATH={!r})'.format(
+        isgx_driver_path), file=sys.stderr)
+    sys.exit(1)
 
+class MesonTemplate(string.Template):
+    pattern = '''
+        @(?:
+            (?P<escaped>@) |
+            (?P<named>[A-Za-z0-9_]+)@ |
+            (?P<braced>[A-Za-z0-9_]+)@ |
+            (?P<invalid>)
+        )
+    '''
 
 def main():
-    """ Find and copy header/device paths from Intel SGX Driver"""
-    header_path, dev_path = find_intel_sgx_driver()
+    '''
+    Find and copy header/device paths from Intel SGX Driver
+    '''
+    try:
+        isgx_driver_path = os.environ['ISGX_DRIVER_PATH']
+    except KeyError:
+        print(
+            'ISGX_DRIVER_PATH environment variable is undefined. You can define\n'
+            'ISGX_DRIVER_PATH="" to use the default in-kernel driver\'s C header.',
+            file=sys.stderr)
+        sys.exit(1)
 
-    this_header_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'sgx.h')
-    shutil.copyfile(header_path, this_header_path)
+    if not isgx_driver_path:
+        # user did not specify any driver path, use default in-kernel driver's C header
+        isgx_driver_path = os.path.dirname(os.path.abspath(__file__))
 
-    with open(this_header_path, 'a') as f:
-        f.write('\n\n#ifndef ISGX_FILE\n#define ISGX_FILE "%s"\n#endif\n' % dev_path)
-        if dev_path == '/dev/sgx' or dev_path == '/dev/sgx/enclave':
-            f.write('\n\n#ifndef SGX_DCAP\n#define SGX_DCAP 1\n#endif\n')
-        if dev_path == '/dev/sgx/enclave':
-            f.write('\n\n#ifndef SGX_DCAP_16_OR_LATER\n#define SGX_DCAP_16_OR_LATER 1\n#endif\n')
+    header_path, dev_path = find_intel_sgx_driver(isgx_driver_path)
+
+    with sys.stdin:
+        template = MesonTemplate(sys.stdin.read())
+
+    sys.stdout.write(template.safe_substitute(
+        DRIVER_SGX_H=header_path,
+        ISGX_FILE=dev_path,
+        DEFINE_DCAP=('#define SGX_DCAP 1' if dev_path == '/dev/sgx/enclave' else '')
+    ))
 
 
-if __name__ == "__main__":
-    sys.exit(main())
+if __name__ == '__main__':
+    main()
--- a/Pal/src/host/Linux-SGX/link-intel-driver.py
+++ b/Pal/src/host/Linux-SGX/link-intel-driver.py
@@ -7,7 +7,6 @@ import sys
 DRIVER_VERSIONS = {
         'sgx_user.h':                 '/dev/isgx',
         'include/uapi/asm/sgx_oot.h': '/dev/sgx/enclave',
-        'include/uapi/asm/sgx.h':     '/dev/sgx/enclave',
         'sgx_in_kernel.h':            '/dev/sgx/enclave',
 }
 
@@ -18,8 +17,6 @@ def find_intel_sgx_driver(isgx_driver_path):
         (https://github.com/intel/linux-sgx-driver)
       - include/uapi/asm/sgx_oot.h for DCAP 1.6+ version of the driver
         (https://github.com/intel/SGXDataCenterAttestationPrimitives)
-      - include/uapi/asm/sgx.h for in-kernel 20+ version of the driver
-        (https://lore.kernel.org/linux-sgx/20190417103938.7762-1-jarkko.sakkinen@linux.intel.com/)
       - default sgx_in_kernel.h for in-kernel 32+ version of the driver
         (https://lore.kernel.org/linux-sgx/20200716135303.276442-1-jarkko.sakkinen@linux.intel.com)
 
--- a/Pal/src/host/Linux-SGX/sgx_framework.c
+++ b/Pal/src/host/Linux-SGX/sgx_framework.c
@@ -146,7 +146,7 @@ int create_enclave(sgx_arch_secs_t* secs, sgx_arch_token_t* token) {
     uint64_t request_mmap_addr = secs->base;
     uint64_t request_mmap_size = secs->size;
 
-#ifdef SGX_DCAP_16_OR_LATER
+#ifdef SGX_DCAP
     /* newer DCAP/in-kernel SGX drivers allow starting enclave address space with non-zero;
      * the below trick to start from DEFAULT_HEAP_MIN is to avoid vm.mmap_min_addr==0 issue */
     if (request_mmap_addr < DEFAULT_HEAP_MIN) {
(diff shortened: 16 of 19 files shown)


mkow (changes requested):
> Reviewed 13 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: 13 of 15 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

___
*[Documentation/building.rst, line 175 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0pmwIDxUjjqyGbjqi:-MN0pmwIDxUjjqyGbjqj:be2jil3) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Documentation/building.rst#L175)):*
> ```reStructuredText
> 
>    git submodule update --init -- Pal/src/host/Linux-SGX/sgx-driver
>    cd Pal/src/host/Linux-SGX/sgx-driver
> ```

What about this?
___
*[Documentation/cloud-deployment.rst, line 43 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0rESJ1XXiK_-sD0wK:-MN0rESJ1XXiK_-sD0wL:b-wbg1f9) ([raw file](https://github.com/oscarlab/graphene/blob/07f62303c8ef878e08cfb7f64952e45080835098/Documentation/cloud-deployment.rst#L43)):*
> ```reStructuredText
>        git clone https://github.com/oscarlab/graphene.git
>        cd graphene
>        git submodule update --init -- Pal/src/host/Linux-SGX/sgx-driver/
> ```

I think this whole line is not needed now.
___
*[Documentation/quickstart.rst, line 58 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0rHe861LTHyBTAHjS:-MN0rHe861LTHyBTAHjT:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/07f62303c8ef878e08cfb7f64952e45080835098/Documentation/quickstart.rst#L58)):*
> ```reStructuredText
>       git clone https://github.com/oscarlab/graphene.git
>       cd graphene
>       git submodule update --init -- Pal/src/host/Linux-SGX/sgx-driver/
> ```

ditto
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MN0syF00HKRG61CDvwc:bjpu4q8) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*
> ```
>  *                    Dmitrii Kuvaiskii <dmitrii.kuvaiskii@intel.com>
>  */
> ```

Isn't this a 1-1 duplication from the gsgx repo? Seems quite dangerous...
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 11 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0umb10GypXL0bY-3G:-MN0umb10GypXL0bY-3H:b-nafiiq) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/link-intel-driver.py#L11)):*
> ```Python
>         'include/uapi/asm/sgx_oot.h': '/dev/sgx/enclave',
>         'include/uapi/asm/sgx.h':     '/dev/sgx/enclave',
>         'sgx_in_kernel.h':            '/dev/sgx/enclave',
> ```

ditto (question to @dimakuv)
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 21 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0uuc-8ncaNEJVWmnO:-MN0uuc02weiLxE2midZ:bvrjp3u) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/link-intel-driver.py#L21)):*
> ```Python
>       - include/uapi/asm/sgx_oot.h for DCAP 1.6+ version of the driver
>         (https://github.com/intel/SGXDataCenterAttestationPrimitives)
>       - include/uapi/asm/sgx.h for in-kernel 20+ version of the driver
> ```

IMO no reason to support this archaic version
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 71 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0vYRe7N9O5rGdZLMG:-MN0vYRe7N9O5rGdZLMH:bstw2yu) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/link-intel-driver.py#L71)):*
> ```Python
> 
> 
> if __name__ == "__main__":
> ```

Please use single quotes for consistency
___
*[Pal/src/host/Linux-SGX/sgx_in_kernel.h, line 10 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0tGB-6WWSJB9ECiI5:-MN0tGB-6WWSJB9ECiI6:b-2qvj9v) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/sgx_in_kernel.h#L10)):*
> ```objc
>  *       with driver versions 32 and 36, it *may* be out of sync with newer
>  *       versions. If possible, use the header found on the system instead of
>  *       this one. */
> ```

@dimakuv Do we actually need this header? There was some reason I think, but I don't remember it now. I myself just used the headers from /usr/src.

dimakuv (changes requested):
> Reviewed 15 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow and @woju)

___
*[LICENSE.addendum.txt, line 4 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2gAf_0yzol28Fqkz0:-MN2gAf_0yzol28Fqkz1:b-ix71du) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/LICENSE.addendum.txt#L4)):*
> ```vimhelp
> 
> Graphene on an SGX host requires a kernel driver, which is separately licensed under
> the GPL.
> ```

Actually, I recommend to remove the whole para. The "kernel driver" here means the GSGX kernel module -- which only does FSGSBASE enabling (insecurely). This GSGX module is deprecated and is in fact not needed in Linux 5.9+ (and we ship patches to Linux 5.4 to remove the need for this GSGX module).

TLDR: The less we mention GSGX kernel module, the better.
___
*[Documentation/building.rst, line 175 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0pmwIDxUjjqyGbjqi:-MN2hBFb53IVkvOPuU6F:b58lqru) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Documentation/building.rst#L175)):*

What about this?
</blockquote></details>

+1, needs to be fixed to `git clone ...`.
___
*[Documentation/cloud-deployment.rst, line 43 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0rESJ1XXiK_-sD0wK:-MN2hJYM43nh2mktUuo6:b-1inlgr) ([raw file](https://github.com/oscarlab/graphene/blob/07f62303c8ef878e08cfb7f64952e45080835098/Documentation/cloud-deployment.rst#L43)):*

I think this whole line is not needed now.
</blockquote></details>

+1, we are left with only one submodule now (LTP tests), and this is not needed for a casual user.
___
*[Documentation/quickstart.rst, line 58 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0rHe861LTHyBTAHjS:-MN2hSBf0NXlp1RX9__1:b32v57) ([raw file](https://github.com/oscarlab/graphene/blob/07f62303c8ef878e08cfb7f64952e45080835098/Documentation/quickstart.rst#L58)):*

ditto
</blockquote></details>

+1
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MN2ilxb73426JS36zI2:b-7onfpa) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*

Isn't this a 1-1 duplication from the gsgx repo? Seems quite dangerous...
</blockquote></details>

Why dangerous? Woju's idea is to deprecate and remove the graphene-sgx-driver repo altogether (some day).
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 19 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2hep7ABx9BUC5TERm:-MN2hep7ABx9BUC5TERn:b-3gr1pt) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L19)):*
> ```
> 
> #define GSGX_FILE  "/dev/gsgx"
> #define GSGX_MINOR MISC_DYNAMIC_MINOR
> ```

These two lines are not needed, remove them.
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 57 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2iyAl3DD7Eu9JxQLC:-MN2iyAl3DD7Eu9JxQLD:b-6sjct) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L57)):*
> ```
> 
> #ifdef SGX_DCAP
> # define SGX_DCAP_16_OR_LATER 1
> ```

So `SGX_DCAP_16_OR_LATER` simply becomes an alias to `SGX_DCAP`. It would be better to remove all Graphene code under the code path of `ifndef SGX_DCAP_16_OR_LATER` (and replace the macro with `SGX_DCAP` where needed). Could you add this change, @woju ? 
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 11 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0umb10GypXL0bY-3G:-MN2jFUg6N-fcEjQdC48:bcm6gf3) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/link-intel-driver.py#L11)):*

ditto (question to @dimakuv)
</blockquote></details>

What is the question? Do we need this header file shipped with Graphene? Yes, we need it because of usability reasons (people want a one-click build).
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 21 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0uuc-8ncaNEJVWmnO:-MN2jWTwAngaGuQb_KZd:bv6v8h4) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/link-intel-driver.py#L21)):*

IMO no reason to support this archaic version
</blockquote></details>

Yes, agreed. Let's remove this comment and the corresponding entry in `DRIVER_VERSIONS`.
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 55 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2ji3K5MC9zp5z9Utn:-MN2ji3K5MC9zp5z9Uto:b-nllxvl) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/link-intel-driver.py#L55)):*
> ```Python
>         sys.exit(1)
> 
>     if not isgx_driver_path:
> ```

But in the above code you exit with failure. You'll never reach this code. Please move this under `except:`.
___
*[Pal/src/host/Linux-SGX/sgx_in_kernel.h, line 10 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0tGB-6WWSJB9ECiI5:-MN2iPMDCOb7o7bQ7npf:ba0ta6a) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/sgx_in_kernel.h#L10)):*

@dimakuv Do we actually need this header? There was some reason I think, but I don't remember it now. I myself just used the headers from /usr/src.
</blockquote></details>

We need this header, at least for some time more. It was an ask from other people: they don't want to bother typing in the path to the SGX driver; and the interface to the SGX driver became quite stable. So casual users can just use this header and expect it to work in 95% of the cases.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: 8 of 17 files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

___
*[LICENSE.addendum.txt, line 4 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2gAf_0yzol28Fqkz0:-MN2t5-s-LQoVdVhmwLE:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/LICENSE.addendum.txt#L4)):*

Actually, I recommend to remove the whole para. The "kernel driver" here means the GSGX kernel module -- which only does FSGSBASE enabling (insecurely). This GSGX module is deprecated and is in fact not needed in Linux 5.9+ (and we ship patches to Linux 5.4 to remove the need for this GSGX module).

TLDR: The less we mention GSGX kernel module, the better.
</blockquote></details>

Done.
___
*[Documentation/building.rst, line 175 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0pmwIDxUjjqyGbjqi:-MN2vjn30FJ0lHcVHplV:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Documentation/building.rst#L175)):*

+1, needs to be fixed to `git clone ...`.
</blockquote></details>

Done.
___
*[Documentation/cloud-deployment.rst, line 43 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0rESJ1XXiK_-sD0wK:-MN2vtRS71r8tdkkKU3-:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/07f62303c8ef878e08cfb7f64952e45080835098/Documentation/cloud-deployment.rst#L43)):*

+1, we are left with only one submodule now (LTP tests), and this is not needed for a casual user.
</blockquote></details>

Done.
___
*[Documentation/quickstart.rst, line 58 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0rHe861LTHyBTAHjS:-MN2vzW6DvdbQQBYQXU1:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/07f62303c8ef878e08cfb7f64952e45080835098/Documentation/quickstart.rst#L58)):*

+1
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2w4jR3sLQ_5Ja7f-f:-MN2w4jR3sLQ_5Ja7f-g:bef7vzz) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L1)):*
> ```
> /* SPDX-License-Identifier: GPL-2.0-only */
> ```

@dimakuv Would you kindly relicence this to LGPL-3.0-or-later?
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MN2wEUM1bJ3gvjnAnHM:bnahhpz) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*

Why dangerous? Woju's idea is to deprecate and remove the graphene-sgx-driver repo altogether (some day).
</blockquote></details>

About the only dangerous thing is `GSGX_FILE` if that would diverge. GSGX_MINOR is *probably* unneeded, as are `#include <linux/*>` (x2), but I wasn't sure, so I didn't mess with it. There were two files that imported it like this and I didn't read them to check if this is really needed.
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 19 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2hep7ABx9BUC5TERm:-MN2wjyDCFLUhiBqHC3V:b8z76t4) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L19)):*

These two lines are not needed, remove them.
</blockquote></details>

`GSGX_FILE` is certainly needed: https://github.com/oscarlab/graphene/blob/8a381fab5c087cf8388a4c9311c888a5562ad3de/Pal/src/host/Linux-SGX/sgx_framework.c#L19.

I removed `GSGX_MINOR`.
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 57 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2iyAl3DD7Eu9JxQLC:-MN2yJqK38U4S51dNUKo:bo8r1q0) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L57)):*

So `SGX_DCAP_16_OR_LATER` simply becomes an alias to `SGX_DCAP`. It would be better to remove all Graphene code under the code path of `ifndef SGX_DCAP_16_OR_LATER` (and replace the macro with `SGX_DCAP` where needed). Could you add this change, @woju ? 
</blockquote></details>

I'm not sure if there isn't anything that relies on this macro being defined, so I'm doing this as separate commit, should we need to revert this.
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 21 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0uuc-8ncaNEJVWmnO:-MN30gvR0IIDRd0SKooP:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/link-intel-driver.py#L21)):*

Yes, agreed. Let's remove this comment and the corresponding entry in `DRIVER_VERSIONS`.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 55 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2ji3K5MC9zp5z9Utn:-MN30kl52yGfkK60_OJp:b-8l7eki) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/link-intel-driver.py#L55)):*

But in the above code you exit with failure. You'll never reach this code. Please move this under `except:`.
</blockquote></details>

Nope, this is testing for "defined but empty", which is a correct, well-defined state. See for yourself:
```sh
ISGX_DRIVER_PATH= ./link-intel-driver.py < gsgx.h.in > gsgx.h
```
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 71 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0vYRe7N9O5rGdZLMG:-MN31evQ5eu3O98j7N2f:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/link-intel-driver.py#L71)):*

Please use single quotes for consistency
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_in_kernel.h, line 10 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0tGB-6WWSJB9ECiI5:-MN31moVBeFhxSl6e7-m:b-pa43q1) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/sgx_in_kernel.h#L10)):*

We need this header, at least for some time more. It was an ask from other people: they don't want to bother typing in the path to the SGX driver; and the interface to the SGX driver became quite stable. So casual users can just use this header and expect it to work in 95% of the cases.
</blockquote></details>

Let's keep it for now.

dimakuv (changes requested):
> Reviewed 9 of 9 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow and @woju)

___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2w4jR3sLQ_5Ja7f-f:-MN4DzQNCxjxhAV_z807:b-2rcsw0) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L1)):*

@dimakuv Would you kindly relicence this to LGPL-3.0-or-later?
</blockquote></details>

Yes, now that it is not used in the Linux kernel module, we should relicense to LGPL 3.0. Feel free to modify this line.
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 19 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2hep7ABx9BUC5TERm:-MN4EPQY64G7AEdJ860c:bmctigb) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L19)):*

`GSGX_FILE` is certainly needed: https://github.com/oscarlab/graphene/blob/8a381fab5c087cf8388a4c9311c888a5562ad3de/Pal/src/host/Linux-SGX/sgx_framework.c#L19.

I removed `GSGX_MINOR`.
</blockquote></details>

OK, true.
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 55 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2ji3K5MC9zp5z9Utn:-MN4Esbg3a9gHrdl0deC:bx0di2a) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/link-intel-driver.py#L55)):*

Nope, this is testing for "defined but empty", which is a correct, well-defined state. See for yourself:
```sh
ISGX_DRIVER_PATH= ./link-intel-driver.py < gsgx.h.in > gsgx.h
```
</blockquote></details>

Ah, so the user *must* specify `ISGX_DRIVER_PATH`, even an empty one (if empty, then the default header is used). Can we add this to the error message above (`You can define an empty ISGX_DRIVER_PATH="" to use the default in-kernel driver's C header`), something like this. Also, doesn't it mean that our documentation (Quick Start, Building) is not really correct?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: 15 of 19 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 55 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN2ji3K5MC9zp5z9Utn:-MN5H4p874HYTuhiLXF2:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/link-intel-driver.py#L55)):*

Ah, so the user *must* specify `ISGX_DRIVER_PATH`, even an empty one (if empty, then the default header is used). Can we add this to the error message above (`You can define an empty ISGX_DRIVER_PATH="" to use the default in-kernel driver's C header`), something like this. Also, doesn't it mean that our documentation (Quick Start, Building) is not really correct?
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @mkow)

jurobystricky (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @mkow and @woju)

___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 10 at r3](https://reviewable.io/reviews/oscarlab/graphene/1997#-MNFTG0E5YPM7r0NpsyL:-MNFTG0FEvlWM5j24xBD:b-3wmuok) ([raw file](https://github.com/oscarlab/graphene/blob/0a0ba2b50df079bf34556f41ae508e79fdbf2cd5/Pal/src/host/Linux-SGX/link-intel-driver.py#L10)):*
> ```Python
>         'sgx_user.h':                 '/dev/isgx',
>         'include/uapi/asm/sgx_oot.h': '/dev/sgx/enclave',
>         'sgx_in_kernel.h':            '/dev/sgx/enclave',
> ```

The latest kernel device (in-tree) node name is /dev/sgx_enclave.  There will be most likely a symlink to maintain backwards compatibility, however DRIVER_VERSIONS should contain an additional entry "'sgx_in_kernel..h :' /dev/sgx_enclave '""

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @jurobystricky and @mkow)

___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 10 at r3](https://reviewable.io/reviews/oscarlab/graphene/1997#-MNFTG0E5YPM7r0NpsyL:-MNNS9hm-2qbUDAH7r2j:b93nt8x) ([raw file](https://github.com/oscarlab/graphene/blob/0a0ba2b50df079bf34556f41ae508e79fdbf2cd5/Pal/src/host/Linux-SGX/link-intel-driver.py#L10)):*
<details><summary><i>Previously, jurobystricky wrote…</i></summary><blockquote>

The latest kernel device (in-tree) node name is /dev/sgx_enclave.  There will be most likely a symlink to maintain backwards compatibility, however DRIVER_VERSIONS should contain an additional entry "'sgx_in_kernel..h :' /dev/sgx_enclave '""
</blockquote></details>

This won't work. There can be no two entries here with the same key. What we can do is to probe explicitly in this Python code whether `dev/sgx/enclave` exists, and if not, set `dev_path` to `/dev/sgx_enclave`. Or we can just postpone it to another PR.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @jurobystricky and @mkow)

___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 11 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0umb10GypXL0bY-3G:-MNOUX8B8gm0VyZyyWX-:bxrxwlc) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/link-intel-driver.py#L11)):*

What is the question? Do we need this header file shipped with Graphene? Yes, we need it because of usability reasons (people want a one-click build).
</blockquote></details>

@mkow ?
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 10 at r3](https://reviewable.io/reviews/oscarlab/graphene/1997#-MNFTG0E5YPM7r0NpsyL:-MNOU7246tORb7PZJES7:b4ijfbv) ([raw file](https://github.com/oscarlab/graphene/blob/0a0ba2b50df079bf34556f41ae508e79fdbf2cd5/Pal/src/host/Linux-SGX/link-intel-driver.py#L10)):*

This won't work. There can be no two entries here with the same key. What we can do is to probe explicitly in this Python code whether `dev/sgx/enclave` exists, and if not, set `dev_path` to `/dev/sgx_enclave`. Or we can just postpone it to another PR.
</blockquote></details>

Definitely postpone. The proper place for such thing would be a config option to meson.

jurobystricky:
I believe this file should be also updated accordingly:
https://github.com/oscarlab/graphene/blob/woju/get-rid-of-sgx-driver-submodule/Tools/gsc/test/Makefile

woju:
> I believe this file should be also updated accordingly:
> https://github.com/oscarlab/graphene/blob/woju/get-rid-of-sgx-driver-submodule/Tools/gsc/test/Makefile

What's wrong with this file?

jurobystricky:
> > I believe this file should be also updated accordingly:
> > https://github.com/oscarlab/graphene/blob/woju/get-rid-of-sgx-driver-submodule/Tools/gsc/test/Makefile
> 
> What's wrong with this file?

If I am reading this patch correctly, the gsc containers are built with in-tree kernel now. So the line:
<pre>
DEVICES_VOLUMES = --device=/dev/gsgx --device=/dev/${SGX_DEVICE} -v /var/run/aesmd/aesm.socket:/var/run/aesmd/aesm.socket
</pre>

could/should be changed to:
<pre>
DEVICES_VOLUMES = --device=/dev/sgx/enclave 
</pre>

aesmd should not be needed for DCAP/in-tree drivers.
In any case, the proof is in the pudding, you can try to run:
<pre>
$ cd graphene/Tools/gsc/test
$ make
$ make test
</pre>

If the tests don't fail, I stand corrected and you can disregard my comments.

dimakuv (approved):
@vahldiek Could you chime in on the GSC discussion?

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997#-:-MNSsn7VAturucLMazjY:b-w4hhhy)* status: all files reviewed, 9 unresolved discussions (waiting on @jurobystricky and @mkow)

___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 10 at r3](https://reviewable.io/reviews/oscarlab/graphene/1997#-MNFTG0E5YPM7r0NpsyL:-MNSsx2z5yOMLwJ3PTdy:b2nhktp) ([raw file](https://github.com/oscarlab/graphene/blob/0a0ba2b50df079bf34556f41ae508e79fdbf2cd5/Pal/src/host/Linux-SGX/link-intel-driver.py#L10)):*

Definitely postpone. The proper place for such thing would be a config option to meson.
</blockquote></details>

OK.

vahldiek (commented):
GSC is build based on what drivers are specified in the configuration file. For the tests we set the values to Jenkins settings. This has not changed in the PR. What may have tricked you into thinking that it changed is the file in images/ which compiles Graphene for a specific Azure Kubernetes Services (they use DCAP + FSGS patches and hence, need other `DEVICES_VOLUMES` parameters).

So in my opinion, everything should stay as is for `DEVICES_VOLUMES`...

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997#-:-MNSzfGU9Md-OVw1vIws:b-i9umfk)* status: all files reviewed, 9 unresolved discussions (waiting on @jurobystricky and @mkow)

mkow (changes requested):
> Reviewed 1 of 15 files at r1, 7 of 9 files at r2, 4 of 4 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 7 unresolved discussions (waiting on @dimakuv and @woju)

___
*[Documentation/building.rst, line 83 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBHUSdBQbzP-RjzAut:-MPBHUSdBQbzP-RjzAuu:b-ocaikt) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Documentation/building.rst#L83)):*
> ```
> it is available as separate patches.
> ```

Could you point out where they can be found?
___
*[Documentation/building.rst, line 85 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBHa-S1E7g0FrI83a4:-MPBHa-S1E7g0FrI83a5:b10k7ll) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Documentation/building.rst#L85)):*
> ```reStructuredText
> patches.
> 
> Enabling FSGSBASE support requires building and installing a custom kernel with
> ```

The new version of the previous sentence doesn't work well with this paragraph.
___
*[Documentation/building.rst, line 141 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBIHS62mmgbeCB-rBU:-MPBIHS62mmgbeCB-rBV:b-6qkc1m) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Documentation/building.rst#L141)):*
> ```reStructuredText
> .. warning::
> 
>    This module is a |~| quick-and-dirty hack with gaping security hole.
> ```

`gaping` -> `dangerous`
+ please add parenthesis: "(allows unauthorized local privilege escalation)"
___
*[Documentation/building.rst, line 143 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBID_cE4VrLbH2c4Ze:-MPBID_cE4VrLbH2c4Zf:b-a2ugx) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Documentation/building.rst#L143)):*
> ```reStructuredText
>    This module is a |~| quick-and-dirty hack with gaping security hole.
>    "Do not use for production" is not a |~| joke. We use it only for testing
>    on very old kernels when the patchset does not apply cleanly.
> ```

`when` -> `where`
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MPBNYu07J7acSnWCjP7:b-p94g40) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*

About the only dangerous thing is `GSGX_FILE` if that would diverge. GSGX_MINOR is *probably* unneeded, as are `#include <linux/*>` (x2), but I wasn't sure, so I didn't mess with it. There were two files that imported it like this and I didn't read them to check if this is really needed.
</blockquote></details>

Ok, if we are going to remove it soon, then please add a comment that this soon will be removed.
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 52 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBLPvD3djVy2n0iKL6:-MPBLPvD3djVy2n0iKL7:b5274ga) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Pal/src/host/Linux-SGX/link-intel-driver.py#L52)):*
> ```Python
>     except KeyError:
>         print('ISGX_DRIVER_PATH environment variable is undefined. You can define an\n'
>             'empty ISGX_DRIVER_PATH="" to use the default in-kernel driver\'s C header.',
> ```

`an empty ISGX_DRIVER_PATH=""` - either `an empty` or `ISGX_DRIVER_PATH=""`
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 52 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBLfZB0B2OZpT374HY:-MPBLfZB0B2OZpT374HZ:bt5y3pf) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Pal/src/host/Linux-SGX/link-intel-driver.py#L52)):*
> ```Python
>     except KeyError:
>         print('ISGX_DRIVER_PATH environment variable is undefined. You can define an\n'
>             'empty ISGX_DRIVER_PATH="" to use the default in-kernel driver\'s C header.',
> ```

this line should be wrapped to the parenthesis (2 more spaces)

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: 17 of 19 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Documentation/building.rst, line 83 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBHUSdBQbzP-RjzAut:-MPBePlfA-kQMe8Ex-_9:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Documentation/building.rst#L83)):*

> ```
> it is available as separate patches.
> ```

Could you point out where they can be found?
</blockquote></details>

Done.
___
*[Documentation/building.rst, line 85 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBHa-S1E7g0FrI83a4:-MPBeQL98aL4T_9bbxS7:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Documentation/building.rst#L85)):*

The new version of the previous sentence doesn't work well with this paragraph.
</blockquote></details>

Done.
___
*[Documentation/building.rst, line 141 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBIHS62mmgbeCB-rBU:-MPBexQ77fpIkJ6wBWym:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Documentation/building.rst#L141)):*

`gaping` -> `dangerous`
+ please add parenthesis: "(allows unauthorized local privilege escalation)"
</blockquote></details>

Done.
___
*[Documentation/building.rst, line 143 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBID_cE4VrLbH2c4Ze:-MPBezeyAzn90PNMCfyj:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Documentation/building.rst#L143)):*

`when` -> `where`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MPBf0oAA7EvyC6nkmOa:bnby0qd) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*

Ok, if we are going to remove it soon, then please add a comment that this soon will be removed.
</blockquote></details>

We're going to keep this file (in some form or another) as long as we support different SGX drivers and it will be actively maintained. For example, as part of going full meson, I'll have `#define GSGX_FILE "@SGX_DRIVER_DEV@"` (or whatever) substituted here to adjust between different drivers' ideas about `/dev`.

If you're talking about removing `sgx.h` from oscarlab/graphene-sgx-driver, I don't think a comment is needed here. I just removed all references to it everywhere and substituted for `#include`ing this file, so there would be no actionable info.
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 52 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBLfZB0B2OZpT374HY:-MPBXr367iPW3M5dkAW9:bbqfm5a) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Pal/src/host/Linux-SGX/link-intel-driver.py#L52)):*

this line should be wrapped to the parenthesis (2 more spaces)
</blockquote></details>

Nope, wrapping to next multiple of 4 spaces is also OK according to PEP-8, and consistent with `print()` in `find_intel_sgx_driver()` above. I broke also the first line, so this aligns nicely.

Also, this file is going to be remove as part of meson change.
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 52 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBLPvD3djVy2n0iKL6:-MPBi44WDQ2_P4PSNRMG:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Pal/src/host/Linux-SGX/link-intel-driver.py#L52)):*

`an empty ISGX_DRIVER_PATH=""` - either `an empty` or `ISGX_DRIVER_PATH=""`
</blockquote></details>

Done.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[Documentation/building.rst, line 141 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBIHS62mmgbeCB-rBU:-MPBoV5cBf1pLn7nayrh:b-wh5agh) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Documentation/building.rst#L141)):*

Done.
</blockquote></details>

`-ised` -> `-ized`
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MPBok0k6cmqLmKapeUi:b-gad80f) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*

We're going to keep this file (in some form or another) as long as we support different SGX drivers and it will be actively maintained. For example, as part of going full meson, I'll have `#define GSGX_FILE "@SGX_DRIVER_DEV@"` (or whatever) substituted here to adjust between different drivers' ideas about `/dev`.

If you're talking about removing `sgx.h` from oscarlab/graphene-sgx-driver, I don't think a comment is needed here. I just removed all references to it everywhere and substituted for `#include`ing this file, so there would be no actionable info.
</blockquote></details>

I don't think we'll support the older version of the patchset once it gets upstreamed.
Even more, I'd actively force people to not use outdated versions - those are kind of "preview/draft versions" which never passed the review - only the final version which get accepted + its future versions should be ever used, because they are what finally passed the review. People are using the patchset right now just because other solutions are also bad, but once it gets upstreamed no one should ever use the old revisions.
___
*[Pal/src/host/Linux-SGX/link-intel-driver.py, line 52 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBLfZB0B2OZpT374HY:-MPBocPv5wKhrcUDJeOc:b-dk05wk) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Pal/src/host/Linux-SGX/link-intel-driver.py#L52)):*

Nope, wrapping to next multiple of 4 spaces is also OK according to PEP-8, and consistent with `print()` in `find_intel_sgx_driver()` above. I broke also the first line, so this aligns nicely.

Also, this file is going to be remove as part of meson change.
</blockquote></details>

Ok, now it looks better.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: 18 of 19 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[Documentation/building.rst, line 141 at r4](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPBIHS62mmgbeCB-rBU:-MPBtn3C-TMXH5G5zReh:b-rvgulv) ([raw file](https://github.com/oscarlab/graphene/blob/6ed3691e2f1fb078ce22d0ef519a05b702eb057b/Documentation/building.rst#L141)):*

`-ised` -> `-ized`
</blockquote></details>

dammit
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MPBqRG9AQ8FADUGEJm6:b-q97yky) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*

I don't think we'll support the older version of the patchset once it gets upstreamed.
Even more, I'd actively force people to not use outdated versions - those are kind of "preview/draft versions" which never passed the review - only the final version which get accepted + its future versions should be ever used, because they are what finally passed the review. People are using the patchset right now just because other solutions are also bad, but once it gets upstreamed no one should ever use the old revisions.
</blockquote></details>

Nope. We're talking about SGX drivers here, not FSGSBASE patchset. There are multiple reasons why people would want to run OOT driver.

First, upstream driver won't support EPID (ever, for political reasons: to avoid potential bait-and-switch/EEE by Intel). Second, the last kernel without upstream SGX support (5.10) is "longterm" release (https://twitter.com/gregkh/status/1320745076566433793), supported for at least 2 years (Greg says it might be closer to 6 years), so it may be picked by distros for LTS purposes. I suspect Debian 11 (bullseye) will have it for example. Facing choice of either compiling just the module (which people have been running for years), or recompiling whole kernel to relatively unstable 5.11.0, I'd install the OOT driver.

I agree DCAP driver might well go away, but that doesn't solve anything, because OOT and inkernel drivers have different dev paths.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @woju)

___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MPBvTGHE9EHYBJ2nwBA:b-u5gy2m) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*
> We're talking about SGX drivers here, not FSGSBASE patchset. There are multiple reasons why people would want to run OOT driver.

I know it's not about FSGSBASE. Seems I've mistaken this file with the other one and was thinking that this one is only for supporting the older revisions of the patchset (see my previous comment in this light now ;) ), sorry.
I re-read it again now and I see it's only for the OOT, so we should keep it for some time - but then, my older comment (the first in this thread) is still valid - if we are going to keep this file for quite some time, then we have a duplication of multiple constants which is not validated during build, so I really don't like it (version mismatches will just cause weird things in runtime). Any ideas how do to this better?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MPC-PZe6JpbfWJ5iuIY:b-uhnkm9) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*

> We're talking about SGX drivers here, not FSGSBASE patchset. There are multiple reasons why people would want to run OOT driver.

I know it's not about FSGSBASE. Seems I've mistaken this file with the other one and was thinking that this one is only for supporting the older revisions of the patchset (see my previous comment in this light now ;) ), sorry.
I re-read it again now and I see it's only for the OOT, so we should keep it for some time - but then, my older comment (the first in this thread) is still valid - if we are going to keep this file for quite some time, then we have a duplication of multiple constants which is not validated during build, so I really don't like it (version mismatches will just cause weird things in runtime). Any ideas how do to this better?
</blockquote></details>

I don't know, I have no timeline. This PR does not add any duplication that wasn't in oscarlab/graphene-sgx-driver. I'd have to go check all available `sgx.h` on the Internet to see when those #ifdefed constants are really needed.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MPC0qev1DfJkUfh42wk:b2jc1b2) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*

I don't know, I have no timeline. This PR does not add any duplication that wasn't in oscarlab/graphene-sgx-driver. I'd have to go check all available `sgx.h` on the Internet to see when those #ifdefed constants are really needed.
</blockquote></details>

Uh, wait, why do we have some of the OOT driver constants defined by ourselves if GSGX includes OOT driver headers already?
Previously I thought you just copied the defines from GSGX driver interface itself, so that you can always build Graphene without the submodule. But now I see the contants are _only_ for ISGX and they are also duplicated in the GSGX repo... Let's leave this for later then, but maybe @dimakuv remembers why are they duplicated?
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 53 at r6](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPC2E6R5fluCJzoQ_aN:-MPC2E6R5fluCJzoQ_aO:b-1ruevw) ([raw file](https://github.com/oscarlab/graphene/blob/b646e0f5f4342e6eb9336768a9a1c60415adddaf/Pal/src/host/Linux-SGX/gsgx.h.in#L53)):*
> ```
> #define ISGX_FILE "@ISGX_FILE@"
> 
> @DEFINE_DCAP@
> ```

Ok, last thing I just noticed: Why using this whole templating instead of just passing defines to the compiler's command line, like it's usually done for passing build parameters to C?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MPC3u5mCN6EPkmOFZJm:b-6jy7lw) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*

Uh, wait, why do we have some of the OOT driver constants defined by ourselves if GSGX includes OOT driver headers already?
Previously I thought you just copied the defines from GSGX driver interface itself, so that you can always build Graphene without the submodule. But now I see the contants are _only_ for ISGX and they are also duplicated in the GSGX repo... Let's leave this for later then, but maybe @dimakuv remembers why are they duplicated?
</blockquote></details>

I suspect this was needed for older version. I don't think there's much to it.
___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 53 at r6](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPC2E6R5fluCJzoQ_aN:-MPC47PL4efmJRuVeAtJ:bvq0kyg) ([raw file](https://github.com/oscarlab/graphene/blob/b646e0f5f4342e6eb9336768a9a1c60415adddaf/Pal/src/host/Linux-SGX/gsgx.h.in#L53)):*

Ok, last thing I just noticed: Why using this whole templating instead of just passing defines to the compiler's command line, like it's usually done for passing build parameters to C?
</blockquote></details>

Because our makefiles stink and I don't want to touch them even with a stick.

In this particular case, you'd have to add it everywhere, including this place: https://github.com/oscarlab/graphene/blob/da14ac0297e4747afbc0ce46363fdcab96624b91/python/graphenelibos/sgx_get_token.py#L84-L86

This is out of scope by a great margin.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 53 at r6](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPC2E6R5fluCJzoQ_aN:-MPC5vZiAYYmmPe4-2xq:b-sju43u) ([raw file](https://github.com/oscarlab/graphene/blob/b646e0f5f4342e6eb9336768a9a1c60415adddaf/Pal/src/host/Linux-SGX/gsgx.h.in#L53)):*

Because our makefiles stink and I don't want to touch them even with a stick.

In this particular case, you'd have to add it everywhere, including this place: https://github.com/oscarlab/graphene/blob/da14ac0297e4747afbc0ce46363fdcab96624b91/python/graphenelibos/sgx_get_token.py#L84-L86

This is out of scope by a great margin.
</blockquote></details>

But aren't you partially getting rid of Makefiles here? Is the problem that in the current construction you can't change the parameters of the Make build from Meson?
If so, can you add a TODO here that once we stop building with Make and fully switch to Meson, then we should get rid of this template and just define the macro in compiler's cmdline invocation in Meson?

Ad. changing sgx_get_token.py: I don't think so, AFAIR the attribute it's checking is generated conditionally on the C macro, but I may be wrong.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 53 at r6](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPC2E6R5fluCJzoQ_aN:-MPC9IJH-h1WhjRPh3wQ:blypec4) ([raw file](https://github.com/oscarlab/graphene/blob/b646e0f5f4342e6eb9336768a9a1c60415adddaf/Pal/src/host/Linux-SGX/gsgx.h.in#L53)):*
> attribute it's checking is generated conditionally on the C macro

Yes, this is from `generated_offsets`. But you get the picture.

> in the current construction you can't change the parameters of the Make build from Meson?

Yes, that's the case, because we call them sequentially: first make, then meson, then ninja. That's intended and agreed as part of transition plan. As long as part of the build is done by make, you have to run all of them in this particular order and pass correct arguments to their respective receivers. I don't see a problem here. Also there is no point of adding TODO, because it cannot be done in full meson without changing this, so it's not possible to forget about this.

I don't want to spend any more time on this. All of this will be removed shortly. If you don't want to merge as-is, I'll abandon this MR and will just put this on the bottom of the next MR in the meson queue. Eventually it'll be reduced to empty diff, because, as I wrote, this will be removed. Now is a chance to review this piece by piece, not as 20kloc diff dump.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 53 at r6](https://reviewable.io/reviews/oscarlab/graphene/1997#-MPC2E6R5fluCJzoQ_aN:-MPCCqUL5rXJ-B17Mfh3:b-1ak5i2) ([raw file](https://github.com/oscarlab/graphene/blob/b646e0f5f4342e6eb9336768a9a1c60415adddaf/Pal/src/host/Linux-SGX/gsgx.h.in#L53)):*
> Also there is no point of adding TODO, because it cannot be done in full meson without changing this, so it's not possible to forget about this.

Ok, I hope you're right here :)

> I don't want to spend any more time on this. All of this will be removed shortly.

I just want to ensure that something can't be accidentally left in such "hacked around" state.

dimakuv (dismissed):
> Reviewed 1 of 2 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MPDvKTg3Ne2Ao0Mu0xY:br63to5) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*

I suspect this was needed for older version. I don't think there's much to it.
</blockquote></details>

@mkow These constants (`SGX_INVALID_*`) are redefined here because of legacy/differing SGX driver versions from https://github.com/intel/linux-sgx-driver/tags. The main (real) reason was that our Jenkins used super-old version 1.8 until very recently, and that version didn't have `SGX_INVALID_EINITTOKEN` (see comment in this file). So to make it uniform across different OOT driver versions, we added all these constants here.

I shudder at the memory of OOT SGX drivers... This all can be purged now, since we moved to recent-ish SGX drivers with stable constants. Not in this PR of course. But some future PR should remove this `gsgx.h` alltogether.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[Pal/src/host/Linux-SGX/gsgx.h.in, line 5 at r1](https://reviewable.io/reviews/oscarlab/graphene/1997#-MN0syF00HKRG61CDvwb:-MPEKCEa5zL3hQrd9mgu:b4c9lyr) ([raw file](https://github.com/oscarlab/graphene/blob/54790dc3a1efa7268ece05e0ec8a3e4db38e75ed/Pal/src/host/Linux-SGX/gsgx.h.in#L5)):*

@mkow These constants (`SGX_INVALID_*`) are redefined here because of legacy/differing SGX driver versions from https://github.com/intel/linux-sgx-driver/tags. The main (real) reason was that our Jenkins used super-old version 1.8 until very recently, and that version didn't have `SGX_INVALID_EINITTOKEN` (see comment in this file). So to make it uniform across different OOT driver versions, we added all these constants here.

I shudder at the memory of OOT SGX drivers... This all can be purged now, since we moved to recent-ish SGX drivers with stable constants. Not in this PR of course. But some future PR should remove this `gsgx.h` alltogether.
</blockquote></details>

:thumbsup:

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1997)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2001, opened 2020-11-27
[Pal] Add SGX profiling support

pwmarcz:
This adds a "do it yourself" profiler that instruments AEX handler. I'm not sure how accurate it is, but the overhead seems to be pretty low (I haven't measured it though).

## Background

Background: I looked into `perf` first, and while it's pretty good for profiling non-SGX code, for SGX it only shows time spent in `async_exit_pointer` on interrupts. Unfortunately it seems pretty hard to extend, as perf events are generated in the kernel (and even if we did that, there's still the matter of resolving addresses to symbols).

First, I wrote some code on `async_exit_pointer` that reads the IP from enclave memory. This can already be dynamically instrumented by perf (using `perf probe`), but the resulting report is a list of events with `ip=0xabcd` and doesn't seem very usable. So I decided to go all the way and implement both event collection and symbol resolution in Graphene.

## Description of the changes 

* `SGX_PROFILE` compilation flag, and `sgx.profile` manifest option
* Code in `async_exit_pointer` to measure IP and collect it in a hash table
* `profile-report` Python tool for generating a report

## How to test this PR? 

Try to follow the instructions in `performance.rst`, and run some non-trivial workload.

## Possible improvements

* Make `perf` do more work somehow? Maybe we could generate synthetic `perf` events from userspace, not sure if it's ever done
* The Python program is pretty bad at symbol resolution: unlike perf and gdb, it doesn't understand that there can be external debug info. What can I use to do it better?
* Recover call graph (or at least a single caller)? I guess we could follow saved `RBP` and get something useful most of the time...

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)

### DIFF
--- a/.ci/ubuntu18.04.dockerfile
+++ b/.ci/ubuntu18.04.dockerfile
@@ -46,6 +46,7 @@ RUN apt-get update && env DEBIAN_FRONTEND=noninteractive apt-get install -y \
     python3-numpy \
     python3-pip \
     python3-protobuf \
+    python3-pyelftools \
     python3-pytest \
     python3-scipy \
     r-base-core \
--- a/Documentation/devel/performance.rst
+++ b/Documentation/devel/performance.rst
@@ -513,7 +513,7 @@ SGX profiling
 There is some experimental support for profiling the code inside the SGX
 enclave. Here is how to use it:
 
-#. Compile Graphene with ``SGX=1 SGX_PROFILE=1 DEBUG=1``.
+#. Compile Graphene with ``SGX=1 DEBUG=1``.
 
 #. Make sure to install Python package ``pyelftools`` (e.g. ``sudo apt-get
    install python-pyelftools``).
--- a/Documentation/devel/performance.rst
+++ b/Documentation/devel/performance.rst
@@ -489,7 +489,7 @@ Some useful options for recording (``perf record``):
 * ``-e cpu-clock``: sample the ``cpu-clock`` event, which will be triggered also
   inside enclave (as opposed to the default ``cpu-cycles`` event). Unfortunately
   such events will be counted towards ``async_exit_pointer`` instead of
-  functions executing inside enclave.
+  functions executing inside enclave (but see also :ref:`sgx-profile`).
 
 Some useful options for displaying the report (``perf report``):
 
@@ -504,6 +504,37 @@ Further reading
 * `Linux perf examples - Brendan Gregg
   <http://www.brendangregg.com/perf.html>`__
 * Man pages: ``man perf record``, ``man perf report`` etc.
+
+.. _sgx-profile:
+
+SGX profiling
+-------------
+
+There is some experimental support for profiling the code inside the SGX
+enclave. Here is how to use it:
+
+#. Compile Graphene with ``SGX=1 SGX_PROFILE=1 DEBUG=1``.
+
+#. Make sure to install Python package ``pyelftools`` (e.g. ``sudo apt-get
+   install python-pyelftools``).
+
+#. Add ``sgx.profile = "root"`` to manifest (to collect data for the main
+   process), or ``sgx.profile = "all"`` (to collect data for all processes).
+
+#. Run your application. It should say something like ``writing profile data to
+   sgx-profile.data`` on process exit (in case of ``sgx.profile = "all"``,
+   multiple files will be written).
+
+#. Run ``profile-report``
+   (``Pal/src/host/Linux-SGX/tools/profile-report/profile-report``) with the
+   data file as parameter. It should display a table of functions (or raw
+   addresses, if symbol information is not available)
+
+*Note*: The accuracy of this tool is unclear. The SGX profiling works by
+measuring the value of instruction pointer on each asynchronous enclave exit
+(AEX), which happen on Linux scheduler interrupts as well as other such as page
+faults. While we attempt to measure time (and not only count occurences), the
+results might be inaccurate.
 
 Other useful tools for profiling
 --------------------------------
--- a/Documentation/devel/performance.rst
+++ b/Documentation/devel/performance.rst
@@ -518,7 +518,7 @@ enclave. Here is how to use it:
 #. Make sure to install Python package ``pyelftools`` (e.g. ``sudo apt-get
    install python-pyelftools``).
 
-#. Add ``sgx.profile = "root"`` to manifest (to collect data for the main
+#. Add ``sgx.profile = "main"`` to manifest (to collect data for the main
    process), or ``sgx.profile = "all"`` (to collect data for all processes).
 
 #. Run your application. It should say something like ``writing profile data to
@@ -528,13 +528,13 @@ enclave. Here is how to use it:
 #. Run ``profile-report``
    (``Pal/src/host/Linux-SGX/tools/profile-report/profile-report``) with the
    data file as parameter. It should display a table of functions (or raw
-   addresses, if symbol information is not available)
+   addresses, if symbol information is not available).
 
 *Note*: The accuracy of this tool is unclear. The SGX profiling works by
 measuring the value of instruction pointer on each asynchronous enclave exit
-(AEX), which happen on Linux scheduler interrupts as well as other such as page
-faults. While we attempt to measure time (and not only count occurences), the
-results might be inaccurate.
+(AEX), which happen on Linux scheduler interrupts, as well as other events such
+as page faults. While we attempt to measure time (and not only count
+occurences), the results might be inaccurate.
 
 Other useful tools for profiling
 --------------------------------
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -458,3 +458,28 @@ This syntax specifies whether to enable SGX enclave-specific statistics:
 *Note:* this option is insecure and cannot be used with production enclaves
 (``sgx.debug = 0``). If the production enclave is started with this option set,
 Graphene will fail initialization of the enclave.
+
+SGX profiling
+^^^^^^^^^^^^^
+
+::
+
+    sgx.profile = ["none"|"root"|"all"]
+    (Default: "none")
+
+This syntax specifies whether to enable SGX profiling. Graphene must be compiled
+with ``DEBUG=1`` and ``SGX_PROFILE=1`` for this option to work.
+
+If this option is set to ``root``, the root process will collect IP samples and
+save them as ``profile-sgx.data``. If it's to ``all``, all processes will
+collect samples and save them to ``profile-sgx-<PID>.data``.
+
+The saved files can be used to generate a report with ``profile-report`` tool,
+located in ``Pal/src/host/Linux-SGX/tools/profile-report/`` directory. It
+requires ``pyelftools`` Python package to be installed.
+
+See :doc:`devel/performance` for more information.
+
+*Note:* this option is insecure and cannot be used with production enclaves
+(``sgx.debug = 0``). If the production enclave is started with this option set,
+Graphene will fail initialization of the enclave.
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -468,7 +468,7 @@ SGX profiling
     (Default: "none")
 
 This syntax specifies whether to enable SGX profiling. Graphene must be compiled
-with ``DEBUG=1`` and ``SGX_PROFILE=1`` for this option to work.
+with ``DEBUG=1`` for this option to work.
 
 If this option is set to ``main``, the main process will collect IP samples and
 save them as ``sgx-profile.data``. If it is set to ``all``, all processes will
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -464,15 +464,15 @@ SGX profiling
 
 ::
 
-    sgx.profile = ["none"|"root"|"all"]
+    sgx.profile = ["none"|"main"|"all"]
     (Default: "none")
 
 This syntax specifies whether to enable SGX profiling. Graphene must be compiled
 with ``DEBUG=1`` and ``SGX_PROFILE=1`` for this option to work.
 
-If this option is set to ``root``, the root process will collect IP samples and
-save them as ``profile-sgx.data``. If it's to ``all``, all processes will
-collect samples and save them to ``profile-sgx-<PID>.data``.
+If this option is set to ``main``, the main process will collect IP samples and
+save them as ``sgx-profile.data``. If it is set to ``all``, all processes will
+collect samples and save them to ``sgx-profile-<PID>.data``.
 
 The saved files can be used to generate a report with ``profile-report`` tool,
 located in ``Pal/src/host/Linux-SGX/tools/profile-report/`` directory. It
--- a/Pal/include/pal/pal_debug.h
+++ b/Pal/include/pal/pal_debug.h
@@ -11,6 +11,7 @@
 #include "pal.h"
 
 int pal_printf(const char* fmt, ...) __attribute__((format(printf, 1, 2)));
+int pal_fdprintf(int fd, const char* fmt, ...) __attribute__((format(printf, 2, 3)));
 void warn(const char* format, ...);
 
 void DkDebugAttachBinary(PAL_STR uri, PAL_PTR start_addr);
--- a/Pal/src/host/Linux-SGX/Makefile
+++ b/Pal/src/host/Linux-SGX/Makefile
@@ -70,6 +70,7 @@ enclave-asm-objs = enclave_entry.o
 
 urts-objs = \
 	clone-x86_64.o \
+	sgx_profile.o \
 	sgx_enclave.o \
 	sgx_exception.o \
 	sgx_framework.o \
--- a/Pal/src/host/Linux-SGX/Makefile
+++ b/Pal/src/host/Linux-SGX/Makefile
@@ -70,7 +70,6 @@ enclave-asm-objs = enclave_entry.o
 
 urts-objs = \
 	clone-x86_64.o \
-	sgx_profile.o \
 	sgx_enclave.o \
 	sgx_exception.o \
 	sgx_framework.o \
@@ -78,6 +77,7 @@ urts-objs = \
 	sgx_main.o \
 	sgx_platform.o \
 	sgx_process.o \
+	sgx_profile.o \
 	sgx_gdb_info.o \
 	sgx_thread.o \
 	quote/aesm.pb-c.o \
--- a/Pal/src/host/Linux-SGX/enclave_framework.c
+++ b/Pal/src/host/Linux-SGX/enclave_framework.c
@@ -38,14 +38,14 @@ bool sgx_is_completely_outside_enclave(const void* addr, size_t size) {
 }
 
 /*
- * When SGX_PROFILE is enabled, we run some code on asynchronous enclave exit (AEX) that uses the
- * stack. Make sure to update URSP so that the AEX handler does not overwrite the part of the stack
- * that we just allocated.
+ * When DEBUG is enabled, we run sgx_profile_sample() during asynchronous enclave exit (AEX), which
+ * uses the stack. Make sure to update URSP so that the AEX handler does not overwrite the part of
+ * the stack that we just allocated.
  *
  * (Recall that URSP is an outside stack pointer, saved by EENTER and restored on AEX by the SGX
  * hardware itself.)
  */
-#if SGX_PROFILE
+#ifdef DEBUG
 
 #define UPDATE_USTACK(_ustack)                           \
     do {                                                 \
--- a/Pal/src/host/Linux-SGX/enclave_framework.c
+++ b/Pal/src/host/Linux-SGX/enclave_framework.c
@@ -41,6 +41,9 @@ bool sgx_is_completely_outside_enclave(const void* addr, size_t size) {
  * When SGX_PROFILE is enabled, we run some code on asynchronous enclave exit (AEX) that uses the
  * stack. Make sure to update URSP so that the AEX handler does not overwrite the part of the stack
  * that we just allocated.
+ *
+ * (Recall that URSP is an outside stack pointer, saved by EENTER and restored on AEX by the SGX
+ * hardware itself.)
  */
 #if SGX_PROFILE
 
--- a/Pal/src/host/Linux-SGX/enclave_framework.c
+++ b/Pal/src/host/Linux-SGX/enclave_framework.c
@@ -37,13 +37,32 @@ bool sgx_is_completely_outside_enclave(const void* addr, size_t size) {
     return g_enclave_base >= addr + size || g_enclave_top <= addr;
 }
 
+/*
+ * When SGX_PROFILE is enabled, we run some code on asynchronous enclave exit (AEX) that uses the
+ * stack. Make sure to update URSP so that the AEX handler does not overwrite the part of the stack
+ * that we just allocated.
+ */
+#if SGX_PROFILE
+
+#define UPDATE_USTACK(_ustack)                           \
+    do {                                                 \
+        SET_ENCLAVE_TLS(ustack, _ustack);                \
+        GET_ENCLAVE_TLS(gpr)->ursp = (uint64_t)_ustack;  \
+    } while(0)
+
+#else
+
+#define UPDATE_USTACK(_ustack) SET_ENCLAVE_TLS(ustack, _ustack)
+
+#endif
+
 void* sgx_prepare_ustack(void) {
     void* old_ustack = GET_ENCLAVE_TLS(ustack);
 
     void* ustack = old_ustack;
     if (ustack != GET_ENCLAVE_TLS(ustack_top))
         ustack -= RED_ZONE_SIZE;
-    SET_ENCLAVE_TLS(ustack, ustack);
+    UPDATE_USTACK(ustack);
 
     return old_ustack;
 }
@@ -55,7 +74,7 @@ void* sgx_alloc_on_ustack_aligned(size_t size, size_t alignment) {
     if (!sgx_is_completely_outside_enclave(ustack, size)) {
         return NULL;
     }
-    SET_ENCLAVE_TLS(ustack, ustack);
+    UPDATE_USTACK(ustack);
     return ustack;
 }
 
@@ -76,7 +95,7 @@ void* sgx_copy_to_ustack(const void* ptr, size_t size) {
 
 void sgx_reset_ustack(const void* old_ustack) {
     assert(old_ustack <= GET_ENCLAVE_TLS(ustack_top));
-    SET_ENCLAVE_TLS(ustack, old_ustack);
+    UPDATE_USTACK(old_ustack);
 }
 
 bool sgx_copy_ptr_to_enclave(void** ptr, void* uptr, size_t size) {
--- a/Pal/src/host/Linux-SGX/sgx_enclave.c
+++ b/Pal/src/host/Linux-SGX/sgx_enclave.c
@@ -64,6 +64,9 @@ static long sgx_ocall_exit(void* pms) {
     /* exit the whole process if exit_group() */
     if (ms->ms_is_exitgroup) {
         update_and_print_stats(/*process_wide=*/true);
+#ifdef SGX_PROFILE
+        sgx_profile_finish();
+#endif
         INLINE_SYSCALL(exit_group, 1, (int)ms->ms_exitcode);
     }
 
@@ -76,6 +79,9 @@ static long sgx_ocall_exit(void* pms) {
     if (!current_enclave_thread_cnt()) {
         /* no enclave threads left, kill the whole process */
         update_and_print_stats(/*process_wide=*/true);
+#ifdef SGX_PROFILE
+        sgx_profile_finish();
+#endif
         INLINE_SYSCALL(exit_group, 1, (int)ms->ms_exitcode);
     }
 
--- a/Pal/src/host/Linux-SGX/sgx_enclave.c
+++ b/Pal/src/host/Linux-SGX/sgx_enclave.c
@@ -64,7 +64,7 @@ static long sgx_ocall_exit(void* pms) {
     /* exit the whole process if exit_group() */
     if (ms->ms_is_exitgroup) {
         update_and_print_stats(/*process_wide=*/true);
-#ifdef SGX_PROFILE
+#ifdef DEBUG
         sgx_profile_finish();
 #endif
         INLINE_SYSCALL(exit_group, 1, (int)ms->ms_exitcode);
@@ -79,7 +79,7 @@ static long sgx_ocall_exit(void* pms) {
     if (!current_enclave_thread_cnt()) {
         /* no enclave threads left, kill the whole process */
         update_and_print_stats(/*process_wide=*/true);
-#ifdef SGX_PROFILE
+#ifdef DEBUG
         sgx_profile_finish();
 #endif
         INLINE_SYSCALL(exit_group, 1, (int)ms->ms_exitcode);
--- a/Pal/src/host/Linux-SGX/sgx_entry.S
+++ b/Pal/src/host/Linux-SGX/sgx_entry.S
@@ -69,7 +69,7 @@ async_exit_pointer:
 	# increment per-thread AEX counter for stats
 	lock incq %gs:PAL_TCB_URTS_AEX_CNT
 
-#ifdef SGX_PROFILE
+#ifdef DEBUG
 	# Save ERESUME parameters
 	pushq %rax
 	.cfi_adjust_cfa_offset 8
--- a/Pal/src/host/Linux-SGX/sgx_entry.S
+++ b/Pal/src/host/Linux-SGX/sgx_entry.S
@@ -78,7 +78,7 @@ async_exit_pointer:
 	pushq %rcx
 	.cfi_adjust_cfa_offset 8
 
-	# Align stack (eequired by System V AMD64 ABI)
+	# Align stack (required by System V AMD64 ABI)
 	movq %rsp, %rbp
 	andq $~0xF, %rsp
 
--- a/Pal/src/host/Linux-SGX/sgx_entry.S
+++ b/Pal/src/host/Linux-SGX/sgx_entry.S
@@ -64,8 +64,42 @@ sgx_ecall:
 	.type async_exit_pointer, @function
 
 async_exit_pointer:
+	.cfi_startproc
+
 	# increment per-thread AEX counter for stats
 	lock incq %gs:PAL_TCB_URTS_AEX_CNT
+
+#ifdef SGX_PROFILE
+	# Save ERESUME parameters
+	pushq %rax
+	.cfi_adjust_cfa_offset 8
+	pushq %rbx
+	.cfi_adjust_cfa_offset 8
+	pushq %rcx
+	.cfi_adjust_cfa_offset 8
+
+	# Align stack (eequired by System V AMD64 ABI)
+	movq %rsp, %rbp
+	andq $~0xF, %rsp
+
+	# Call sgx_profile_sample with %rdi = TCS
+	movq %rbx, %rdi
+	call sgx_profile_sample
+
+	# Restore stack
+	movq %rbp, %rsp
+
+	# Restore ERESUME parameters
+	popq %rcx
+	.cfi_adjust_cfa_offset -8
+	popq %rbx
+	.cfi_adjust_cfa_offset -8
+	popq %rax
+	.cfi_adjust_cfa_offset -8
+#endif
+
+	.cfi_endproc
+
 	# fall-through to ERESUME
 
 	.global eresume_pointer
--- a/Pal/src/host/Linux-SGX/sgx_graphene.c
+++ b/Pal/src/host/Linux-SGX/sgx_graphene.c
@@ -29,15 +29,26 @@ static int fputch(void* f, int ch, struct printbuf* b) {
     return 0;
 }
 
-static int vprintf(const char* fmt, va_list ap) {
+static int vprintf(int fd, const char* fmt, va_list ap) {
     struct printbuf b;
 
     b.idx = 0;
     b.cnt = 0;
     vfprintfmt((void*)&fputch, NULL, &b, fmt, ap);
-    INLINE_SYSCALL(write, 3, 2, b.buf, b.idx);
+    INLINE_SYSCALL(write, 3, fd, b.buf, b.idx);
 
     return b.cnt;
+}
+
+int pal_fdprintf(int fd, const char* fmt, ...) {
+    va_list ap;
+    int cnt;
+
+    va_start(ap, fmt);
+    cnt = vprintf(fd, fmt, ap);
+    va_end(ap);
+
+    return cnt;
 }
 
 int pal_printf(const char* fmt, ...) {
@@ -45,7 +56,7 @@ int pal_printf(const char* fmt, ...) {
     int cnt;
 
     va_start(ap, fmt);
-    cnt = vprintf(fmt, ap);
+    cnt = vprintf(2, fmt, ap);
     va_end(ap);
 
     return cnt;
--- a/Pal/src/host/Linux-SGX/sgx_graphene.c
+++ b/Pal/src/host/Linux-SGX/sgx_graphene.c
@@ -29,7 +29,7 @@ static int fputch(void* f, int ch, struct printbuf* b) {
     return 0;
 }
 
-static int vprintf(int fd, const char* fmt, va_list ap) {
+static int vfdprintf(int fd, const char* fmt, va_list ap) {
     struct printbuf b;
 
     b.idx = 0;
@@ -45,7 +45,7 @@ int pal_fdprintf(int fd, const char* fmt, ...) {
     int cnt;
 
     va_start(ap, fmt);
-    cnt = vprintf(fd, fmt, ap);
+    cnt = vfdprintf(fd, fmt, ap);
     va_end(ap);
 
     return cnt;
@@ -56,7 +56,7 @@ int pal_printf(const char* fmt, ...) {
     int cnt;
 
     va_start(ap, fmt);
-    cnt = vprintf(2, fmt, ap);
+    cnt = vfdprintf(2, fmt, ap);
     va_end(ap);
 
     return cnt;
--- a/Pal/src/host/Linux-SGX/sgx_internal.h
+++ b/Pal/src/host/Linux-SGX/sgx_internal.h
@@ -147,7 +147,7 @@ int block_async_signals(bool block);
 
 void update_debugger(void);
 
-#ifdef SGX_PROFILE
+#ifdef DEBUG
 int sgx_profile_init(bool all);
 void sgx_profile_finish(void);
 void sgx_profile_sample(void* tcs);
--- a/Pal/src/host/Linux-SGX/sgx_internal.h
+++ b/Pal/src/host/Linux-SGX/sgx_internal.h
@@ -147,4 +147,10 @@ int block_async_signals(bool block);
 
 void update_debugger(void);
 
+#ifdef SGX_PROFILE
+int sgx_profile_init(bool all);
+void sgx_profile_finish(void);
+void sgx_profile_sample(void* tcs);
+#endif
+
 #endif
--- a/Pal/src/host/Linux-SGX/sgx_main.c
+++ b/Pal/src/host/Linux-SGX/sgx_main.c
@@ -301,17 +301,17 @@ static int initialize_enclave(struct pal_enclave* enclave, bool first_process) {
     }
     g_sgx_enable_stats = !!enable_stats_int64;
 
-    char* profile_str;
+    char* profile_str = NULL;
     ret = toml_string_in(enclave->manifest_root, "sgx.profile", &profile_str);
     if (ret < 0) {
-        SGX_DBG(DBG_E, "Cannot parse \'sgx.profile\' (the value must be 'none', 'root' or 'all')\n");
+        SGX_DBG(DBG_E, "Cannot parse \'sgx.profile\' (the value must be \"none\", \"main\" or \"all\")\n");
         ret = -EINVAL;
         goto out;
     }
 #ifdef SGX_PROFILE
     if (!profile_str || !strcmp(profile_str, "none")) {
         // do not enable
-    } else if (!strcmp(profile_str, "root")) {
+    } else if (!strcmp(profile_str, "main")) {
         if (first_process) {
             ret = sgx_profile_init(/*all=*/false);
             if (ret < 0)
@@ -322,15 +322,15 @@ static int initialize_enclave(struct pal_enclave* enclave, bool first_process) {
         if (ret < 0)
             goto out;
     } else {
-        SGX_DBG(DBG_E, "Invalid \'sgx.profile\' (the value must be 'none', 'root' or 'all')\n");
-        ret = EINVAL;
+        SGX_DBG(DBG_E, "Invalid \'sgx.profile\' (the value must be \"none', \"main\" or \"all\")\n");
+        ret = -EINVAL;
         goto out;
     }
 #else
     __UNUSED(first_process);
     if (profile_str && strcmp(profile_str, "none")) {
         SGX_DBG(DBG_E, "Cannot use \'sgx.profile\' when compiled without SGX_PROFILE\n");
-        ret = EINVAL;
+        ret = -EINVAL;
         goto out;
     }
 #endif
@@ -684,6 +684,7 @@ static int initialize_enclave(struct pal_enclave* enclave, bool first_process) {
     }
 
     ret = 0;
+
 out:
     if (enclave_image >= 0)
         INLINE_SYSCALL(close, 1, enclave_image);
--- a/Pal/src/host/Linux-SGX/sgx_main.c
+++ b/Pal/src/host/Linux-SGX/sgx_main.c
@@ -301,6 +301,13 @@ static int initialize_enclave(struct pal_enclave* enclave, bool first_process) {
     }
     g_sgx_enable_stats = !!enable_stats_int64;
 
+    ret = read_enclave_token(enclave->token, &enclave_token);
+    if (ret < 0) {
+        SGX_DBG(DBG_E, "Reading enclave token failed: %d\n", -ret);
+        goto out;
+    }
+    enclave->pal_sec.enclave_attributes = enclave_token.body.attributes;
+
     char* profile_str = NULL;
     ret = toml_string_in(enclave->manifest_root, "sgx.profile", &profile_str);
     if (ret < 0) {
@@ -309,22 +316,34 @@ static int initialize_enclave(struct pal_enclave* enclave, bool first_process) {
         goto out;
     }
 #ifdef DEBUG
+    bool enable_profile = false;
+    bool enable_profile_all = false;
+
     if (!profile_str || !strcmp(profile_str, "none")) {
         // do not enable
     } else if (!strcmp(profile_str, "main")) {
         if (first_process) {
-            ret = sgx_profile_init(/*all=*/false);
-            if (ret < 0)
-                goto out;
+            enable_profile = true;
         }
     } else if (!strcmp(profile_str, "all")) {
-        ret = sgx_profile_init(/*all=*/true);
-        if (ret < 0)
-            goto out;
+        enable_profile = true;
+        enable_profile_all = true;
     } else {
         SGX_DBG(DBG_E, "Invalid \'sgx.profile\' (the value must be \"none', \"main\" or \"all\")\n");
         ret = -EINVAL;
         goto out;
+    }
+
+    if (enable_profile) {
+        if (!(enclave->pal_sec.enclave_attributes.flags & SGX_FLAGS_DEBUG)) {
+            SGX_DBG(DBG_E, "Cannot use \'sgx.profile\' with a production enclave\n");
+            ret = -EINVAL;
+            goto out;
+        }
+
+        ret = sgx_profile_init(/*all=*/enable_profile_all);
+        if (ret < 0)
+            goto out;
     }
 #else
     __UNUSED(first_process);
@@ -334,13 +353,6 @@ static int initialize_enclave(struct pal_enclave* enclave, bool first_process) {
         goto out;
     }
 #endif
-
-    ret = read_enclave_token(enclave->token, &enclave_token);
-    if (ret < 0) {
-        SGX_DBG(DBG_E, "Reading enclave token failed: %d\n", -ret);
-        goto out;
-    }
-    enclave->pal_sec.enclave_attributes = enclave_token.body.attributes;
 
     ret = read_enclave_sigstruct(enclave->sigfile, &enclave_sigstruct);
     if (ret < 0) {
--- a/Pal/src/host/Linux-SGX/sgx_main.c
+++ b/Pal/src/host/Linux-SGX/sgx_main.c
@@ -172,7 +172,7 @@ static int load_enclave_binary(sgx_arch_secs_t* secs, int fd, unsigned long base
     return 0;
 }
 
-static int initialize_enclave(struct pal_enclave* enclave) {
+static int initialize_enclave(struct pal_enclave* enclave, bool first_process) {
     int ret = 0;
     int enclave_image = -1;
     sgx_arch_token_t enclave_token;
@@ -300,6 +300,40 @@ static int initialize_enclave(struct pal_enclave* enclave) {
         goto out;
     }
     g_sgx_enable_stats = !!enable_stats_int64;
+
+    char* profile_str;
+    ret = toml_string_in(enclave->manifest_root, "sgx.profile", &profile_str);
+    if (ret < 0) {
+        SGX_DBG(DBG_E, "Cannot parse \'sgx.profile\' (the value must be 'none', 'root' or 'all')\n");
+        ret = -EINVAL;
+        goto out;
+    }
+#ifdef SGX_PROFILE
+    if (!profile_str || !strcmp(profile_str, "none")) {
+        // do not enable
+    } else if (!strcmp(profile_str, "root")) {
+        if (first_process) {
+            ret = sgx_profile_init(/*all=*/false);
+            if (ret < 0)
+                goto out;
+        }
+    } else if (!strcmp(profile_str, "all")) {
+        ret = sgx_profile_init(/*all=*/true);
+        if (ret < 0)
+            goto out;
+    } else {
+        SGX_DBG(DBG_E, "Invalid \'sgx.profile\' (the value must be 'none', 'root' or 'all')\n");
+        ret = EINVAL;
+        goto out;
+    }
+#else
+    __UNUSED(first_process);
+    if (profile_str && strcmp(profile_str, "none")) {
+        SGX_DBG(DBG_E, "Cannot use \'sgx.profile\' when compiled without SGX_PROFILE\n");
+        ret = EINVAL;
+        goto out;
+    }
+#endif
 
     ret = read_enclave_token(enclave->token, &enclave_token);
     if (ret < 0) {
@@ -650,7 +684,6 @@ static int initialize_enclave(struct pal_enclave* enclave) {
     }
 
     ret = 0;
-
 out:
     if (enclave_image >= 0)
         INLINE_SYSCALL(close, 1, enclave_image);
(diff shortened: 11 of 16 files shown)


dimakuv (changes requested):
> Reviewed 14 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: all files reviewed, 27 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNrG4u1Rh1NGLOTp4J:-MNNrG4u1Rh1NGLOTp4K:bs9i9k2) (no related file):*
Do we want to add `SGX_PROFILE=1` to Jenkins on debug builds? Here: https://github.com/oscarlab/graphene/blob/master/.ci/lib/config-debug.jenkinsfile. Just to test that it compiles correctly.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNyhZuCiFuDk7L_g9e:-MNNyhZuCiFuDk7L_g9f:bcybehr) (no related file):*
> The Python program is pretty bad at symbol resolution: unlike perf and gdb, it doesn't understand that there can be external debug info. What can I use to do it better?

How does it work in GDB? I think there are some default paths that are looked for, with `[binaryname].debug` file? Cannot you do the same in your `profile-report` script, maybe even with additional arguments (i.e., the user must find the appropriate `.debug` files and explicitly specify them as arguments)?

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNzcmvEPwHj_JAhHFX:-MNNzcmvEPwHj_JAhHFY:b-o3793f) (no related file):*
> Recover call graph (or at least a single caller)? I guess we could follow saved RBP and get something useful most of the time...

This would be super-useful, even with a single caller. I'm afraid we'll end up with `spinlock_lock` or something less than helpful in the profiles, and without a call graph that's not helpful.

From what I see in our Makefiles and `enclave_entry.S`, yes, we can use RBP. And even if not, we should be able to change our Makefiles and fix the asm files to track it.

---
___
*[Documentation/manifest-syntax.rst, line 448 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNo7lu7MGnXcayfRTF:-MNNo7lu7MGnXcayfRTG:b60hre0) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Documentation/manifest-syntax.rst#L448)):*
> ```reStructuredText
> This syntax specifies whether to enable SGX enclave-specific statistics:
> 
> #. ``TCS.FLAGS.DBGOPTIN`` flag. This flag is set in all enclave threads and
> ```

How is it possible that you don't need to set this flag on `sgx.profile`? Or you do, but forgot to mention this in your added text?

Also, now we have two ways of checking stats. Maybe we should merge them into one? Like `sgx.profile = "sgx_stats"`?
___
*[Documentation/manifest-syntax.rst, line 473 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNkUVl32W3ZMw2Uyec:-MNNkUVl32W3ZMw2Uyed:b383kgy) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Documentation/manifest-syntax.rst#L473)):*
> ```reStructuredText
> with ``DEBUG=1`` and ``SGX_PROFILE=1`` for this option to work.
> 
> If this option is set to ``root``, the root process will collect IP samples and
> ```

Can we specify `root (main) process` as a synonym? (I typically use "main process" and haven't heard the term "root process".)
___
*[Documentation/manifest-syntax.rst, line 474 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNk9K02C4M-XdmzKVp:-MNNk9K02C4M-XdmzKVq:b-oxq368) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Documentation/manifest-syntax.rst#L474)):*
> ```reStructuredText
> 
> If this option is set to ``root``, the root process will collect IP samples and
> save them as ``profile-sgx.data``. If it's to ``all``, all processes will
> ```

`If it's to` -> `If it is set to`
___
*[Documentation/devel/performance.rst, line 531 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNkmNf5ku5OtNdHaeS:-MNNkmNf5ku5OtNdHaeT:b-6w324r) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Documentation/devel/performance.rst#L531)):*
> ```reStructuredText
>    (``Pal/src/host/Linux-SGX/tools/profile-report/profile-report``) with the
>    data file as parameter. It should display a table of functions (or raw
>    addresses, if symbol information is not available)
> ```

Add a dot at the end.
___
*[Documentation/devel/performance.rst, line 535 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNkxVw2o1QJllq1-TX:-MNNkxVw2o1QJllq1-TY:bdl70w4) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Documentation/devel/performance.rst#L535)):*
> ```reStructuredText
> *Note*: The accuracy of this tool is unclear. The SGX profiling works by
> measuring the value of instruction pointer on each asynchronous enclave exit
> (AEX), which happen on Linux scheduler interrupts as well as other such as page
> ```

`as well as other such` -> `as well as other events`
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 43 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNnaceDnfQlt99aOGK:-MNNnaceDnfQlt99aOGL:b-8h7j7h) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/enclave_framework.c#L43)):*
> ```C
>  * When SGX_PROFILE is enabled, we run some code on asynchronous enclave exit (AEX) that uses the
>  * stack. Make sure to update URSP so that the AEX handler does not overwrite the part of the stack
>  * that we just allocated.
> ```

Please add smth like: `(recall that URSP is an outside stack pointer saved by EENTER and restored on AEX by the SGX hardware itself)`
___
*[Pal/src/host/Linux-SGX/Makefile, line 73 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNnmtR7d4r6MatCru_:-MNNnmtR7d4r6MatCrua:b97bcjv) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/Makefile#L73)):*
> ```Makefile
> urts-objs = \
> 	clone-x86_64.o \
> 	sgx_profile.o \
> ```

This is not sorted.
___
*[Pal/src/host/Linux-SGX/sgx_entry.S, line 81 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNpH-M5yFMULeLLnwj:-MNNpH-M5yFMULeLLnwk:bv58gli) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_entry.S#L81)):*
> ```
> 	.cfi_adjust_cfa_offset 8
> 
> 	# Align stack (eequired by System V AMD64 ABI)
> ```

`required`
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 304 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNpxZuBI9s2l9VYPUI:-MNNpxZuBI9s2l9VYPUJ:bytpqpv) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_main.c#L304)):*
> ```C
>     g_sgx_enable_stats = !!enable_stats_int64;
> 
>     char* profile_str;
> ```

Could you set it to `= NULL` for sanity?
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 307 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNqT9FEmH-8MCoMgei:-MNNqT9FEmH-8MCoMgej:bje96tq) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_main.c#L307)):*
> ```C
>     ret = toml_string_in(enclave->manifest_root, "sgx.profile", &profile_str);
>     if (ret < 0) {
>         SGX_DBG(DBG_E, "Cannot parse \'sgx.profile\' (the value must be 'none', 'root' or 'all')\n");
> ```

Can you use double-quotes for values? This may be confusing to a casual user (they will try to literally set `sgx.profile = 'none'` and won't be happy with the result).
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 318 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNt9Yd-uD6_jXY_8MP:-MNNt9Yd-uD6_jXY_8MQ:bsg8vje) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_main.c#L318)):*
> ```C
>             ret = sgx_profile_init(/*all=*/false);
>             if (ret < 0)
>                 goto out;
> ```

`sgx_profile_init()` may return ENOSYS on older machines (because invariant-TSC is a rather new CPU feature). Basically, we cannot profile on machines older than Icelake (I think). We should either handle this gracefully in the C code, or add this disclaimer to the documentation.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 323 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNtQU44LmK30nBbOoz:-MNNtQU44LmK30nBbOp-:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_main.c#L323)):*
> ```C
>         ret = sgx_profile_init(/*all=*/true);
>         if (ret < 0)
>             goto out;
> ```

ditto
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 325 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNqdgd5oAitQ19oP7H:-MNNqdgd5oAitQ19oP7I:b-qubziu) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_main.c#L325)):*
> ```C
>             goto out;
>     } else {
>         SGX_DBG(DBG_E, "Invalid \'sgx.profile\' (the value must be 'none', 'root' or 'all')\n");
> ```

Can you use double-quotes for values?
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 326 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNqnrI500wKkbf9Hm9:-MNNqnrI500wKkbf9HmA:btp45av) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_main.c#L326)):*
> ```C
>     } else {
>         SGX_DBG(DBG_E, "Invalid \'sgx.profile\' (the value must be 'none', 'root' or 'all')\n");
>         ret = EINVAL;
> ```

`-EINVAL`
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNrgRWAaiZOGA67Qfs:-MNNrgRWAaiZOGA67Qft:b3cyff0) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L26)):*
> ```C
>     void* ip;
>     uint64_t count;
>     UT_hash_handle hh;
> ```

You use UTHash but don't add it in includes?
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 53 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNtifB9pD6unXFAPV1:-MNNtifB9pD6unXFAPV2:b-k6z6xj) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L53)):*
> ```C
>         if (IS_ERR(ret)) {
>             SGX_DBG(DBG_E, "debug_read: error reading %lu bytes at %p: %d\n", size, addr, ERRNO(ret));
>             INLINE_SYSCALL(exit_group, 1, ERRNO(ret));
> ```

Exiting the whole process on a bad debug-read is a bit extreme... Is this intentional?
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 58 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNtny41jgcPdnyOohm:-MNNtny41jgcPdnyOohn:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L58)):*
> ```C
>         if (ret == 0) {
>             SGX_DBG(DBG_E, "debug_read: EOF reading %lu bytes at %p\n", size, addr);
>             INLINE_SYSCALL(exit_group, 1, 255);
> ```

ditto
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 86 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNsFY37uh6RTsKuRLV:-MNNsFY37uh6RTsKuRLW:bmlo2c6) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L86)):*
> ```C
> }
> 
> #define CPUID_LEAF_TSC_FREQ 0x15
> ```

Now that we use this define in two files, let's move it to a common header (maybe `pal_host.h`)? Note that Linux-SGX PAL is x86-specific anyway, so we may just plug it in a header file as-is.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 88 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNsqK_655KC_sw4qiq:-MNNsqK_655KC_sw4qir:b-xg0qi8) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L88)):*
> ```C
> #define CPUID_LEAF_TSC_FREQ 0x15
> 
> static uint64_t get_tsc_hz(void) {
> ```

A bit annoying to duplicate this function, but I understand that there is a slight difference in the code (`_DkCpuIdRetrieve` vs `cpuid`). So nothing that can be done here.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 113 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNu_bY1wnl3Oe9smGx:-MNNu_bY1wnl3Oe9smGy:b8fpwq) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L113)):*
> ```C
>     }
> 
>     // Assume Linux scheduler will normally interrupt the enclave 4 ms, or 250 times per second.
> ```

`each 4 ms`
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 147 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNuxPL9XucBjwKJWVL:-MNNuxPL9XucBjwKJWVM:b-j0yy8w) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L147)):*
> ```C
>     SGX_DBG(DBG_I, "writing profile data to %s\n", buf);
> 
>     int fd = INLINE_SYSCALL(open, 3, buf, O_WRONLY | O_TRUNC | O_CREAT, PERM_rw_______);
> ```

Could we allow other users/groups to read this file? Sometimes useful.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 225 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNvulC7vudyu8DEDdf:-MNNvulC7vudyu8DEDdg:bmcfpz) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L225)):*
> ```C
>         HASH_FIND_PTR(g_counters, &ip, counter);
>         if (counter) {
>             counter->count += dt;
> ```

Why do we increase by ticks and not by `1`? I don't understand the high-level comment of this function as well...
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 229 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNw1RFAdki9JCkafK8:-MNNw1RFAdki9JCkafK9:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L229)):*
> ```C
>             counter = malloc(sizeof(*counter));
>             counter->ip = ip;
>             counter->count = dt;
> ```

ditto
___
*[Pal/src/host/Linux-SGX/tools/profile-report/profile-report, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNxWD0E3dlAUUwXRmi:-MNNxWD0E3dlAUUwXRmj:b-wyabfc) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/tools/profile-report/profile-report#L15)):*
> ```
> import os
> import sys
> import bisect
> ```

Can we at least partially sort?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: 6 of 16 files reviewed, 28 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @pwmarcz, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNrG4u1Rh1NGLOTp4J:-MNOhUhx8pb-TxKexbPt:b13g43j) (no related file):*

Do we want to add `SGX_PROFILE=1` to Jenkins on debug builds? Here: https://github.com/oscarlab/graphene/blob/master/.ci/lib/config-debug.jenkinsfile. Just to test that it compiles correctly.
</blockquote></details>

Sounds good to me. Strictly speaking it could be tested as a separate configuration, but it's probably not worth the resources.

(Actually, do we have debug builds for SGX yet? @woju ?)

Alternatively, maybe we could have this under `DEBUG` and remove `SGX_PROFILE`? I was afraid of adding the code in `async_exit_pointer` without a feature flag, but the overhead should be negligible when profiling is disabled, and maybe in `DEBUG` it's okay to have it. What do you think? 

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNyhZuCiFuDk7L_g9e:-MNOkcuQD9UNluZxFJ9d:bnbz1my) (no related file):*

> The Python program is pretty bad at symbol resolution: unlike perf and gdb, it doesn't understand that there can be external debug info. What can I use to do it better?

How does it work in GDB? I think there are some default paths that are looked for, with `[binaryname].debug` file? Cannot you do the same in your `profile-report` script, maybe even with additional arguments (i.e., the user must find the appropriate `.debug` files and explicitly specify them as arguments)?
</blockquote></details>

Okay, I implemented [what GDB specifies](https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html) and it seems to work okay. For example, I installed `python3.6-dbg` and now the symbols in Python binary resolve correctly.

I think that's enough for now - I would rather not add a separate option until somebody needs it; I think being able to install a `*-dbg` package or provide `binary.debug` / `.debug/binary.debug` file should be enough.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNzcmvEPwHj_JAhHFX:-MNOzZmQ64vk3lvexjfC:b-gzcu04) (no related file):*

> Recover call graph (or at least a single caller)? I guess we could follow saved RBP and get something useful most of the time...

This would be super-useful, even with a single caller. I'm afraid we'll end up with `spinlock_lock` or something less than helpful in the profiles, and without a call graph that's not helpful.

From what I see in our Makefiles and `enclave_entry.S`, yes, we can use RBP. And even if not, we should be able to change our Makefiles and fix the asm files to track it.
</blockquote></details>

Okay. I can try it in next PR.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNParlz9FcL5WRPwNUn:-MNParlz9FcL5WRPwNUo:b-dllkoa) (no related file):*
I realized keeping a global variable for last timestamp is incorrect with multiple threads, so I moved it to thread-local storage (`sgx_tls`).

---
___
*[Documentation/manifest-syntax.rst, line 448 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNo7lu7MGnXcayfRTF:-MNP103M2lC8b-6FdMyh:b5yhmzg) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Documentation/manifest-syntax.rst#L448)):*
>How is it possible that you don't need to set this flag on sgx.profile? Or you do, but forgot to mention this in your added text?

Apparently `TCS.FLAGS.DBGOPTIN` controls these additional features but not whether to allow access to memory. This is controlled by DEBUG flag in SECS structure, which is enabled by default, and can be disabled by `sgx.debug = 0` in manifest.

>Also, now we have two ways of checking stats. Maybe we should merge them into one? Like sgx.profile = "sgx_stats"?

These are separate subsystems, and you can configure them independently (enable one of them, or enable both, and also configure `sgx.profile` as `root/all`), so I think they should stay as separate manifest options.
___
*[Documentation/manifest-syntax.rst, line 473 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNkUVl32W3ZMw2Uyec:-MNP1vOZ1Yj3LUKN3LxW:b-4ynoge) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Documentation/manifest-syntax.rst#L473)):*

Can we specify `root (main) process` as a synonym? (I typically use "main process" and haven't heard the term "root process".)
</blockquote></details>

Sure, I invented this on the spot. Changed to `main`.
___
*[Documentation/manifest-syntax.rst, line 474 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNk9K02C4M-XdmzKVp:-MNP2Tm66C5jG6WECNhj:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Documentation/manifest-syntax.rst#L474)):*

`If it's to` -> `If it is set to`
</blockquote></details>

Done.
___
*[Documentation/devel/performance.rst, line 531 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNkmNf5ku5OtNdHaeS:-MNP2_k31axU2K9GPJpD:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Documentation/devel/performance.rst#L531)):*

Add a dot at the end.
</blockquote></details>

Done.
___
*[Documentation/devel/performance.rst, line 535 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNkxVw2o1QJllq1-TX:-MNP2g0r5OmGDvUSMheP:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Documentation/devel/performance.rst#L535)):*

`as well as other such` -> `as well as other events`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 43 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNnaceDnfQlt99aOGK:-MNP2wTa6KSo4INwRK9f:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/enclave_framework.c#L43)):*

Please add smth like: `(recall that URSP is an outside stack pointer saved by EENTER and restored on AEX by the SGX hardware itself)`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/Makefile, line 73 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNnmtR7d4r6MatCru_:-MNP30PICR8uf5pOQ20g:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/Makefile#L73)):*

This is not sorted.
</blockquote></details>

Fixed.
___
*[Pal/src/host/Linux-SGX/sgx_entry.S, line 81 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNpH-M5yFMULeLLnwj:-MNP38Y01DRRM84BCpZQ:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_entry.S#L81)):*

`required`
</blockquote></details>

Fixed.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 304 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNpxZuBI9s2l9VYPUI:-MNP3UVa67CY8RifagiA:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_main.c#L304)):*

Could you set it to `= NULL` for sanity?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 307 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNqT9FEmH-8MCoMgei:-MNP3kqk0bOPBBWRnoIh:b-24zulu) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_main.c#L307)):*

Can you use double-quotes for values? This may be confusing to a casual user (they will try to literally set `sgx.profile = 'none'` and won't be happy with the result).
</blockquote></details>

Oh, I didn't know TOML used double-quotes only. That makes sense (and I like TOML more now!)

Fixed.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 318 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNt9Yd-uD6_jXY_8MP:-MNPCB1r4I0dcSqh32lR:b-cm7bws) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_main.c#L318)):*

`sgx_profile_init()` may return ENOSYS on older machines (because invariant-TSC is a rather new CPU feature). Basically, we cannot profile on machines older than Icelake (I think). We should either handle this gracefully in the C code, or add this disclaimer to the documentation.
</blockquote></details>

I started looking into that deeper, and realized that there is no point in using RDTSC if we are already issuing multiple system calls to read the enclave memory. So I changed that to `clock_gettime(CLOCK_MONOTONIC)`. The error condition should not occur anymore.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 325 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNqdgd5oAitQ19oP7H:-MNPT22b7Pf93ORTL0Vn:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_main.c#L325)):*

Can you use double-quotes for values?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 326 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNqnrI500wKkbf9Hm9:-MNPT6aG1V3-vb5j2UHV:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_main.c#L326)):*

`-EINVAL`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNrgRWAaiZOGA67Qfs:-MNPTFpt9zA-ebHtRntC:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L26)):*

You use UTHash but don't add it in includes?
</blockquote></details>

Fixed.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 53 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNtifB9pD6unXFAPV1:-MNPW2WmEPl3q_du1wkc:boj66x5) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L53)):*

Exiting the whole process on a bad debug-read is a bit extreme... Is this intentional?
</blockquote></details>

Yeah, my bad. I replaced it with a normal return.

It's still not a great experience if the reads fail (a lot of error messages on trying to read the IP), but all things considered, I would rather keep this log message because it makes it easier to see what's wrong.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 86 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNsFY37uh6RTsKuRLV:-MNPTR6R4aFFeYFRnmnL:bqvt945) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L86)):*

Now that we use this define in two files, let's move it to a common header (maybe `pal_host.h`)? Note that Linux-SGX PAL is x86-specific anyway, so we may just plug it in a header file as-is.
</blockquote></details>

Replaced with Linux clock, see above.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 88 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNsqK_655KC_sw4qiq:-MNPTNJoBSHaohpImIf0:bqvt945) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L88)):*

A bit annoying to duplicate this function, but I understand that there is a slight difference in the code (`_DkCpuIdRetrieve` vs `cpuid`). So nothing that can be done here.
</blockquote></details>

Replaced with Linux clock, see above.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 113 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNu_bY1wnl3Oe9smGx:-MNPTLwZDMZFNpil68nJ:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L113)):*

`each 4 ms`
</blockquote></details>

Fixed.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 147 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNuxPL9XucBjwKJWVL:-MNPXh2k7F0v80W9WdAF:b-ii41o) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L147)):*

Could we allow other users/groups to read this file? Sometimes useful.
</blockquote></details>

Sure, this is all debug-mode anyway, so no need to keep it extra secret, I guess. Changed to 0644.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 225 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNvulC7vudyu8DEDdf:-MNPYiks8jEZsnrVjkrB:b-jy4q5l) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L225)):*

Why do we increase by ticks and not by `1`? I don't understand the high-level comment of this function as well...
</blockquote></details>

Maybe I'll start by showing what happened when I tried that.

This is the report for a short program that spends a few seconds in a loop (20e9 iterations):
```
   TIME  OBJ                   SYMBOL              

 93.91%  bootstrap             spin                
  4.57%  libpal.so             memcpy              
  0.62%  libpal.so             mbedtls_internal_sha256_process
  0.49%  libpal.so             cmac_xor_block      
  0.12%  libpal.so             mbedtls_aesni_crypt_ecb
  0.08%  libpal.so             memset              
  0.07%  libpal.so             _DkMutexLockTimeout 
```

This is the report for the same program when we increase by 1:
```
   TIME  OBJ                   SYMBOL              

 46.29%  libpal.so             memcpy              
 43.52%  bootstrap             spin                
  4.50%  libpal.so             memset              
  1.43%  libpal.so             mbedtls_internal_sha256_process
  0.84%  libpal.so             cmac_xor_block      
  0.69%  libpal.so             _DkExceptionHandler 
  0.24%  libpal.so             maybe_enlarge_slab_mgr
  0.22%  libpal.so             mbedtls_aesni_crypt_ecb
  0.17%  libpal.so             create_slab_mgr     
  0.09%  libpal.so             pal_linux_main      
  0.09%  libpal.so             elf_machine_rela_relative
  0.09%  libpal.so             strnlen             
  0.09%  libpal.so             _DkMutexLockTimeout 
  0.06%  libpal.so             handle_ecall        
  0.06%  libpal.so             (2a0ea)             
  0.06%  libpal.so             elf_dynamic_do_rela 
  0.06%  libpal.so             slab_alloc          
```

I think the difference is that there are a lot of AEXs caused by page faults, which inflates the count for `memcpy` and other functions, even though much less real time is spent in them than in `spin`. (I checked the exact IP value and it points right at the `rep movsb` instruction in memcpy, with 2138 occurences).

That's why I think measuring time is more accurate than just measuring number of AEXs (which is what increasing by `1` would do). Is that more clear? If so, do you have any suggestion how I could reword the explanation in the comment?
___
*[Pal/src/host/Linux-SGX/tools/profile-report/profile-report, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNxWD0E3dlAUUwXRmi:-MNPXwzCCqiw7FbsWY1R:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/tools/profile-report/profile-report#L15)):*

Can we at least partially sort?
</blockquote></details>

Done.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: 6 of 16 files reviewed, 28 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNrG4u1Rh1NGLOTp4J:-MNPsQ-80aysOxhFWI2w:b-nbt2qu) (no related file):*
> Actually, do we have debug builds for SGX yet?
No, we don't. Right now we lack capacity, but I'm working on disabling LTP on SGX, so we could gain some time here, which we could expend on SGX debug builds. Right now this is stuck as part of #2003, which isn't ready yet.

---

dimakuv (changes requested):
> Reviewed 10 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow, @pwmarcz, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNrG4u1Rh1NGLOTp4J:-MNSbV_VBj_-119A7wKo:btjzg51) (no related file):*

> Actually, do we have debug builds for SGX yet?
No, we don't. Right now we lack capacity, but I'm working on disabling LTP on SGX, so we could gain some time here, which we could expend on SGX debug builds. Right now this is stuck as part of #2003, which isn't ready yet.
</blockquote></details>

I see no problem in tucking the `SGX_PROFILE` functionality under `DEBUG`. The introduced AEX code is quite small and fast. And having a separate IFDEF logic for it is bad for development and testing (of other features, since developers will forget to test with `SGX_PROFILE=1` and discrepancies will creep in).

We'll need @mkow and @yamahata to chime in.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNyhZuCiFuDk7L_g9e:-MNSc8WkAnafOgDZcKIN:b-vifnll) (no related file):*

Okay, I implemented [what GDB specifies](https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html) and it seems to work okay. For example, I installed `python3.6-dbg` and now the symbols in Python binary resolve correctly.

I think that's enough for now - I would rather not add a separate option until somebody needs it; I think being able to install a `*-dbg` package or provide `binary.debug` / `.debug/binary.debug` file should be enough.
</blockquote></details>

Ok, this is awesome!

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNParlz9FcL5WRPwNUn:-MNSlGdm3GANpGHrwX8j:b2eca5z) (no related file):*

I realized keeping a global variable for last timestamp is incorrect with multiple threads, so I moved it to thread-local storage (`sgx_tls`).
</blockquote></details>

OK. You marked this comment as "blocking", please unblock yourself.

---
___
*[Documentation/manifest-syntax.rst, line 448 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNo7lu7MGnXcayfRTF:-MNScJQt7HPRi05uQH81:bkdwrpw) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Documentation/manifest-syntax.rst#L448)):*

>How is it possible that you don't need to set this flag on sgx.profile? Or you do, but forgot to mention this in your added text?

Apparently `TCS.FLAGS.DBGOPTIN` controls these additional features but not whether to allow access to memory. This is controlled by DEBUG flag in SECS structure, which is enabled by default, and can be disabled by `sgx.debug = 0` in manifest.

>Also, now we have two ways of checking stats. Maybe we should merge them into one? Like sgx.profile = "sgx_stats"?

These are separate subsystems, and you can configure them independently (enable one of them, or enable both, and also configure `sgx.profile` as `root/all`), so I think they should stay as separate manifest options.
</blockquote></details>

OK, agreed in both points.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 225 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNvulC7vudyu8DEDdf:-MNSiJV7ANdb26EZGGbD:b-r13gpl) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L225)):*

Maybe I'll start by showing what happened when I tried that.

This is the report for a short program that spends a few seconds in a loop (20e9 iterations):
```
   TIME  OBJ                   SYMBOL              

 93.91%  bootstrap             spin                
  4.57%  libpal.so             memcpy              
  0.62%  libpal.so             mbedtls_internal_sha256_process
  0.49%  libpal.so             cmac_xor_block      
  0.12%  libpal.so             mbedtls_aesni_crypt_ecb
  0.08%  libpal.so             memset              
  0.07%  libpal.so             _DkMutexLockTimeout 
```

This is the report for the same program when we increase by 1:
```
   TIME  OBJ                   SYMBOL              

 46.29%  libpal.so             memcpy              
 43.52%  bootstrap             spin                
  4.50%  libpal.so             memset              
  1.43%  libpal.so             mbedtls_internal_sha256_process
  0.84%  libpal.so             cmac_xor_block      
  0.69%  libpal.so             _DkExceptionHandler 
  0.24%  libpal.so             maybe_enlarge_slab_mgr
  0.22%  libpal.so             mbedtls_aesni_crypt_ecb
  0.17%  libpal.so             create_slab_mgr     
  0.09%  libpal.so             pal_linux_main      
  0.09%  libpal.so             elf_machine_rela_relative
  0.09%  libpal.so             strnlen             
  0.09%  libpal.so             _DkMutexLockTimeout 
  0.06%  libpal.so             handle_ecall        
  0.06%  libpal.so             (2a0ea)             
  0.06%  libpal.so             elf_dynamic_do_rela 
  0.06%  libpal.so             slab_alloc          
```

I think the difference is that there are a lot of AEXs caused by page faults, which inflates the count for `memcpy` and other functions, even though much less real time is spent in them than in `spin`. (I checked the exact IP value and it points right at the `rep movsb` instruction in memcpy, with 2138 occurences).

That's why I think measuring time is more accurate than just measuring number of AEXs (which is what increasing by `1` would do). Is that more clear? If so, do you have any suggestion how I could reword the explanation in the comment?
</blockquote></details>

Ok, now I understand this. What I was missing in my mental model is that we count only on AEX events, and these events do *not* have a fixed sampling period (unlike `perf record`). So we'll give more weight to the samples (and their corresponding IPs) after longer execution periods, and less weight to the samples after shorter execution periods. But we need to cap the duration of one execution period, otherwise a blocking/sleeping thread may seriously skew the results.

In other words, if AEX events would happen only due to timer interrupts (every 4ms), then we would have a fixed sampling period and could use `1` (or any fixed value like the clock `dt`, which would be approx. equal to 4ms). But since AEX events can happen also due to other events, we need to lower the weight of these "interjecting" events.

Well, I'm not sure how to describe it succinctly in a comment. Maybe `Samples are collected on AEX events, thus this profiling method doesn't have a fixed sampling period (unlike perf-record). But on modern Linux, we know that timer interrupts happen every 4ms (unless explicitly modified by admin), so we may assume that at least one AEX event happens every 4ms. However, AEX events happen on other interrupts/exceptions as well (such as page faults), thus we must lower the weight of these interjecting AEXs. We use "elapsed clock cycles" as a simple and sufficient way to control the weights of samples happening with different frequencies.`
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 63 at r2](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNSd5Yp03PpN749ETIK:-MNSd5Yp03PpN749ETIL:b-jnj20v) ([raw file](https://github.com/oscarlab/graphene/blob/78eae2e6a168fec876845522bffc7e4bd643f3ef/Pal/src/host/Linux-SGX/sgx_profile.c#L63)):*
> ```C
>         if (ret == 0) {
>             SGX_DBG(DBG_E, "debug_read: EOF reading %lu bytes at %p\n", size, addr);
>             return ret;
> ```

This returns `0` (success). Do you want to change to smth like `-EINVAL`? But I'm not blocking on this.

yamahata (changes requested):
> Reviewed 6 of 14 files at r1, 10 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: 15 of 16 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, @pwmarcz, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNrG4u1Rh1NGLOTp4J:-MNT8MKn4zmKDq6Q52Dg:bnjv9v6) (no related file):*

I see no problem in tucking the `SGX_PROFILE` functionality under `DEBUG`. The introduced AEX code is quite small and fast. And having a separate IFDEF logic for it is bad for development and testing (of other features, since developers will forget to test with `SGX_PROFILE=1` and discrepancies will creep in).

We'll need @mkow and @yamahata to chime in.
</blockquote></details>

+1 for DEBUG because people won’t test many ifdef combinations.

---
___
*[Pal/src/host/Linux-SGX/sgx_graphene.c, line 32 at r2](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNT9xxY3exGXdZqU5wK:-MNT9ufxEoc9b8QSOSQf:b-44p6km) ([raw file](https://github.com/oscarlab/graphene/blob/78eae2e6a168fec876845522bffc7e4bd643f3ef/Pal/src/host/Linux-SGX/sgx_graphene.c#L32)):*
> ```C
> }
> 
> static int vprintf(int fd, const char* fmt, va_list ap) {
> ```

nit: can the function be renamed?
because vprintf is wildly accepted name and we’re deviating from it.
___
*[Pal/src/host/Linux-SGX/tools/profile-report/profile-report, line 108 at r2](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNTtcKi2MKQqYi6Lwta:-MNTtcKi2MKQqYi6Lwtb:b-mi6lmg) ([raw file](https://github.com/oscarlab/graphene/blob/78eae2e6a168fec876845522bffc7e4bd643f3ef/Pal/src/host/Linux-SGX/tools/profile-report/profile-report#L108)):*
> ```
>         for addr, file_name in self.files:
>             self.symbols.extend(load_symbols(file_name, addr))
> ```

nitpick:
files.symbolts = [load_symbols(file_name, addr) if _, file_name in self.files]

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: 15 of 16 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, @pwmarcz, and @yamahata)

___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 225 at r1](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNvulC7vudyu8DEDdf:-MNTrn2G8qWvqhBJO2hQ:bwh1yj6) ([raw file](https://github.com/oscarlab/graphene/blob/96fc1f7c629b06fe49af7120c69c4ba97cf13cc4/Pal/src/host/Linux-SGX/sgx_profile.c#L225)):*

Ok, now I understand this. What I was missing in my mental model is that we count only on AEX events, and these events do *not* have a fixed sampling period (unlike `perf record`). So we'll give more weight to the samples (and their corresponding IPs) after longer execution periods, and less weight to the samples after shorter execution periods. But we need to cap the duration of one execution period, otherwise a blocking/sleeping thread may seriously skew the results.

In other words, if AEX events would happen only due to timer interrupts (every 4ms), then we would have a fixed sampling period and could use `1` (or any fixed value like the clock `dt`, which would be approx. equal to 4ms). But since AEX events can happen also due to other events, we need to lower the weight of these "interjecting" events.

Well, I'm not sure how to describe it succinctly in a comment. Maybe `Samples are collected on AEX events, thus this profiling method doesn't have a fixed sampling period (unlike perf-record). But on modern Linux, we know that timer interrupts happen every 4ms (unless explicitly modified by admin), so we may assume that at least one AEX event happens every 4ms. However, AEX events happen on other interrupts/exceptions as well (such as page faults), thus we must lower the weight of these interjecting AEXs. We use "elapsed clock cycles" as a simple and sufficient way to control the weights of samples happening with different frequencies.`
</blockquote></details>

Thanks. I reworded the comment based on your suggestion.

I also changed the clock to `CLOCK_THREAD_CPUTIME_ID`. It should give better results in case the thread sleeps for some time, due to high load or blocking system calls (while these do not generate AEX, the next AEX will be delayed, so we would count too much time anyway). On my toy example from above, the result is pretty similar but `_DkMutexLockTimeout` went down from 0.07% to about 0.03%.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 63 at r2](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNSd5Yp03PpN749ETIK:-MNTnYaOASNtf0T_CvRD:bdwcyr5) ([raw file](https://github.com/oscarlab/graphene/blob/78eae2e6a168fec876845522bffc7e4bd643f3ef/Pal/src/host/Linux-SGX/sgx_profile.c#L63)):*

This returns `0` (success). Do you want to change to smth like `-EINVAL`? But I'm not blocking on this.
</blockquote></details>

Good catch. I'm changing it to `-EINVAL` as we always expect to read the full size.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 216 at r3](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNU1oNB0gEBY-rwXCvi:-MNU1oNC-MLFgKUdfg-2:b-mkww0u) ([raw file](https://github.com/oscarlab/graphene/blob/2aea517515de5c5874be94c9b5842138e3263c9f/Pal/src/host/Linux-SGX/sgx_profile.c#L216)):*
> ```C
>     // Check current CPU time
>     struct timespec ts;
>     int res = INLINE_SYSCALL(clock_gettime, 2, CLOCK_THREAD_CPUTIME_ID, &ts);
> ```

Good idea to use this clock!

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: 5 of 15 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNNrG4u1Rh1NGLOTp4J:-MNXxAgg0UxDUoIY-LK5:b-58xudh) (no related file):*

+1 for DEBUG because people won’t test many ifdef combinations.
</blockquote></details>

Done. I removed `SGX_PROFILE` and used `DEBUG` instead.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNXzzKs3TgYB5kNYCpT:-MNXzzKs3TgYB5kNYCpU:baggw3x) (no related file):*
First success already! I used this to figure out a performance degradation in Tensorflow. The report said that a suspicious amount of time is spent in `memset`, which turned out to be zeroing memory for `_DkVirtualMemoryAlloc`, because the code allocated a lot of memory using `mmap`.

I think this confirms that some callers information would be super helpful (I added some code to get to `*(RBP + 8)` to find out about `_DkVirtualMemoryAlloc`). I talked with @mkow and he pointed out that using RBP will be inaccurate (the code might not use frame pointers, or have different conventions, etc.) But I still want to try that out, maybe adding some sanity checks for the value of RBP relative to RSP.

---
___
*[Pal/src/host/Linux-SGX/sgx_graphene.c, line 32 at r2](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNT9xxY3exGXdZqU5wK:-MNXxZnhEIgHTFAjWpL8:bcyxdl6) ([raw file](https://github.com/oscarlab/graphene/blob/78eae2e6a168fec876845522bffc7e4bd643f3ef/Pal/src/host/Linux-SGX/sgx_graphene.c#L32)):*

nit: can the function be renamed?
because vprintf is wildly accepted name and we’re deviating from it.
</blockquote></details>

I renamed it to `vfdprintf`, to match the `pal_fdprintf` below.
___
*[Pal/src/host/Linux-SGX/tools/profile-report/profile-report, line 108 at r2](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNTtcKi2MKQqYi6Lwta:-MNXxfaD3n3dQPjC9M-P:b-2764nu) ([raw file](https://github.com/oscarlab/graphene/blob/78eae2e6a168fec876845522bffc7e4bd643f3ef/Pal/src/host/Linux-SGX/tools/profile-report/profile-report#L108)):*

> ```
>         for addr, file_name in self.files:
>             self.symbols.extend(load_symbols(file_name, addr))
> ```

nitpick:
files.symbolts = [load_symbols(file_name, addr) if _, file_name in self.files]
</blockquote></details>

That would create a list of lists, not a flat list (I'm using `extend` here, not `append`). I could use something like
```
self.symbols = itertools.chain(*(load_symbols(file_name, addr) for addr, file_name in self.files))
```
but I think this is too much complexity for a single line, and the current way is more readable.

yamahata (approved):
> Reviewed 10 of 11 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/tools/profile-report/profile-report, line 108 at r2](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNTtcKi2MKQqYi6Lwta:-MN_oFvv6_mB0BjlcCPo:b-cx07dc) ([raw file](https://github.com/oscarlab/graphene/blob/78eae2e6a168fec876845522bffc7e4bd643f3ef/Pal/src/host/Linux-SGX/tools/profile-report/profile-report#L108)):*

That would create a list of lists, not a flat list (I'm using `extend` here, not `append`). I could use something like
```
self.symbols = itertools.chain(*(load_symbols(file_name, addr) for addr, file_name in self.files))
```
but I think this is too much complexity for a single line, and the current way is more readable.

</blockquote></details>

Oops. I scratch that.

dimakuv:
Jenkins, retest Jenkins-Debug-18.04 please (`sendfile08` LTP test failed/timed out, never seen this before, must be unrelated to this PR)

dimakuv (changes requested):
> Reviewed 9 of 11 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 485 at r4](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNboiTR2MPPnfXYDrwV:-MNboiTR2MPPnfXYDrwW:bnvs28c) ([raw file](https://github.com/oscarlab/graphene/blob/4f95892b315f01517e57d83dc313731a7889901c/Documentation/manifest-syntax.rst#L485)):*
> ```reStructuredText
> *Note:* this option is insecure and cannot be used with production enclaves
> (``sgx.debug = 0``). If the production enclave is started with this option set,
> Graphene will fail initialization of the enclave.
> ```

Is this true? I don't see any enforcement of this rule in your code.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 247 at r4](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNboZVvCQspmMzRDCHL:-MNboZVvCQspmMzRDCHM:b86gedd) ([raw file](https://github.com/oscarlab/graphene/blob/4f95892b315f01517e57d83dc313731a7889901c/Pal/src/host/Linux-SGX/sgx_profile.c#L247)):*
> ```C
>             counter->count += dt;
>         } else {
>             counter = malloc(sizeof(*counter));
> ```

Can we add error checking with "out of memory" debug message?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNbpeAzAtRUl3NDzOeq:-MNbpeAzAtRUl3NDzOer:b-bzdeur) (no related file):*
FYI: Jenkins-SGX-gsc pipeline fails because this PR was unfortunate to be submitted during the Jenkins transitioning period. The gsc failure must go away after the rebase of this PR.

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 485 at r4](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNboiTR2MPPnfXYDrwV:-MNcmpIu1QpJgRj_d5yp:b3i5ev9) ([raw file](https://github.com/oscarlab/graphene/blob/4f95892b315f01517e57d83dc313731a7889901c/Documentation/manifest-syntax.rst#L485)):*

Is this true? I don't see any enforcement of this rule in your code.
</blockquote></details>

Not true anymore: an error in `debug_read` is not fatal.

I added a check to the initialization code.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 247 at r4](https://reviewable.io/reviews/oscarlab/graphene/2001#-MNboZVvCQspmMzRDCHL:-MNcqEZgBKxVFLQDX3PB:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/4f95892b315f01517e57d83dc313731a7889901c/Pal/src/host/Linux-SGX/sgx_profile.c#L247)):*

Can we add error checking with "out of memory" debug message?
</blockquote></details>

Done.

dimakuv (approved):
> Reviewed 1 of 11 files at r4, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MO_Oy28CdFA2LDrYJly:-MO_Oy28CdFA2LDrYJlz:b-mpvql) (no related file):*
How does this compare with #2026 now?

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MO_Oy28CdFA2LDrYJly:-MO_PT0-5PxgAu-oZ2Hr:b-s5igkv) (no related file):*

How does this compare with #2026 now?
</blockquote></details>

This PR does not use `perf` so #2026 does not help.

(I have a next version of this profiler, which does use `perf` and has stack information. But I was hoping to get this PR merged first, so that the next PR can be on top of it).

---

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2001)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2001#-MO_Oy28CdFA2LDrYJly:-MOaD_63-93sy95DCVgh:bl34lks) (no related file):*

This PR does not use `perf` so #2026 does not help.

(I have a next version of this profiler, which does use `perf` and has stack information. But I was hoping to get this PR merged first, so that the next PR can be on top of it).
</blockquote></details>

Ok, good.

---

pwmarcz:
Okay... I wanted to wait for this one to be merged, and then make a new PR. But I see that it has non-trivial conflicts with master (manifest parsing code got moved) so I won't bother and make a new PR independently.

Consider this PR superseded by the new one: #2027

PR #1991, opened 2020-11-24
[Pal/Linux-SGX] Allow ocall_mmap_untrusted() with fixed address

dimakuv:
## Description of the changes 

This PR allows `ocall_mmap_untrusted()` to be used with a fixed address (aka `MAP_FIXED`). This can be used to allocate untrusted memory at specific untrusted addresses.

This functionality is currently not used by Linux-SGX PAL but will be required in the IOCTL emulation in future PRs. (Some devices require allocating shared untrusted memory with `MAP_FIXED`.)

## How to test this PR? 

All tests must pass.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1991)

PR #1959, opened 2020-11-12
[Pal/{Linux,Linux-SGX}] Allow communication with host devices

dimakuv:
## Description of the changes 

Previously, Graphene supported only `/dev/tty` (stdin/stdout terminal) device. This PR introduces support for other, arbitrary host devices. The devices must be explicitly allowed in the manifest via a mount point like this:
```
      fs.mount.devkmsg.type = chroot
      fs.mount.devkmsg.path = /dev/kmsg
      fs.mount.devkmsg.uri  = dev:/dev/kmsg
```

Currently only `open/read/write/close/fstat` are supported. Support for IOCTLs and mmaps will be introduced in future PRs.

~~Needs to be rebased on top of #1951.~~

## How to test this PR? 

A simple test that opens `/dev/kmsg` and reads from it is added to LibOS regression tests.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/1959)

PR #2002, opened 2020-11-30
[Pal/{Linux,Linux-SGX}] Allow IOCTLs and mmaps backed by host devices

dimakuv:
This PR is more of a **work-in-progress** PR for **discussion**. Even though it should work just fine, it is a bit controversial.

## Description of the changes 

This PR adds support for `mmap(.., fd=device-fd, ..)` and `ioctl(fd=device-fd, ..)`.

- The `ioctl()` syscalls on device-backed allowed file descriptors are pass-through. On the Linux-SGX PAL, the allowed IOCTL requests must be explicitly allowed in the manifest via the new option `sgx.allowed_ioctls.[id].request`.  Also on the Linux-SGX PAL, the allowed IOCTLs' arguments (typically pointers to complex nested objects) must be explicitly described in the manifest via the new options `sgx.ioctl_structs.[id]` and a corresponding reference `sgx.allowed_ioctls.[id].struct`; see docs for explanation of the IOCTL struct format.

- The `mmap()` syscalls on device-backed allowed file descriptors are pass-through. On the LinuxSGX PAL, all mmapped memory regions reside in untrusted memory, and the enclave logic must be careful when accessing it (typically the app developer must use this mmapped memory region as an encrypted bounce buffer).

Device-backed `ioctl()` and `mmap()` syscalls are insecure by themselves since they only pass the arguments to and from the untrusted memory. It is the responsibility of the app developer to correctly use them for secure communication with the host devices.

~~Must be rebased on top of #1959.~~

Must be rebased on top of #1991.

Must be rebased on top of #2016.

UPDATE: Changed the manifest syntax according to the feedback from maintainers (mainly Michal).

## How to test this PR? 

This PR adds a `device_enclave` LibOS regression test that asks the Intel SGX driver to create a new enclave and add enclave pages. This test stresses the new `mmap()`/`ioctl()` device-backed syscalls. Unfortunately, there are several versions of the Intel SGX driver lying around, so this test was tested only with some versions of the legacy SGX driver (`/dev/isgx`) and the newer DCAP/in-kernel driver (`/dev/sgx/enclave`).

TODO: Other maintainers want to add a FUSE dummy driver, and test against it (instead of the more complicated and volatile use of the Intel SGX driver).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2002)


dimakuv:
For reference, I will copy-paste my comment on the memory-layout TOML syntax used in this PR (to describe nested objects that are deep-copied from/to the enclave):

```c
/*
 * Code below describes the deep-copy syntax in the TOML manifest used for copying complex nested
 * objects out and in the SGX enclave. This syntax is currently used for IOCTL emulation. This
 * syntax is generic enough to describe any memory layout for deep copy.
 *
 * The following example describes the main implementation details:
 *
 *   struct pascal_str { uint8_t len; char str[]; };
 *   struct c_str { char str[]; };
 *   struct root { struct pascal_str* s1; struct c_str* s2; uint64_t s2_len; int8_t x; int8_t y; };
 *
 *   alignas(128) struct root obj;
 *   ioctl(devfd, _IOWR(DEVICE_MAGIC, DEVICE_FUNC, struct root), &obj);
 *
 * The example IOCTL takes as a third argument a pointer to an object of type `struct root` that
 * contains two pointers to other objects (pascal-style string and a C-style string) and embeds two
 * integers `x` and `y`. The two strings reside in separate memory regions in enclave memory. Note
 * that the length of the C-style string is stored in the `s2_len` field of the root object. The
 * `pascal_str` string is an input to the IOCTL, the `c_str` string is both input and output of the
 * IOCTL, and the integers `x` and `y` are both outputs of the IOCTL. Also note that the root
 * object is 128B-aligned (for illustration purposes).
 *
 * The corresponding deep-copy syntax in TOML looks like this:
 *
 *   sgx.ioctl.DEVICE_FUNC.struct = [
 *     { align = 128, ptr = [ {name="pascal-str-len", size=1, type="out"},
 *                            {name="pascal-str", size="pascal-str-len", type="out"} ] },
 *     { ptr = [ {name="c-str", size="c-str-len", type="inout"} ] },
 *     { name = "c-str-len", size = 8, type = "in" },
 *     { size = 2, type = "in" }
 *   ]
 *
 * One can observe the following rules in this TOML syntax:
 * 1. Each separate memory region is represented as a TOML array (`[]`).
 * 2. Each sub-region of one memory region is represented as a TOML table (`{}`).
 * 3. Each sub-region may be a pointer (`ptr`) to another memory region. In this case, the value of
 *    `ptr` is a TOML-array representation of that other memory region. The `ptr` sub-region always
 *    has size of 8B (assuming x86-64) and doesn't have an in/out type.
 * 4. Sub-regions can be fixed-size (like the last sub-region containing two bytes `x` and `y`) or
 *    can be flexible-size (like the two strings). In the latter case, the `size` field contains a
 *    name of a sub-region where the actual size is stored.
 * 5. Sub-regions that store the size of another sub-region must be 1, 2, 4, or 8 bytes in size.
 * 6. Sub-regions may have a name for ease of identification; this is required for "size"
 *    sub-regions but may be omitted for all other kinds of sub-regions.
 * 7. Sub-regions may have one of the three types: "out" to copy contents of the sub-region outside
 *    the enclave to untrusted memory, "in" to copy from untrusted memory to inside the enclave,
 *    "inout" to copy in both directions. Note that pointer sub-regions do not have a type.
 * 8. The first sub-region (and only the first!) may specify the alignment of the memory region.
 *
 * The diagram below shows how this complex object is copied from enclave memory (left side) to
 * untrusted memory (right side). MR stands for "memory region", SR stands for "sub-region". Note
 * how enclave pointers are copied and rewired to point to untrusted memory regions.
 *
 *       struct root (MR1)            |       deep-copied struct (aligned at 128B)
 *      +------------------+          |     +------------------------+
 *  +----+ pascal_str* s1  |     SR1  |  +----+ pascal_str* s1  (MR1)|
 *  |   |                  |          |  |  |                        |
 *  |   |  c_str* s2 +-------+   SR2  |  |  |   c_str* s2 +-------------+
 *  |   |                  | |        |  |  |                        |  |
 *  |   |  uint64_t s2_len | |   SR3  |  |  |   uint64_t s2_len      |  |
 *  |   |                  | |        |  |  |                        |  |
 *  |   |  int8_t x, y     | |   SR4  |  |  |   int8_t x=0, y=0      |  |
 *  |   +------------------+ |        |  |  +------------------------+  |
 *  |                        |        |  +->|   uint8_t len     (MR2)|  |
 *  v (MR2)                  |        |     |                        |  |
 * +-------------+           |        |     |   char str[]           |  |
 * | uint8_t len |           |   SR5  |     +------------------------+  |
 * |             |           |        |     |   char str[]      (MR3)|<-+
 * | char str[]  |           |   SR6  |     +------------------------+
 * +-------------+           |        |
 *                  (MR3)    v        |
 *                +----------+-+      |
 *                | char str[] | SR7  |
 *                +------------+      |
 *
 */
```
```

PR #2015, opened 2020-12-07
[LibOS] Fix following the link in FS subsystem

dimakuv:
## Description of the changes 

Found out that following a link was broken. Looks like it was a typo. Fixing the typo made my rework of `/proc` FS work.

## How to test this PR? 

All tests must pass.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2015)

PR #2016, opened 2020-12-07
[LibOS] Refactor /proc/[tid|pid] logic

### LINKED ISSUE #1387: [LibOS] Targets of links are always assumed to be under chroot
## Description of the problem:

Current logic in Graphene assumes that target file is under `chroot` (actual file system). This misses cases like `/dev/stdout` which is a link to `/proc/self/fd/1` (i.e., Graphene currently fails if target is under pseudo-FS). This is the offending line:

https://github.com/oscarlab/graphene/blob/80030b1157da371251f632e325df71e9454b4e55/LibOS/shim/src/fs/shim_namei.c#L394

## Expected results:

```
[     1] ---- shim_openat(AT_FDCWD,"/dev/stdout",O_WRONLY|O_CREAT|O_TRUNC,0666) = 0
```

## Actual results:

```
[     1] ---- shim_openat(AT_FDCWD,"/dev/stdout",O_WRONLY|O_CREAT|O_TRUNC,0666) = -21
```

## Additional information: 

Generally, the whole FS subsystem must be rewritten in Graphene.

dimakuv:
## Description of the changes 

Graphene emulates `/proc/[tid|pid]` subdirectory using the pseudo-FS logic under `fs/proc/thread.c` and `fs/proc/ipc-thread.c`. This PR refactors these files: name changes, better structure, using string.h functions, small bug fixes. The IPC-thread logic is significantly simplified and partly commented out (it was never used and probably will never be used, but still keeping it just in case).

~~Needs to be rebased on top of #2015 (for the `proc_common` test to work correctly).~~

## How to test this PR? 

The LibOS `proc_common` test is augmented to test more functionality.

Fixes #1387.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2016)


dimakuv:
Jenkins, retest Jenkins-18.04 please (`kill11` LTP test failed; apparently there are some data races/bugs still lurking in Graphene)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2016)* status: 0 of 8 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2016#-MRLv3CP38RH4jHBxGkJ:-MRLv3CQ4rbsRtvJgJfN:b-5jktcb) (no related file):*
FYI: Rebased to the latest master (no fixups were needed).

---

pwmarcz (commented):
I'm on it, need a bit more time to read everything.

FYI, I talked with @mkow and @boryspoplawski and I want to open a bigger discussion of filesystem rewrite and multi-process synchronization soon. But it won't hurt to get this merged already :)

> Reviewed 4 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2016#-:-MTLxv92Afzf4-i4AOe2:bucsuqn)* status: 4 of 8 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2016#-MRLv3CP38RH4jHBxGkJ:-MTM22ZL0IRJG3NHUaXd:b365o6i) (no related file):*

FYI: Rebased to the latest master (no fixups were needed).
</blockquote></details>

Unfortunately master has drifted somewhat since then, there is `/proc/PID/task/` and some smaller changes...

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2016#-MTLlELw9i8J-1STJJaU:-MTLlELw9i8J-1STJJaV:b-l39c6t) (no related file):*
I tested this with `busybox` example, and `./pal_loader busybox ls /proc/1`  works, but `./pal_loader busybox` and then `ls /proc/1` doesn't (only `ls /proc/self`). Is it because the `ls` inside spawns a separate process, and we try to handle that using `ipc-thread.c` code which is stubbed out?)

Another thing is that "symlinks" like `/proc/self` are not really symlinks: `readlink` reports the target, but I think `lstat` claims it's a regular file. Graphene FS doesn't fully recognize the concept of symlinks, so probably that's not something to fix here.

---
___
*[LibOS/shim/include/shim_fs.h, line 486 at r1](https://reviewable.io/reviews/oscarlab/graphene/2016#-MTLy45n2WEF7kw1hhZS:-MTLy45n2WEF7kw1hhZT:bakmllc) ([raw file](https://github.com/oscarlab/graphene/blob/dd56fbe267e97bee8b7248284466dfee4dfdd8ff/LibOS/shim/include/shim_fs.h#L486)):*
> ```objc
> 
> struct pseudo_name_ops {
>     int (*match_path)(const char* relpath);
> ```

This is confusing... If I understand correctly, if this structure is attached to a directory `foo/bar`, and we look up a path such as `foo/bar/baz/quux`, then

* we call `match_path` with `"foo/bar/baz/quux"`,
* `"foo/bar"` is a prefix that we already validated, but the function may use it for determining context,
* `"baz"` is the part that we are asking about, and previously was passed to `match_name`,
* `"quux"` is a suffix that we disregard right now.

How about a function like that instead:

```c
int (*match_name)(const char* relpath, const char* name);
```

It would be called with a prefix and current part, i.e. `match_name("foo/bar", "baz")` in my example.

This way, it would be clear that `match_name(relpath, name)` and `list_dirents(relpath)` are expected to be consistent, and in principle you could implement `match_name` using `list_dirents` (event though the actual implementation should be more efficient).

Correct me if I'm wrong, it seems to me that the implementation would be clearer but I haven't looked very deeply.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2016)* status: 4 of 8 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2016#-MRLv3CP38RH4jHBxGkJ:-MTaBVD65c7VK3yP8x_u:b67hprk) (no related file):*

Unfortunately master has drifted somewhat since then, there is `/proc/PID/task/` and some smaller changes...
</blockquote></details>

I'm not sure how to proceed with this PR. It got quite stale, and we also added another pseudo-FS `/sys/` now, using the old-style code that I'll need to change as well.

I don't think it is that hard to rebase to current master, but should I do it if we're re-writing the FS subsystem completely? @pwmarcz I'll rebase if you recommend to merge this PR before your FS changes.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2016#-MTLlELw9i8J-1STJJaU:-MTaBLFJAk_d4UsqC93O:b-ao0jzf) (no related file):*

I tested this with `busybox` example, and `./pal_loader busybox ls /proc/1`  works, but `./pal_loader busybox` and then `ls /proc/1` doesn't (only `ls /proc/self`). Is it because the `ls` inside spawns a separate process, and we try to handle that using `ipc-thread.c` code which is stubbed out?)

Another thing is that "symlinks" like `/proc/self` are not really symlinks: `readlink` reports the target, but I think `lstat` claims it's a regular file. Graphene FS doesn't fully recognize the concept of symlinks, so probably that's not something to fix here.

</blockquote></details>

Yes, I think the second example doesn't work because there is no shared FS between the two separate processes (I stubbed it out).

---
___
*[LibOS/shim/include/shim_fs.h, line 486 at r1](https://reviewable.io/reviews/oscarlab/graphene/2016#-MTLy45n2WEF7kw1hhZS:-MTaClQ35I2EZFnHfDBR:bszglxm) ([raw file](https://github.com/oscarlab/graphene/blob/dd56fbe267e97bee8b7248284466dfee4dfdd8ff/LibOS/shim/include/shim_fs.h#L486)):*

This is confusing... If I understand correctly, if this structure is attached to a directory `foo/bar`, and we look up a path such as `foo/bar/baz/quux`, then

* we call `match_path` with `"foo/bar/baz/quux"`,
* `"foo/bar"` is a prefix that we already validated, but the function may use it for determining context,
* `"baz"` is the part that we are asking about, and previously was passed to `match_name`,
* `"quux"` is a suffix that we disregard right now.

How about a function like that instead:

```c
int (*match_name)(const char* relpath, const char* name);
```

It would be called with a prefix and current part, i.e. `match_name("foo/bar", "baz")` in my example.

This way, it would be clear that `match_name(relpath, name)` and `list_dirents(relpath)` are expected to be consistent, and in principle you could implement `match_name` using `list_dirents` (event though the actual implementation should be more efficient).

Correct me if I'm wrong, it seems to me that the implementation would be clearer but I haven't looked very deeply.
</blockquote></details>

Yes, you are correct. I can change to your suggested function signature, it will be clearer. I just followed the current code and tried to keep the interfaces roughly the same but with better names and types.

I can change the func as you suggested. But only if you think we should merge this PR first, instead of burning all this code and writing something new from scratch.

dimakuv:
Closing this PR as it was superseded by Pawel's work on FS refactoring. @pwmarcz There is nothing useful left in this PR, right? If something can be cherry-picked there, please tell and I can create a new PR.

pwmarcz:
Thanks. It looks like we don't need anything - I thought about using some of the code while refactoring, but I ended up rewriting these files to not use any caching for simplicity.

PR #2020, opened 2020-12-09
[Pal/Linux-SGX] Fix type cast issue during async signal handling and also retry OCALLs on EINTR

dimakuv:
## Description of the changes 

Untrusted Linux-SGX PAL handles host-level asynchronous signals by emulating the interrupt (-EINTR) of the pending OCALL. Unfortunately, there was a type cast issue such that int32_t -EINTR (`-4`) was casted to a positive uint64_t and then OCALL consumed this positive number instead of erroring out on -EINTR. This PR adds explicit type casting to fix this bug.

After the above was fixed (in the first commit of this PR), I debugged a small data race and fixed it. It goes like this. Graphene internal logic as well as application logic relies on most OCALLs being "uninterruptible", e.g., `recv()` OCALL is assumed to be interruptible but `mmap()` OCALL is assumed to be uninterruptible. At the same time, Linux-SGX PAL always injects `-EINTR` in OCALLs when an async signal arrives from host OS, no matter what OCALL. This led to spontaneous failures with debug message "ocall returned -4". The second commit of this PR adds retrying on all "uninterruptible" OCALLs.

## How to test this PR? 

All tests must pass. This bug was found on a large workload that spawned several child processes. Sometimes it so happened that one child process finished and sent SIGCHLD signal to the main (parent) process, and the main process simultaneously was sending a checkpoint to another spawned child. This resulted in `ocall_read()` being interrupted and then returning a ridiculous big number instead of `-EINTR`, and the whole application crashed.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)

### DIFF
--- a/Pal/include/arch/x86_64/Linux/ucontext.h
+++ b/Pal/include/arch/x86_64/Linux/ucontext.h
@@ -157,21 +157,15 @@ static inline uint64_t pal_ucontext_get_ip(ucontext_t* uc) {
     return uc->uc_mcontext.gregs[REG_RIP];
 }
 
-static inline void pal_ucontext_set_function_parameters(ucontext_t* uc, void* func, size_t count,
-                                                        ...) {
+static inline void pal_ucontext_set_function_parameters(ucontext_t* uc, void* func,
+                                                        size_t func_args_num, greg_t* func_args) {
     const unsigned int param_regs[] = {REG_RDI, REG_RSI, REG_RDX, REG_RCX};
-    va_list ap;
 
-    assert(count <= ARRAY_SIZE(param_regs));
+    assert(func_args_num <= ARRAY_SIZE(param_regs));
 
     uc->uc_mcontext.gregs[REG_RIP] = (greg_t)func;
-
-    va_start(ap, count);
-
-    for (size_t i = 0; i < count; i++)
-        uc->uc_mcontext.gregs[param_regs[i]] = va_arg(ap, greg_t);
-
-    va_end(ap);
+    for (size_t i = 0; i < func_args_num; i++)
+        uc->uc_mcontext.gregs[param_regs[i]] = func_args[i];
 }
 
 #else /* __WORDSIZE == 32 */
--- a/Pal/src/host/Linux-SGX/enclave_ocalls.c
+++ b/Pal/src/host/Linux-SGX/enclave_ocalls.c
@@ -1,5 +1,21 @@
 /*
  * This is for enclave to make ocalls to untrusted runtime.
+ *
+ * Most ocall implementations retry the host-level operations on -EINTR, except for a few ocalls
+ * that are expected to be able to return -EINTR (read, write, recv, send, accept, connect, sleep,
+ * futex). The -EINTR error happens when an async signal arrives from the host OS, with two
+ * sub-cases: (a) signal arrives during a slow host-level syscall or (b) signal arrives during other
+ * untrusted-PAL code execution. In both cases, the untrusted-PAL signal handler injects -EINTR and
+ * forces an ocall return. To prevent ocalls from returning -EINTR to unsuspecting LibOS/user app,
+ * here we retry a host-level syscall. In some cases, this may lead to FD leaks or incorrect
+ * semantics (e.g., a retried open() syscall may have succeeded the first time, but a signal arrived
+ * right-after this syscall and forced -EINTR, thus leaving an FD from the first try open and
+ * leaking). See also `man 7 signal` and `sgx_exception.c: handle_async_signal()`.
+ *
+ * FIXME: Ideally, the untrusted-PAL signal handler must inspect the interrupted RIP and unwind/fix
+ *        the untrusted state and decide whether to inject -EINTR or report success (e.g., if a
+ *        signal arrives right-after an open() syscall, the signal handler must update ocall return
+ *        values and report success instead of injecting -EINTR).
  */
 
 #include "enclave_ocalls.h"
@@ -158,7 +174,9 @@ int ocall_mmap_untrusted(int fd, uint64_t offset, size_t size, unsigned short pr
     WRITE_ONCE(ms->ms_size, size);
     WRITE_ONCE(ms->ms_prot, prot);
 
-    retval = sgx_exitless_ocall(OCALL_MMAP_UNTRUSTED, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_MMAP_UNTRUSTED, ms);
+    } while (retval == -EINTR);
 
     if (!retval) {
         if (!sgx_copy_ptr_to_enclave(mem, READ_ONCE(ms->ms_mem), size)) {
@@ -190,7 +208,9 @@ int ocall_munmap_untrusted(const void* mem, size_t size) {
     WRITE_ONCE(ms->ms_mem, mem);
     WRITE_ONCE(ms->ms_size, size);
 
-    retval = sgx_exitless_ocall(OCALL_MUNMAP_UNTRUSTED, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_MUNMAP_UNTRUSTED, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -282,7 +302,9 @@ int ocall_cpuid(unsigned int leaf, unsigned int subleaf, unsigned int values[4])
     WRITE_ONCE(ms->ms_leaf, leaf);
     WRITE_ONCE(ms->ms_subleaf, subleaf);
 
-    retval = bypass_exitless ? sgx_ocall(OCALL_CPUID, ms) : sgx_exitless_ocall(OCALL_CPUID, ms);
+    do {
+        retval = bypass_exitless ? sgx_ocall(OCALL_CPUID, ms) : sgx_exitless_ocall(OCALL_CPUID, ms);
+    } while (retval == -EINTR);
 
     if (!retval) {
         values[0] = READ_ONCE(ms->ms_values[0]);
@@ -316,7 +338,9 @@ int ocall_open(const char* pathname, int flags, unsigned short mode) {
     }
     WRITE_ONCE(ms->ms_pathname, untrusted_pathname);
 
-    retval = sgx_exitless_ocall(OCALL_OPEN, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_OPEN, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -335,6 +359,7 @@ int ocall_close(int fd) {
 
     WRITE_ONCE(ms->ms_fd, fd);
 
+    /* note that close() must not be retried on -EINTR, see e.g. https://lwn.net/Articles/576478 */
     retval = sgx_exitless_ocall(OCALL_CLOSE, ms);
 
     sgx_reset_ustack(old_ustack);
@@ -577,7 +602,9 @@ int ocall_fstat(int fd, struct stat* buf) {
 
     WRITE_ONCE(ms->ms_fd, fd);
 
-    retval = sgx_exitless_ocall(OCALL_FSTAT, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_FSTAT, ms);
+    } while (retval == -EINTR);
 
     if (!retval) {
         memcpy(buf, &ms->ms_stat, sizeof(struct stat));
@@ -600,7 +627,9 @@ int ocall_fionread(int fd) {
 
     WRITE_ONCE(ms->ms_fd, fd);
 
-    retval = sgx_exitless_ocall(OCALL_FIONREAD, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_FIONREAD, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -620,7 +649,9 @@ int ocall_fsetnonblock(int fd, int nonblocking) {
     WRITE_ONCE(ms->ms_fd, fd);
     WRITE_ONCE(ms->ms_nonblocking, nonblocking);
 
-    retval = sgx_exitless_ocall(OCALL_FSETNONBLOCK, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_FSETNONBLOCK, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -640,7 +671,9 @@ int ocall_fchmod(int fd, unsigned short mode) {
     WRITE_ONCE(ms->ms_fd, fd);
     WRITE_ONCE(ms->ms_mode, mode);
 
-    retval = sgx_exitless_ocall(OCALL_FCHMOD, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_FCHMOD, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -659,7 +692,9 @@ int ocall_fsync(int fd) {
 
     WRITE_ONCE(ms->ms_fd, fd);
 
-    retval = sgx_exitless_ocall(OCALL_FSYNC, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_FSYNC, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -679,7 +714,9 @@ int ocall_ftruncate(int fd, uint64_t length) {
     WRITE_ONCE(ms->ms_fd, fd);
     WRITE_ONCE(ms->ms_length, length);
 
-    retval = sgx_exitless_ocall(OCALL_FTRUNCATE, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_FTRUNCATE, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -705,7 +742,9 @@ int ocall_mkdir(const char* pathname, unsigned short mode) {
     }
     WRITE_ONCE(ms->ms_pathname, untrusted_pathname);
 
-    retval = sgx_exitless_ocall(OCALL_MKDIR, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_MKDIR, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -731,7 +770,9 @@ int ocall_getdents(int fd, struct linux_dirent64* dirp, size_t dirp_size) {
     }
     WRITE_ONCE(ms->ms_dirp, untrusted_dirp);
 
-    retval = sgx_exitless_ocall(OCALL_GETDENTS, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_GETDENTS, ms);
+    } while (retval == -EINTR);
 
     if (retval > 0) {
         size_t size = (size_t)retval;
@@ -767,12 +808,21 @@ out:
 }
 
 int ocall_resume_thread(void* tcs) {
-    return sgx_exitless_ocall(OCALL_RESUME_THREAD, tcs);
+    int retval = 0;
+    do {
+        retval = sgx_exitless_ocall(OCALL_RESUME_THREAD, tcs);
+    } while (retval == -EINTR);
+    return retval;
 }
 
 int ocall_clone_thread(void) {
+    int retval = 0;
     void* dummy = NULL;
-    return sgx_exitless_ocall(OCALL_CLONE_THREAD, dummy);
+    do {
+        /* FIXME: if there was an EINTR, there may be an untrusted thread left over */
+        retval = sgx_exitless_ocall(OCALL_CLONE_THREAD, dummy);
+    } while (retval == -EINTR);
+    return retval;
 }
 
 int ocall_create_process(const char* uri, size_t nargs, const char** args, int* stream_fd,
@@ -807,7 +857,10 @@ int ocall_create_process(const char* uri, size_t nargs, const char** args, int* 
         WRITE_ONCE(ms->ms_args[i], unstrusted_arg);
     }
 
-    retval = sgx_exitless_ocall(OCALL_CREATE_PROCESS, ms);
+    do {
+        /* FIXME: if there was an EINTR, there may be an untrusted process left over */
+        retval = sgx_exitless_ocall(OCALL_CREATE_PROCESS, ms);
+    } while (retval == -EINTR);
 
     if (!retval) {
         if (pid)
@@ -861,7 +914,9 @@ int ocall_socketpair(int domain, int type, int protocol, int sockfds[2]) {
     WRITE_ONCE(ms->ms_type, type);
     WRITE_ONCE(ms->ms_protocol, protocol);
 
-    retval = sgx_exitless_ocall(OCALL_SOCKETPAIR, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_SOCKETPAIR, ms);
+    } while (retval == -EINTR);
 
     if (!retval) {
         sockfds[0] = READ_ONCE(ms->ms_sockfds[0]);
@@ -897,7 +952,9 @@ int ocall_listen(int domain, int type, int protocol, int ipv6_v6only, struct soc
     }
     WRITE_ONCE(ms->ms_addr, untrusted_addr);
 
-    retval = sgx_exitless_ocall(OCALL_LISTEN, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_LISTEN, ms);
+    } while (retval == -EINTR);
 
     if (retval >= 0) {
         if (addr && len) {
@@ -1192,7 +1249,9 @@ int ocall_setsockopt(int sockfd, int level, int optname, const void* optval, siz
         WRITE_ONCE(ms->ms_optval, untrusted_optval);
     }
 
-    retval = sgx_exitless_ocall(OCALL_SETSOCKOPT, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_SETSOCKOPT, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -1212,7 +1271,9 @@ int ocall_shutdown(int sockfd, int how) {
     WRITE_ONCE(ms->ms_sockfd, sockfd);
     WRITE_ONCE(ms->ms_how, how);
 
-    retval = sgx_exitless_ocall(OCALL_SHUTDOWN, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_SHUTDOWN, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -1332,7 +1393,9 @@ int ocall_rename(const char* oldpath, const char* newpath) {
     WRITE_ONCE(ms->ms_oldpath, untrusted_oldpath);
     WRITE_ONCE(ms->ms_newpath, untrusted_newpath);
 
-    retval = sgx_exitless_ocall(OCALL_RENAME, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_RENAME, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -1357,7 +1420,9 @@ int ocall_delete(const char* pathname) {
     }
     WRITE_ONCE(ms->ms_pathname, untrusted_pathname);
 
-    retval = sgx_exitless_ocall(OCALL_DELETE, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_DELETE, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -1375,7 +1440,10 @@ int ocall_update_debugger(struct debug_map* _Atomic* debug_map) {
     }
 
     WRITE_ONCE(ms->ms_debug_map, debug_map);
-    retval = sgx_exitless_ocall(OCALL_UPDATE_DEBUGGER, ms);
+
+    do {
+        retval = sgx_exitless_ocall(OCALL_UPDATE_DEBUGGER, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -1423,7 +1491,9 @@ int ocall_eventfd(unsigned int initval, int flags) {
     WRITE_ONCE(ms->ms_initval, initval);
     WRITE_ONCE(ms->ms_flags, flags);
 
-    retval = sgx_exitless_ocall(OCALL_EVENTFD, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_EVENTFD, ms);
+    } while (retval == -EINTR);
 
     sgx_reset_ustack(old_ustack);
     return retval;
@@ -1454,7 +1524,9 @@ int ocall_get_quote(const sgx_spid_t* spid, bool linkable, const sgx_report_t* r
     memcpy(&ms->ms_nonce, nonce, sizeof(*nonce));
     WRITE_ONCE(ms->ms_linkable, linkable);
 
-    retval = sgx_exitless_ocall(OCALL_GET_QUOTE, ms);
+    do {
+        retval = sgx_exitless_ocall(OCALL_GET_QUOTE, ms);
+    } while (retval == -EINTR);
 
     if (!IS_ERR(retval)) {
         ms_ocall_get_quote_t ms_copied;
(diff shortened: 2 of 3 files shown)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 178 at r1](https://reviewable.io/reviews/oscarlab/graphene/2020#-MO7yTxa2HlK2WS1syAo:-MO7yTxbCJHX6P2BCIMj:bee50tm) ([raw file](https://github.com/oscarlab/graphene/blob/4ae33fead790076669f0069888a11e9b43cbfdab/Pal/src/host/Linux-SGX/sgx_exception.c#L178)):*
> ```C
>     /* signal arrived while in untrusted PAL code (during syscall handling), emulate as if syscall
>      * was interrupted by calling sgx_entry_return(syscall_return_value=-EINTR, event); NOTE: must
>      * always explicitly type cast arguments to the `greg_t` type! */
> ```

please explain why (because `pal_ucontext_set_function_parameters` is variadic and reads this argument as `greg_t`, not `int`)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @dimakuv)

yamahata (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 178 at r1](https://reviewable.io/reviews/oscarlab/graphene/2020#-MO7yTxa2HlK2WS1syAo:-MO8QtIa5pwWEpBnRhBy:b-xzfvki) ([raw file](https://github.com/oscarlab/graphene/blob/4ae33fead790076669f0069888a11e9b43cbfdab/Pal/src/host/Linux-SGX/sgx_exception.c#L178)):*

please explain why (because `pal_ucontext_set_function_parameters` is variadic and reads this argument as `greg_t`, not `int`)
</blockquote></details>

+1. This is the only use.
Maybe we can have explicit 4 parameters with size.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 178 at r1](https://reviewable.io/reviews/oscarlab/graphene/2020#-MO7yTxa2HlK2WS1syAo:-MO8YVU3ElZf-T1qsxX8:b4n7jt4) ([raw file](https://github.com/oscarlab/graphene/blob/4ae33fead790076669f0069888a11e9b43cbfdab/Pal/src/host/Linux-SGX/sgx_exception.c#L178)):*

+1. This is the only use.
Maybe we can have explicit 4 parameters with size.
</blockquote></details>

+1 to what @yamahata said: please change it to non-variadic function (and change name accordingly).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 178 at r1](https://reviewable.io/reviews/oscarlab/graphene/2020#-MO7yTxa2HlK2WS1syAo:-MOB1Fju64NagMAk4xzu:bobz2o5) ([raw file](https://github.com/oscarlab/graphene/blob/4ae33fead790076669f0069888a11e9b43cbfdab/Pal/src/host/Linux-SGX/sgx_exception.c#L178)):*

+1 to what @yamahata said: please change it to non-variadic function (and change name accordingly).
</blockquote></details>

Done. I opted for a fixed-size array instead of for explicit arguments. And I don't know how to change the function name. Please suggest.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: 0 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOB1_c3ClWNFWpOx32e:-MOB1_c3ClWNFWpOx32f:bx8ra9s) (no related file):*
@stefanberger Does this change make sense to you? Is it fine for your purposes?

---

boryspoplawski (changes requested):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Pal/include/arch/x86_64/Linux/ucontext.h, line 161 at r3](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOBjj5uBDYOVnFWSGqV:-MOBjj5uBDYOVnFWSGqW:b-prx4qv) ([raw file](https://github.com/oscarlab/graphene/blob/5cfcc323c3809d30595460664e1dc3cec9b2ab07/Pal/include/arch/x86_64/Linux/ucontext.h#L161)):*
> ```objc
> 
> static inline void pal_ucontext_set_function_parameters(ucontext_t* uc, void* func, size_t count,
>                                                         greg_t func_args[4]) {
> ```

Please change to `greg_t*`. It's the same thing, but since we have a `count` parameter it will look less confusing.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: 0 of 2 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[Pal/include/arch/x86_64/Linux/ucontext.h, line 161 at r3](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOBjj5uBDYOVnFWSGqV:-MOC6kSk3dyVTta8_gGz:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/5cfcc323c3809d30595460664e1dc3cec9b2ab07/Pal/include/arch/x86_64/Linux/ucontext.h#L161)):*

Please change to `greg_t*`. It's the same thing, but since we have a `count` parameter it will look less confusing.
</blockquote></details>

Done.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

stefanberger:
> @stefanberger Does this change make sense to you? Is it fine for your purposes?

[ I cannot reply via reviewable due to some 3rd party permission problem that oscarlabs has to resolve ]

@dimakuv Are you referring to ppc64 work? I don't see a problem with this.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 181 at r4](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOXlpLZ16MbM46MgRdO:-MOXlpLZ16MbM46MgRdP:b-jgobnn) ([raw file](https://github.com/oscarlab/graphene/blob/df51e9722204d31d7e4291fb201e7df78f72cbf2/Pal/src/host/Linux-SGX/sgx_exception.c#L181)):*
> ```
> /*count=*/
> ```

It's not `count` anymore.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOB1_c3ClWNFWpOx32e:-MOXpa7yCqY27rMcQK1l:bn3mdlt) (no related file):*
> [ I cannot reply via reviewable due to some 3rd party permission problem that oscarlabs has to resolve ]

Should be fixed now, sorry.

@dimakuv FYI (if you missed it): Stefan's response from GitHub: "@dimakuv Are you referring to ppc64 work? I don't see a problem with this.".

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 181 at r4](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOXlpLZ16MbM46MgRdO:-MO_Kgzz1nbsqlycGg17:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/df51e9722204d31d7e4291fb201e7df78f72cbf2/Pal/src/host/Linux-SGX/sgx_exception.c#L181)):*

> ```
> /*count=*/
> ```

It's not `count` anymore.
</blockquote></details>

Done.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: 1 of 2 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2020#-MO_ZFKI6eZOCRsizMa7:-MO_ZFKI6eZOCRsizMa8:bk1g00t) (no related file):*
Blocking myself. Need to debug this error of `pipe_ocloexec`:
```
Standard Output
pipe_ocloexec: child received an unexpected signal: 11

Standard Error
WARNING: Using insecure argv source. Don't use this configuration in production!
Accessing file:../../../../Runtime/ld-linux-x86-64.so.2 is denied (Operation interrupted). This file is not trusted or allowed. Trusted files should be regular files (seekable).
```

Looks like the child process got interrupted by some signal during startup, when the child is still reading startup files.

---

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion

yamahata (approved):
> Reviewed 2 of 2 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: 2 of 3 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2020#-MO_ZFKI6eZOCRsizMa7:-MOb2NnV72W4Telq9-pb:b-9u1yfi) (no related file):*

Blocking myself. Need to debug this error of `pipe_ocloexec`:
```
Standard Output
pipe_ocloexec: child received an unexpected signal: 11

Standard Error
WARNING: Using insecure argv source. Don't use this configuration in production!
Accessing file:../../../../Runtime/ld-linux-x86-64.so.2 is denied (Operation interrupted). This file is not trusted or allowed. Trusted files should be regular files (seekable).
```

Looks like the child process got interrupted by some signal during startup, when the child is still reading startup files.
</blockquote></details>

Done. Please see the second commit (and file `enclave_ocalls.c`).

---

boryspoplawski (changes requested):
> Reviewed 1 of 1 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1454 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOb5hKN7nYe54Ct4J1x:-MOb5hKN7nYe54Ct4J1y:bkary6j) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1454)):*
> ```C
> 
>     do {
>         retval = sgx_exitless_ocall(OCALL_EVENTFD, ms);
> ```

Why don't we allow `EINTR` here?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1454 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOb5hKN7nYe54Ct4J1x:-MOb7vfQ3gp_CKoKKCp3:bwzkr8c) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1454)):*

Why don't we allow `EINTR` here?
</blockquote></details>

This OCALL issues `INLINE_SYSCALL(eventfd2, 2, ms->ms_initval, ms->ms_flags);`. I don't see any mention of EINTR as a possible return value in `man 2 eventfd2`.

Why would we want to allow `-EINTR` here? It's just a "create an FD of type eventfd" syscall.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1454 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOb5hKN7nYe54Ct4J1x:-MObHb079pZHiuY3TXAN:b1rhzw1) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1454)):*

This OCALL issues `INLINE_SYSCALL(eventfd2, 2, ms->ms_initval, ms->ms_flags);`. I don't see any mention of EINTR as a possible return value in `man 2 eventfd2`.

Why would we want to allow `-EINTR` here? It's just a "create an FD of type eventfd" syscall.
</blockquote></details>

Yeah, my bad.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 326 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MObNSHABUzcBnZlsbb8:-MObNSHABUzcBnZlsbb9:bbt17v8) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L326)):*
> ```C
> 
>     /* this open() ocall is used only for files and devices (never to open pipes), so it must never
>      * return -EINTR (which may happen when untrusted PAL receives an async signal from host OS) */
> ```

can we assert this somewhere, or adding it is not easy? looks like an assumption which can be easily broken in the future and go unnoticed in review

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 326 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MObNSHABUzcBnZlsbb8:-MObR6G79fq8fju7Tvgw:bqv2hlr) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L326)):*

can we assert this somewhere, or adding it is not easy? looks like an assumption which can be easily broken in the future and go unnoticed in review
</blockquote></details>

I can't figure out a way how to assert this. There is nothing in the arguments here we can assert on...

Yeah, it's a tricky one. Not sure what we could do here.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 326 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MObNSHABUzcBnZlsbb8:-MObYVfF9VcHatLFWkO_:bu13bpb) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L326)):*

I can't figure out a way how to assert this. There is nothing in the arguments here we can assert on...

Yeah, it's a tricky one. Not sure what we could do here.
</blockquote></details>

I will change the name to `ocall_open_uninterruptible()`. Or even `ocall_open_with_retry()`.

yamahata (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 2 unresolved discussions (waiting on @dimakuv and @mkow)

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 349 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MObVedNAI-H_6-5TTKW:-MObVedNAI-H_6-5TTKX:b3wxnwj) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L349)):*
> ```C
> 
>     do {
>         retval = sgx_exitless_ocall(OCALL_CLOSE, ms);
> ```

close case is subtle. Please don't retry on close case.
For reference, https://lwn.net/Articles/576478/

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 2 unresolved discussions (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOB1_c3ClWNFWpOx32e:-MOMcnZa62wjyDqTRUP7:b2464he) (no related file):*

@stefanberger Does this change make sense to you? Is it fine for your purposes?
</blockquote></details>

@dimakuv Are you referring to ppc64 work? I don't see a problem with this (unless you point one out to me).

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: 2 of 10 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOB1_c3ClWNFWpOx32e:-MOed3sb7hhG5c-wKPfe:bvu3k7v) (no related file):*

@dimakuv Are you referring to ppc64 work? I don't see a problem with this (unless you point one out to me).
</blockquote></details>

Thanks, @stefanberger! Yes, I meant your ppc64 work. Thanks for acking that this PR is ok.

---
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 326 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MObNSHABUzcBnZlsbb8:-MOeicuq-nwWQeOeTnQH:b-5rz4d9) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L326)):*

I will change the name to `ocall_open_uninterruptible()`. Or even `ocall_open_with_retry()`.
</blockquote></details>

Changed to `ocall_open_with_retry()` because we already have a similar name in LibOS (`object_wait_with_retry()`).
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 349 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MObVedNAI-H_6-5TTKW:-MOedxgpCAzc4pKnGqyl:b-5ipav6) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L349)):*

close case is subtle. Please don't retry on close case.
For reference, https://lwn.net/Articles/576478/

</blockquote></details>

Done. Thanks for info!

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: 2 of 10 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOf0SY10papp_IE2mq7:-MOf0SY21W6XHEKL5g6Q:be311i3) (no related file):*
Blocking myself: during final rebase, fix typo: `uninterruptable` -> `uninterruptible` in the commit message.

---

boryspoplawski (changes requested):
> Reviewed 8 of 8 files at r7.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @yamahata)

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 349 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MObVedNAI-H_6-5TTKW:-MOfhsDk2GG1-fLhs8xb:b-ap6wjc) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L349)):*

Done. Thanks for info!
</blockquote></details>

Otoh there is also `EINTR` which gets artificially injected (is *not* returned by actual syscall on host) on which we should retry ...

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @yamahata)

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 349 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MObVedNAI-H_6-5TTKW:-MOfnMObDA2uAxSHcIyG:blhttm3) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L349)):*

Otoh there is also `EINTR` which gets artificially injected (is *not* returned by actual syscall on host) on which we should retry ...
</blockquote></details>

@boryspoplawski True. But this is a rare case and the current code has an FD leak, whereas retrying (as in my previous version of code) could break Graphene process (?). I still think that we shouldn't retry per Isaku's suggestion.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow and @yamahata)

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 349 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MObVedNAI-H_6-5TTKW:-MOfvpJnCMTF81h3d6XW:bhpifba) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L349)):*

@boryspoplawski True. But this is a rare case and the current code has an FD leak, whereas retrying (as in my previous version of code) could break Graphene process (?). I still think that we shouldn't retry per Isaku's suggestion.
</blockquote></details>

I guess, you are right, let's leave it as it is.
Some notes:
- I think we should not just inject `EINTR`, but should be able to distinguish interrupted enclave<->untrusted transition from host syscall returning `EINTR`, but that would probably require heavy redesign.
- I also think that `close` returning `EINTR` is much more rare (idk when it could happen, nfs failure or something) than us getting this artificial `EINTR` from Graphene code ...

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow and @yamahata)

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 349 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MObVedNAI-H_6-5TTKW:-MOfzNgO1n1m2OCRT8Aj:b-coijo1) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L349)):*

I guess, you are right, let's leave it as it is.
Some notes:
- I think we should not just inject `EINTR`, but should be able to distinguish interrupted enclave<->untrusted transition from host syscall returning `EINTR`, but that would probably require heavy redesign.
- I also think that `close` returning `EINTR` is much more rare (idk when it could happen, nfs failure or something) than us getting this artificial `EINTR` from Graphene code ...
</blockquote></details>

@boryspoplawski I agree with both your notes. The original idea (or hmmm... wish?) was to "continue/fix the untrusted logic of each OCALL", as documented here: https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/sgx_exception.c#L178-L179

For example, for the `sgx_ocall_close()`, we would compare interrupted-at-RIP with "RIP-before-host-close-syscall" and "RIP-after-host-close-syscall" and based on the case, either perform the `close()` syscall again or inject `RAX = 0` to kinda "roll RIP forward to good state".

This is quite heavy to implement though...

yamahata (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 349 at r6](https://reviewable.io/reviews/oscarlab/graphene/2020#-MObVedNAI-H_6-5TTKW:-MOgdAT586dMmdbNAndp:b-y4uoh3) ([raw file](https://github.com/oscarlab/graphene/blob/8bc976a1645a63d392fb7f581ffe106f9a0b6e9d/Pal/src/host/Linux-SGX/enclave_ocalls.c#L349)):*

@boryspoplawski I agree with both your notes. The original idea (or hmmm... wish?) was to "continue/fix the untrusted logic of each OCALL", as documented here: https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/sgx_exception.c#L178-L179

For example, for the `sgx_ocall_close()`, we would compare interrupted-at-RIP with "RIP-before-host-close-syscall" and "RIP-after-host-close-syscall" and based on the case, either perform the `close()` syscall again or inject `RAX = 0` to kinda "roll RIP forward to good state".

This is quite heavy to implement though...
</blockquote></details>

I agree that we should able to distingush host system call error code and artificial injection and that it would require redesign. (or even we should stop eintr injection)
For now I unblock this.

mkow (changes requested):
> Reviewed 8 of 8 files at r7.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/db_files.c, line 592 at r7](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOgo35U2KdN4XUvLM2g:-MOgo35U2KdN4XUvLM2h:b-sxoqdx) ([raw file](https://github.com/oscarlab/graphene/blob/a4cbf2d30082d0b2f80e75968283f07cfdefca61/Pal/src/host/Linux-SGX/db_files.c#L592)):*
> ```C
>     /* open the file with O_NONBLOCK to avoid blocking the current thread if it is actually a FIFO
>      * pipe; O_NONBLOCK will be reset below if it is a regular file */
>     int fd = ocall_open_with_retry(uri, O_NONBLOCK, /*mode=*/0);
> ```

Umm, this callsites looks now like a bug (see the comment above). The semantic of `ocall_open_with_retry` is really weird now - could you update the comment above and describe the peculiarity?
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 325 at r7](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOgoSqn7VHPi1IwHvR9:-MOgoSqn7VHPi1IwHvRA:b-pex77v) ([raw file](https://github.com/oscarlab/graphene/blob/a4cbf2d30082d0b2f80e75968283f07cfdefca61/Pal/src/host/Linux-SGX/enclave_ocalls.c#L325)):*
> ```C
>     WRITE_ONCE(ms->ms_pathname, untrusted_pathname);
> 
>     /* this ocall is used only for files and devices (never to open pipes), so it must never return
> ```

This stands contrary to the callsite I marked above.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: 2 of 3 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[Pal/src/host/Linux-SGX/db_files.c, line 592 at r7](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOgo35U2KdN4XUvLM2g:-MOjyjbs1idPl1pbZjsv:bkae3op) ([raw file](https://github.com/oscarlab/graphene/blob/a4cbf2d30082d0b2f80e75968283f07cfdefca61/Pal/src/host/Linux-SGX/db_files.c#L592)):*

Umm, this callsites looks now like a bug (see the comment above). The semantic of `ocall_open_with_retry` is really weird now - could you update the comment above and describe the peculiarity?
</blockquote></details>

Done. Reverted all these callsites. I don't see a particular issue here (EINTR and non-blocking are two orthogonal things), but I can see how the naming became confusing. I reverted to having a "normal" ocall name and explaining the situation in a top-level comment.
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 325 at r7](https://reviewable.io/reviews/oscarlab/graphene/2020#-MOgoSqn7VHPi1IwHvR9:-MOjySK16SuVj4o55veJ:b-gwtr2g) ([raw file](https://github.com/oscarlab/graphene/blob/a4cbf2d30082d0b2f80e75968283f07cfdefca61/Pal/src/host/Linux-SGX/enclave_ocalls.c#L325)):*

This stands contrary to the callsite I marked above.
</blockquote></details>

Done. I reverted the rename `ocall_open_with_retry()` back to `ocall_open()` and removed this comment. Now I have a top-level comment at the beginning of this file explaining the retry logic in these ocalls. This should cover everything and explain the current situation with async signals and EINTR.

boryspoplawski (approved):
> Reviewed 2 of 8 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (dismissed):
> Reviewed 8 of 8 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners

dimakuv:
Jenkins, retest Jenkins-18.04 please (`writev07` LTP test failed, unrelated since it was not under SGX)

mkow (approved):
> Reviewed 1 of 4 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: :shipit: complete! all files reviewed, all discussions resolved

boryspoplawski (approved):
> Reviewed 1 of 4 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2020)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2038, opened 2020-12-22
[LibOS,Pal] Use GCC's stack protector in LibOS and PAL functions

dimakuv:
## Description of the changes 

GCC (and other compilers, e.g. Clang) provide a **stack protector** feature to detect stack corruptions. This is achieved by storing a 64-bit canary value on the stack frame on function entry and verifying this value on function exit.

Previously, Graphene disabled stack protector completely. This PR enables it in LibOS and PAL code (only if `-mstack-protector` feature is supported by compiler).

The stack protector uses a random per-thread canary stored in the TLS/TCB of each thread. Each PAL implementation must follow the rule that TLS/TCB is accessed via the GS register and that the offset of canary in TLS/TCB is 0x8. Since LibOS re-uses TLS/TCB of the PAL, there is no need for additional enabling at the LibOS layer.

Since `-mstack-protector` feature is architecture-specific, it is currently enabled only for x86-64.

Co-authored-by: Isaku Yamahata <isaku.yamahata@gmail.com>

Recreation (and simplification) of #774. Closes #774.

**NOTE:** The `-mstack-protector` options were introduced in GCC8 (the default GCC of Ubuntu 18.04 is GCC7, so update to GCC8 to build with stack protector).

Useful links:
- https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html (see `no_stack_protector`)
- https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html (see `-fstack-protector`)
- https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html#x86-Options (see `-mstack-protector-guard`)
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=29838
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94722
- https://embeddedartistry.com/blog/2020/05/18/implementing-stack-smashing-protection-for-microcontrollers-and-embedded-artistrys-libc/

## How to test this PR? 

All tests must pass. You can manually verify that LibOS and PAL binaries now have stack checks:
```
cd graphene/Runtime
objdump -D libsysdb.so | grep __stack_chk_fail
objdump -D pal-Linux | grep __stack_chk_fail
objdump -D libpal-Linux-SGX.so | grep __stack_chk_fail

# as a bonus, verify that the untrusted part of Linux-SGX PAL doesn't have stack protector
objdump -D pal-Linux-SGX | grep __stack_chk_fail
```

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)

### DIFF
--- a/LibOS/shim/src/Makefile
+++ b/LibOS/shim/src/Makefile
@@ -6,7 +6,7 @@ CFLAGS += -fPIC -Winline -Wwrite-strings \
 	  -fmerge-all-constants -Wstrict-prototypes $(call cc-option,-Wtrampolines) \
 	  -Werror=implicit-function-declaration \
 	  $(cc-option, -Wnull-dereference) \
-	  -fno-stack-protector -fno-builtin -Wno-inline \
+	  -fno-builtin -Wno-inline \
 	  -I../include \
 	  -I../include/arch/$(ARCH) \
 	  -I../../../Pal/include \
@@ -17,6 +17,16 @@ CFLAGS += -fPIC -Winline -Wwrite-strings \
 	  -I../../../Pal/include/pal
 
 CFLAGS += -Wextra
+
+# use TLS-based stack protector of GCC (we rely on the fact that LibOS reuses the same TCB as its
+# underlying PAL which must have a canary in its PAL_TCB at offset 0x8, so no additional enabling
+# is required in the LibOS code);
+# not all compilers support mstack-protector-guard, so use stack protector only if supported
+CFLAGS += -fno-stack-protector
+ifeq ($(ARCH),x86_64)
+CFLAGS += $(call cc-option, -fstack-protector-strong -mstack-protector-guard=tls \
+	                        -mstack-protector-guard-reg=%gs -mstack-protector-guard-offset=8)
+endif
 
 # Some of the code uses alignof on expressions, which is a GNU extension.
 # Silence Clang - it complains but does support it.
--- a/LibOS/shim/src/shim_init.c
+++ b/LibOS/shim/src/shim_init.c
@@ -58,6 +58,13 @@ noreturn void __abort(void) {
     DkProcessExit(1);
 }
 
+/* we use GCC's stack protector; when it detects corrupted stack, it calls __stack_chk_fail() */
+noreturn void __stack_chk_fail(void); /* to suppress GCC's warning "no previous prototype" */
+noreturn void __stack_chk_fail(void) {
+    debug("Stack protector: Graphene LibOS internal stack corruption detected\n");
+    __abort();
+}
+
 static int pal_errno_to_unix_errno[PAL_ERROR_NATIVE_COUNT + 1] = {
     [PAL_ERROR_SUCCESS]         = 0,
     [PAL_ERROR_NOTIMPLEMENTED]  = ENOSYS,
--- a/Pal/include/arch/x86_64/Linux/pal_host-arch.h
+++ b/Pal/include/arch/x86_64/Linux/pal_host-arch.h
@@ -17,7 +17,9 @@
 
 #include "sysdep-arch.h"
 
-static inline int pal_set_tcb(PAL_TCB* tcb) {
+/* Graphene uses GCC's stack protector that looks for canary at gs:[0x8], but this function changes
+ * the GS register value, so we disable stack protector here (even though it is mostly inlined) */
+__attribute__((__optimize__("-fno-stack-protector"))) static inline int pal_set_tcb(PAL_TCB* tcb) {
     return INLINE_SYSCALL(arch_prctl, 2, ARCH_SET_GS, tcb);
 }
 
--- a/Pal/include/arch/x86_64/pal-arch.h
+++ b/Pal/include/arch/x86_64/pal-arch.h
@@ -13,6 +13,7 @@
 #ifndef PAL_ARCH_H
 #define PAL_ARCH_H
 
+#include <assert.h>
 #include <stdint.h>
 
 #include "cpu.h"
@@ -25,12 +26,20 @@ typedef struct pal_tcb PAL_TCB;
 
 #define PAL_LIBOS_TCB_SIZE 256
 
+#define STACK_PROTECTOR_CANARY_DEFAULT  0xbadbadbadbadUL
+
 typedef struct pal_tcb {
     struct pal_tcb* self;
+    /* random per-thread canary used by GCC's stack protector; must be at %gs:[0x8]
+     * because we specify `-mstack-protector-guard-reg=%gs -mstack-protector-guard-offset=8` */
+    uint64_t stack_protector_canary;
     /* uint64_t for alignment */
     uint64_t libos_tcb[(PAL_LIBOS_TCB_SIZE + sizeof(uint64_t) - 1) / sizeof(uint64_t)];
     /* data private to PAL implementation follows this struct. */
 } PAL_TCB;
+
+static_assert(offsetof(PAL_TCB, stack_protector_canary) == 0x8,
+              "unexpected offset of stack_protector_canary in PAL_TCB struct");
 
 #include "pal_host-arch.h"
 
--- a/Pal/include/host/Linux-common/linux_utils.h
+++ b/Pal/include/host/Linux-common/linux_utils.h
@@ -8,5 +8,8 @@ char* get_main_exec_path(void);
 
 int read_text_file_to_cstr(const char* path, char** out);
 
+/* called only from GCC-emitted code; declare here to suppress GCC warn "no previous prototype" */
+noreturn void __stack_chk_fail(void);
+
 #endif // _LINUX_UTILS_H
 
--- a/Pal/lib/Makefile
+++ b/Pal/lib/Makefile
@@ -2,6 +2,11 @@ include ../../Scripts/Makefile.configs
 include ../../Scripts/Makefile.rules
 include ../src/host/$(PAL_HOST)/Makefile.am
 
+# FIXME: currently disable GCC's stack protector in the common libs because they are linked with
+#        different libraries/executables including those that do not provide __stack_chk_fail(),
+#        e.g. PAL regression tests (in particular, avl_tree_test.c)
+CFLAGS += -fno-stack-protector
+
 CFLAGS += \
 	-I../include \
 	-I../include/arch/$(ARCH) \
--- a/Pal/regression/Makefile
+++ b/Pal/regression/Makefile
@@ -1,7 +1,7 @@
 include ../../Scripts/Makefile.configs
 include ../../Scripts/Makefile.rules
 
-CFLAGS	+= -Wp,-U_FORTIFY_SOURCE -fno-builtin -nostdlib \
+CFLAGS	+= -Wp,-U_FORTIFY_SOURCE -fno-stack-protector -fno-builtin -nostdlib \
 	  -I../src \
 	  -I../include \
 	  -I../include/pal \
--- a/Pal/src/Makefile
+++ b/Pal/src/Makefile
@@ -28,6 +28,14 @@ CFLAGS += \
 	-I../include/arch/$(ARCH)/$(PAL_HOST) \
 	-I../include/lib
 
+# use TLS-based stack protector of GCC (all PALs must have a canary in its PAL_TCB at offset 0x8);
+# not all compilers support mstack-protector-guard, so use stack protector only if supported
+CFLAGS += -fno-stack-protector
+ifeq ($(ARCH),x86_64)
+CFLAGS += $(call cc-option, -fstack-protector-strong -mstack-protector-guard=tls \
+	                        -mstack-protector-guard-reg=%gs -mstack-protector-guard-offset=8)
+endif
+
 PAL_HOST_MACRO = $(shell echo $(PAL_HOST) | tr '[:lower:]' '[:upper:]' | tr '-' '_')
 
 # Build Targets:
--- a/Pal/src/host/Linux-SGX/Makefile
+++ b/Pal/src/host/Linux-SGX/Makefile
@@ -31,7 +31,7 @@ defs	= -DIN_PAL
 CFLAGS += $(defs)
 ASFLAGS += $(defs)
 
-commons_objs_encl = bogomips.o
+commons_objs_encl = bogomips.o stack_protector.o
 commons_objs_urts = file_utils.o main_exec_path.o
 
 enclave-objs = \
@@ -85,6 +85,16 @@ urts-objs = \
 urts-asm-objs = sgx_entry.o
 
 graphene_lib = .lib/graphene-lib.a
+
+# use TLS-based stack protector of GCC (trusted PAL must have a canary in its PAL_TCB at offset
+# 0x8, untrusted PAL uses the default GCC scheme)
+# not all compilers support mstack-protector-guard, so use stack protector only if supported
+CFLAGS += -fno-stack-protector
+ifeq ($(ARCH),x86_64)
+$(enclave-objs): CFLAGS += $(call cc-option, -fstack-protector-strong -mstack-protector-guard=tls \
+                             -mstack-protector-guard-reg=%gs -mstack-protector-guard-offset=8)
+$(urts-objs): CFLAGS += $(call cc-option, -fstack-protector-strong)
+endif
 
 .PHONY: all
 all: gsgx.h $(host_files) tools
--- a/Pal/src/host/Linux-SGX/Makefile.am
+++ b/Pal/src/host/Linux-SGX/Makefile.am
@@ -1,7 +1,6 @@
 # Add host-specific compilation rules here
 
-CFLAGS += -fPIC -maes -Wp,-U_FORTIFY_SOURCE \
-	  -fno-stack-protector -fno-builtin $(call cc-option,-Wtrampolines)
+CFLAGS += -fPIC -maes -Wp,-U_FORTIFY_SOURCE -fno-builtin $(call cc-option,-Wtrampolines)
 
 CFLAGS += -Wextra $(call cc-option,-Wnull-dereference)
 
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -172,6 +172,9 @@ fail:
 extern void* g_enclave_base;
 extern void* g_enclave_top;
 
+/* Graphene uses GCC's stack protector that looks for a canary at gs:[0x8], but this function starts
+ * with a default canary and then updates it to a random one, so we disable stack protector here */
+__attribute__((__optimize__("-fno-stack-protector")))
 noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char* uptr_args,
                              size_t args_size, char* uptr_env, size_t env_size,
                              struct pal_sec* uptr_sec_info) {
@@ -423,6 +426,14 @@ noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char*
     first_thread->thread.tid = 1;
     g_pal_control.first_thread = first_thread;
     SET_ENCLAVE_TLS(thread, &first_thread->thread);
+
+    uint64_t stack_protector_canary;
+    ret = _DkRandomBitsRead(&stack_protector_canary, sizeof(stack_protector_canary));
+    if (ret < 0) {
+        SGX_DBG(DBG_E, "_DkRandomBitsRead failed: %d\n", ret);
+        ocall_exit(1, /*is_exitgroup=*/true);
+    }
+    pal_set_tcb_stack_canary(stack_protector_canary);
 
     /* call main function */
     pal_main(g_pal_sec.instance_id, exec, g_pal_sec.exec_addr, parent, first_thread, arguments,
--- a/Pal/src/host/Linux-SGX/db_threading.c
+++ b/Pal/src/host/Linux-SGX/db_threading.c
@@ -50,7 +50,11 @@ static PAL_IDX pal_assign_tid(void) {
     return __atomic_add_fetch(&tid.counter, 1, __ATOMIC_SEQ_CST);
 }
 
-void pal_start_thread(void) {
+/* Initialization wrapper of a newly-created thread. This function finds a newly-created thread in
+ * g_thread_list, initializes its TCB/TLS, and jumps into the callback-to-run. Graphene uses GCC's
+ * stack protector that looks for a canary at gs:[0x8], but this function starts with a default
+ * canary and then updates it to a random one, so we disable stack protector here. */
+__attribute__((__optimize__("-fno-stack-protector"))) void pal_start_thread(void) {
     struct pal_handle_thread *new_thread = NULL, *tmp;
 
     _DkInternalLock(&g_thread_list_lock);
@@ -72,8 +76,14 @@ void pal_start_thread(void) {
     const void* param = thread_param->param;
     free(thread_param);
     new_thread->param = NULL;
+
     SET_ENCLAVE_TLS(thread, new_thread);
     SET_ENCLAVE_TLS(ready_for_exceptions, 1UL);
+
+    /* each newly-created thread (including the first thread) has its own random stack canary */
+    uint64_t stack_protector_canary;
+    _DkRandomBitsRead(&stack_protector_canary, sizeof(stack_protector_canary));
+    pal_set_tcb_stack_canary(stack_protector_canary);
     PAL_TCB* pal_tcb = pal_get_tcb();
     memset(&pal_tcb->libos_tcb, 0, sizeof(pal_tcb->libos_tcb));
     callback((void*)param);
--- a/Pal/src/host/Linux-SGX/generated-offsets.c
+++ b/Pal/src/host/Linux-SGX/generated-offsets.c
@@ -24,6 +24,9 @@ __attribute__((__used__)) static void dummy(void) {
     DEFINE(SGX_MISCSELECT_EXINFO, SGX_MISCSELECT_EXINFO);
     DEFINE(SE_KEY_SIZE, SE_KEY_SIZE);
 
+    /* defines in pal-arch.h */
+    DEFINE(STACK_PROTECTOR_CANARY_DEFAULT, STACK_PROTECTOR_CANARY_DEFAULT);
+
     /* sgx_measurement_t */
     DEFINE(SGX_HASH_SIZE, sizeof(sgx_measurement_t));
 
@@ -73,6 +76,7 @@ __attribute__((__used__)) static void dummy(void) {
 
     /* struct enclave_tls */
     OFFSET(SGX_COMMON_SELF, enclave_tls, common.self);
+    OFFSET(SGX_COMMON_STACK_PROTECTOR_CANARY, enclave_tls, common.stack_protector_canary);
     OFFSET(SGX_ENCLAVE_SIZE, enclave_tls, enclave_size);
     OFFSET(SGX_TCS_OFFSET, enclave_tls, tcs_offset);
     OFFSET(SGX_INITIAL_STACK_OFFSET, enclave_tls, initial_stack_offset);
--- a/Pal/src/host/Linux-SGX/sgx_main.c
+++ b/Pal/src/host/Linux-SGX/sgx_main.c
@@ -489,6 +489,7 @@ static int initialize_enclave(struct pal_enclave* enclave, const char* manifest_
                 memset(gs, 0, g_page_size);
                 assert(sizeof(*gs) <= g_page_size);
                 gs->common.self = (PAL_TCB*)(tls_area->addr + g_page_size * t + enclave_secs.base);
+                gs->common.stack_protector_canary = STACK_PROTECTOR_CANARY_DEFAULT;
                 gs->enclave_size = enclave->size;
                 gs->tcs_offset = tcs_area->addr + g_page_size * t;
                 gs->initial_stack_offset = stack_areas[t].addr + ENCLAVE_STACK_SIZE;
--- a/Pal/src/host/Linux-SGX/sgx_tls.h
+++ b/Pal/src/host/Linux-SGX/sgx_tls.h
@@ -81,7 +81,16 @@ static inline struct enclave_tls* get_tcb_trts(void) {
                 : "ir"(value), "i"(offsetof(struct enclave_tls, member))                       \
                 : "memory");                                                                   \
     } while (0)
-#else
+
+
+__attribute__((__optimize__("-fno-stack-protector")))
+static inline void pal_set_tcb_stack_canary(uint64_t canary) {
+    ((char*)&canary)[0] = 0; /* prevent C-string-based stack leaks from exposing the cookie */
+    SET_ENCLAVE_TLS(common.stack_protector_canary, canary);
+}
+
+#else /* IN_ENCLAVE */
+
 /* private to untrusted Linux PAL, unique to each untrusted thread */
 typedef struct pal_tcb_urts {
     struct pal_tcb_urts* self;
@@ -109,6 +118,6 @@ static inline PAL_TCB_URTS* get_tcb_urts(void) {
 
 extern bool g_sgx_enable_stats;
 void update_and_print_stats(bool process_wide);
-#endif
+#endif /* IN_ENCLAVE */
 
 #endif /* __SGX_TLS_H__ */
--- a/Pal/src/host/Linux-common/stack_protector.c
+++ b/Pal/src/host/Linux-common/stack_protector.c
@@ -1,0 +1,12 @@
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
+/* Copyright (C) 2020 Intel Corporation */
+
+#include "api.h"
+#include "linux_utils.h"
+#include "pal_internal.h"
+
+/* we use GCC's stack protector; when it detects corrupted stack, it calls __stack_chk_fail() */
+noreturn void __stack_chk_fail(void) {
+    printf("Stack protector: Graphene PAL internal stack corruption detected\n");
+    _DkProcessExit(1);
+}
--- a/Pal/src/host/Linux/Makefile
+++ b/Pal/src/host/Linux/Makefile
@@ -1,7 +1,7 @@
 include ../../../../Scripts/Makefile.configs
 include Makefile.am
 
-CFLAGS	+= \
+CFLAGS += \
 	-I. \
 	-Iinclude \
 	-I../.. \
@@ -12,6 +12,14 @@ CFLAGS	+= \
 	-I../../../include/host/Linux \
 	-I../../../include/host/Linux-common \
 	-I../../../include/lib
+
+# use TLS-based stack protector of GCC (Linux PAL must have a canary in its PAL_TCB at offset 0x8);
+# not all compilers support mstack-protector-guard, so use stack protector only if supported
+CFLAGS += -fno-stack-protector
+ifeq ($(ARCH),x86_64)
+CFLAGS += $(call cc-option, -fstack-protector-strong -mstack-protector-guard=tls \
+	                        -mstack-protector-guard-reg=%gs -mstack-protector-guard-offset=8)
+endif
 
 ASFLAGS += \
 	-I. \
@@ -29,7 +37,8 @@ ASFLAGS += $(defs)
 commons_objs = \
 	bogomips.o \
 	file_utils.o \
-	main_exec_path.o
+	main_exec_path.o \
+	stack_protector.o
 
 objs = \
 	clone-$(ARCH).o \
--- a/Pal/src/host/Linux/Makefile.am
+++ b/Pal/src/host/Linux/Makefile.am
@@ -1,8 +1,7 @@
 # Add host-specific compilation rules here
 SEC_DIR = security/$(PAL_HOST)
 
-CFLAGS += -fPIC -Wp,-U_FORTIFY_SOURCE \
-	  -fno-stack-protector -fno-builtin $(call cc-option,-Wtrampolines)
+CFLAGS += -fPIC -Wp,-U_FORTIFY_SOURCE -fno-builtin $(call cc-option,-Wtrampolines)
 
 CFLAGS += -Wextra $(call cc-option,-Wnull-dereference)
 
--- a/Pal/src/host/Linux/db_main.c
+++ b/Pal/src/host/Linux/db_main.c
@@ -142,15 +142,26 @@ noreturn static void print_usage_and_exit(const char* argv_0) {
     _DkProcessExit(1);
 }
 
+/* Graphene uses GCC's stack protector that looks for a canary at gs:[0x8], but this function starts
+ * with no TCB in the GS register, so we disable stack protector here */
+__attribute__((__optimize__("-fno-stack-protector")))
 noreturn void pal_linux_main(void* initial_rsp, void* fini_callback) {
     __UNUSED(fini_callback);  // TODO: We should call `fini_callback` at the end.
     int ret;
+
+    /* we don't yet have a TCB in the GS register, but GCC's stack protector will look for a canary
+     * at gs:[0x8] in functions called below, so let's install a dummy TCB with a default canary */
+    PAL_TCB_LINUX dummy_tcb_for_stack_protector = { 0 };
+    dummy_tcb_for_stack_protector.common.self = &dummy_tcb_for_stack_protector.common;
+    pal_set_tcb_stack_canary(&dummy_tcb_for_stack_protector, STACK_PROTECTOR_CANARY_DEFAULT);
+    ret = pal_set_tcb(&dummy_tcb_for_stack_protector.common);
+    if (ret < 0)
+        INIT_FAIL(unix_to_pal_error(-ret), "pal_set_tcb() failed");
 
     uint64_t start_time;
     ret = _DkSystemTimeQuery(&start_time);
     if (ret < 0)
         INIT_FAIL(unix_to_pal_error(-ret), "_DkSystemTimeQuery() failed");
-
 
     /* Initialize alloc_align as early as possible, a lot of PAL APIs depend on this being set. */
     g_pal_state.alloc_align = _DkGetAllocationAlignment();
--- a/Pal/src/host/Linux/db_threading.c
+++ b/Pal/src/host/Linux/db_threading.c
@@ -78,19 +78,29 @@ out:
     return ret;
 }
 
-/*
- * pal_thread_init(): An initialization wrapper of a newly-created thread (including
- * the first thread). This function accepts a TCB pointer to be set to the GS register
- * of the thread. The rest of the TCB is used as the alternative stack for signal
- * handling.
- */
-int pal_thread_init(void* tcbptr) {
+/* Initialization wrapper of a newly-created thread (including the first thread). This function
+ * accepts a TCB pointer to be set to the GS register (on x86-64) of the thread. The rest of the TCB
+ * is used as the alternate stack for signal handling. Since Graphene uses GCC's stack protector,
+ * and this function modifies the stack protector's GS register, we disable stack protector here. */
+__attribute__((__optimize__("-fno-stack-protector"))) int pal_thread_init(void* tcbptr) {
     PAL_TCB_LINUX* tcb = tcbptr;
     int ret;
 
+    /* we inherited the parent's GS register which we shouldn't use in the child thread, but GCC's
+     * stack protector will look for a canary at gs:[0x8] in functions called below (e.g.,
+     * _DkRandomBitsRead), so let's install a default canary in the child's TCB */
+    pal_set_tcb_stack_canary(tcb, STACK_PROTECTOR_CANARY_DEFAULT);
     ret = pal_set_tcb(&tcb->common);
     if (IS_ERR(ret))
         return -ERRNO(ret);
+
+    /* each newly-created thread (including the first thread) has its own random stack canary */
+    uint64_t stack_protector_canary;
+    ret = _DkRandomBitsRead(&stack_protector_canary, sizeof(stack_protector_canary));
+    if (IS_ERR(ret))
+        return -EPERM;
+
+    pal_set_tcb_stack_canary(tcb, stack_protector_canary);
 
     if (tcb->alt_stack) {
         stack_t ss = {
@@ -224,8 +234,8 @@ noreturn void _DkThreadExit(int* clear_child_tid) {
         ss.ss_flags = SS_DISABLE;
         ss.ss_size  = 0;
 
-        // Take precautions to unset the TCB and alternative stack first.
-        pal_set_tcb(NULL);
+        /* take precautions to unset alternate stack; note that we cannot unset the TCB because
+         * GCC's stack protector still uses the GS register until the end of this function */
         INLINE_SYSCALL(sigaltstack, 2, &ss, NULL);
     }
 
--- a/Pal/src/host/Linux/pal_linux.h
+++ b/Pal/src/host/Linux/pal_linux.h
@@ -160,4 +160,14 @@ static inline PAL_TCB_LINUX* get_tcb_linux(void) {
     return (PAL_TCB_LINUX*)pal_get_tcb();
 }
 
+__attribute__((__optimize__("-fno-stack-protector")))
+static inline void pal_set_tcb_stack_canary(PAL_TCB_LINUX* tcbptr, uint64_t canary) {
+    ((char*)&canary)[0] = 0; /* prevent C-string-based stack leaks from exposing the cookie */
+#ifdef __x86_64__
+    tcbptr->common.stack_protector_canary = canary;
+#else
+#error "unsupported architecture"
+#endif
+}
+
 #endif /* PAL_LINUX_H */
--- a/python/graphenelibos/sgx_sign.py
+++ b/python/graphenelibos/sgx_sign.py
@@ -464,6 +464,8 @@ def gen_area_content(attr, areas, enclave_base_addr, enclave_heap_min):
 
         set_tls_field(t, offs.SGX_COMMON_SELF,
                       tls_area.addr + offs.PAGESIZE * t + enclave_base_addr)
+        set_tls_field(t, offs.SGX_COMMON_STACK_PROTECTOR_CANARY,
+                      offs.STACK_PROTECTOR_CANARY_DEFAULT)
         set_tls_field(t, offs.SGX_ENCLAVE_SIZE, attr['enclave_size'])
         set_tls_field(t, offs.SGX_TCS_OFFSET, tcs_area.addr + offs.TCS_SIZE * t)
         set_tls_field(t, offs.SGX_INITIAL_STACK_OFFSET,


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: 0 of 19 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2038#-MP9xUTL3thq25uozx9t:-MP9xUTL3thq25uozx9u:b-3823in) (no related file):*
BTW, I observed about 1-2% throughput overhead on the worst-case Redis benchmark with enabled Stack Protector in Graphene.

At the same time, I observed *no* noticeable increase in binary sizes (recall that Stack Protector adds a couple instructions in each function prologue and epilogue).

---

yamahata (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: 0 of 19 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2038#-MP9xUTL3thq25uozx9t:-MPAq_zGERphlZDrGo87:b62mkeb) (no related file):*

BTW, I observed about 1-2% throughput overhead on the worst-case Redis benchmark with enabled Stack Protector in Graphene.

At the same time, I observed *no* noticeable increase in binary sizes (recall that Stack Protector adds a couple instructions in each function prologue and epilogue).
</blockquote></details>

Great! now the patch is much simpler.

---
___
*[Pal/src/host/Linux/db_main.c, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPAiv6e-T1lRF8j1yYa:-MPAiv6f-c7-eAZyVnR7:b-9caj4z) ([raw file](https://github.com/oscarlab/graphene/blob/d6561d404234f8a8d629182c3150f885915676a3/Pal/src/host/Linux/db_main.c#L165)):*
> ```C
>     dummy_tcb_for_stack_protector.common.self = &dummy_tcb_for_stack_protector.common;
>     dummy_tcb_for_stack_protector.common.stack_protector_canary = STACK_PROTECTOR_CANARY_DEFAULT;
>     ret = pal_set_tcb(&dummy_tcb_for_stack_protector.common);
> ```

Pranoid: Add -fno-stack-protector to pal_set_tcb().
I think usually gcc inlines the function and it doesn't matter. But no guarantee for gcc to do that.
___
*[Pal/src/host/Linux/db_threading.c, line 91 at r1](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPAjJrtCM1rrFEzwlkS:-MPAjJrtCM1rrFEzwlkT:buuw2j5) ([raw file](https://github.com/oscarlab/graphene/blob/d6561d404234f8a8d629182c3150f885915676a3/Pal/src/host/Linux/db_threading.c#L91)):*
> ```C
>     /* each newly-created thread (including the first thread) has its own random stack canary */
>     tcb->common.stack_protector_canary = STACK_PROTECTOR_CANARY_DEFAULT;
>     _DkRandomBitsRead(&tcb->common.stack_protector_canary,
> ```

_DkRandomBitsRead() is compiled with stack protector and here %gs has parent's value.
So
- initialize the value by thread parent or
- or no-stack-protector version of _DkRandomBitsRead().

On the other hand, Pal/Linux-SGX case, %gsbase is known value on enclave entry.
___
*[Pal/src/host/Linux/db_threading.c, line 94 at r1](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPAjt578G8VFV1Jz2vw:-MPAjt578G8VFV1Jz2vx:b-lfod3j) ([raw file](https://github.com/oscarlab/graphene/blob/d6561d404234f8a8d629182c3150f885915676a3/Pal/src/host/Linux/db_threading.c#L94)):*
> ```C
>                       sizeof(tcb->common.stack_protector_canary));
> 
>     ret = pal_set_tcb(&tcb->common);
> ```

Pranoid: no-stack-protector to pal_set_tcb()

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: 0 of 20 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPE4LnH5jugWuOSmTaL:-MPE4LnH5jugWuOSmTaM:b-x7n8ps) (no related file):*
A question to reviewers: in this PR, I'm using the `-fstack-protector-strong` variant of the Stack Protector. This variant enables rather strong checks, see comparison e.g. here: https://www.keil.com/support/man/docs/armclang_ref/armclang_ref_cjh1548250046139.htm. This is used at least by Google ChromiumOS, and they claim it's the best performance-security option.

So, should we also use this `strong` version? We could also go for `-fstack-protector` but this looks quite useless and for `-fstack-protector-all` which seems to be an overkill.

---
___
*[Pal/lib/Makefile, line 8 at r1](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPE5daP5GHNffO9bdGn:-MPE5daP5GHNffO9bdGo:b-ln2n9q) ([raw file](https://github.com/oscarlab/graphene/blob/d6561d404234f8a8d629182c3150f885915676a3/Pal/lib/Makefile#L8)):*
> ```Makefile
> #        different libraries/executables including those that do not provide __stack_chk_fail(),
> #        e.g. PAL regression tests (in particular, avl_tree_test.c)
> CFLAGS += -fno-stack-protector
> ```

BTW, after we move to Meson and away from Makefiles, I definitely want to:
- revisit the way we detect compiler options (right now we basically append these options, and sometimes we use the trick of multiple compiler options overriding each other, depending on their order in the command line)
- move our "common library" from `/Pal/lib` to outside of PAL, maybe to its own directory `/Common`.
___
*[Pal/src/host/Linux/db_main.c, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPAiv6e-T1lRF8j1yYa:-MPE6aP69rMTyaQc0QGt:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d6561d404234f8a8d629182c3150f885915676a3/Pal/src/host/Linux/db_main.c#L165)):*

Pranoid: Add -fno-stack-protector to pal_set_tcb().
I think usually gcc inlines the function and it doesn't matter. But no guarantee for gcc to do that.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_threading.c, line 91 at r1](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPAjJrtCM1rrFEzwlkS:-MPE8ljr0XbayjebAE0U:b-eh5156) ([raw file](https://github.com/oscarlab/graphene/blob/d6561d404234f8a8d629182c3150f885915676a3/Pal/src/host/Linux/db_threading.c#L91)):*

_DkRandomBitsRead() is compiled with stack protector and here %gs has parent's value.
So
- initialize the value by thread parent or
- or no-stack-protector version of _DkRandomBitsRead().

On the other hand, Pal/Linux-SGX case, %gsbase is known value on enclave entry.

</blockquote></details>

Done. Similarly to the case of the main thread, I now install a dummy TCB.

As you correctly mentioned, we already have a `gsbase` value set in the Linux-SGX case, so nothing to be done there.
___
*[Pal/src/host/Linux/db_threading.c, line 94 at r1](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPAjt578G8VFV1Jz2vw:-MPE6en03SDZkYfwa8Ta:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d6561d404234f8a8d629182c3150f885915676a3/Pal/src/host/Linux/db_threading.c#L94)):*

Pranoid: no-stack-protector to pal_set_tcb()
</blockquote></details>

Done.

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: 0 of 20 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @yamahata)

___
*[Pal/src/host/Linux/db_threading.c, line 100 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPucgQH5Dj02ED7DphI:-MPucgQH5Dj02ED7DphJ:b-mi56g5) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L100)):*
> ```C
> 
>     /* each newly-created thread (including the first thread) has its own random stack canary */
>     tcb->common.stack_protector_canary = STACK_PROTECTOR_CANARY_DEFAULT;
> ```

You should be able to remove the above line. It's overwritten immediately after.

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: 0 of 20 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPud1pO6J_K9HJ5jiF8:-MPud1pO6J_K9HJ5jiF9:blpdh2i) (no related file):*
Great stuff! I am looking forward to (trying to) enable this on ppc64 as well. It should be possible to enable this on the PAL layer independently of the LibOS layer since we are setting the canary really early... We only have one TLS pointer on ppc64 (I think x86_64 has 2 wit %fs and %gs), so trying to think what must not happen to cause failures. glibc seems to set its own stack protector and I suppose it will never unroll the stack and run into the canary set by either PAL or LibOS and error because of that then... I guess it's going to be a matter of trying.

---

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: 0 of 20 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @yamahata)

___
*[Pal/src/host/Linux/db_threading.c, line 94 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPugH925G54l5-Usors:-MPugH925G54l5-Usort:bc7fx9l) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L94)):*
> ```C
>     static PAL_TCB_LINUX dummy_tcb_for_stack_protector = { 0 };
>     dummy_tcb_for_stack_protector.common.self = &dummy_tcb_for_stack_protector.common;
>     dummy_tcb_for_stack_protector.common.stack_protector_canary = STACK_PROTECTOR_CANARY_DEFAULT;
> ```

I think we will need an architecture-specific function to initialize the PAL_TCB_LINUX.

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: 0 of 20 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @yamahata)

___
*[Pal/src/host/Linux/db_threading.c, line 94 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPugH925G54l5-Usors:-MPujnsk-Tqu_6b9Zo0C:bx1e43c) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L94)):*

I think we will need an architecture-specific function to initialize the PAL_TCB_LINUX.
</blockquote></details>

For PPC64 we will need to change pal_tcb_linux so we can have the canary  at the right location:

```
typedef struct pal_tcb_linux {
    PAL_TCB common;
    struct {
        /* private to Linux PAL */
        int        pending_events[MAX_SIGNAL_LOG];
        int        pending_events_num;
        PAL_HANDLE handle;
        void*      alt_stack;
        int        (*callback)(void*);
        void*      param;
    };
    PAL_TCB_ARCH_TRAILER trailer;   <---- for PPC64
} PAL_TCB_LINUX;


typedef struct pal_tcb_arch_trailer {
    uintptr_t stack_protector_canary; // must be previous to last in TCB
    void *dtv; // must be last in TCB
} PAL_TCB_ARCH_TRAILER;
```

On ppc64 the stack_protector_canary must be the 2nd to last field in the thread control block with the last field being a pointer.

see also: http://openpowerfoundation.org/wp-content/uploads/resources/leabi/content/dbdoclet.50655241_50444.html

stefanberger (commented):
> Reviewed 17 of 19 files at r1, 2 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: 19 of 20 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @yamahata)

stefanberger (commented):
> Reviewed 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @yamahata)

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

yamahata (dismissed):
> Reviewed 17 of 19 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @stefanberger)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPud1pO6J_K9HJ5jiF8:-MQBSr5W3j5QMWmnOtVa:b-yy4j1g) (no related file):*

Great stuff! I am looking forward to (trying to) enable this on ppc64 as well. It should be possible to enable this on the PAL layer independently of the LibOS layer since we are setting the canary really early... We only have one TLS pointer on ppc64 (I think x86_64 has 2 wit %fs and %gs), so trying to think what must not happen to cause failures. glibc seems to set its own stack protector and I suppose it will never unroll the stack and run into the canary set by either PAL or LibOS and error because of that then... I guess it's going to be a matter of trying.
</blockquote></details>

I guess this is thinking aloud, right? :) Or is there any action I need to do on this PR?

In Graphene, we download and build Glibc *without* stack protector: https://github.com/oscarlab/graphene/blob/master/LibOS/Makefile#L48. Not sure if this is what you meant in your comment.

---
___
*[Pal/src/host/Linux/db_threading.c, line 94 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPugH925G54l5-Usors:-MQBRfE329vMnW0sbwi5:bmtz3fd) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L94)):*

For PPC64 we will need to change pal_tcb_linux so we can have the canary  at the right location:

```
typedef struct pal_tcb_linux {
    PAL_TCB common;
    struct {
        /* private to Linux PAL */
        int        pending_events[MAX_SIGNAL_LOG];
        int        pending_events_num;
        PAL_HANDLE handle;
        void*      alt_stack;
        int        (*callback)(void*);
        void*      param;
    };
    PAL_TCB_ARCH_TRAILER trailer;   <---- for PPC64
} PAL_TCB_LINUX;


typedef struct pal_tcb_arch_trailer {
    uintptr_t stack_protector_canary; // must be previous to last in TCB
    void *dtv; // must be last in TCB
} PAL_TCB_ARCH_TRAILER;
```

On ppc64 the stack_protector_canary must be the 2nd to last field in the thread control block with the last field being a pointer.

see also: http://openpowerfoundation.org/wp-content/uploads/resources/leabi/content/dbdoclet.50655241_50444.html
</blockquote></details>

I can create an arch-specific new function `pal_set_tcb_stack_canary()`, would that work?

But I don't want to move `PAL_TCB::stack_protector_canary` from its offset `0x8` in the `PAL_TCB` struct. Would it be fine with you? You'll then have an unused x86-specific field in the common TCB...
___
*[Pal/src/host/Linux/db_threading.c, line 100 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPucgQH5Dj02ED7DphI:-MQBRd9c7Y6mScXP0xBl:bhy9ot6) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L100)):*

You should be able to remove the above line. It's overwritten immediately after.
</blockquote></details>

Will do.

mkow (changes requested):
> Reviewed 17 of 19 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: all files reviewed, 27 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @stefanberger)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPE4LnH5jugWuOSmTaL:-MQIciXrBRV4aqNJewqj:bnkc7df) (no related file):*

A question to reviewers: in this PR, I'm using the `-fstack-protector-strong` variant of the Stack Protector. This variant enables rather strong checks, see comparison e.g. here: https://www.keil.com/support/man/docs/armclang_ref/armclang_ref_cjh1548250046139.htm. This is used at least by Google ChromiumOS, and they claim it's the best performance-security option.

So, should we also use this `strong` version? We could also go for `-fstack-protector` but this looks quite useless and for `-fstack-protector-all` which seems to be an overkill.
</blockquote></details>

`-fstack-protector-strong` sounds like a good choice.
btw. the docs you linked are for clang, GCC is free to have different semantic for this option.

---
___
*[LibOS/shim/src/Makefile, line 21 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIeCpa8kVQ6KcynLID:-MQIeCpa8kVQ6KcynLIE:b-dzbjm7) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/LibOS/shim/src/Makefile#L21)):*
> ```Makefile
> CFLAGS += -Wextra
> 
> # use TLS-based stack protector of GCC (we rely on the fact LibOS re-uses the same TCB as its
> ```

`on the fact LibOS` -> `on the fact that LibOS`
`re-uses` -> `reuses`
___
*[LibOS/shim/src/shim_init.c, line 63 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIhAxY7rRMrXExDQ3Y:-MQIhAxc3vYZEYi1XFKu:b2nclf8) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/LibOS/shim/src/shim_init.c#L63)):*
> ```C
> 
> /* we use GCC's stack protector; when it detects corrupted stack, it calls __stack_chk_fail() */
> noreturn void __stack_chk_fail(void); /* to supress GCC's warning "no previous prototype" */
> ```

`supress` -> `suppress`
___
*[LibOS/shim/src/shim_init.c, line 64 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIhI4q-KVY1-S9dqlD:-MQIhI4q-KVY1-S9dqlE:b-k662tk) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/LibOS/shim/src/shim_init.c#L64)):*
> ```C
> /* we use GCC's stack protector; when it detects corrupted stack, it calls __stack_chk_fail() */
> noreturn void __stack_chk_fail(void); /* to supress GCC's warning "no previous prototype" */
> void __stack_chk_fail(void) {
> ```

no `noreturn` here?
___
*[LibOS/shim/src/shim_init.c, line 65 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIhtylDXO9Zmy3JZlt:-MQIhtym82q0STD5n7Ic:b-ei826z) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/LibOS/shim/src/shim_init.c#L65)):*
> ```C
> noreturn void __stack_chk_fail(void); /* to supress GCC's warning "no previous prototype" */
> void __stack_chk_fail(void) {
>     debug("Stack protector: Graphene LibOS internal stack is corrupted at %p\n",
> ```

`is corrupted at %p` -> `corruption detected` (I don't think printing the stack pointer is anyhow helpful, it's random and impossible to analyze after the crash)
___
*[Pal/include/arch/x86_64/pal-arch.h, line 32 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIibK2AgsCWICKtYvv:-MQIibK2AgsCWICKtYvw:br20dxe) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/include/arch/x86_64/pal-arch.h#L32)):*
> ```objc
> typedef struct pal_tcb {
>     struct pal_tcb* self;
>     /* random per-thread canary used by GCC's stack protector; must be in %gs reg and at offset 0x8
> ```

`must be in %gs reg and at offset 0x8` -> `must be at gs:[0x8]`
___
*[Pal/include/arch/x86_64/pal-arch.h, line 34 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIiNwD6nvJDPfVr5uj:-MQIiNwD6nvJDPfVr5uk:bpdfp8w) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/include/arch/x86_64/pal-arch.h#L34)):*
> ```objc
>     /* random per-thread canary used by GCC's stack protector; must be in %gs reg and at offset 0x8
>      * because we specify `-mstack-protector-guard-reg=%gs -mstack-protector-guard-offset=8` */
>     uint64_t stack_protector_canary;
> ```

This struct should either be packed or we should have a `static_assert` for this field having offset == 8.
___
*[Pal/include/arch/x86_64/Linux/pal_host-arch.h, line 22 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIjQ_c5DTyVAYcsr1Y:-MQIjQ_c5DTyVAYcsr1Z:b-kq0yj8) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/include/arch/x86_64/Linux/pal_host-arch.h#L22)):*
> ```objc
> /* Graphene uses GCC's stack protector that looks for a canary at %gs:0x8, but this function changes
>  * the GS register value, so we disable stack protector here (even though it is mostly inlined) */
> __attribute__ ((__optimize__("-fno-stack-protector"))) static inline int pal_set_tcb(PAL_TCB* tcb) {
> ```

no space after `__attribute__`
___
*[Pal/src/host/Linux/db_exception.c, line 45 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIkk0W07dvh8W93aKr:-MQIkk0W07dvh8W93aKs:b2nclf8) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_exception.c#L45)):*
> ```C
> 
> /* we use GCC's stack protector; when it detects corrupted stack, it calls __stack_chk_fail() */
> noreturn void __stack_chk_fail(void); /* to supress GCC's warning "no previous prototype" */
> ```

`supress` -> `suppress`
___
*[Pal/src/host/Linux/db_exception.c, line 46 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQInoFSEVjBYoYtjKMa:-MQInoFSEVjBYoYtjKMb:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_exception.c#L46)):*
> ```C
> /* we use GCC's stack protector; when it detects corrupted stack, it calls __stack_chk_fail() */
> noreturn void __stack_chk_fail(void); /* to supress GCC's warning "no previous prototype" */
> void __stack_chk_fail(void) {
> ```

ditto
___
*[Pal/src/host/Linux/db_exception.c, line 47 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIklj2Albmjzd7L5Fr:-MQIklj2Albmjzd7L5Fs:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_exception.c#L47)):*
> ```C
> noreturn void __stack_chk_fail(void); /* to supress GCC's warning "no previous prototype" */
> void __stack_chk_fail(void) {
>     printf("Stack protector: Graphene PAL internal stack is corrupted at %p\n",
> ```

ditto
___
*[Pal/src/host/Linux/db_main.c, line 155 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIkyqb2E6fISumrMQH:-MQIkyqcAnDlkeoE0FJ2:b-kq0yj8) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_main.c#L155)):*
> ```C
> /* Graphene uses GCC's stack protector that looks for a canary at %gs:0x8, but this function starts
>  * with no TCB in the GS register, so we disable stack protector here */
> __attribute__ ((__optimize__("-fno-stack-protector")))
> ```

no space after `__attribute__`
___
*[Pal/src/host/Linux/db_threading.c, line 85 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIlmcSEPRZo0zRQgWW:-MQIlmcSEPRZo0zRQgWX:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L85)):*
> ```C
>  * is used as the alternate stack for signal handling. Since Graphene uses GCC's stack protector,
>  * and this function modifies the stack protector's GS register, we disable stack protector here. */
> __attribute__ ((__optimize__("-fno-stack-protector"))) int pal_thread_init(void* tcbptr) {
> ```

ditto
___
*[Pal/src/host/Linux/db_threading.c, line 100 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQImmWz6o7gCIw3eTZX:-MQImmWz6o7gCIw3eTZY:b2z8kji) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L100)):*
> ```C
> 
>     /* each newly-created thread (including the first thread) has its own random stack canary */
>     tcb->common.stack_protector_canary = STACK_PROTECTOR_CANARY_DEFAULT;
> ```

If we'd start with this and setting TCB then we could skip the dummy TCB (and correct the cookie later, ensuring that `_DkRandomBitsRead` reads to a temporary variable, not the real cookie).
___
*[Pal/src/host/Linux/db_threading.c, line 105 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQImIclD_yPrml4FVgW:-MQImIclD_yPrml4FVgX:b-ml1s3p) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L105)):*
> ```C
>     if (IS_ERR(ret))
>         return -EPERM;
> ```

Could we also zero the first byte of the canary? (i.e. make `(char*)&stack_protector_canary[0] == 0`) This prevents C-string-based stack leaks from exposing the cookie.
___
*[Pal/src/host/Linux-SGX/db_exception.c, line 32 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIndstDZYrXHZa0Gyz:-MQIndstDZYrXHZa0Gz-:b2nclf8) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_exception.c#L32)):*
> ```C
> 
> /* we use GCC's stack protector; when it detects corrupted stack, it calls __stack_chk_fail() */
> noreturn void __stack_chk_fail(void); /* to supress GCC's warning "no previous prototype" */
> ```

`supress` -> `suppress`
___
*[Pal/src/host/Linux-SGX/db_exception.c, line 33 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQInZHp3tpq4aJfrRjY:-MQInZHp3tpq4aJfrRjZ:b-994d05) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_exception.c#L33)):*
> ```C
> /* we use GCC's stack protector; when it detects corrupted stack, it calls __stack_chk_fail() */
> noreturn void __stack_chk_fail(void); /* to supress GCC's warning "no previous prototype" */
> void __stack_chk_fail(void) {
> ```

ditto (no `noreturn`?)
___
*[Pal/src/host/Linux-SGX/db_exception.c, line 33 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIoNvhDx9KpXITbLa4:-MQIoNvhDx9KpXITbLa5:b-f16z7c) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_exception.c#L33)):*
> ```C
> /* we use GCC's stack protector; when it detects corrupted stack, it calls __stack_chk_fail() */
> noreturn void __stack_chk_fail(void); /* to supress GCC's warning "no previous prototype" */
> void __stack_chk_fail(void) {
> ```

Can't we move it to host/Linux-common and deduplicate?
___
*[Pal/src/host/Linux-SGX/db_exception.c, line 34 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQInhak20je5kAdD5Qr:-MQInhak20je5kAdD5Qs:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_exception.c#L34)):*
> ```C
> noreturn void __stack_chk_fail(void); /* to supress GCC's warning "no previous prototype" */
> void __stack_chk_fail(void) {
>     printf("Stack protector: Graphene PAL internal stack is corrupted at %p\n",
> ```

ditto
___
*[Pal/src/host/Linux-SGX/db_main.c, line 177 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQInxr-75FplrhNmroj:-MQInxr-75FplrhNmrok:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_main.c#L177)):*
> ```C
> /* Graphene uses GCC's stack protector that looks for a canary at %gs:0x8, but this function starts
>  * with a default canary and then updates it to a random one, so we disable stack protector here */
> __attribute__ ((__optimize__("-fno-stack-protector")))
> ```

ditto
___
*[Pal/src/host/Linux-SGX/db_main.c, line 455 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIo0Zw7LHZpDhM_-hd:-MQIo0Zw7LHZpDhM_-he:bsj228f) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_main.c#L455)):*
> ```C
>         ocall_exit(1, /*is_exitgroup=*/true);
>     }
>     SET_ENCLAVE_TLS(common.stack_protector_canary, stack_protector_canary);
> ```

ditto (zeroing the first byte)
___
*[Pal/src/host/Linux-SGX/db_threading.c, line 57 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIoG8uA0FvVhRhYfUN:-MQIoG8uA0FvVhRhYfUO:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_threading.c#L57)):*
> ```C
>  * stack protector that looks for a canary at %gs:0x8, but this function starts with a default
>  * canary and then updates it to a random one, so we disable stack protector here. */
> __attribute__ ((__optimize__("-fno-stack-protector"))) void pal_start_thread(void) {
> ```

ditto
___
*[Pal/src/host/Linux-SGX/db_threading.c, line 85 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIoiRJ2WXPdHksgawf:-MQIoiRJ2WXPdHksgawg:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_threading.c#L85)):*
> ```C
>     /* each newly-created thread (including the first thread) has its own random stack canary */
>     uint64_t stack_protector_canary;
>     _DkRandomBitsRead(&stack_protector_canary, sizeof(stack_protector_canary));
> ```

ditto
___
*[Pal/src/host/Linux-SGX/Makefile, line 109 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIpWPE3kktm7vPL_mj:-MQIpWPF-dfICCCmUX_D:b-sltnyt) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/Makefile#L109)):*
> ```Makefile
> $(patsubst %.o,%.s,$(enclave-asm-objs)): ASFLAGS += -DIN_ENCLAVE
> 
> # there is no point in having stack protector in the untrusted part of Linux-SGX PAL
> ```

it can be useful for detecting bugs (not as a security measure, just an aid in early detection of bugs)

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: all files reviewed, 27 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @stefanberger)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPud1pO6J_K9HJ5jiF8:-MQJDgaVAAS4yHU18fhW:bs98vlg) (no related file):*

I guess this is thinking aloud, right? :) Or is there any action I need to do on this PR?

In Graphene, we download and build Glibc *without* stack protector: https://github.com/oscarlab/graphene/blob/master/LibOS/Makefile#L48. Not sure if this is what you meant in your comment.
</blockquote></details>

No, just thinking out loud...

---

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: all files reviewed, 27 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/host/Linux/db_threading.c, line 94 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPugH925G54l5-Usors:-MQJDoFPAxY0TWU0SnEL:b-17hsof) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L94)):*

I can create an arch-specific new function `pal_set_tcb_stack_canary()`, would that work?

But I don't want to move `PAL_TCB::stack_protector_canary` from its offset `0x8` in the `PAL_TCB` struct. Would it be fine with you? You'll then have an unused x86-specific field in the common TCB...
</blockquote></details>

`pal_set_tcb_stack_canary` should then take `pal_tcb_linux` as first parameter so x86 can then use `PAL_TCB` and ppc64 `PAL_TCB_ARCH` to actually set it. Maybe the function should be called `pal_tcb_linux_set_stack_canary` ?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: 8 of 22 files reviewed, 25 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow, @stefanberger, and @yamahata)

___
*[LibOS/shim/src/Makefile, line 21 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIeCpa8kVQ6KcynLID:-MQLzlnd-M7B3n1cr3zc:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/LibOS/shim/src/Makefile#L21)):*

`on the fact LibOS` -> `on the fact that LibOS`
`re-uses` -> `reuses`
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_init.c, line 63 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIhAxY7rRMrXExDQ3Y:-MQM-56bBWofPFaMjbY5:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/LibOS/shim/src/shim_init.c#L63)):*

`supress` -> `suppress`
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_init.c, line 64 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIhI4q-KVY1-S9dqlD:-MQM-9dA3K2eKK9Zu8dE:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/LibOS/shim/src/shim_init.c#L64)):*

no `noreturn` here?
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_init.c, line 65 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIhtylDXO9Zmy3JZlt:-MQM-JyJCkSAyzuRoWZp:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/LibOS/shim/src/shim_init.c#L65)):*

`is corrupted at %p` -> `corruption detected` (I don't think printing the stack pointer is anyhow helpful, it's random and impossible to analyze after the crash)
</blockquote></details>

Done.
___
*[Pal/include/arch/x86_64/pal-arch.h, line 32 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIibK2AgsCWICKtYvv:-MQM-USy7jc-8fOwpcfx:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/include/arch/x86_64/pal-arch.h#L32)):*

`must be in %gs reg and at offset 0x8` -> `must be at gs:[0x8]`
</blockquote></details>

Done.
___
*[Pal/include/arch/x86_64/pal-arch.h, line 34 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIiNwD6nvJDPfVr5uj:-MQM08VE0Y5RaV263ACF:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/include/arch/x86_64/pal-arch.h#L34)):*

This struct should either be packed or we should have a `static_assert` for this field having offset == 8.
</blockquote></details>

Done.
___
*[Pal/include/arch/x86_64/Linux/pal_host-arch.h, line 22 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIjQ_c5DTyVAYcsr1Y:-MQM1YBk1rxWPVMd3eZ-:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/include/arch/x86_64/Linux/pal_host-arch.h#L22)):*

no space after `__attribute__`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_exception.c, line 45 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIkk0W07dvh8W93aKr:-MQM1ts12gKcAyDWwDNn:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_exception.c#L45)):*

`supress` -> `suppress`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_exception.c, line 46 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQInoFSEVjBYoYtjKMa:-MQM1uQxD42DciUXjGHC:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_exception.c#L46)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_exception.c, line 47 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIklj2Albmjzd7L5Fr:-MQM1unt1AtlZ6krcYbn:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_exception.c#L47)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_main.c, line 155 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIkyqb2E6fISumrMQH:-MQM1y779Bxnfk-jrwRk:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_main.c#L155)):*

no space after `__attribute__`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_threading.c, line 85 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIlmcSEPRZo0zRQgWW:-MQM21Bf3Y2QlFPGvqw2:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L85)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_threading.c, line 94 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPugH925G54l5-Usors:-MQMGmcv6DG3fDno3THx:buy78j4) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L94)):*

`pal_set_tcb_stack_canary` should then take `pal_tcb_linux` as first parameter so x86 can then use `PAL_TCB` and ppc64 `PAL_TCB_ARCH` to actually set it. Maybe the function should be called `pal_tcb_linux_set_stack_canary` ?
</blockquote></details>

Done. But I still chose the name `pal_set_tcb_stack_canary()`.
___
*[Pal/src/host/Linux/db_threading.c, line 100 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQImmWz6o7gCIw3eTZX:-MQM2t808xvs--brm6OY:bgey0fc) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L100)):*

If we'd start with this and setting TCB then we could skip the dummy TCB (and correct the cookie later, ensuring that `_DkRandomBitsRead` reads to a temporary variable, not the real cookie).
</blockquote></details>

I don't see how we can skip dummy TCB (unless we mark `_DkRandomBitsRead()` and all its called functions as `-fno-stack-protector`).

Note that when we're in this function, Linux just created a new thread, and there is nothing stored at `gs:[0x8]` (it may be even inaccessible). This function is fine because it is marked as `-fno-stack-protector`, but we need to call `_DkRandomBitsRead()` which is not marked (and we don't want to mark it because this is really hard to maintain if we ever change that func). So at runtime, there will be an access to `gs:[0x8]` in the `_DkRandomBitsRead()` prologue, and this may lead to SEGFAULT.

So it looks like we need this dummy TCB. I don't see how to circumvent this.
___
*[Pal/src/host/Linux/db_threading.c, line 100 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MPucgQH5Dj02ED7DphI:-MQM3glT7VlD6CJzbgtk:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L100)):*

Will do.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_threading.c, line 105 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQImIclD_yPrml4FVgW:-MQMIB2b4fDEGcNnHU4a:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L105)):*

Could we also zero the first byte of the canary? (i.e. make `(char*)&stack_protector_canary[0] == 0`) This prevents C-string-based stack leaks from exposing the cookie.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_exception.c, line 32 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIndstDZYrXHZa0Gyz:-MQM4Xys6nJThsU4EBK_:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_exception.c#L32)):*

`supress` -> `suppress`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_exception.c, line 33 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQInZHp3tpq4aJfrRjY:-MQM4d8_6g7QrgyI5kEz:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_exception.c#L33)):*

ditto (no `noreturn`?)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_exception.c, line 33 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIoNvhDx9KpXITbLa4:-MQM7rrcApVajRFUpWxU:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_exception.c#L33)):*

Can't we move it to host/Linux-common and deduplicate?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_exception.c, line 34 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQInhak20je5kAdD5Qr:-MQM4e0x6ARKkhgUGkTs:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_exception.c#L34)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 177 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQInxr-75FplrhNmroj:-MQM4i3p63CDqZhwQo-C:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_main.c#L177)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 455 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIo0Zw7LHZpDhM_-hd:-MQMIDNLA7NJboHbRY3M:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_main.c#L455)):*

ditto (zeroing the first byte)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_threading.c, line 57 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIoG8uA0FvVhRhYfUN:-MQM4s28CzBBeWyiFai6:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_threading.c#L57)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_threading.c, line 85 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIoiRJ2WXPdHksgawf:-MQMICiK5v5bpDQy7kfU:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/db_threading.c#L85)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/Makefile, line 109 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQIpWPE3kktm7vPL_mj:-MQM50XdBJN3yB2nfht9:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux-SGX/Makefile#L109)):*

it can be useful for detecting bugs (not as a security measure, just an aid in early detection of bugs)
</blockquote></details>

Done.

mkow (changes requested):
> Reviewed 16 of 16 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/include/arch/x86_64/pal-arch.h, line 29 at r3](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQOnjIv2xmSQHxS9vls:-MQOnjIv2xmSQHxS9vlt:b-5x5uqv) ([raw file](https://github.com/oscarlab/graphene/blob/833c4e83222e0e2fdd4b2e6f5ba11fee3fc255f2/Pal/include/arch/x86_64/pal-arch.h#L29)):*
> ```objc
> #define PAL_LIBOS_TCB_SIZE 256
> 
> #define STACK_PROTECTOR_CANARY_DEFAULT  0xbadbadbadbad00UL
> ```

No point in having `00` in this hardcoded cookie, no one needs to leak it :)
___
*[Pal/src/host/Linux/db_threading.c, line 100 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQImmWz6o7gCIw3eTZX:-MQOm5hx6hcvmdlaRac3:bhgpwb5) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L100)):*

I don't see how we can skip dummy TCB (unless we mark `_DkRandomBitsRead()` and all its called functions as `-fno-stack-protector`).

Note that when we're in this function, Linux just created a new thread, and there is nothing stored at `gs:[0x8]` (it may be even inaccessible). This function is fine because it is marked as `-fno-stack-protector`, but we need to call `_DkRandomBitsRead()` which is not marked (and we don't want to mark it because this is really hard to maintain if we ever change that func). So at runtime, there will be an access to `gs:[0x8]` in the `_DkRandomBitsRead()` prologue, and this may lead to SEGFAULT.

So it looks like we need this dummy TCB. I don't see how to circumvent this.
</blockquote></details>

What I meant is doing `stack_protector_canary = STACK_PROTECTOR_CANARY_DEFAULT` and `pal_set_tcb(&tcb->common);` at the very beginning of this function and then set up the final canary later (so, `_DkRandomBitsRead` would be called with a dummy canary and after its return we'd assign the random value to the canary).
___
*[Pal/src/host/Linux/pal_linux.h, line 170 at r3](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQOn06V-aw-w2R1W2Gq:-MQOn06V-aw-w2R1W2Gr:b63vfcr) ([raw file](https://github.com/oscarlab/graphene/blob/833c4e83222e0e2fdd4b2e6f5ba11fee3fc255f2/Pal/src/host/Linux/pal_linux.h#L170)):*
> ```objc
> 
> static inline void pal_set_tcb_stack_canary(PAL_TCB_LINUX* tcbptr, uint64_t canary) {
>     canary &= ~0xffUL; /* prevent C-string-based stack leaks from exposing the cookie */
> ```

You should actually cast it to `uint8_t*`, otherwise this snippet won't work on big endian.
___
*[Pal/src/host/Linux-SGX/sgx_tls.h, line 87 at r3](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQOndzh71QIxQqVTVqs:-MQOndzh71QIxQqVTVqt:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/833c4e83222e0e2fdd4b2e6f5ba11fee3fc255f2/Pal/src/host/Linux-SGX/sgx_tls.h#L87)):*
> ```objc
> 
> static inline void pal_set_tcb_stack_canary(uint64_t canary) {
>     canary &= ~0xffUL; /* prevent C-string-based stack leaks from exposing the cookie */
> ```

ditto

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: 18 of 22 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/include/arch/x86_64/pal-arch.h, line 29 at r3](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQOnjIv2xmSQHxS9vls:-MQRGvOLBIRNbKKKKRcI:b-whwz66) ([raw file](https://github.com/oscarlab/graphene/blob/833c4e83222e0e2fdd4b2e6f5ba11fee3fc255f2/Pal/include/arch/x86_64/pal-arch.h#L29)):*

No point in having `00` in this hardcoded cookie, no one needs to leak it :)
</blockquote></details>

Done. Reverted.
___
*[Pal/src/host/Linux/db_threading.c, line 100 at r2](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQImmWz6o7gCIw3eTZX:-MQRIajK0OJjOzj5joFA:b-2m4r8s) ([raw file](https://github.com/oscarlab/graphene/blob/256ef53592f2d4967119c2d3e66a102299d0fac5/Pal/src/host/Linux/db_threading.c#L100)):*

What I meant is doing `stack_protector_canary = STACK_PROTECTOR_CANARY_DEFAULT` and `pal_set_tcb(&tcb->common);` at the very beginning of this function and then set up the final canary later (so, `_DkRandomBitsRead` would be called with a dummy canary and after its return we'd assign the random value to the canary).
</blockquote></details>

Done. Understood now.
___
*[Pal/src/host/Linux/pal_linux.h, line 170 at r3](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQOn06V-aw-w2R1W2Gq:-MQRJ02WBcp_BeWw-Kfr:b46rjs9) ([raw file](https://github.com/oscarlab/graphene/blob/833c4e83222e0e2fdd4b2e6f5ba11fee3fc255f2/Pal/src/host/Linux/pal_linux.h#L170)):*

You should actually cast it to `uint8_t*`, otherwise this snippet won't work on big endian.
</blockquote></details>

Done. I have a feeling we have quite a few other places in Graphene where big-endian is broken...
___
*[Pal/src/host/Linux-SGX/sgx_tls.h, line 87 at r3](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQOndzh71QIxQqVTVqs:-MQRJSC1CBqXp8OKHiwy:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/833c4e83222e0e2fdd4b2e6f5ba11fee3fc255f2/Pal/src/host/Linux-SGX/sgx_tls.h#L87)):*

ditto
</blockquote></details>

Done.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux/pal_linux.h, line 170 at r3](https://reviewable.io/reviews/oscarlab/graphene/2038#-MQOn06V-aw-w2R1W2Gq:-MQRlJ1aE4Gy06DNoeNb:b-bnwvnv) ([raw file](https://github.com/oscarlab/graphene/blob/833c4e83222e0e2fdd4b2e6f5ba11fee3fc255f2/Pal/src/host/Linux/pal_linux.h#L170)):*

Done. I have a feeling we have quite a few other places in Graphene where big-endian is broken...
</blockquote></details>

Definitely, but let's try to keep this at minimum :)

mkow (approved):
> Reviewed 12 of 12 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2038)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2044, opened 2020-12-23
[Pal] Unify debug maps

pwmarcz:
## Description of the changes 

- Use the same mechanism (debug_map) in Pal/Linux and Pal/Linux-SGX.
  Previously, Pal/Linux emulated the _r_debug structure, normally
  maintained by ld.so, but that cannot be done in SGX outer PAL,
  because it's loaded by ld.so already.
- Maintain the debug maps outside of SGX enclave. This allows
  initializing them before enclave start, and potentially makes
  them easier to use.
- Initialize PAL debug map before enclave start. Previously, this
  was done from inside the enclave, so you couldn't set a
  breakpoint too early (e.g. in pal_linux_main).
- Store only load address, without list of sections. This is to
  avoid parsing the list of sections just to report them to the
  debugger. Unfortunately, the GDB version that we support still
  needs these sections, but we can retrieve them in GDB plugin.
- Move Python GDB code related to debug maps to a common file.

## How to test this PR? 

GDB integration should still work and display full stack traces; both in Linux and Linux-SGX.

In addition, it should be possible to set a breakpoint anywhere during enclave start (e.g. `handle_ecall`, `pal_linux_main`).

Note that the new integration requires Python `elftools` package (e.g. `apt install python3-pyelftools`).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)


yamahata (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: 0 of 21 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPF_1f1CdOaIiCeXmL2:-MPF_1f1CdOaIiCeXmL3:b-d6qza1) (no related file):*
some files seem missing. forgot git add?

---
___
*[LibOS/shim/test/regression/bootstrap.c, line 7 at r1](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPFZNdgEVI818RC2L7C:-MPFZNdh7cTpCLXet_i6:b-2ky6gu) ([raw file](https://github.com/oscarlab/graphene/blob/e5a35662ed51b808317c2b908fb176f3c6bb92e5/LibOS/shim/test/regression/bootstrap.c#L7)):*
> ```C
> int main(int argc, const char** argv, const char** envp) {
>     printf("User Program Started\n");
>     getchar();
> ```

leftover from debug?
___
*[Pal/lib/Makefile, line 107 at r1](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPFZTpW5KsFBZ_xN7wj:-MPFZTpW5KsFBZ_xN7wk:bweym72) ([raw file](https://github.com/oscarlab/graphene/blob/e5a35662ed51b808317c2b908fb176f3c6bb92e5/Pal/lib/Makefile#L107)):*
> ```Makefile
> 	avl_tree.o \
> 	crypto/udivmodti4.o \
> 	debug_map.o \
> ```

debug_map.c is missing.
___
*[Runtime/pal_loader, line 80 at r1](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPFZqeZ8sAXIrxGNfzU:-MPFZqeZ8sAXIrxGNfzV:bmprbce) ([raw file](https://github.com/oscarlab/graphene/blob/e5a35662ed51b808317c2b908fb176f3c6bb92e5/Runtime/pal_loader#L80)):*
> ```
> 	fi
> 	if [ 0"$SGX" -gt 0 ]; then
> 		PREFIX+=("-x" "$HOST_PAL_PATH/../../../gdb_integration/debug_map_gdb.py")
> ```

the file is mossong.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: 0 of 23 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPF_1f1CdOaIiCeXmL2:-MPFaP3B2kQGofg2wYBt:bnd2aas) (no related file):*

some files seem missing. forgot git add?
</blockquote></details>

Ouch. Right. Added now.

---
___
*[LibOS/shim/test/regression/bootstrap.c, line 7 at r1](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPFZNdgEVI818RC2L7C:-MPFaX9X26lkQ31DrSHe:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/e5a35662ed51b808317c2b908fb176f3c6bb92e5/LibOS/shim/test/regression/bootstrap.c#L7)):*

leftover from debug?
</blockquote></details>

Fixed.
___
*[Pal/lib/Makefile, line 107 at r1](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPFZTpW5KsFBZ_xN7wj:-MPFaZ5w3SLdpKB24R-a:bqmjwcz) ([raw file](https://github.com/oscarlab/graphene/blob/e5a35662ed51b808317c2b908fb176f3c6bb92e5/Pal/lib/Makefile#L107)):*

debug_map.c is missing.
</blockquote></details>

Added.
___
*[Runtime/pal_loader, line 80 at r1](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPFZqeZ8sAXIrxGNfzU:-MPFbam35es3Zqf2dD-n:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/e5a35662ed51b808317c2b908fb176f3c6bb92e5/Runtime/pal_loader#L80)):*

the file is mossong.
</blockquote></details>

Fixed.

pwmarcz (commented):
Looks like the branch conflicted with a change on master (66d8e61453da7bec667936feaf494672eaded4d8), I rebased it.

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044#-:-MPFg5XhCUMJfCu_l3Jt:bgqkv2b)* status: 0 of 23 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @yamahata)

dimakuv (changes requested):
> Reviewed 18 of 21 files at r1, 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPclBNT2jF371lCiG72:-MPclBNT2jF371lCiG73:bsayjrb) (no related file):*
Jenkins fails on trying to build the Skeleton PAL:
```
make[1]: Entering directory '/home/jenkins/workspace/graphene-18.04/Pal/src/host/Skeleton'
  ...
  [ db_rtld.o ]
db_rtld.c:8:6: error: no previous prototype for '_DkDebugMapAdd' [-Werror=missing-prototypes]
 void _DkDebugMapAdd(const char* name, void* addr) {}
      ^~~~~~~~~~~~~~
db_rtld.c:10:6: error: no previous prototype for '_DkDebugMapRemove' [-Werror=missing-prototypes]
 void _DkDebugMapRemove(void* addr) {}
      ^~~~~~~~~~~~~~~~~
  [ db_misc.o ]
```

Either remove these functions or prepend them with `static`.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcxx-754mQf8DKFp6G:-MPcxx-8D2haDrn9VN2e:b-nux7m4) (no related file):*
I dislike the new `Pal/gdb_integration/` subdirectory. It contains a single Python script. Maybe we should move it elsewhere? The candidates are:
1. `Pal/lib/` -- this is where we dump all "common" C code (bundled in a `graphene-lib.a` static lib). The GDB integration is "common" code though it has nothing to do with C and the lib...
2. `python/` -- this is where we have Python scripts (currently only SGX-specific scripts). I suggest to move `debug_map_gdb.py` here.

@mkow @woju @yamahata Any suggestions on where to put this file?

---
___
*[Pal/gdb_integration/debug_map_gdb.py, line 96 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcoKEY6S4p--sjIt1e:-MPcoKEZ11PW60Q8N3Ko:bjge8mx) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/gdb_integration/debug_map_gdb.py#L96)):*
> ```Python
>                 continue
> 
>             # Note that this doesn't escape the file names.
> ```

What is the point of this comment? Why would we care about escaping filenames?
___
*[Pal/gdb_integration/debug_map_gdb.py, line 127 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcq6OK11QQfHVL0OMw:-MPcq6OK11QQfHVL0OMx:b-x5w181) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/gdb_integration/debug_map_gdb.py#L127)):*
> ```Python
>     def stop(self):
>         gdb.execute('update-debug-maps')
>         return False
> ```

Can you add a comment that returning False means `continue automatically, without actual stopping`
___
*[Pal/gdb_integration/debug_map_gdb.py, line 141 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcqg_DBFpTWS24c9my:-MPcqg_DBFpTWS24c9mz:b-xvrjzs) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/gdb_integration/debug_map_gdb.py#L141)):*
> ```Python
> 
> def debug_map_clear_objfiles_handler(event):
>     # Record that symbol files has been cleared on GDB's side (e.g. on program exit), so that we do
> ```

`has` -> `have`
___
*[Pal/include/lib/debug_map.h, line 10 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcrRKX2KfFtxW96Z-b:-MPcrRKX2KfFtxW96Z-c:bgcrvm2) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/include/lib/debug_map.h#L10)):*
> ```objc
>  *
>  * Note that this is a separate module (and not part of libpal) because in case of SGX it is used in
>  * the outer pal-sgx process, not in libpal itself.
> ```

Separate module? I don't understand this comment. Either remove this comment altogether or rephrase (do you mean that this module is part of the "common library" under `Pal/lib` and not a Pal-specific file?).

Also, this comment is too SGX-specific. We can potentially have many different PALs, but here you only differentiate between Linux and Linux-SGX.
___
*[Pal/include/lib/debug_map.h, line 16 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcruEX-1rJ8O3i9qaR:-MPcruEYBXm8HFxaltex:b-990qq8) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/include/lib/debug_map.h#L16)):*
> ```objc
> #define PAL_DEBUG_MAP_H
> 
> #include <stdint.h>
> ```

This seems to be not needed.
___
*[Pal/src/host/Linux/db_rtld.c, line 33 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPctWTr090Tuo-LzB_z:-MPctWTs3j4rj9-v2Oj2:b-pjy59r) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/src/host/Linux/db_rtld.c#L33)):*
> ```C
> }
> 
> void setup_pal_map(struct link_map* pal_map) {
> ```

Do we even need this function and corresponding `g_pal_map` and `g_loaded_maps` now? Because we already have `g_debug_map`. Or is this needed to inform the LibOS layer about the loaded files and thus cannot be removed?

I understand that this is for another PR. Just curious if we can remove even more RTLD code.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 9 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcuFGIBiYhzQQG6Njq:-MPcuFGIBiYhzQQG6Njr:b6uzcg6) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/src/host/Linux-SGX/db_rtld.c#L9)):*
> ```C
>  * Overview of ELF files used in this host:
>  * - pal-sgx and libraries it uses (outside enclave) - handled by ld.so and reported by it (through
>  *   _r_debug mechanism)
> ```

Do we have these differences between Linux's `libpal.so` and Linux-SGX's `pal-sgx` because the former is statically linked and the latter is dynamically linked against Glibc?

I'm confused why we have `_r_debug` here (I understand that it is implicitly added by GCC when building) and don't have it in Linux PAL.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 43 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcumkcDtjD5-9xfddL:-MPcumkcDtjD5-9xfddM:b2aikof) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/src/host/Linux-SGX/db_rtld.c#L43)):*
> ```C
> }
> 
> void setup_pal_map(struct link_map* pal_map) {
> ```

Same question as in Linux PAL: why do we need this?
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 255 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcv9Ai7462EI435l5p:-MPcv9Ai7462EI435l5q:bpfmlai) ([raw file](https://github.com/oscarlab/graphene/blob/0472a678c52d1082b164f8feca72cb70eb6433e7/Pal/src/host/Linux-SGX/db_rtld.c#L255)):*
> ```C
>         goto fail;
> 
>     debug_map_add(debug_map);
> ```

For my own understanding: why didn't we have this "add all sections" in the Linux PAL in the first place? Was it because of that `_r_debug` mechanism?
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.h, line 94 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcvf3YAdwdn2sjDW1F:-MPcvf3YAdwdn2sjDW1G:b87flnt) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/src/host/Linux-SGX/enclave_ocalls.h#L94)):*
> ```objc
> int ocall_delete(const char* pathname);
> 
> int ocall_report_mmap(const char* filename, uint64_t addr, uint64_t len, uint64_t offset);
> ```

Putting new `ocall_debug_map_{add/remove}` before `ocall_report_mmap` confused git diff/Reviewable, and now it's hard to review the changes :)

Could you reorder these functions everywhere in this PR, such that `ocall_report_mmap` is the last one? This way it will be easier to review and track historical changes. And I don't think the reordering will hurt readability.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPdPPhM7xZSel1gEaOn:-MPdPPhN5XpzK-HYO4yn:bmk42fc) (no related file):*
Why doesn't this PR remove all mentions of `_r_debug`, in particular in `Pal/src/host/Linux/pal_security.h`?

---

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 18 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow, @pwmarcz, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcxx-754mQf8DKFp6G:-MPi1dRBCtGtp6w2BWx2:b-461z9r) (no related file):*

I dislike the new `Pal/gdb_integration/` subdirectory. It contains a single Python script. Maybe we should move it elsewhere? The candidates are:
1. `Pal/lib/` -- this is where we dump all "common" C code (bundled in a `graphene-lib.a` static lib). The GDB integration is "common" code though it has nothing to do with C and the lib...
2. `python/` -- this is where we have Python scripts (currently only SGX-specific scripts). I suggest to move `debug_map_gdb.py` here.

@mkow @woju @yamahata Any suggestions on where to put this file?
</blockquote></details>

`/gdb` together with other plugins/configs for gdb and call it a day. I'd advise to reduce the level of nested subdirectories as well as the verbosity of `gdb_integration`. It'll be easier to install.

Python keeps it like this: https://github.com/python/cpython/tree/c56988b88fecf6dc70f039704fda6051a0754db1/Tools/gdb.

---
___
*[Runtime/pal_loader, line 80 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPi3Rc78bK8kkVpOj87:-MPi3Rc78bK8kkVpOj88:b-f67jfe) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Runtime/pal_loader#L80)):*
> ```
> 	fi
> 	if [ 0"$SGX" -gt 0 ]; then
> 		PREFIX+=("-x" "$HOST_PAL_PATH/../../../gdb_integration/debug_map_gdb.py")
> ```

Does this still work when installed from meson?

mkow (changes requested):
> Reviewed 7 of 21 files at r1, 1 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcxx-754mQf8DKFp6G:-MPgHV397zzPATAgdJ5y:b4vjz1k) (no related file):*

`/gdb` together with other plugins/configs for gdb and call it a day. I'd advise to reduce the level of nested subdirectories as well as the verbosity of `gdb_integration`. It'll be easier to install.

Python keeps it like this: https://github.com/python/cpython/tree/c56988b88fecf6dc70f039704fda6051a0754db1/Tools/gdb.
</blockquote></details>

@dimakuv: I like the directory added in this PR, it looks more organized to me, even if it just has a single scripts. Otherwise it will be mixed will all other random files.

Ad. woju's proposal: I don't like /gdb, but no strong feelings. In case of Python, it's "Tools/gdb", not just a freestanding "gdb". If we want it toplevel then the name should be IMO more verbose than just "gdb". Btw., @woju: Please consider, that the GDB integration is not just a freestanding plugin which integrates with GDB, it's the plugin plus a lot of support code in Graphene source (which is GDB- and our integration-specific). So, having half of the implementation in Pal/ and the other matching half in a totally different place sounds bad to me.

---
___
*[Pal/src/db_rtld.c, line 891 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPgJCak5O2tphMqBFzA:-MPgJCalC89NGKbt7ryt:bfcm7qw) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/src/db_rtld.c#L891)):*
> ```C
> 
>     _DkDebugMapAdd(realname, start_addr);
>     return;
> ```

not needed?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 18 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcxx-754mQf8DKFp6G:-MPiSLn643jpBbqjUAbQ:b-3ekddi) (no related file):*

@dimakuv: I like the directory added in this PR, it looks more organized to me, even if it just has a single scripts. Otherwise it will be mixed will all other random files.

Ad. woju's proposal: I don't like /gdb, but no strong feelings. In case of Python, it's "Tools/gdb", not just a freestanding "gdb". If we want it toplevel then the name should be IMO more verbose than just "gdb". Btw., @woju: Please consider, that the GDB integration is not just a freestanding plugin which integrates with GDB, it's the plugin plus a lot of support code in Graphene source (which is GDB- and our integration-specific). So, having half of the implementation in Pal/ and the other matching half in a totally different place sounds bad to me.
</blockquote></details>

I also don't like woju's proposal on the top-level `/gdb` dir. After Michal's comment, I also agree that putting this script under `python/` is not a good idea. So I guess the current dir is fine.

---

yamahata (commented):
> Reviewed 18 of 21 files at r1, 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz and @yamahata)

___
*[Pal/gdb_integration/debug_map_gdb.py, line 137 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPkCghGESdQa1QKq86j:-MPkCghGESdQa1QKq86k:b-blom3e) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/gdb_integration/debug_map_gdb.py#L137)):*
> ```
> 'updated'
> ```
nitpick: updated is a bit terse. Can you please be a bit meaningful. (or delete this print)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: 14 of 29 files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, @woju, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPclBNT2jF371lCiG72:-MQHGyOu0zwRQtAT-ME4:b33b21s) (no related file):*

Jenkins fails on trying to build the Skeleton PAL:
```
make[1]: Entering directory '/home/jenkins/workspace/graphene-18.04/Pal/src/host/Skeleton'
  ...
  [ db_rtld.o ]
db_rtld.c:8:6: error: no previous prototype for '_DkDebugMapAdd' [-Werror=missing-prototypes]
 void _DkDebugMapAdd(const char* name, void* addr) {}
      ^~~~~~~~~~~~~~
db_rtld.c:10:6: error: no previous prototype for '_DkDebugMapRemove' [-Werror=missing-prototypes]
 void _DkDebugMapRemove(void* addr) {}
      ^~~~~~~~~~~~~~~~~
  [ db_misc.o ]
```

Either remove these functions or prepend them with `static`.
</blockquote></details>

I added the `#include` that declares these (`pal_rtld.h`).

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPdPPhM7xZSel1gEaOn:-MQHI2Vp8cF_acy8xi9r:bm0yv8b) (no related file):*

Why doesn't this PR remove all mentions of `_r_debug`, in particular in `Pal/src/host/Linux/pal_security.h`?
</blockquote></details>

Right. I removed that.

(There is also the `r_debug` structure in LibOS - as far as I can tell, it's used to keep track of the debug maps and re-reporting it to PAL after a fork. It could be simpler, but it doesn't hurt too much, and it's basically unrelated to this change.)

---
___
*[Pal/gdb_integration/debug_map_gdb.py, line 96 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcoKEY6S4p--sjIt1e:-MQHL5wXCBIyhg8J2JE_:bqhlgi6) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/gdb_integration/debug_map_gdb.py#L96)):*

What is the point of this comment? Why would we care about escaping filenames?
</blockquote></details>

We pass the file name as argument to `add-symbol-file`, so files with spaces and special characters (like `#`) might not work correctly.

Actually, from what I can tell, GDB uses a shell-like syntax, and whatever `shlex.quote` does works, at least for simple cases that I checked (spaces, `#`, `"`, `'`).  Updated.
___
*[Pal/gdb_integration/debug_map_gdb.py, line 127 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcq6OK11QQfHVL0OMw:-MQHN6_sBbYdx28K5uZD:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/gdb_integration/debug_map_gdb.py#L127)):*

Can you add a comment that returning False means `continue automatically, without actual stopping`
</blockquote></details>

Done.
___
*[Pal/gdb_integration/debug_map_gdb.py, line 137 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPkCghGESdQa1QKq86j:-MQHN7TvDECbzoMfnYi-:b9xn0u6) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/gdb_integration/debug_map_gdb.py#L137)):*

> ```
> 'updated'
> ```
nitpick: updated is a bit terse. Can you please be a bit meaningful. (or delete this print)
</blockquote></details>

Deleted, this was a debug print.
___
*[Pal/gdb_integration/debug_map_gdb.py, line 141 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcqg_DBFpTWS24c9my:-MQHNCTFCw8BG3dtvKyj:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/gdb_integration/debug_map_gdb.py#L141)):*

`has` -> `have`
</blockquote></details>

Done.
___
*[Pal/include/lib/debug_map.h, line 10 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcrRKX2KfFtxW96Z-b:-MQHNGAR2er1YdPIUtXp:burjmzb) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/include/lib/debug_map.h#L10)):*
>do you mean that this module is part of the "common library" under Pal/lib and not a Pal-specific file?

Yes. I reworded the comment.
___
*[Pal/include/lib/debug_map.h, line 16 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcruEX-1rJ8O3i9qaR:-MQHNzZADzl2sy5VQJMV:bw0lrkx) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/include/lib/debug_map.h#L16)):*

This seems to be not needed.
</blockquote></details>

Yes, it's a leftover from an intermediate version. Removed.
___
*[Pal/src/db_rtld.c, line 891 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPgJCak5O2tphMqBFzA:-MQHOGY43_8kM5hsrBVz:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/src/db_rtld.c#L891)):*

not needed?
</blockquote></details>

Fixed.
___
*[Pal/src/host/Linux/db_rtld.c, line 33 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPctWTr090Tuo-LzB_z:-MQHOIl04N-Ipdm5Uing:b-1i68qf) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/src/host/Linux/db_rtld.c#L33)):*

Do we even need this function and corresponding `g_pal_map` and `g_loaded_maps` now? Because we already have `g_debug_map`. Or is this needed to inform the LibOS layer about the loaded files and thus cannot be removed?

I understand that this is for another PR. Just curious if we can remove even more RTLD code.
</blockquote></details>

I think `g_loaded_map` is currently used because the code actually does some dynamic linking, I see it in the symbol lookup code (`do_lookup`). Presumably we at least need to resolve symbols from PAL when loading LibOS.

But yes, maybe that code can be simplified - my goal is to remove the glibc-originated code and replace it with something simpler. This PR is just the first step to remove some dependencies out of the way.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 9 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcuFGIBiYhzQQG6Njq:-MQHPdns2Xd_m5328KbN:bypuea8) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/src/host/Linux-SGX/db_rtld.c#L9)):*
>Do we have these differences between Linux's `libpal.so` and Linux-SGX's `pal-sgx` because the former is statically linked and the latter is dynamically linked against Glibc?

Yes. Also, Linux has just `libpal.so`, and Linux-SGX is split between two binaries (outer `pal-sgx` and inner `libpal.so`).

>I understand that it is implicitly added by GCC when building

No, but it comes from ld.so (and that's why you don't have it when loading a static binary):

```
(gdb) p &_r_debug
$1 = (struct r_debug *) 0x7754d0c73140 <_r_debug>
(gdb) info files
	0x00007754d0c72fe0 - 0x00007754d0c73170 is .bss in /lib64/ld-linux-x86-64.so.2
```

Previously, Linux `libpal.so` exported its own `_r_debug` so that GDB could use it. But you can't really do that in case of `pal-sgx` (since the `_r_debug` is already there, and adding your own maps to it would probably break things), so you need a custom mechanism anyway. This PR simplifies things by using only that custom mechanism.

(And I kind of feel better about the custom mechanism than about emulating `_r_debug`; despite the amount of Python code it feels more straightforward.)
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 43 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcumkcDtjD5-9xfddL:-MQHUtNM32FMewcOsdiJ:b2xgv9v) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/src/host/Linux-SGX/db_rtld.c#L43)):*

Same question as in Linux PAL: why do we need this?
</blockquote></details>

Same as above - the common `db_rtld.c` seems to use it.
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 255 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcv9Ai7462EI435l5p:-MQHUysUC8TjIStctWP6:bszcr80) ([raw file](https://github.com/oscarlab/graphene/blob/0472a678c52d1082b164f8feca72cb70eb6433e7/Pal/src/host/Linux-SGX/db_rtld.c#L255)):*

For my own understanding: why didn't we have this "add all sections" in the Linux PAL in the first place? Was it because of that `_r_debug` mechanism?
</blockquote></details>

Yes. From what I understand, the sections were needed only for GDB's `add-symbol-file` command, and the other PAL was added using `_r_debug` mechanism.
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.h, line 94 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcvf3YAdwdn2sjDW1F:-MQHVZOVCIYWePouqxsc:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/src/host/Linux-SGX/enclave_ocalls.h#L94)):*

Putting new `ocall_debug_map_{add/remove}` before `ocall_report_mmap` confused git diff/Reviewable, and now it's hard to review the changes :)

Could you reorder these functions everywhere in this PR, such that `ocall_report_mmap` is the last one? This way it will be easier to review and track historical changes. And I don't think the reordering will hurt readability.
</blockquote></details>

Done.
___
*[Runtime/pal_loader, line 80 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPi3Rc78bK8kkVpOj87:-MQHVg6l01QOVX10sgfe:b-t06zbv) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Runtime/pal_loader#L80)):*

Does this still work when installed from meson?
</blockquote></details>

I see, meson copies the `gdb_integration` directory but not this one.

For now, I modified it so that
- both files (`debug_maps.py` and `graphene_sgx_gdb.py`) are expected in host `gdb_integration` directory,
- when installing from meson, we copy from both directories,
- when running Graphene directly from working directory, the`debug_maps.py` file is a symlink.

I realize this is hacky and I'm open to better suggestions.

dimakuv (changes requested):
> Reviewed 15 of 15 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow, @pwmarcz, @woju, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPclBNT2jF371lCiG72:-MQHmFDm0KKz3ieBESaP:b-nsw0c7) (no related file):*

I added the `#include` that declares these (`pal_rtld.h`).
</blockquote></details>

Now Meson doesn't build properly:
```
[1/1] Installing files.
Installing subdir /home/jenkins/workspace/graphene/Runtime/../Pal/src/host/Linux/gdb_integration to /home/jenkins/workspace/graphene/install/usr/local/libexec/graphene/linux/gdb_integration
Installing subdir /home/jenkins/workspace/graphene/Runtime/../Pal/gdb_integration to /home/jenkins/workspace/graphene/install/usr/local/libexec/graphene/linux/gdb_integration
Traceback (most recent call last):
  File "/usr/local/bin/meson", line 26, in <module>
    sys.exit(main())
  File "/usr/local/bin/meson", line 23, in main
    return mesonmain.run(sys.argv[1:], launcher)
  File "/usr/local/lib/python3.5/dist-packages/mesonbuild/mesonmain.py", line 318, in run
    sys.exit(run_script_command(args[1:]))
  File "/usr/local/lib/python3.5/dist-packages/mesonbuild/mesonmain.py", line 274, in run_script_command
    return cmdfunc(cmdargs)
  File "/usr/local/lib/python3.5/dist-packages/mesonbuild/scripts/meson_install.py", line 393, in run
    do_install(datafilename)
  File "/usr/local/lib/python3.5/dist-packages/mesonbuild/scripts/meson_install.py", line 211, in do_install
    install_subdirs(d) # Must be first, because it needs to delete the old subtree.
  File "/usr/local/lib/python3.5/dist-packages/mesonbuild/scripts/meson_install.py", line 224, in install_subdirs
    do_copydir(d, src_dir, full_dst_dir, exclude)
  File "/usr/local/lib/python3.5/dist-packages/mesonbuild/scripts/meson_install.py", line 193, in do_copydir
    shutil.copy2(abs_src, abs_dst, follow_symlinks=False)
  File "/usr/lib/python3.5/shutil.py", line 251, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.5/shutil.py", line 115, in copyfile
    with open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: '/home/jenkins/workspace/graphene/install/usr/local/libexec/graphene/linux/gdb_integration/debug_map_gdb.py'
FAILED: /usr/bin/python3 /usr/local/bin/meson --internal install /home/jenkins/workspace/graphene/build/meson-private/install.dat
ninja: build stopped: subcommand failed.
```

---
___
*[Pal/src/host/Linux-SGX/db_rtld.c, line 9 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPcuFGIBiYhzQQG6Njq:-MQHnX2P3xTAdHyR8C4t:b30hhvz) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Pal/src/host/Linux-SGX/db_rtld.c#L9)):*

>Do we have these differences between Linux's `libpal.so` and Linux-SGX's `pal-sgx` because the former is statically linked and the latter is dynamically linked against Glibc?

Yes. Also, Linux has just `libpal.so`, and Linux-SGX is split between two binaries (outer `pal-sgx` and inner `libpal.so`).

>I understand that it is implicitly added by GCC when building

No, but it comes from ld.so (and that's why you don't have it when loading a static binary):

```
(gdb) p &_r_debug
$1 = (struct r_debug *) 0x7754d0c73140 <_r_debug>
(gdb) info files
	0x00007754d0c72fe0 - 0x00007754d0c73170 is .bss in /lib64/ld-linux-x86-64.so.2
```

Previously, Linux `libpal.so` exported its own `_r_debug` so that GDB could use it. But you can't really do that in case of `pal-sgx` (since the `_r_debug` is already there, and adding your own maps to it would probably break things), so you need a custom mechanism anyway. This PR simplifies things by using only that custom mechanism.

(And I kind of feel better about the custom mechanism than about emulating `_r_debug`; despite the amount of Python code it feels more straightforward.)
</blockquote></details>

OK. Thanks for these explanations!

yamahata (approved):
> Reviewed 9 of 15 files at r3, 11 of 11 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow, @pwmarcz, and @woju)

dimakuv (approved):
> Reviewed 11 of 11 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

dimakuv (approved):
> Reviewed 7 of 7 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow and @woju)

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow, @pwmarcz, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MQvvxnoA4y4XSPNcTb1:-MQvvxnoA4y4XSPNcTb2:b6s9661) (no related file):*
This PR fails Jenkins-Debug (on Ubuntu 16.04) on the following tests:
```
test_libos.TC_50_GDB.test_000_gdb_backtrace
test_libos.TC_50_GDB.test_010_regs_x86_64
```

For some details, here is the log for the `gdb_backtrace` failure:
```
AssertionError: None is not true : backtrace 1 not found in GDB output

[graphene_gdb.py] Loading /home/jenkins/workspace/graphene-debug/Pal/src/host/Linux/gdb_integration/graphene.gdb...
Breakpoint 1 at 0x17720: file pal_start-x86_64.S, line 13.
Breakpoint 2 at 0x17780: file clone-x86_64.S, line 86.
Catchpoint 3 (vfork)
Temporary breakpoint 4 (func) pending.
add symbol table from file "/home/jenkins/workspace/graphene-debug/LibOS/shim/src/libsysdb.so" at
	.text_addr = 0x7ff6ffc68700
	...
add symbol table from file "/home/jenkins/workspace/graphene-debug/LibOS/shim/test/regression/debug" at
	.text_addr = 0x8004b0
	...
add symbol table from file "/home/jenkins/workspace/graphene-debug/Runtime/ld-linux-x86-64.so.2" at
	.text_addr = 0x2db5944b70f0
	...
[New LWP 20911]
[Switching to LWP 20911]
add symbol table from file "/home/jenkins/workspace/graphene-debug/Runtime/libc.so.6" at
	.text_addr = 0x2db5941222b0
	...
quit
A debugging session is active.

	Inferior 1 [process 20907] will be killed.

Quit anyway? (y or n) [answered Y; input not from terminal]


Standard Error
Function "func" not defined.
Error in re-setting breakpoint 4: Cannot access memory at address 0x8005a6
Error in re-setting breakpoint 4: Cannot access memory at address 0x8005a6
Error in re-setting breakpoint 4: Cannot access memory at address 0x8005a6
``` 

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: 30 of 32 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2044#-MQvvxnoA4y4XSPNcTb1:-MQwGXsgBSzX0aSHCq0o:b-85i61g) (no related file):*

This PR fails Jenkins-Debug (on Ubuntu 16.04) on the following tests:
```
test_libos.TC_50_GDB.test_000_gdb_backtrace
test_libos.TC_50_GDB.test_010_regs_x86_64
```

For some details, here is the log for the `gdb_backtrace` failure:
```
AssertionError: None is not true : backtrace 1 not found in GDB output

[graphene_gdb.py] Loading /home/jenkins/workspace/graphene-debug/Pal/src/host/Linux/gdb_integration/graphene.gdb...
Breakpoint 1 at 0x17720: file pal_start-x86_64.S, line 13.
Breakpoint 2 at 0x17780: file clone-x86_64.S, line 86.
Catchpoint 3 (vfork)
Temporary breakpoint 4 (func) pending.
add symbol table from file "/home/jenkins/workspace/graphene-debug/LibOS/shim/src/libsysdb.so" at
	.text_addr = 0x7ff6ffc68700
	...
add symbol table from file "/home/jenkins/workspace/graphene-debug/LibOS/shim/test/regression/debug" at
	.text_addr = 0x8004b0
	...
add symbol table from file "/home/jenkins/workspace/graphene-debug/Runtime/ld-linux-x86-64.so.2" at
	.text_addr = 0x2db5944b70f0
	...
[New LWP 20911]
[Switching to LWP 20911]
add symbol table from file "/home/jenkins/workspace/graphene-debug/Runtime/libc.so.6" at
	.text_addr = 0x2db5941222b0
	...
quit
A debugging session is active.

	Inferior 1 [process 20907] will be killed.

Quit anyway? (y or n) [answered Y; input not from terminal]


Standard Error
Function "func" not defined.
Error in re-setting breakpoint 4: Cannot access memory at address 0x8005a6
Error in re-setting breakpoint 4: Cannot access memory at address 0x8005a6
Error in re-setting breakpoint 4: Cannot access memory at address 0x8005a6
``` 
</blockquote></details>

Looks like the wrong address was passed from LibOS. Should be fixed now.

---

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

mkow (changes requested):
Awesome PR!

> Reviewed 3 of 21 files at r1, 2 of 6 files at r2, 6 of 15 files at r3, 10 of 11 files at r4, 6 of 7 files at r5, 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044#-:-MR2ca8I836AsDMH5yZ5:b-51d0kj)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz and @woju)

___
*[Pal/gdb_integration/debug_map_gdb.py, line 4 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MQync_H68F2A-Jfb9Ko:-MQync_H68F2A-Jfb9Kp:bj7ao3v) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/gdb_integration/debug_map_gdb.py#L4)):*
> ```Python
> # SPDX-License-Identifier: LGPL-3.0-or-later */
> # Copyright (C) 2020 Intel Corporation
> #                    Paweł Marczewski <mkow@invisiblethingslab.com>
> ```

wrong email :)
___
*[Pal/gdb_integration/debug_map_gdb.py, line 31 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MQys0143TSlkQCXr_ok:-MQys0143TSlkQCXr_ol:b-9dll9y) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/gdb_integration/debug_map_gdb.py#L31)):*
> ```
> {name: addr} elements
> ```

This sounds like a bad idea, AFAIK there's no guarantee that section names are unique. This probably won't happen with the workloads we run, but I'd prefer to not assume this.
Does GDB really look at the section names to find .text section? Not some attributes? I thought that both section and segment names are only for the user and are irrelevant?
___
*[Pal/gdb_integration/debug_map_gdb.py, line 109 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR16jODBel2l6BQ1zr6:-MR16jODBel2l6BQ1zr7:beyzq3t) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/gdb_integration/debug_map_gdb.py#L109)):*
> ```Python
>                 file_name, sections = old[load_addr]
>                 print("Removing symbol file (was {}) from addr: 0x{:x}".format(
>                     file_name, load_addr))
> ```

please align this to `(`
___
*[Pal/gdb_integration/debug_map_gdb.py, line 133 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR17C2X1rhPdxQ5aXHH:-MR17C2X1rhPdxQ5aXHI:b-mmm6n0) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/gdb_integration/debug_map_gdb.py#L133)):*
> ```
> internal=1
> ```

Not `True`? OTOH, I can't find valid values for this argument in the GDB docs :/
___
*[Pal/include/lib/debug_map.h, line 31 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR19Ypz8zwPVvF1mcQu:-MR19Yq-D6sfxadkKUln:b-ym0jfv) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/include/lib/debug_map.h#L31)):*
> ```objc
> int debug_map_remove(void* addr);
> 
> #endif
> ```

`#endif` -> `#endif /* PAL_DEBUG_MAP_H */`
___
*[Pal/lib/debug_map.c, line 8 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR2YJqp-h6P9mRHLDaR:-MR2YJqp-h6P9mRHLDaS:b1cy4rx) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/lib/debug_map.c#L8)):*
> ```C
> #include <asm/errno.h>
> 
> #include "debug_map.h"
> ```

Please move this to the very beginning (the matching header should be first, in its own group)
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1471 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR2eLsPEuKNQfFQyO_s:-MR2eLsQ7CfvBNRxp9vS:b-9aptxc) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1471)):*
> ```C
> }
> 
> int ocall_debug_map_add(const char* name, void* addr) {
> ```

Just to be sure that we don't introduce accidental leakage, could you no-op both this and `ocall_debug_map_remove` in non-debug builds?
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1477 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR2dG7HBL6sYO9s5__D:-MR2dG7HBL6sYO9s5__E:b-vg3oky) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1477)):*
> ```C
> 
>      void* old_ustack = sgx_prepare_ustack();
>      ms = sgx_alloc_on_ustack_aligned(sizeof(*ms), alignof(*ms));
> ```

Accidental change?

woju (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, 8 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Runtime/pal_loader, line 80 at r2](https://reviewable.io/reviews/oscarlab/graphene/2044#-MPi3Rc78bK8kkVpOj87:-MR2g-Iz9mbO_rUvX_Ri:bi67s9o) ([raw file](https://github.com/oscarlab/graphene/blob/999032e6466e25a70af7bde8e2799004b97292b8/Runtime/pal_loader#L80)):*

I see, meson copies the `gdb_integration` directory but not this one.

For now, I modified it so that
- both files (`debug_maps.py` and `graphene_sgx_gdb.py`) are expected in host `gdb_integration` directory,
- when installing from meson, we copy from both directories,
- when running Graphene directly from working directory, the`debug_maps.py` file is a symlink.

I realize this is hacky and I'm open to better suggestions.
</blockquote></details>

The current solution is OK, because those will be installed from two different `.deb` packages. If you'd try to deduplicate this, we'd have to introduce a third, common package for both of them.

Alternatively `debug_maps.py` could live in `/python/graphenelibos` and be installed into system-wide python library, so it could be just imported from inside gdb.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: 28 of 32 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @yamahata)

___
*[Pal/gdb_integration/debug_map_gdb.py, line 4 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MQync_H68F2A-Jfb9Ko:-MR4fK-x74br-L9XSZ2_:b-knu9jr) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/gdb_integration/debug_map_gdb.py#L4)):*

wrong email :)
</blockquote></details>

Too much copy-paste... Fixed.
___
*[Pal/gdb_integration/debug_map_gdb.py, line 31 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MQys0143TSlkQCXr_ok:-MR4fTRe3bU6nL59MQ2O:b-q2scm1) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/gdb_integration/debug_map_gdb.py#L31)):*

> ```
> {name: addr} elements
> ```

This sounds like a bad idea, AFAIK there's no guarantee that section names are unique. This probably won't happen with the workloads we run, but I'd prefer to not assume this.
Does GDB really look at the section names to find .text section? Not some attributes? I thought that both section and segment names are only for the user and are irrelevant?
</blockquote></details>

I didn't read the code too deeply but [apparently it does look at the sections by name](https://github.com/bminor/binutils-gdb/blob/gdb-8.2-release/gdb/symfile.c#L223). The GDB command `info files` will also show you a list of sections mapped in address space, so I assume they're pretty important for GDB.

And the first offset we provide to the command is indeed [treated as the offset of `.text` section, by name](https://github.com/bminor/binutils-gdb/blob/gdb-8.2-release/gdb/symfile.c#L2165). (As I understand it, the problem is that before GDB 8.2 you can either manually provide all section addresses, or no address at all - there is no option to provide a global offset of `load_addr` which is what we need).

Anyway, I changed the dictionary to a list of tuples.
___
*[Pal/gdb_integration/debug_map_gdb.py, line 109 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR16jODBel2l6BQ1zr6:-MR4kd-K3r4Kcz7btelo:bcuqesm) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/gdb_integration/debug_map_gdb.py#L109)):*

please align this to `(`
</blockquote></details>

Are you sure? That would suggest `file_name, load_addr` are arguments to `print()`, not to the nested `format()` invocation.

Asking mostly because that's the indentation suggested by my editor, I'm not sure but I think [PEP8 also supports me](https://www.python.org/dev/peps/pep-0008/#indentation) (this is a "hanging indent" which should be indented by 4 spaces). But I won't insist, linters don't complain either way.
___
*[Pal/gdb_integration/debug_map_gdb.py, line 133 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR17C2X1rhPdxQ5aXHH:-MR4mE6dE-tjsaFKniU1:b-feew55) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/gdb_integration/debug_map_gdb.py#L133)):*

> ```
> internal=1
> ```

Not `True`? OTOH, I can't find valid values for this argument in the GDB docs :/
</blockquote></details>

Yeah, it looks like that should be `True` (1 was there before, and Python doesn't care either way). Fixed.
___
*[Pal/include/lib/debug_map.h, line 31 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR19Ypz8zwPVvF1mcQu:-MR4n47tAPXK2sBEIIGa:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/include/lib/debug_map.h#L31)):*

`#endif` -> `#endif /* PAL_DEBUG_MAP_H */`
</blockquote></details>

Done.
___
*[Pal/lib/debug_map.c, line 8 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR2YJqp-h6P9mRHLDaR:-MR4nZfU2IsBbJzK24Fj:b-nvbf93) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/lib/debug_map.c#L8)):*

Please move this to the very beginning (the matching header should be first, in its own group)
</blockquote></details>

I didn't know that, thanks. Done.
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1471 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR2eLsPEuKNQfFQyO_s:-MR4oCbx5Tx2xVNyXQDV:b-kmwpw5) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1471)):*

Just to be sure that we don't introduce accidental leakage, could you no-op both this and `ocall_debug_map_remove` in non-debug builds?
</blockquote></details>

Done, and I did the same for `ocall_report_mmap` as it serves a similar purpose.
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1477 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR2dG7HBL6sYO9s5__D:-MR4olif5R8vuby3CkLu:bck71j7) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1477)):*

Accidental change?
</blockquote></details>

You mean the indentation? Fixed, thanks.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[Pal/gdb_integration/debug_map_gdb.py, line 31 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MQys0143TSlkQCXr_ok:-MRGeIcmA-K4YipQrQ3Q:b-4ls36d) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/gdb_integration/debug_map_gdb.py#L31)):*

I didn't read the code too deeply but [apparently it does look at the sections by name](https://github.com/bminor/binutils-gdb/blob/gdb-8.2-release/gdb/symfile.c#L223). The GDB command `info files` will also show you a list of sections mapped in address space, so I assume they're pretty important for GDB.

And the first offset we provide to the command is indeed [treated as the offset of `.text` section, by name](https://github.com/bminor/binutils-gdb/blob/gdb-8.2-release/gdb/symfile.c#L2165). (As I understand it, the problem is that before GDB 8.2 you can either manually provide all section addresses, or no address at all - there is no option to provide a global offset of `load_addr` which is what we need).

Anyway, I changed the dictionary to a list of tuples.
</blockquote></details>

Huh, interesting. Thanks for changing this.
___
*[Pal/gdb_integration/debug_map_gdb.py, line 109 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR16jODBel2l6BQ1zr6:-MRGfGHk2jEOQ7FHyIug:btzjmud) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/gdb_integration/debug_map_gdb.py#L109)):*

Are you sure? That would suggest `file_name, load_addr` are arguments to `print()`, not to the nested `format()` invocation.

Asking mostly because that's the indentation suggested by my editor, I'm not sure but I think [PEP8 also supports me](https://www.python.org/dev/peps/pep-0008/#indentation) (this is a "hanging indent" which should be indented by 4 spaces). But I won't insist, linters don't complain either way.
</blockquote></details>

Ah, right, I missed the `.format()` invocation. Resolving then.
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1477 at r6](https://reviewable.io/reviews/oscarlab/graphene/2044#-MR2dG7HBL6sYO9s5__D:-MRGfcbo0HVmdsxxybjX:bb7867z) ([raw file](https://github.com/oscarlab/graphene/blob/0505ae576388b4d3f6da718a45cb41779120687f/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1477)):*

You mean the indentation? Fixed, thanks.
</blockquote></details>

Yup.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (approved):
> Reviewed 5 of 5 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> Reviewed 5 of 5 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2044)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 1549 at r8](https://reviewable.io/reviews/oscarlab/graphene/2044#-MRJKpZu3iAU9pH7B9uw:-MRJKpZvEDmhHsSHRSkj:bb13ydp) ([raw file](https://github.com/oscarlab/graphene/blob/5ef9bdc8618b614ec402e27ab42bbe514ddb671f/Pal/src/host/Linux-SGX/enclave_ocalls.c#L1549)):*
> ```C
>     WRITE_ONCE(ms->ms_offset, offset);
> 
>     retval = sgx_exitless_ocall(OCALL_REPORT_MMAP, ms);
> ```

FYI: I just noticed that this OCALL is not wrapped in a `do while -EINTR` loop, but it's a debugging feature and not too important. I don't want to fix and rebase this PR because of this, so let this pass (but probably fix in a later PR).

PR #2049, opened 2020-12-29
[Pal/Linux-SGX] Enable ENCLAVE_ADD_PAGES on SGX drivers than cap to 1MB

dimakuv:
## Description of the changes 

Some versions of the Intel SGX driver have a 1MB (256 pages) cap implicitly (and erroneously) enforced on `ioctl(SGX_IOC_ENCLAVE_ADD_PAGES)`. This PR lets Graphene support these subpar versions by always
requesting no more than 1MB of enclave pages to be added at a time.

Two arguments against this PR:
1. This is a fix for basically "broken temporary" SGX driver versions (I don't even know the exact versions/commits, somewhere around v40-v41). One day everyone will migrate to the stable version and this will be over...
2. This fix slightly regresses all Graphene enclave startups: previously, one `ioctl(ADD_PAGES)` could add the whole 8GB enclave, now you'll need a bunch of such ioctls. On the other hand, the overhead of 1MB-capped ioctl is negligible anyway. The real bottleneck is adding the pages themselves at the kernel/HW level.

One argument for this PR:
1. We won't be asked why Graphene fails with error `EADD returned -16` on seemingly good systems.

## How to test this PR? 

All tests must pass. For fun, try all intermediate versions of the in-kernel Intel SGX driver and verify they all work with Graphene.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2049)

### DIFF
--- a/Pal/src/host/Linux-SGX/sgx_framework.c
+++ b/Pal/src/host/Linux-SGX/sgx_framework.c
@@ -283,7 +283,7 @@ int add_pages_to_enclave(sgx_arch_secs_t* secs, void* addr, void* user_addr, uns
     struct sgx_enclave_add_pages param = {
         .offset  = (uint64_t)addr,
         .src     = (uint64_t)(user_addr ?: g_zero_pages),
-        .length  = size,
+        .length  = 0 /* dummy, will be calculated below */,
         .secinfo = (uint64_t)&secinfo,
         .flags   = skip_eextend ? 0 : SGX_PAGE_MEASURE,
         .count   = 0, /* output parameter, will be checked after IOCTL */
@@ -300,7 +300,10 @@ int add_pages_to_enclave(sgx_arch_secs_t* secs, void* addr, void* user_addr, uns
      * field of struct and thus may stay redundant (and unused by driver v39). We hope that this
      * contrived logic won't be needed when the SGX driver stabilizes its ioctl interface.
      * (https://git.kernel.org/pub/scm/linux/kernel/git/jarkko/linux-sgx.git/tag/?h=v39) */
-    while (param.length > 0) {
+    uint64_t rest_size = size;
+    while (rest_size > 0) {
+        param.length = rest_size > SGX_DRIVER_MAX_ADDPAGES_SIZE ? SGX_DRIVER_MAX_ADDPAGES_SIZE
+                                                                : rest_size;
         ret = INLINE_SYSCALL(ioctl, 3, g_isgx_device, SGX_IOC_ENCLAVE_ADD_PAGES, &param);
         if (IS_ERR(ret)) {
             if (ret == -EINTR)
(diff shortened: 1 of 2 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2049)* status: 0 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[Pal/src/host/Linux-SGX/sgx_framework.c, line 317 at r1](https://reviewable.io/reviews/oscarlab/graphene/2049#-MPhrwtpD8GovrcYq3ln:-MPhrwtpD8GovrcYq3lo:b38pdiu) ([raw file](https://github.com/oscarlab/graphene/blob/5ab84012dd4c017f9daee70d43216b813e888486/Pal/src/host/Linux-SGX/sgx_framework.c#L317)):*
> ```C
>         uint64_t added_size = ret > 0 ? (uint64_t)ret : param.count;
>         if (!added_size) {
>             /* some versions of Intel SGX driver do not report how many pages/bytes were added */
> ```

Yes, don't be surprised, it is a bug in that particular version of the Intel SGX driver: it doesn't report an error if you supply more than 1MB worth of enclave pages but simply **ignores** the pages that exceed this cap of 1MB.

yamahata (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2049)* status: 0 of 2 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2049#-MPiDMwJDEgvMNZ80QJU:-MPiDMwJDEgvMNZ80QJV:bl8zlp3) (no related file):*
Can the error  be distinguishable from other error?
If yes, we can emit warning that the driver is development quality and upgrade the SGX driver.  And graphene exits.
If no, this change seems reasonable to avoid such query.
(-16=-EBUSY seems like the answer is no.)

---
___
*[Pal/src/host/Linux-SGX/sgx_framework.c, line 305 at r1](https://reviewable.io/reviews/oscarlab/graphene/2049#-MPiDFNeEIj7eFHK2Hgr:-MPiDFNeEIj7eFHK2Hgs:bxhyrd3) ([raw file](https://github.com/oscarlab/graphene/blob/5ab84012dd4c017f9daee70d43216b813e888486/Pal/src/host/Linux-SGX/sgx_framework.c#L305)):*
> ```
>         param.length = rest_size > SGX_DRIVER_MAX_ADDPAGES_SIZE ? SGX_DRIVER_MAX_ADDPAGES_SIZE
>                                                                 : rest_size;
> ```

nitpick: MIN(SGX_DRIVER_MAX_ADDPAGS_SIZE, rest_size)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2049)* status: 0 of 2 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/sgx_framework.c, line 317 at r1](https://reviewable.io/reviews/oscarlab/graphene/2049#-MPhrwtpD8GovrcYq3ln:-MQ96iq17hhceCb4aRr2:btcamuc) ([raw file](https://github.com/oscarlab/graphene/blob/5ab84012dd4c017f9daee70d43216b813e888486/Pal/src/host/Linux-SGX/sgx_framework.c#L317)):*

Yes, don't be surprised, it is a bug in that particular version of the Intel SGX driver: it doesn't report an error if you supply more than 1MB worth of enclave pages but simply **ignores** the pages that exceed this cap of 1MB.
</blockquote></details>

As discussed on Slack: My proposal is to just add some more info to the error we print when Graphene runs on this buggy driver, instead of working around the bug.

Issue #2058, opened 2021-01-05
[Pal/Linux-SGX] Calling send() from multiple threads causes error in mbedTLS


lejunzhu:
## Description of the problem

When the application creates a socket pair and calls send() from more than one thread simultaneously, sometimes recv() at the other end of the pipe will receive invalid data and returns -13 (permission denied).

Tracing the problem leads to:

https://github.com/oscarlab/graphene/blob/345d271e663725d58ce941fc49f08ebc02c32fa2/Pal/lib/crypto/adapters/mbedtls_adapter.c#L504-L509

Here mbedtls_ssl_write() is called without any lock protection. Because writing to the same mbedTLS context is not thread-safe, it triggers a race condition inside mbedTLS and causes encrypted data to be sent prematurely. 

Adding a spinlock (perhaps should be a futex?) around mbedtls_ssl_write() makes the problem go away.

This is found when testing Python asyncio, which internally uses one-byte write to signal the waiting thread. Errors like this appear in the log, which is caused by this same problem:

```
Exception in callback BaseSelectorEventLoop._read_from_self()
handle: <Handle BaseSelectorEventLoop._read_from_self()>
Traceback (most recent call last):
  File "/usr/lib/python3.7/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/lib/python3.7/asyncio/selector_events.py", line 119, in _read_from_self
    data = self._ssock.recv(4096)
PermissionError: [Errno 13] Permission denied
```

## Steps to reproduce

Run the C program attached, with and without Graphene SGX 

[test.txt](https://github.com/oscarlab/graphene/files/5767402/test.txt)

## Expected results

When running without Graphene, the program will report:

c1=100000, c2=100000, e13=0

## Actual results

Running with Graphene SGX, the result shows a lot of error -13 and few successful recv():

c1=186, c2=64, e13=199750

dimakuv:
Thanks @lejunzhu for the awesome analysis, as always!

I was vaguely aware of this issue when adding mbedTLS send/recv to pipes/UNIX domain sockets, but thought that applications don't really do this (`call send() from more than one thread simultaneously`).

> Adding a spinlock (perhaps should be a futex?) around mbedtls_ssl_write() makes the problem go away.

[...]

PR #2059, opened 2021-01-05
[Pal/lib] Add spinlocks to mbedTLS-specific SSL recv/send

dimakuv:
## Description of the changes 

Linux-SGX PAL wraps all pipe/UNIX domain socket communication in TLS sessions. Previously, Graphene assumed that only one thread at a time accesses one TLS session (i.e., no multi-threading support). This commit adds spinlocks to `lib_SSL*` functions to support such (rare) multi-threading scenarios.

Note that we cannot rely on pthreads and/or mutexes so we use simple spinlocks. Because this is in the "common library" used by PAL and LibOS, with no "complex" dependencies allowed.

Also, there is a prerequisite (first) commit in this PR: remove `api.h` from `spinlock.h`. See commit messages for explanations.

## How to test this PR? 

This PR adds a corresponding LibOS test `pipe_multithread` (kudos to @lejunzhu for the initial version of this test and debugging this).

### DIFF
--- a/LibOS/shim/test/regression/.gitignore
+++ b/LibOS/shim/test/regression/.gitignore
@@ -65,6 +65,7 @@
 /multi_pthread_exitless
 /openmp
 /pipe
+/pipe_multithread
 /pipe_nonblocking
 /pipe_ocloexec
 /poll
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -53,6 +53,7 @@ c_executables = \
 	multi_pthread \
 	openmp \
 	pipe \
+	pipe_multithread \
 	pipe_nonblocking \
 	pipe_ocloexec \
 	poll \
@@ -151,26 +152,27 @@ extra_rules = \
 include ../../../../Scripts/Makefile.manifest
 include ../../../../Scripts/Makefile.Test
 
-CFLAGS-bootstrap_static = -static
+CFLAGS-abort_multithread = -pthread
 CFLAGS-bootstrap_pie = -fPIC -pie
+CFLAGS-bootstrap_static = -static
 CFLAGS-debug = -g3
 CFLAGS-debug_regs-x86_64 = -g3
-CFLAGS-exec_same = -pthread
-CFLAGS-shared_object = -fPIC -pie
-CFLAGS-syscall += -I$(PALDIR)/../include -I$(PALDIR)/host/$(PAL_HOST) -I$(PALDIR)/../include/arch/$(ARCH)/Linux
-CFLAGS-openmp = -fopenmp
-CFLAGS-multi_pthread = -pthread
-CFLAGS-exit_group = -pthread
-CFLAGS-abort_multithread = -pthread
 CFLAGS-eventfd = -pthread
+CFLAGS-exec_same = -pthread
+CFLAGS-exit_group = -pthread
 CFLAGS-futex_bitset = -pthread
 CFLAGS-futex_requeue = -pthread
 CFLAGS-futex_wake_op = -pthread
+CFLAGS-multi_pthread = -pthread
+CFLAGS-openmp = -fopenmp
+CFLAGS-pipe_multithread = -pthread
 CFLAGS-proc_common = -pthread
-CFLAGS-spinlock += -I$(PALDIR)/../include/lib -I$(PALDIR)/../include/arch/$(ARCH) -pthread
+CFLAGS-pthread_set_get_affinity += -pthread
+CFLAGS-shared_object = -fPIC -pie
 CFLAGS-sigaction_per_process += -pthread
 CFLAGS-signal_multithread += -pthread
-CFLAGS-pthread_set_get_affinity += -pthread
+CFLAGS-spinlock += -I$(PALDIR)/../include/lib -I$(PALDIR)/../include/arch/$(ARCH) -pthread
+CFLAGS-syscall += -I$(PALDIR)/../include -I$(PALDIR)/host/$(PAL_HOST) -I$(PALDIR)/../include/arch/$(ARCH)/Linux
 
 CFLAGS-attestation += -I$(PALDIR)/../lib/crypto/mbedtls/crypto/include \
                       -I$(PALDIR)/host/Linux-SGX \
--- a/LibOS/shim/test/regression/pipe_multithread.c
+++ b/LibOS/shim/test/regression/pipe_multithread.c
@@ -1,0 +1,81 @@
+/* test creates two threads simulteneously writing on the same pipe */
+
+#include <err.h>
+#include <errno.h>
+#include <pthread.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+
+#define ITERATIONS 100000
+
+int fds[2];
+
+static void* thread_run(void* arg) {
+    char c = (char)(uintptr_t)arg;
+    for (int i = 0; i < ITERATIONS; i++) {
+        ssize_t bytes = 0;
+        while (bytes < sizeof(c)) {
+            bytes = send(fds[1], &c, sizeof(c), /*flags=*/0);
+            if (bytes < 0) {
+                if (errno == EAGAIN || errno == EINTR)
+                    continue;
+                err(1, "send");
+            }
+        }
+    }
+    return NULL;
+}
+
+int main(int argc, char** argv) {
+    int ret;
+    pthread_t threads[2];
+    char thread_ids[2]   = {42, 24};
+    int thread_bytes[2]  = {0, 0};
+
+    ret = socketpair(AF_UNIX, SOCK_STREAM, 0, fds);
+    if (ret) {
+        err(1, "socketpair");
+    }
+
+    ret = pthread_create(&threads[0], NULL, &thread_run, (void*)(uintptr_t)thread_ids[0]);
+    if (ret) {
+        errno = ret;
+        err(1, "pthread_create");
+    }
+
+    ret = pthread_create(&threads[1], NULL, &thread_run, (void*)(uintptr_t)thread_ids[1]);
+    if (ret) {
+        errno = ret;
+        err(1, "pthread_create");
+    }
+
+    for (int i = 0; i < 2 * ITERATIONS; i++) {
+        char c = 0;
+        ssize_t bytes = 0;
+        while (bytes < sizeof(c)) {
+            bytes = recv(fds[0], &c, sizeof(c), /*flags=*/0);
+            if (bytes < 0) {
+                if (errno == EAGAIN || errno == EINTR)
+                    continue;
+                err(1, "recv");
+            }
+        }
+
+        if (c == thread_ids[0])
+            thread_bytes[0] += bytes;
+        else if (c == thread_ids[1])
+            thread_bytes[1] += bytes;
+        else
+            errx(1, "received unrecognized thread ID");
+    }
+
+    printf("received total bytes from threads: %d and %d\n", thread_bytes[0], thread_bytes[1]);
+
+    if (thread_bytes[0] != ITERATIONS || thread_bytes[1] != ITERATIONS)
+        errx(1, "received wrong number of bytes from threads");
+
+    puts("TEST OK");
+    return 0;
+}
--- a/Pal/include/lib/pal_crypto.h
+++ b/Pal/include/lib/pal_crypto.h
@@ -15,6 +15,8 @@
 #include <stdint.h>
 #include <unistd.h>
 
+#include "spinlock.h"
+
 #define SHA256_DIGEST_LEN 32
 
 #ifdef CRYPTO_USE_MBEDTLS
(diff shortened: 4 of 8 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2059)* status: 0 of 8 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[LibOS/shim/test/regression/Makefile, line 173 at r1](https://reviewable.io/reviews/oscarlab/graphene/2059#-MQHIAuDE1lZm36RpO9b:-MQHIAuDE1lZm36RpO9c:b-tbvve1) ([raw file](https://github.com/oscarlab/graphene/blob/345d271e663725d58ce941fc49f08ebc02c32fa2/LibOS/shim/test/regression/Makefile#L173)):*
> ```Makefile
> CFLAGS-sigaction_per_process += -pthread
> CFLAGS-signal_multithread += -pthread
> CFLAGS-pthread_set_get_affinity += -pthread
> ```

FYI: This wasn't sorted, so I just added sorting.
___
*[Pal/lib/crypto/adapters/mbedtls_adapter.c, line 388 at r1](https://reviewable.io/reviews/oscarlab/graphene/2059#-MQHIPEq8fR0bRXTEzHa:-MQHIPEq8fR0bRXTEzHb:b-6sfzku) ([raw file](https://github.com/oscarlab/graphene/blob/4d016c0d0684cd4845f3495287ab111b3d8fdad5/Pal/lib/crypto/adapters/mbedtls_adapter.c#L388)):*
> ```C
>      *       recv() and the other will spin and burn CPU cycles. We consider "shared SSL context"
>      *       a rare case and use simple spinlocks instead of mutexes. */
>     assert(spinlock_is_locked(&ssl_ctx->lock));
> ```

FYI: I consider the performance overhead of spinlocks here minimal. We never hit multi-threaded pipes until #2058 (actually, I know that Erlang does the same), so most apps won't notice this. And the apps that do use this multi-threaded pattern... well, I just hope it's not on critical path because these are just pipes for intra-process communication.
___
*[Pal/src/pal_internal.h, line 12 at r1](https://reviewable.io/reviews/oscarlab/graphene/2059#-MQHJ0BIBPbamvoWycYd:-MQHJ0BJD1DOVNolIeLW:b8ot0cp) ([raw file](https://github.com/oscarlab/graphene/blob/4d016c0d0684cd4845f3495287ab111b3d8fdad5/Pal/src/pal_internal.h#L12)):*
> ```objc
> 
> #include <stdarg.h>
> #include <sys/types.h>
> ```

FYI: That's because one function here uses `ssize_t` type.

dimakuv:
Jenkins, retest Jenkins-Debug please (LTP test `alarm03` failed, cannot reproduce locally and generally unrelated)

lejunzhu:
I think protecting mbedtls_ssl_read and mbedtls_ssl_write with the same lock may cause other problems. Read usually blocks for a long time. If the application is trying to send when recv'ing in another thread, it will cause a deadlock.

For example, this test C program runs fine under Linux and Graphene SGX before this PR, but with this PR it will hang.
[test.txt](https://github.com/oscarlab/graphene/files/5773255/test.txt)

I don't know a real world example like this, but I guess it is possible.

Looking at mbedTLS code, read and write does not seem to cause conflict, except when handshake or re-negotiation is pending. And I can't think why an application will call recv() from two threads. Therefore we can probably call mbedtls_ssl_read without a lock.

dimakuv:
>I don't know a real world example like this, but I guess it is possible.

The attached test program reads and writes on the same file descriptor from two threads. This looks like an unrealistic scenario. Typically applications create a socket pair or a pipe pair and use one end for recv and another end for send.

> ...And I can't think why an application will call recv() from two threads. Therefore we can probably call mbedtls_ssl_read without a lock.

I definitely don't like the idea of having `mbedtls_ssl_read()` not protected with a lock. What happens if we do `mbedtls_ssl_read(my_ssl_ctx)` and a simultaneous `mbedlts_ssl_write(my_ssl_ctx)`? Even though the write is protected with a lock, it modifies `my_ssl_ctx`. If the read looks at `my_ssl_ctx` simultaneously, then it may observe inconsistent state of the SSL context, and all hell breaks loose. It's better to deadlock in such situations than to have memory/SSL state corruptions.

lejunzhu:
It seems that calling fork() after a thread have started recv() will also cause deadlock. See:
[test.txt](https://github.com/oscarlab/graphene/files/5778865/test.txt)

Perhaps we can drop the lock before ocall_recv, then re-acquire it after it returns? This should be fine as long as re-negotiation is disabled.

Or, at least we can use spinlock with a timeout and print some warning message, so users can see and report the problem when it happens.

yamahata (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2059)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Pal/lib/crypto/adapters/mbedtls_adapter.c, line 417 at r1](https://reviewable.io/reviews/oscarlab/graphene/2059#-MQQGtRz0SWvcQV_1KAz:-MQQGtS-8aEVUE8TTdw2:bh5etcg) ([raw file](https://github.com/oscarlab/graphene/blob/4d016c0d0684cd4845f3495287ab111b3d8fdad5/Pal/lib/crypto/adapters/mbedtls_adapter.c#L417)):*
> ```C
>      *       a rare case and use simple spinlocks instead of mutexes. */
>     assert(spinlock_is_locked(&ssl_ctx->lock));
>     ssize_t ret = ssl_ctx->pal_send_cb(fd, buf, buf_size);
> ```

pal_send_cb eventually calls blocking host system call?
it would cause troubles.
Anyway I don't know about MT-requirement for mbedtls. So I leave this comment non-blocking.

dimakuv (commented):
> Perhaps we can drop the lock before ocall_recv, then re-acquire it after it returns? This should be fine as long as re-negotiation is disabled.

But this doesn't help with the SSL context... mbedTLS internally changes its own metadata before and after the `recv()` syscall/ocall. So there can be a situation when two threads try to recv on the same SSL context, and one unlocks the SSL context and blocks on `recv()`, and the other thread now can proceed, and observes inconsistent state of the SSL context... and stuff breaks.

I googled around but couldn't find a good solution. I have a feeling that it's just not possible to escape the issue of "blocking recv/send syscall while holding the lock". @mkow @boryspoplawski @yamahata Maybe you know how to circumvent this?

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2059#-:-MQRkDlA-22Fmi8gFRni:bvckd30)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @lejunzhu)

___
*[Pal/lib/crypto/adapters/mbedtls_adapter.c, line 417 at r1](https://reviewable.io/reviews/oscarlab/graphene/2059#-MQQGtRz0SWvcQV_1KAz:-MQRlL8xEx4AvgCR73SB:brc8rz4) ([raw file](https://github.com/oscarlab/graphene/blob/4d016c0d0684cd4845f3495287ab111b3d8fdad5/Pal/lib/crypto/adapters/mbedtls_adapter.c#L417)):*

pal_send_cb eventually calls blocking host system call?
it would cause troubles.
Anyway I don't know about MT-requirement for mbedtls. So I leave this comment non-blocking.
</blockquote></details>

Yes, it will cause troubles (deadlocks in scenarios @lejunzhu mentioned)... I can't come up with a way to fix this.

lejunzhu:
> But this doesn't help with the SSL context... mbedTLS internally changes its own metadata before and after the `recv()` syscall/ocall. So there can be a situation when two threads try to recv on the same SSL context, and one unlocks the SSL context and blocks on `recv()`, and the other thread now can proceed, and observes inconsistent state of the SSL context... and stuff breaks.

I can't think why an application needs to use two recv() simultaneously. If it does happen, perhaps we can add a flag in ssl_ctx indicating the first mbedtls_ssl_read() is already running, and just return fail to the second one before it even enters mbedTLS.

boryspoplawski (commented):
I see one solution: create two TLS contexts, one for each communication direction. Then either:
1) `recv` changes only context internals, but does not send anything on the socket (like some metadata, idk)
2) `recv` also sends some data on the socket.
If it's 1) then we could wrap the very same `fd` into both context, otherwise (2) we need to create two separate connections to emulate each way of the original one.

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2059#-:-MQbwcHI0FlkgOeKo-Y9:beei1yn)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @lejunzhu)

boryspoplawski (commented):
Btw I assumed mbedTLS does not give us any way to implement this (but I'm not that familiar with it, so this needs to be confirmed).

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2059#-:-MQbxWPXELXEZJcm5fhk:b-990hzv)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @lejunzhu)

mkow (commented):
1. sounds like a risky assumption, I guess TLS may sometimes send some data when receiving, e.g. to rotate symmetric keys, in case of long-lasting connections.

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2059#-:-MQitnVcCxM6U9-BETOI:b-dj6c8o)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @lejunzhu)

lejunzhu (commented):
Maybe non-blocking I/O can solve this. Roughly, calls to mbedtls_ssl_handshake(), mbedtls_ssl_read() and mbedtls_ssl_write() will look like this:

```
while (1) {
  lock();
  ret = mbedtls_ssl_...(); /* handshake/read/write */
  unlock();
  if (ret == MBEDTLS_ERR_SSL_WANT_WRITE) {
    ocall_poll(stream_fd, POLLOUT); /* wait until we can write */
    continue;
  } else if (ret == MBEDTLS_ERR_SSL_WANT_READ) {
    ocall_poll(stream_fd, POLLIN); /* wait until we can read */
    continue;
  } else {
    break; /* Go on and process the return value as usual */
  }
}
```

Threads will only block at ocall_poll(). All calls to mbedTLS API are serialized, therefore the internal state must be consistent. 

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2059#-:-MQtbR4x3Z81RtnbGYt3:bq4swfe)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @lejunzhu)

dimakuv (commented):
To be honest, I didn't understand Borys's idea. (As Michal mentioned, 1 is a risky assumption so we could go with 2, but I don't really understand 2. Sounds very complicated.)

I like @lejunzhu's workaround. @lejunzhu Do I understand correctly that we do this:
1. We *always* create *non-blocking* host `stream_fd` for TLS connections
2. If the user app wants non-blocking pipes/sockets, then nothing to be done additionally
3. But if the user app wants blocking pipes/sockets, then we additionally do `ocall_poll()` to wait for data to be available for read/write outside of the lock.

Is my understanding correct? From the first glance, this looks like a good workaround. @boryspoplawski @mkow Any drawbacks to this hack?

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2059#-:-MQvz4H31eRVgycuSY67:b-ro0f3)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @lejunzhu)

dimakuv (commented):
Also summoning @yamahata 

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2059#-:-MQw-4qi5-6Xn__UQHXF:b1b4ott)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @lejunzhu)

lejunzhu:
> To be honest, I didn't understand Borys's idea. (As Michal mentioned, 1 is a risky assumption so we could go with 2, but I don't really understand 2. Sounds very complicated.)
> 
> I like @lejunzhu's workaround. @lejunzhu Do I understand correctly that we do this:
> 
> 1. We _always_ create _non-blocking_ host `stream_fd` for TLS connections
> 2. If the user app wants non-blocking pipes/sockets, then nothing to be done additionally
> 3. But if the user app wants blocking pipes/sockets, then we additionally do `ocall_poll()` to wait for data to be available for read/write outside of the lock.
> 
> Is my understanding correct? From the first glance, this looks like a good workaround. @boryspoplawski @mkow Any drawbacks to this hack?

Yes, this is what I'm suggesting.

llly:
I worked out a commit https://github.com/llly/graphene/commit/e4dc52e036faf7e8628c3772c5c4a6e0858d6ba7 based on this PR. It's similar to @lejunzhu 's idea,
The test program in this PR can work.
The test program in #2058 doesn't return errno13 anymore, but the counts are incorrect. 
```
c1=99994, c2=100006, e13=0
c1=100009, c2=99991, e13=0
``` 
I don't know what caused this behavior.

Issue #790, opened 2019-06-24
Deprecate shim_do_sandbox_attach


donporter:
This call appears to be a no-op.  Do we still need it?

mkow:
Doesn't exist anymore on master, so I'm closing this issue.

PR #2063, opened 2021-01-07
WIP Decouple downloads

woju:
## Description of the changes 
In Real Build Systems™ (obs, mock, ...) there is no way to download anything from the Internet. Instead all distfiles should be declared in `debian/`/`SPEC`/whatever the build system likes and are predownloaded (hashes checked etc.) and afterwards the chroot is configured without external access. This is by design to ensure reliable and repeatable builds.

Therefore we should decouple the download from our build harness and eventually retire `Scripts/download` altogether.

This PR will remove downloading of at least some of the things that we currently do download:
- [x] uthash
- [ ] cjson
- [ ] mbedtls
- [ ] tomlc99
- [ ] glibc

#790

## How to test this PR? 
Nothing should regress in build. The Documentation should accurately reflect the correct build procedure.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2063)

### DIFF
--- a/.ci/ubuntu16.04.dockerfile
+++ b/.ci/ubuntu16.04.dockerfile
@@ -50,6 +50,7 @@ RUN apt-get update \
        r-base-core \
        shellcheck \
        texinfo \
+       uthash-dev \
        wget \
        zlib1g \
        zlib1g-dev \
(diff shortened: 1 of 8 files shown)


woju:
It *might* be that this will break xenial (16.04), simply because the packages there are too old. I don't know yet. If it is really broken, I'll consider the options, but I'll likely advise to retire support for 16.04.

mkow:
@woju said that this will be obsolete after Meson rework, so I'm closing it.

PR #2064, opened 2021-01-07
WIP Test what's installed, not from repo

woju:
## Description of the changes 

This includes meson into testing. It should be possible to properly install graphene, so we should fail the tests if graphene is not properly installed. Therefore as part of tests we install graphene and test only what is installed.

Although in practice this PR doesn't break anything, it theoretically makes meson mandatory, because we won't test anything not installed.

This is intermediate measure, in reality we should install from `.deb`/`.rpm`. But installing directly from meson is good enough and actually easier to do without root.

#760

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2064)


woju:
This PR is replaced by #2288 and #2298.

PR #2075, opened 2021-01-12
New logging system

pwmarcz:
New version of #1979 (closes #1979). The breaking change is `loader.log_level`.

See also issue #1947 (Allow redirecting debug logs to a file).

## Description of the changes 

Instead of 'loader.debug_type', introduce 'loader.log_level'
and 'loader.log_file', along with a set of definitions for
logging at a chosen level.

For now, the call sites keep using the legacy macros (SGX_DBG and
debug()), because converting them all will conflict with other
big changes in the code base. The existing LibOS calls are
assumed to be at 'info' level.

## Planned follow-up

* Change default stream to stderr (will need adjusting some tests)
* Convert call sites (maybe not while there are big changes in pipeline)
* Adjust levels (for example, system calls probably belong at "debug" or "trace")
* Make PAL use the same prefix for logs (thread number etc.)

## How to test this PR? 

Existing tests should be enough, I expect Jenkins will put up a fight.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)

### DIFF
--- a/Documentation/devel/performance.rst
+++ b/Documentation/devel/performance.rst
@@ -370,7 +370,7 @@ particular, build Graphene in non-debug configuration (simple ``make SGX=1``
 defaults to non-debug configuration). Also build the application itself in
 non-debug configuration (again, typically simple ``make SGX=1`` is sufficient).
 Finally, disable the debug log of Graphene by specifying the manifest option
-``loader.debug_type = "none"``.
+``loader.log_level = "none"``.
 
 If your application periodically fails and complains about seemingly irrelevant
 things, it may be due to insufficient enclave memory. Please try to increase
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -23,21 +23,21 @@ Comments can be inlined in a |~| manifest by starting them with a |~| hash sign
 Common syntax
 -------------
 
-Debug type
-^^^^^^^^^^
+Log level
+^^^^^^^^^
 
 ::
 
-    loader.debug_type = "[none|inline|file]"
-    (Default: "none")
+    loader.log_level = "[none|error|info|debug|trace|all]"
+    (Default: "error")
 
-    loader.debug_file = "[PATH]"
+    loader.log_file = "[PATH]"
 
-This specifies the debug option while running the library OS. If the debug type
-is ``none``, no debug output will be printed to standard output. If the debug
-type is ``inline``, a dmesg-like debug output will be printed inline with
-standard output. If the debug type is ``file``, debug output will be written to
-the file specified in ``loader.debug_file``.
+This configures Graphene's debug log. The ``log_level`` option specifies what
+messages to enable (e.g. ``loader.log_level = "debug"`` will enable all messages
+of type ``error``, ``info`` and ``debug``). By default, the messages are printed
+to the standard output. If ``log_file`` is specified, the messages will be
+appended to that file.
 
 Preloaded libraries
 ^^^^^^^^^^^^^^^^^^^
--- a/Examples/apache/Makefile
+++ b/Examples/apache/Makefile
@@ -32,9 +32,9 @@ GRAPHENEDIR ?= $(THIS_DIR)../..
 SGX_SIGNER_KEY ?= $(GRAPHENEDIR)/Pal/src/host/Linux-SGX/signer/enclave-key.pem
 
 ifeq ($(DEBUG),1)
-GRAPHENEDEBUG = inline
+GRAPHENE_LOG_LEVEL = debug
 else
-GRAPHENEDEBUG = none
+GRAPHENE_LOG_LEVEL = error
 endif
 
 .PHONY: all
@@ -116,7 +116,7 @@ httpd-trusted-libs: httpd-deps httpd-modules $(INSTALL_DIR)/bin/httpd
 
 httpd.manifest: httpd.manifest.template httpd-trusted-mods httpd-trusted-libs
 	sed -e 's|$$(GRAPHENEDIR)|'"$(GRAPHENEDIR)"'|g' \
-	    -e 's|$$(GRAPHENEDEBUG)|'"$(GRAPHENEDEBUG)"'|g' \
+	    -e 's|$$(GRAPHENE_LOG_LEVEL)|'"$(GRAPHENE_LOG_LEVEL)"'|g' \
 	    -e 's|$$(INSTALL_DIR)|'"$(INSTALL_DIR)"'|g' \
 	    -e 's|$$(INSTALL_DIR_ABSPATH)|'"$(abspath $(INSTALL_DIR))"'|g' \
 	    -e 's|$$(HTTPD_TRUSTED_MODS)|'"`cat httpd-trusted-mods`"'|g' \
--- a/Examples/apache/httpd.manifest.template
+++ b/Examples/apache/httpd.manifest.template
@@ -9,7 +9,7 @@ loader.argv0_override = "httpd"
 # Read application arguments directly from the command line. Don't use this on production!
 loader.insecure__use_cmdline_argv = 1
 
-loader.debug_type = "$(GRAPHENEDEBUG)"
+loader.log_level = "$(GRAPHENE_LOG_LEVEL)"
 
 # Environment variables
 loader.env.LD_LIBRARY_PATH = "/lib:$(ARCH_LIBDIR):/usr/lib:/usr/$(ARCH_LIBDIR)"
--- a/Examples/bash/Makefile
+++ b/Examples/bash/Makefile
@@ -12,9 +12,9 @@ GRAPHENEDIR ?= ../..
 SGX_SIGNER_KEY ?= $(GRAPHENEDIR)/Pal/src/host/Linux-SGX/signer/enclave-key.pem
 
 ifeq ($(DEBUG),1)
-GRAPHENEDEBUG = inline
+GRAPHENE_LOG_LEVEL = debug
 else
-GRAPHENEDEBUG = none
+GRAPHENE_LOG_LEVEL = error
 endif
 
 .PHONY: all
@@ -35,7 +35,7 @@ trusted-libs: ../common_tools/get_deps.sh
 
 bash.manifest: manifest.template trusted-libs
 	(sed -e 's|$$(GRAPHENEDIR)|'"$(GRAPHENEDIR)"'|g' \
-	     -e 's|$$(GRAPHENEDEBUG)|'"$(GRAPHENEDEBUG)"'|g' \
+	     -e 's|$$(GRAPHENE_LOG_LEVEL)|'"$(GRAPHENE_LOG_LEVEL)"'|g' \
 	     -e 's|$$(EXECDIR)|'"$(shell dirname $(shell which bash))"'|g' \
 	     -e 's|$$(ARCH_LIBDIR)|'"$(ARCH_LIBDIR)"'|g' \
 	     $<; \
--- a/Examples/bash/manifest.template
+++ b/Examples/bash/manifest.template
@@ -8,7 +8,7 @@ libos.entrypoint = "file:$(EXECDIR)/bash"
 # Read application arguments directly from the command line. Don't use this on production!
 loader.insecure__use_cmdline_argv = 1
 
-loader.debug_type = "$(GRAPHENEDEBUG)"
+loader.log_level = "$(GRAPHENE_LOG_LEVEL)"
 
 # Environment variables
 loader.env.LD_LIBRARY_PATH = "/lib:$(ARCH_LIBDIR):/usr$(ARCH_LIBDIR)"
--- a/Examples/blender/Makefile
+++ b/Examples/blender/Makefile
@@ -22,9 +22,9 @@ $(error This example requires Ubuntu 16.04, 18.04 or 20.04)
 endif
 
 ifeq ($(DEBUG),1)
-GRAPHENE_DEBUG = inline
+GRAPHENE_LOG_LEVEL = debug
 else
-GRAPHENE_DEBUG = none
+GRAPHENE_LOG_LEVEL = error
 endif
 
 
@@ -51,7 +51,7 @@ trusted-libs: ../common_tools/get_deps.sh $(BLENDER_DIR)/blender
 
 blender.manifest: blender.manifest.template trusted-libs | $(RUN_DIR)
 	(sed -e 's|$$(GRAPHENE_DIR)|'"$(GRAPHENE_DIR)"'|g' \
-	     -e 's|$$(GRAPHENE_DEBUG)|'"$(GRAPHENE_DEBUG)"'|g' \
+	     -e 's|$$(GRAPHENE_LOG_LEVEL)|'"$(GRAPHENE_LOG_LEVEL)"'|g' \
 	     -e 's|$$(DATA_DIR)|'"$(DATA_DIR)"'|g' \
 	     -e 's|$$(BLENDER_DIR)|'"$(BLENDER_DIR)"'|g' \
 	     -e 's|$$(BLENDER_VER)|'"$(BLENDER_VER)"'|g' \
--- a/Examples/blender/blender.manifest.template
+++ b/Examples/blender/blender.manifest.template
@@ -17,7 +17,7 @@ libos.entrypoint = "file:$(BLENDER_DIR)/blender"
 loader.insecure__use_cmdline_argv = 1
 
 loader.preload = "file:$(GRAPHENE_DIR)/Runtime/libsysdb.so"
-loader.debug_type = "$(GRAPHENE_DEBUG)"
+loader.log_level = "$(GRAPHENE_LOG_LEVEL)"
 
 loader.env.LD_LIBRARY_PATH = "/graphene_lib:/blender_lib:/usr/$(ARCH_LIBDIR):$(ARCH_LIBDIR)"
 loader.env.HOME = "/home/user"
--- a/Examples/busybox/Makefile
+++ b/Examples/busybox/Makefile
@@ -21,9 +21,9 @@ BUSYBOX_URL ?= https://busybox.net/downloads/busybox-1.32.0.tar.bz2
 BUSYBOX_SHA256 ?= c35d87f1d04b2b153d33c275c2632e40d388a88f19a9e71727e0bbbff51fe689
 
 ifeq ($(DEBUG),1)
-GRAPHENEDEBUG = inline
+GRAPHENE_LOG_LEVEL = debug
 else
-GRAPHENEDEBUG = none
+GRAPHENE_LOG_LEVEL = error
 endif
 
 
@@ -69,7 +69,7 @@ $(SRCDIR)/busybox: $(SRCDIR)/.config
 
 busybox.manifest: busybox.manifest.template
 	sed -e 's|$$(GRAPHENEDIR)|'"$(GRAPHENEDIR)"'|g' \
-		-e 's|$$(GRAPHENEDEBUG)|'"$(GRAPHENEDEBUG)"'|g' \
+		-e 's|$$(GRAPHENE_LOG_LEVEL)|'"$(GRAPHENE_LOG_LEVEL)"'|g' \
 		-e 's|$$(ARCH_LIBDIR)|'"$(ARCH_LIBDIR)"'|g' \
 		$< > $@
 
--- a/Examples/busybox/busybox.manifest.template
+++ b/Examples/busybox/busybox.manifest.template
@@ -16,10 +16,11 @@ loader.insecure__use_cmdline_argv = 1
 # build process.
 loader.preload = "file:$(GRAPHENEDIR)/Runtime/libsysdb.so"
 
-# Show/hide debug log of Graphene ('inline' or 'none' respectively). Note that
-# GRAPHENEDEBUG macro is expanded to inline/none in the Makefile as part of the
-# build process.
-loader.debug_type = "$(GRAPHENEDEBUG)"
+# Verbosity of Graphene debug log (none/error/info/debug/trace/all). Note that
+# GRAPHENE_LOG_LEVEL macro is expanded in the Makefile as part of the building
+# process: the default is "error" for non-debug builds, and "debug" for debug
+# builds.
+loader.log_level = "$(GRAPHENE_LOG_LEVEL)"
 
 ################################# ENV VARS  ###################################
 
--- a/Examples/capnproto/Makefile
+++ b/Examples/capnproto/Makefile
@@ -25,9 +25,9 @@ ADDRESSBOOK_CAP_URL    ?= $(CAPNPROTO_SRC_URL)/c%2B%2B/samples/addressbook.capnp
 ADDRESSBOOK_CAP_SHA256 ?= cf309b8b277881c95aaa785f4d0154f46f82575fdb9dea49fb10ee072b7c3ae6
 
 ifeq ($(DEBUG),1)
-GRAPHENEDEBUG = inline
+GRAPHENE_LOG_LEVEL = debug
 else
-GRAPHENEDEBUG = none
+GRAPHENE_LOG_LEVEL = error
 endif
 
 .PHONY: all
@@ -63,7 +63,7 @@ trusted-libs: ../common_tools/get_deps.sh $(SRCDIR)/addressbook
 
 addressbook.manifest: addressbook.manifest.template trusted-libs
 	(sed -e 's|$$(GRAPHENEDIR)|'"$(GRAPHENEDIR)"'|g' \
-		-e 's|$$(GRAPHENEDEBUG)|'"$(GRAPHENEDEBUG)"'|g' \
+		-e 's|$$(GRAPHENE_LOG_LEVEL)|'"$(GRAPHENE_LOG_LEVEL)"'|g' \
 		-e 's|$$(ARCH_LIBDIR)|'"$(ARCH_LIBDIR)"'|g' \
 		$<; \
 	cat trusted-libs) > $@
--- a/Examples/capnproto/addressbook.manifest.template
+++ b/Examples/capnproto/addressbook.manifest.template
@@ -8,8 +8,8 @@ libos.entrypoint = "file:addressbook"
 # LibOS layer library of Graphene (currently only one implementation, libsysdb)
 loader.preload = "file:$(GRAPHENEDIR)/Runtime/libsysdb.so"
 
-# Show/hide debug log of Graphene ('inline' or 'none' respectively)
-loader.debug_type = "$(GRAPHENEDEBUG)"
+# Graphene log level
+loader.log_level = "$(GRAPHENE_LOG_LEVEL)"
 
 # Read application arguments directly from the command line. Don't use this on production!
 loader.insecure__use_cmdline_argv = 1
--- a/Examples/curl/Makefile
+++ b/Examples/curl/Makefile
@@ -15,9 +15,9 @@ GRAPHENEDIR ?= ../..
 SGX_SIGNER_KEY ?= $(GRAPHENEDIR)/Pal/src/host/Linux-SGX/signer/enclave-key.pem
 
 ifeq ($(DEBUG),1)
-GRAPHENEDEBUG = inline
+GRAPHENE_LOG_LEVEL = debug
 else
-GRAPHENEDEBUG = none
+GRAPHENE_LOG_LEVEL = error
 endif
 
 .PHONY: all
@@ -37,7 +37,7 @@ trusted-libs: ../common_tools/get_deps.sh
 
 curl.manifest: curl.manifest.template trusted-libs
 	(sed -e 's|$$(GRAPHENEDIR)|'"$(GRAPHENEDIR)"'|g' \
-	     -e 's|$$(GRAPHENEDEBUG)|'"$(GRAPHENEDEBUG)"'|g' \
+	     -e 's|$$(GRAPHENE_LOG_LEVEL)|'"$(GRAPHENE_LOG_LEVEL)"'|g' \
 	     -e 's|$$(CURL_DIR)|'"$(CURL_DIR)"'|g' \
 	     -e 's|$$(ARCH_LIBDIR)|'"$(ARCH_LIBDIR)"'|g' \
 	     $<; \
--- a/Examples/curl/curl.manifest.template
+++ b/Examples/curl/curl.manifest.template
@@ -9,8 +9,8 @@ loader.argv0_override = "curl"
 # so it is always set to libsysdb.so. 
 loader.preload = "file:$(GRAPHENEDIR)/Runtime/libsysdb.so"
 
-# Show/hide debug log of Graphene ('inline' or 'none' respectively).
-loader.debug_type = "$(GRAPHENEDEBUG)"
+# Graphene log level
+loader.log_level = "$(GRAPHENE_LOG_LEVEL)"
 
 # Read application arguments directly from the command line. Don't use this on production!
 loader.insecure__use_cmdline_argv = 1
--- a/Examples/gcc/Makefile
+++ b/Examples/gcc/Makefile
@@ -11,9 +11,9 @@ GZIP_MIRRORS ?= https://people.csail.mit.edu/smcc/projects/single-file-programs/
 GZIP_HASH ?= 7ec7d87822e6497648580064756f64e47dbd085727910ebbc52a7c240a88dc27
 
 ifeq ($(DEBUG),1)
-GRAPHENEDEBUG = inline
+GRAPHENE_LOG_LEVEL = debug
 else
-GRAPHENEDEBUG = none
+GRAPHENE_LOG_LEVEL = error
 endif
 
 include ../../Scripts/Makefile.configs
@@ -43,7 +43,7 @@ trusted-libs: ../common_tools/get_deps.sh
 
 %.manifest: %.manifest.template trusted-libs
 	(sed -e 's|$$(GRAPHENEDIR)|'"$(GRAPHENEDIR)"'|g' \
-		-e 's|$$(GRAPHENEDEBUG)|'"$(GRAPHENEDEBUG)"'|g' \
+		-e 's|$$(GRAPHENE_LOG_LEVEL)|'"$(GRAPHENE_LOG_LEVEL)"'|g' \
 		-e 's|$$(BINUTILS_VERSION)|'"$(BINUTILS_VERSION)"'|g' \
 		-e 's|$$(GCC_MAJOR_VERSION)|'"$(GCC_MAJOR_VERSION)"'|g' \
 		-e 's|$$(GCC_LIB_PATH)|'"$(GCC_LIB_PATH)"'|g' \
--- a/Examples/gcc/gcc.manifest.template
+++ b/Examples/gcc/gcc.manifest.template
@@ -1,7 +1,7 @@
 loader.preload = "file:$(GRAPHENEDIR)/Runtime/libsysdb.so"
 loader.env.LD_LIBRARY_PATH = "/lib:$(ARCH_LIBDIR):/usr/lib:/usr/$(ARCH_LIBDIR)"
 loader.env.PATH = "/bin:/usr/bin"
-loader.debug_type = "$(GRAPHENEDEBUG)"
+loader.log_level = "$(GRAPHENE_LOG_LEVEL)"
 libos.entrypoint = "file:/usr/bin/gcc"
 
 # Read application arguments directly from the command line. Don't use this on production!
--- a/Examples/lighttpd/Makefile
+++ b/Examples/lighttpd/Makefile
@@ -17,9 +17,9 @@ GRAPHENEDIR ?= $(THIS_DIR)../..
 SGX_SIGNER_KEY ?= $(GRAPHENEDIR)/Pal/src/host/Linux-SGX/signer/enclave-key.pem
 
 ifeq ($(DEBUG),1)
-GRAPHENEDEBUG = inline
+GRAPHENE_LOG_LEVEL = debug
 else
-GRAPHENEDEBUG = none
+GRAPHENE_LOG_LEVEL = error
 endif
 
 CONF_FILES = lighttpd-server.conf lighttpd.conf
@@ -62,7 +62,7 @@ trusted-libs: ../common_tools/get_deps.sh $(INSTALL_DIR)/sbin/lighttpd
 
 lighttpd.manifest: lighttpd.manifest.template trusted-libs
 	(sed -e 's|$$(GRAPHENEDIR)|'"$(GRAPHENEDIR)"'|g' \
-	     -e 's|$$(GRAPHENEDEBUG)|'"$(GRAPHENEDEBUG)"'|g' \
+	     -e 's|$$(GRAPHENE_LOG_LEVEL)|'"$(GRAPHENE_LOG_LEVEL)"'|g' \
 	     -e 's|$$(INSTALL_DIR)|'"$(INSTALL_DIR)"'|g' \
 	     -e 's|$$(INSTALL_DIR_ABSPATH)|'"$(abspath $(INSTALL_DIR))"'|g' \
 	     -e 's|$$(ARCH_LIBDIR)|'"$(ARCH_LIBDIR)"'|g' \
--- a/Examples/lighttpd/lighttpd.manifest.template
+++ b/Examples/lighttpd/lighttpd.manifest.template
@@ -8,10 +8,11 @@
 
 libos.entrypoint = "file:$(INSTALL_DIR)/sbin/lighttpd"
 
-# Graphene environment, including the path of the library OS and the debug
-# option (inline/none).
+# Path to the library OS
 loader.preload = "file:$(GRAPHENEDIR)/Runtime/libsysdb.so"
-loader.debug_type = "$(GRAPHENEDEBUG)"
+
+# Graphene log level
+loader.log_level = "$(GRAPHENE_LOG_LEVEL)"
 
 # Read application arguments directly from the command line. Don't use this on production!
 loader.insecure__use_cmdline_argv = 1
--- a/Examples/memcached/Makefile
+++ b/Examples/memcached/Makefile
@@ -28,9 +28,9 @@ $(error This example requires Ubuntu 16.04 or 18.04)
 endif
 
 ifeq ($(DEBUG),1)
-GRAPHENEDEBUG = inline
+GRAPHENE_LOG_LEVEL = debug
 else
-GRAPHENEDEBUG = none
+GRAPHENE_LOG_LEVEL = error
 endif
 
 .PHONY: all
@@ -65,7 +65,7 @@ $(SRCDIR)/memcached: $(SRCDIR)/configure
 
 memcached.manifest: memcached.manifest.template
 	sed -e 's|$$(GRAPHENEDIR)|'"$(GRAPHENEDIR)"'|g' \
-		-e 's|$$(GRAPHENEDEBUG)|'"$(GRAPHENEDEBUG)"'|g' \
+		-e 's|$$(GRAPHENE_LOG_LEVEL)|'"$(GRAPHENE_LOG_LEVEL)"'|g' \
 		-e 's|$$(ARCH_LIBDIR)|'"$(ARCH_LIBDIR)"'|g' \
 		-e 's|# \['"$(UBUNTU_VER)"'\] ||g' \
 		$< > $@
--- a/Examples/memcached/memcached.manifest.template
+++ b/Examples/memcached/memcached.manifest.template
@@ -26,10 +26,11 @@ loader.preload = "file:$(GRAPHENEDIR)/Runtime/libsysdb.so"
 # Binary to run.
 libos.entrypoint = "file:memcached"
 
-# Show/hide debug log of Graphene ("inline" or "none" respectively). Note that
-# GRAPHENEDEBUG macro is expanded to inline/none in the Makefile as part of the
-# build process.
-loader.debug_type = "$(GRAPHENEDEBUG)"
+# Verbosity of Graphene debug log (none/error/info/debug/trace/all). Note that
+# GRAPHENE_LOG_LEVEL macro is expanded in the Makefile as part of the building
+# process: the default is "error" for non-debug builds, and "debug" for debug
+# builds.
+loader.log_level = "$(GRAPHENE_LOG_LEVEL)"
 
 ################################# ARGUMENTS ###################################
 
--- a/Examples/nginx/Makefile
+++ b/Examples/nginx/Makefile
@@ -30,9 +30,9 @@ GRAPHENEDIR ?= $(THIS_DIR)../..
 SGX_SIGNER_KEY ?= $(GRAPHENEDIR)/Pal/src/host/Linux-SGX/signer/enclave-key.pem
 
 ifeq ($(DEBUG),1)
-GRAPHENEDEBUG = inline
+GRAPHENE_LOG_LEVEL = debug
 else
-GRAPHENEDEBUG = none
+GRAPHENE_LOG_LEVEL = error
 endif
 
 .PHONY: all
@@ -72,7 +72,7 @@ trusted-libs: ../common_tools/get_deps.sh $(INSTALL_DIR)/sbin/nginx
 
 nginx.manifest: nginx.manifest.template trusted-libs
 	(sed -e 's|$$(GRAPHENEDIR)|'"$(GRAPHENEDIR)"'|g' \
-	     -e 's|$$(GRAPHENEDEBUG)|'"$(GRAPHENEDEBUG)"'|g' \
+	     -e 's|$$(GRAPHENE_LOG_LEVEL)|'"$(GRAPHENE_LOG_LEVEL)"'|g' \
 	     -e 's|$$(INSTALL_DIR)|'"$(INSTALL_DIR)"'|g' \
 	     -e 's|$$(INSTALL_DIR_ABSPATH)|'"$(abspath $(INSTALL_DIR))"'|g' \
 	     -e 's|$$(ARCH_LIBDIR)|'"$(ARCH_LIBDIR)"'|g' \
--- a/Examples/nginx/nginx.manifest.template
+++ b/Examples/nginx/nginx.manifest.template
@@ -4,10 +4,11 @@
 
 libos.entrypoint = "file:$(INSTALL_DIR)/sbin/nginx"
 
-# Graphene environment, including the path to the library OS and the debug
-# option (inline/none).
+# Path to the library OS
 loader.preload = "file:$(GRAPHENEDIR)/Runtime/libsysdb.so"
-loader.debug_type = "$(GRAPHENEDEBUG)"
+
+# Graphene log level
+loader.log_level = "$(GRAPHENE_LOG_LEVEL)"
 
 # Read application arguments directly from the command line. Don't use this on production!
 loader.insecure__use_cmdline_argv = 1
--- a/Examples/nodejs-express-server/Makefile
+++ b/Examples/nodejs-express-server/Makefile
@@ -15,9 +15,9 @@ GRAPHENEDIR ?= ../..
 SGX_SIGNER_KEY ?= $(GRAPHENEDIR)/Pal/src/host/Linux-SGX/signer/enclave-key.pem
 
 ifeq ($(DEBUG),1)
-GRAPHENEDEBUG = inline
+GRAPHENE_LOG_LEVEL = debug
 else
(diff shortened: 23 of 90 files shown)


yamahata (approved):
> Reviewed 111 of 111 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

dimakuv (changes requested):
> Reviewed 111 of 111 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, 17 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvfcnvAlro3RPPTaZD:-MQvfcnwEbHraXT4PNDY:b-qvnjrq) (no related file):*
Shouldn't you merge two separate commits in this PR into one? Otherwise the example manifests will fail if we bisect between these two commits.

---
___
*[Examples/gcc/as.manifest.template, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvjxhy4VEPVdlmOEVP:-MQvjxhzByvM4pPGrCzk:b-20okfp) ([raw file](https://github.com/oscarlab/graphene/blob/303528131c67f58aeee677397ade9593f222ae88/Examples/gcc/as.manifest.template#L1)):*
> ```
> loader.preload = "file:$(GRAPHENEDIR)/Runtime/libsysdb.so"
> ```

This file won't exist after the rebase to the latest master.
___
*[Examples/gcc/cc1.manifest.template, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvk1E--lCD3kAfWFcH:-MQvk1E--lCD3kAfWFcI:b-20okfp) ([raw file](https://github.com/oscarlab/graphene/blob/303528131c67f58aeee677397ade9593f222ae88/Examples/gcc/cc1.manifest.template#L1)):*
> ```
> loader.preload = "file:$(GRAPHENEDIR)/Runtime/libsysdb.so"
> ```

This file won't exist after the rebase to the latest master.
___
*[Examples/gcc/collect2.manifest.template, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvk28J25heueT4nX25:-MQvk28J25heueT4nX26:b-20okfp) ([raw file](https://github.com/oscarlab/graphene/blob/303528131c67f58aeee677397ade9593f222ae88/Examples/gcc/collect2.manifest.template#L1)):*
> ```
> loader.preload = "file:$(GRAPHENEDIR)/Runtime/libsysdb.so"
> ```

This file won't exist after the rebase to the latest master.
___
*[Examples/gcc/ld.manifest.template, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvk3um8PGphPydtN_K:-MQvk3um8PGphPydtN_L:b-20okfp) ([raw file](https://github.com/oscarlab/graphene/blob/303528131c67f58aeee677397ade9593f222ae88/Examples/gcc/ld.manifest.template#L1)):*
> ```
> loader.preload = "file:$(GRAPHENEDIR)/Runtime/libsysdb.so"
> ```

This file won't exist after the rebase to the latest master.
___
*[Examples/r/rm.manifest.template, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvkQnTB9QuaJO9TLT2:-MQvkQnTB9QuaJO9TLT3:b-20okfp) ([raw file](https://github.com/oscarlab/graphene/blob/303528131c67f58aeee677397ade9593f222ae88/Examples/r/rm.manifest.template#L1)):*
> ```
> # rm manifest example
> ```

This file won't exist after the rebase to the latest master.
___
*[Examples/r/sh.manifest.template, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvkRv2Ddn4ykaSmi6l:-MQvkRv2Ddn4ykaSmi6m:b-20okfp) ([raw file](https://github.com/oscarlab/graphene/blob/303528131c67f58aeee677397ade9593f222ae88/Examples/r/sh.manifest.template#L1)):*
> ```
> # sh manifest example
> ```

This file won't exist after the rebase to the latest master.
___
*[LibOS/shim/test/ltp/manifest.template, line 6 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvl7IcC4oQkvWZiUSu:-MQvl7IcC4oQkvWZiUSv:b-5xikvf) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/LibOS/shim/test/ltp/manifest.template#L6)):*
> ```
> loader.env.LD_PRELOAD = "$(COREUTILS_LIBDIR)/libstdbuf.so"
> loader.env._STDBUF_O = "L"
> loader.log_level = "error"
> ```

Why do you need it explicitly? Just remove it.
___
*[LibOS/shim/test/native/exec_fork.manifest.template, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvlKW1EftO3oJQlIVV:-MQvlKW1EftO3oJQlIVW:ba98sy5) ([raw file](https://github.com/oscarlab/graphene/blob/303528131c67f58aeee677397ade9593f222ae88/LibOS/shim/test/native/exec_fork.manifest.template#L1)):*
> ```
> loader.preload = "file:$(SHIMPATH)"
> ```

Some of the below files won't exist after the rebase to the latest master. Don't be surprised.
___
*[Pal/regression/Bootstrap3.manifest.template, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvlpM83PBWkS1YVJTt:-MQvlpM83PBWkS1YVJTu:ba98sy5) ([raw file](https://github.com/oscarlab/graphene/blob/303528131c67f58aeee677397ade9593f222ae88/Pal/regression/Bootstrap3.manifest.template#L1)):*
> ```
> loader.debug_type = "inline"
> ```

Some of the below files won't exist after the rebase to the latest master. Don't be surprised.
___
*[Pal/src/db_main.c, line 160 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvmLAL-NhUnavqN4vD:-MQvmLAL-NhUnavqN4vE:b-x00hqg) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/db_main.c#L160)):*
> ```C
>     ret = toml_string_in(g_pal_state.manifest_root, "loader.debug_type", &debug_type);
>     if (ret < 0)
>         INIT_FAIL_MANIFEST(PAL_ERROR_DENIED, "Cannot parse \'loader.debug_type\'");
> ```

There is no need for `\'`, just `'` is enough. I erroneously added this escaping in the previous PRs. Can you remove it from your PR everywhere, including this line?
___
*[Pal/src/db_main.c, line 166 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvm98S1xck_1ckCvg5:-MQvm98TCPqlYHDErnkq:bszhwbc) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/db_main.c#L166)):*
> ```C
>             PAL_ERROR_DENIED,
>             "'loader.debug_type' has been replaced by 'loader.log_level' "
>             "and 'loader.log_file'");
> ```

This indentation is a bit weird, can you fix it to occupy two 100-char lines fully?
___
*[Pal/src/db_main.c, line 168 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvmssE65mP1UHvvE9-:-MQvmssF5gVyp-cTe_Ma:bbi765v) ([raw file](https://github.com/oscarlab/graphene/blob/303528131c67f58aeee677397ade9593f222ae88/Pal/src/db_main.c#L168)):*
> ```C
>     if (!strcmp(debug_type, "inline")) {
>         // TODO: use /dev/stderr instead?
>         ret = _DkInitDebugStream("/dev/stdout");
> ```

Aren't you supposed to do this if there is no `loader.log_file` in the manifest?

UPDATE: Nevermind, I see you introduced `PAL_LOG_DEFAULT_FD 1`.
___
*[Pal/src/db_main.c, line 204 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvpCWH-4wSi1cyp9Gt:-MQvpCWH-4wSi1cyp9Gu:bv8i6cu) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/db_main.c#L204)):*
> ```C
> 
>         if (ret < 0)
>             INIT_FAIL(-ret, "Cannot open debug stream");
> ```

Why do you write `debug stream`? Isn't it more like `debug log file`?
___
*[Pal/src/printf.c, line 67 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvnRtC3dxXZIfQJM_Q:-MQvnRtC3dxXZIfQJM_R:bezmr39) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/printf.c#L67)):*
> ```C
>     return cnt;
> }
> EXTERN_ALIAS(printf);
> ```

Btw, we should remove this `printf()` and co after this PR. There must be no places that just want to do a `printf()` in our Graphene code.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 851 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvojS011ajg58TecVe:-MQvojS011ajg58TecVf:b-bmgz2t) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/host/Linux-SGX/sgx_main.c#L851)):*
> ```C
> #endif
> 
>     int log_level = PAL_LOG_ERROR;
> ```

Shouldn't it be `PAL_LOG_DEFAULT_LEVEL`?
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 855 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvomqKB_iTrvq7Ma4N:-MQvomqKB_iTrvq7Ma4O:byspopd) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/host/Linux-SGX/sgx_main.c#L855)):*
> ```C
>     ret = toml_string_in(manifest_root, "loader.log_level", &log_level_str);
>     if (ret < 0) {
>         SGX_DBG(DBG_E, "Cannot parse \'loader.log_level\'\n");
> ```

Please remove escaping (`\'` -> `'`), here and below
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 894 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvpJbz6xJx-vzNn93I:-MQvpJbz6xJx-vzNn93J:bv8i6cu) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/host/Linux-SGX/sgx_main.c#L894)):*
> ```C
> 
>         if (ret < 0) {
>             SGX_DBG(DBG_E, "Cannot open debug stream: %d\n", ret);
> ```

Why do you write `debug stream`? Isn't it more like `debug log file`?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: 15 of 94 files reviewed, 16 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @pwmarcz, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvfcnvAlro3RPPTaZD:-MQwIG7_Dv4mntAAPc4v:bsjh9vh) (no related file):*

Shouldn't you merge two separate commits in this PR into one? Otherwise the example manifests will fail if we bisect between these two commits.
</blockquote></details>

Makes sense. I made it into a single commit.

---
___
*[Examples/gcc/as.manifest.template, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvjxhy4VEPVdlmOEVP:-MQwIYU59hdZMgK-or2I:bn5vmux) ([raw file](https://github.com/oscarlab/graphene/blob/303528131c67f58aeee677397ade9593f222ae88/Examples/gcc/as.manifest.template#L1)):*

This file won't exist after the rebase to the latest master.
</blockquote></details>

Rebased, all these files are gone now.
___
*[Pal/src/db_main.c, line 160 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvmLAL-NhUnavqN4vD:-MQwIpvb4ldHotU_3Szf:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/db_main.c#L160)):*

There is no need for `\'`, just `'` is enough. I erroneously added this escaping in the previous PRs. Can you remove it from your PR everywhere, including this line?
</blockquote></details>

Done.
___
*[Pal/src/db_main.c, line 166 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvm98S1xck_1ckCvg5:-MQwJ3D94l-vjGqFIbqp:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/db_main.c#L166)):*

This indentation is a bit weird, can you fix it to occupy two 100-char lines fully?
</blockquote></details>

Done.
___
*[Pal/src/db_main.c, line 204 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvpCWH-4wSi1cyp9Gt:-MQwJAEvApTKuWWc9C84:be2vh8r) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/db_main.c#L204)):*

Why do you write `debug stream`? Isn't it more like `debug log file`?
</blockquote></details>

Makes sense. Changed to just "log file".
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 851 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvojS011ajg58TecVe:-MQwJhjn3BOauCQpAi1M:bidrs83) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/host/Linux-SGX/sgx_main.c#L851)):*

Shouldn't it be `PAL_LOG_DEFAULT_LEVEL`?
</blockquote></details>

Right. Fixed.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 855 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvomqKB_iTrvq7Ma4N:-MQwJiraEZgirZ8CJGAK:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/host/Linux-SGX/sgx_main.c#L855)):*

Please remove escaping (`\'` -> `'`), here and below
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 894 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvpJbz6xJx-vzNn93I:-MQwJk474U2zSkyboPDZ:b-9uuokx) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/host/Linux-SGX/sgx_main.c#L894)):*

Why do you write `debug stream`? Isn't it more like `debug log file`?
</blockquote></details>

Updated same as above.

dimakuv (changes requested):
> Reviewed 86 of 116 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz)

___
*[Examples/tensorflow-lite/label_image.manifest.template, line 5 at r2](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQzs7aECZ3r_-27pJ8V:-MQzs7aECZ3r_-27pJ8W:b-bfh39g) ([raw file](https://github.com/oscarlab/graphene/blob/3d31f2d18d1a395d76d2ee0a9fea313bedd3c842/Examples/tensorflow-lite/label_image.manifest.template#L5)):*
> ```
> loader.env.LD_LIBRARY_PATH = "/lib:."
> loader.env.PATH = "/bin:/usr/bin"
> loader.debug_type = "none"
> ```

Can you actually replace `loader.log_level = "$(GRAPHENE_LOG_LEVEL)"` here? It looks like a bug that we hard-coded `none` in this example.
___
*[LibOS/shim/test/ltp/manifest.template, line 6 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvl7IcC4oQkvWZiUSu:-MQzsL1N4Ji84pm-3wY2:bomt3ft) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/LibOS/shim/test/ltp/manifest.template#L6)):*

Why do you need it explicitly? Just remove it.
</blockquote></details>

My comment still stands.
___
*[Pal/src/db_main.c, line 166 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvm98S1xck_1ckCvg5:-MQzsimG1snUxJRGniL0:b-y8cvyc) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/db_main.c#L166)):*

Done.
</blockquote></details>

I also meant to change to this:
```
        INIT_FAIL_MANIFEST(PAL_ERROR_DENIED,
            "'loader.debug_type' has been replaced by 'loader.log_level' and 'loader.log_file'");
```

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: 91 of 94 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Examples/tensorflow-lite/label_image.manifest.template, line 5 at r2](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQzs7aECZ3r_-27pJ8V:-MR-P6vzCt_qw3Xn8rPR:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3d31f2d18d1a395d76d2ee0a9fea313bedd3c842/Examples/tensorflow-lite/label_image.manifest.template#L5)):*

Can you actually replace `loader.log_level = "$(GRAPHENE_LOG_LEVEL)"` here? It looks like a bug that we hard-coded `none` in this example.
</blockquote></details>

Done.
___
*[LibOS/shim/test/ltp/manifest.template, line 6 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvl7IcC4oQkvWZiUSu:-MR-PCUw9ISMSFVMvGwX:bqmbug5) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/LibOS/shim/test/ltp/manifest.template#L6)):*

My comment still stands.
</blockquote></details>

Right, I must have missed that.. Removed.
___
*[Pal/src/db_main.c, line 166 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvm98S1xck_1ckCvg5:-MR-PUP21jQyC6Cnssq1:b-l75zro) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/db_main.c#L166)):*

I also meant to change to this:
```
        INIT_FAIL_MANIFEST(PAL_ERROR_DENIED,
            "'loader.debug_type' has been replaced by 'loader.log_level' and 'loader.log_file'");
```
</blockquote></details>

Okay. I haven't considered it because the surrounding code doesn't follow this convention (and instead indents the next line same as opening bracket). But I don't think it's a big deal. Updated.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 8 of 111 files at r1, 103 of 116 files at r2, 3 of 3 files at r3, 16 of 16 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQ7DHz9kqMum8VIpAR:-MRQ7DHz9kqMum8VIpAS:b-pdk3k) (no related file):*
> Adjust levels (for example, system calls probably belong at "debug" or "trace")

Putting syscalls into a separate "trace" level sounds like an awesome idea, will clean up the log significantly (and we could stop printing this on Jenkins and only print on debug level, this way our logs won't occupy almost 1GB from a single debug run).

---
___
*[Documentation/manifest-syntax.rst, line 31 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQAsN15zOmilfpJbJM:-MRQAsN15zOmilfpJbJN:bozt4ts) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Documentation/manifest-syntax.rst#L31)):*
> ```reStructuredText
> ::
> 
>     loader.log_level = "[none|error|info|debug|trace|all]"
> ```

What's the point of having `all`? Does it have anything more than `trace`?
___
*[Examples/apache/Makefile, line 35 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQBbuQ3PHqFocnKiw1:-MRQBbuQ3PHqFocnKiw2:bh63gr1) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Examples/apache/Makefile#L35)):*
> ```Makefile
> 
> ifeq ($(DEBUG),1)
> GRAPHENE_LOG_LEVEL = info
> ```

Not `debug`? Why? That's what I would expect on debug builds.
ditto for the whole PR
___
*[Examples/busybox/busybox.manifest.template, line 20 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQCESd8nNridYU_Ogt:-MRQCESd8nNridYU_Ogu:b-lylzx8) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Examples/busybox/busybox.manifest.template#L20)):*
> ```
> GRAPHENEDEBUG macro is expanded
> ```

This is stale
___
*[Examples/memcached/memcached.manifest.template, line 30 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQCP6T1sXqnh6UvssQ:-MRQCP6U1hSgotUJXQbX:b-wcjjfb) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Examples/memcached/memcached.manifest.template#L30)):*
> ```
> GRAPHENEDEBUG macro is expanded
> ```

ditto
___
*[Examples/openvino/object_detection_sample_ssd.manifest.template, line 8 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQDBuyDCLeW4MZLbH-:-MRQDBuyDCLeW4MZLbH0:bpk6625) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Examples/openvino/object_detection_sample_ssd.manifest.template#L8)):*
> ```
> inline/none
> ```

ditto + please grep the whole repo for `'inline'`, `inline/none` and `'none'`, there are still multiple occurrences.
___
*[Examples/redis/redis-server.manifest.template, line 15 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQCcLXEDzGVHceogc8:-MRQCcLYDRapUzX_n-wk:b-wcjjfb) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Examples/redis/redis-server.manifest.template#L15)):*
> ```
> GRAPHENEDEBUG macro is expanded
> ```

ditto
___
*[LibOS/shim/include/shim_internal.h, line 52 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQGbee6cvsHoRIKi82:-MRQGbee6cvsHoRIKi83:b6rvufi) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/LibOS/shim/include/shim_internal.h#L52)):*
> ```objc
>         if ((level) <= g_log_level)                  \
>             debug_printf(fmt);                       \
>     }  while(0)
> ```

accidental whitespace change? should be `} while (0)`
___
*[Pal/src/db_main.c, line 199 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQJq4A8cQDH5qOgl2q:-MRQJq4BAM1RmSE8aGKs:b-e1zfcs) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/db_main.c#L199)):*
> ```C
>     if (log_level > PAL_LOG_NONE && log_file) {
>         ret = _DkInitDebugStream(log_file);
>         free(log_file);
> ```

This may leak if `log_level <= PAL_LOG_NONE && log_file` holds, you should just free it outside the `if` (using the property that `free` on `NULL` is a no-op).
___
*[Pal/src/printf.c, line 67 at r1](https://reviewable.io/reviews/oscarlab/graphene/2075#-MQvnRtC3dxXZIfQJM_Q:-MRQL6kB6ej6Vz23FMgN:b32v57) ([raw file](https://github.com/oscarlab/graphene/blob/67f8e42c196b23da691ae5e75b39aa48021922a5/Pal/src/printf.c#L67)):*

Btw, we should remove this `printf()` and co after this PR. There must be no places that just want to do a `printf()` in our Graphene code.
</blockquote></details>

+1
___
*[Pal/src/printf.c, line 51 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQKobSDkQU0tEuIbGD:-MRQKobSDkQU0tEuIbGE:b-3atxgx) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/printf.c#L51)):*
> ```C
>     b.idx = 0;
>     b.cnt = 0;
>     vfprintfmt((void*)&fputch, NULL, &b, fmt, ap);
> ```

this function pointer cast is shady, but I see that this problem existed already before this PR, so I'm not blocking
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 312 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQL_qrDI0Vi2HAaDe0:-MRQL_qrDI0Vi2HAaDe1:bktlel4) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/pal_linux.h#L312)):*
> ```
> sgx_log_* (outside of enclave)
> ```

We should get rid of this confusing naming (that `sgx_` prefix means "outside the enclave"), or at least not introduce new places which are using it.
Any ideas for a better prefix here?
___
*[Pal/src/host/Linux-SGX/sgx_log.h, line 11 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQMFeN69gIajVVh6B4:-MRQMFeN69gIajVVh6B5:b-s0ryji) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/sgx_log.h#L11)):*
> ```objc
>  */
> 
> #ifndef _SGX_LOG_H
> ```

`_SGX_LOG_H` -> `SGX_LOG_H_` (reserved name)
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 869 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQN68p9OeY87MU-lke:-MRQN68p9OeY87MU-lkf:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/sgx_main.c#L869)):*
> ```C
>     if (log_level > PAL_LOG_NONE && log_file) {
>         ret = sgx_log_init(log_file);
>         free(log_file);
> ```

ditto

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow and @pwmarcz)

___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 312 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQL_qrDI0Vi2HAaDe0:-MRUG3fk30TePzCheWcq:b-soxbek) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/pal_linux.h#L312)):*

> ```
> sgx_log_* (outside of enclave)
> ```

We should get rid of this confusing naming (that `sgx_` prefix means "outside the enclave"), or at least not introduce new places which are using it.
Any ideas for a better prefix here?
</blockquote></details>

Why do we even need to have two different names (`log_*` for inside the enclave and `sgx_log_*` for outside)? Can't we just use the same function name `log_*`?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: 42 of 90 files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @yamahata)

___
*[Documentation/manifest-syntax.rst, line 31 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQAsN15zOmilfpJbJM:-MRU8ywb-T3AuxEseAL6:btnlc0z) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Documentation/manifest-syntax.rst#L31)):*

What's the point of having `all`? Does it have anything more than `trace`?
</blockquote></details>

The idea was that `all` and `none` are not used by any messages, but used for configuring the system: in case someone adds more levels above or below the existing ones, `all` and `none` will still mean all messages / no messages.

For comparison: [Java log4j](https://github.com/apache/logging-log4j2/blob/master/log4j-api/src/main/java/org/apache/logging/log4j/Level.java) (`OFF` and `ALL`), [Python logging](https://docs.python.org/3/library/logging.html#logging-levels) (`NOTSET`).
___
*[Examples/apache/Makefile, line 35 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQBbuQ3PHqFocnKiw1:-MRUAY3HCEuGJq0xyxc0:b-aacgo0) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Examples/apache/Makefile#L35)):*

Not `debug`? Why? That's what I would expect on debug builds.
ditto for the whole PR
</blockquote></details>

I kept it at `info` because I kept LibOS messages by default at `info`. But I agree, `debug` is a better default (and it still works the same for these LibOS messages for now). Changed.
___
*[Examples/busybox/busybox.manifest.template, line 20 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQCESd8nNridYU_Ogt:-MRUD9BGEtg9EwhGPw2s:b-bw8e14) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Examples/busybox/busybox.manifest.template#L20)):*

> ```
> GRAPHENEDEBUG macro is expanded
> ```

This is stale
</blockquote></details>

Updated.
___
*[Examples/memcached/memcached.manifest.template, line 30 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQCP6T1sXqnh6UvssQ:-MRUDBHd97QyjwN9sPBv:b-bw8e14) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Examples/memcached/memcached.manifest.template#L30)):*

> ```
> GRAPHENEDEBUG macro is expanded
> ```

ditto
</blockquote></details>

Updated.
___
*[Examples/openvino/object_detection_sample_ssd.manifest.template, line 8 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQDBuyDCLeW4MZLbH-:-MRUMdLe60_DjqsU0V67:b-6tfkun) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Examples/openvino/object_detection_sample_ssd.manifest.template#L8)):*

> ```
> inline/none
> ```

ditto + please grep the whole repo for `'inline'`, `inline/none` and `'none'`, there are still multiple occurrences.
</blockquote></details>

Ouch. The duplication in these examples is pretty awful, with several generations of these comments. Maybe it shouldn't be one makefile and one manifest template?

Anyway, I updated all the comments I could find.
___
*[Examples/redis/redis-server.manifest.template, line 15 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQCcLXEDzGVHceogc8:-MRUFPSB00-WJQJ_94qI:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Examples/redis/redis-server.manifest.template#L15)):*

> ```
> GRAPHENEDEBUG macro is expanded
> ```

ditto
</blockquote></details>

Done.
___
*[LibOS/shim/include/shim_internal.h, line 52 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQGbee6cvsHoRIKi82:-MRUFZ8W5iAU5qMsmxED:bidrs83) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/LibOS/shim/include/shim_internal.h#L52)):*

accidental whitespace change? should be `} while (0)`
</blockquote></details>

Right. Fixed.
___
*[Pal/src/db_main.c, line 199 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQJq4A8cQDH5qOgl2q:-MRUGYBV4EdAAlmAzA14:bm5tk02) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/db_main.c#L199)):*

This may leak if `log_level <= PAL_LOG_NONE && log_file` holds, you should just free it outside the `if` (using the property that `free` on `NULL` is a no-op).
</blockquote></details>

Done (and I moved the `free` above for consistency).
___
*[Pal/src/printf.c, line 51 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQKobSDkQU0tEuIbGD:-MRUHeto9VWIQawdn1gT:bncly7z) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/printf.c#L51)):*

this function pointer cast is shady, but I see that this problem existed already before this PR, so I'm not blocking
</blockquote></details>

This is the reason:

```c
void vfprintfmt(int (*_fputch)(void*, int, void*), void* f, void* putdat, const char* fmt,
                va_list ap) __attribute__((format(printf, 4, 0))); 
```

It's simple to fix (the `fputch` should have a matching signature and do a cast), so I did it.
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 312 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQL_qrDI0Vi2HAaDe0:-MRUIQOE0IYMZ6SkAMjr:b-eeobdr) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/pal_linux.h#L312)):*
>We should get rid of this confusing naming (that sgx_ prefix means "outside the enclave"), or at least not introduce new places which are using it.

I agree the naming is confusing and I would rather change sgx_ prefix to something else. But I think we should change it globally, not introduce a new one in one place and have two prefixes in use at the same time.

>Why do we even need to have two different names (log_* for inside the enclave and sgx_log_* for outside)? Can't we just use the same function name log_*?

This is because the headers are a mess, and outer-PAL includes inner-PAL headers pretty freely. The `log_*` macros are defined in `pal_internal.h`, and if we redefine them in an outer-PAL-specific header, then we are dependent on the order in which these two headers are included.

There are probably ways of solving it, but I'm afraid it would still be pretty brittle. So while having to use `sgx_log_*` is a bit harder on the user, it's also clearer and less magical (these are two separate logging subsystems, after all).
___
*[Pal/src/host/Linux-SGX/sgx_log.h, line 11 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQMFeN69gIajVVh6B4:-MRUKcNG2ZGqMo1LOcv0:b-w2nbd7) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/sgx_log.h#L11)):*

`_SGX_LOG_H` -> `SGX_LOG_H_` (reserved name)
</blockquote></details>

Changed. But we have absolutely no consistency in these.
___
*[Pal/src/host/Linux-SGX/sgx_main.c, line 869 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQN68p9OeY87MU-lke:-MRUKrCl9KGYdN54nYue:b-hawh6y) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/sgx_main.c#L869)):*

ditto
</blockquote></details>

Changed same as above.

dimakuv (changes requested):
> Reviewed 7 of 16 files at r4, 48 of 48 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[Examples/nginx/nginx.manifest.template, line 8 at r5](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRUb_h--Zsx03nTa9vT:-MRUb_h--Zsx03nTa9vU:b-scl8do) ([raw file](https://github.com/oscarlab/graphene/blob/726372ab5a6a831e451b96b0ed89663a8c084eda/Examples/nginx/nginx.manifest.template#L8)):*
> ```
> 
> # Path to the library OS
> loader.preload = "file:$(GRAPHENE_DIR)/Runtime/libsysdb.so"
> ```

With an underscore? Looks like a bug crept in.
___
*[LibOS/shim/include/shim_internal.h, line 52 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQGbee6cvsHoRIKi82:-MRUc0iOAgbFv5LFgX71:bulygo3) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/LibOS/shim/include/shim_internal.h#L52)):*

Right. Fixed.
</blockquote></details>

Is it?
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 312 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQL_qrDI0Vi2HAaDe0:-MRUcHJuC2GyUqD_J7fB:bkq973u) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/pal_linux.h#L312)):*

>We should get rid of this confusing naming (that sgx_ prefix means "outside the enclave"), or at least not introduce new places which are using it.

I agree the naming is confusing and I would rather change sgx_ prefix to something else. But I think we should change it globally, not introduce a new one in one place and have two prefixes in use at the same time.

>Why do we even need to have two different names (log_* for inside the enclave and sgx_log_* for outside)? Can't we just use the same function name log_*?

This is because the headers are a mess, and outer-PAL includes inner-PAL headers pretty freely. The `log_*` macros are defined in `pal_internal.h`, and if we redefine them in an outer-PAL-specific header, then we are dependent on the order in which these two headers are included.

There are probably ways of solving it, but I'm afraid it would still be pretty brittle. So while having to use `sgx_log_*` is a bit harder on the user, it's also clearer and less magical (these are two separate logging subsystems, after all).
</blockquote></details>

Ok, got it.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: 87 of 90 files reviewed, 12 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Examples/nginx/nginx.manifest.template, line 8 at r5](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRUb_h--Zsx03nTa9vT:-MRUdMvY-780MiEBT3Py:b-xmm9eh) ([raw file](https://github.com/oscarlab/graphene/blob/726372ab5a6a831e451b96b0ed89663a8c084eda/Examples/nginx/nginx.manifest.template#L8)):*

With an underscore? Looks like a bug crept in.
</blockquote></details>

Good catch... I copied it from another template, but it looks like some files use `GRAPHENEDIR`, and some use `GRAPHENE_DIR` :(
Fixed (here and in two other places).
___
*[LibOS/shim/include/shim_internal.h, line 52 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQGbee6cvsHoRIKi82:-MRUe9Nh9tMl2UdhuLUb:b-a0oe7u) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/LibOS/shim/include/shim_internal.h#L52)):*

Is it?
</blockquote></details>

This is what I did:

```diff
-    }  while(0)
+    } while(0)
```

Am I missing something?

mkow (commented):
> Reviewed 45 of 48 files at r5, 3 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 31 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQAsN15zOmilfpJbJM:-MRUmQ4J2nAKeseTPHVO:b-rsu3mq) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Documentation/manifest-syntax.rst#L31)):*

The idea was that `all` and `none` are not used by any messages, but used for configuring the system: in case someone adds more levels above or below the existing ones, `all` and `none` will still mean all messages / no messages.

For comparison: [Java log4j](https://github.com/apache/logging-log4j2/blob/master/log4j-api/src/main/java/org/apache/logging/log4j/Level.java) (`OFF` and `ALL`), [Python logging](https://docs.python.org/3/library/logging.html#logging-levels) (`NOTSET`).
</blockquote></details>

ok, makes sense
___
*[Examples/openvino/object_detection_sample_ssd.manifest.template, line 8 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQDBuyDCLeW4MZLbH-:-MRUqBGY3pStpS5162Xt:b-3n2q7) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Examples/openvino/object_detection_sample_ssd.manifest.template#L8)):*

Ouch. The duplication in these examples is pretty awful, with several generations of these comments. Maybe it shouldn't be one makefile and one manifest template?

Anyway, I updated all the comments I could find.
</blockquote></details>

Yeah, the duplication here is really unbearable. I felt it the most when fixing them up in #2050.
We need to somehow make them much more concise, but still quite self-contained (i.e. no dependencies on the Makefile terror of the core Graphene).
___
*[LibOS/shim/include/shim_internal.h, line 52 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQGbee6cvsHoRIKi82:-MRUrOt-E4ruAYA_WCTh:b32kk9v) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/LibOS/shim/include/shim_internal.h#L52)):*

This is what I did:

```diff
-    }  while(0)
+    } while(0)
```

Am I missing something?
</blockquote></details>

There should be a space after `while` keyword (we put it after `if`, `for`, etc.)
___
*[Pal/src/printf.c, line 51 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQKobSDkQU0tEuIbGD:-MRUs3BhDH6dmEdZEA5Z:b-6mhyx0) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/printf.c#L51)):*

This is the reason:

```c
void vfprintfmt(int (*_fputch)(void*, int, void*), void* f, void* putdat, const char* fmt,
                va_list ap) __attribute__((format(printf, 4, 0))); 
```

It's simple to fix (the `fputch` should have a matching signature and do a cast), so I did it.
</blockquote></details>

Great, thanks!
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 312 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQL_qrDI0Vi2HAaDe0:-MRUsRGw8AKz9fn1YxxC:b7dyyc2) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/pal_linux.h#L312)):*
> I agree the naming is confusing and I would rather change sgx_ prefix to something else. But I think we should change it globally, not introduce a new one in one place and have two prefixes in use at the same time.

Yeah, but something like "untrusted_log" or similar would not be confusing, and IMO would be much better as a temporary measure than the extremely (even I get confused by it sometimes, despite knowing about it) confusing `sgx_`.
___
*[Pal/src/host/Linux-SGX/sgx_log.h, line 11 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQMFeN69gIajVVh6B4:-MRUt19H0L-ZLiJjy4jo:bbmeqqt) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/sgx_log.h#L11)):*

Changed. But we have absolutely no consistency in these.
</blockquote></details>

Yes, there's little consistency in this project, unfortunately.
Btw. I personally prefer `#pragma once`, but that's a long discussion for another time ;)

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[LibOS/shim/include/shim_internal.h, line 52 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQGbee6cvsHoRIKi82:-MRUx7lx9gQdHT99XjAL:by7biuu) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/LibOS/shim/include/shim_internal.h#L52)):*

There should be a space after `while` keyword (we put it after `if`, `for`, etc.)
</blockquote></details>

Oh, right. Fixed.
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 312 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQL_qrDI0Vi2HAaDe0:-MRUxTdk4gqaa_qmQlY1:b-5q72ll) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/pal_linux.h#L312)):*

> I agree the naming is confusing and I would rather change sgx_ prefix to something else. But I think we should change it globally, not introduce a new one in one place and have two prefixes in use at the same time.

Yeah, but something like "untrusted_log" or similar would not be confusing, and IMO would be much better as a temporary measure than the extremely (even I get confused by it sometimes, despite knowing about it) confusing `sgx_`.
</blockquote></details>

Okay. Are you sure about "untrusted_log" though? It suggests crossing some trust boundary (it's used this way in inner PAL when referring to outer memory) or signals that the user has to be somehow more careful (similar to common conventions like unsafe_*). Do we have any better name for the outer PAL?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 312 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQL_qrDI0Vi2HAaDe0:-MRV1cAb2VhZIPUhdLd4:bdwxiu6) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/pal_linux.h#L312)):*

Okay. Are you sure about "untrusted_log" though? It suggests crossing some trust boundary (it's used this way in inner PAL when referring to outer memory) or signals that the user has to be somehow more careful (similar to common conventions like unsafe_*). Do we have any better name for the outer PAL?
</blockquote></details>

Yeah, true. Maybe "urts_"? Harder to decipher, but we use it sometimes in Graphene. It means "untrusted runtime".

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: 86 of 90 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @yamahata)

___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 312 at r4](https://reviewable.io/reviews/oscarlab/graphene/2075#-MRQL_qrDI0Vi2HAaDe0:-MRVFv6lBO1M7gXF4Hq4:bcqgbhp) ([raw file](https://github.com/oscarlab/graphene/blob/3e82aaf1654afeff7c9f00cef107ba019f635856/Pal/src/host/Linux-SGX/pal_linux.h#L312)):*

Yeah, true. Maybe "urts_"? Harder to decipher, but we use it sometimes in Graphene. It means "untrusted runtime".
</blockquote></details>

OK. Changed.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> Reviewed 1 of 1 files at r7, 4 of 4 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2075)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2079, opened 2021-01-13
[Pal/Linux-SGX] Remove loader.nonpie_binary manifest option

mkow:
## Description of the changes 

Non-PIE binaries support requires ELRANGE to start at low addresses, which on older SGX drivers required root permissions or reconfiguring the host, so this mapping strategy was made optional by us. This issue was fixed long time ago, so we can drop this option and always start enclaves at 0.

## How to test this PR? 

Run this on various SGX drivers and check if everything still works.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2079)


mkow:
Too bad, seems that out-of-tree driver still doesn't support this and the issue was fixed only in DCAP. We'll need to wait for better times with this PR.

PR #2102, opened 2021-01-19
LibOS: Return -EFAULT on name = NULL for sethostname/setdomainname (UBSAN)

stefanberger:
## Description of the changes 

This patch fixes an error detected with UBSAN where we have to return -EFAULT on name = NULL for sethostname and setdomainname syscalls.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)

### DIFF
--- a/LibOS/shim/src/sys/shim_getrandom.c
+++ b/LibOS/shim/src/sys/shim_getrandom.c
@@ -20,7 +20,7 @@ long shim_do_getrandom(char* buf, size_t count, unsigned int flags) {
     if (count > INT_MAX)
         count = INT_MAX;
 
-    if (test_user_memory(buf, count, /*write=*/true))
+    if (!buf || test_user_memory(buf, count, /*write=*/true))
         return -EFAULT;
 
     /* In theory, DkRandomBitsRead may block on some PALs (which conflicts with GRND_NONBLOCK flag),
--- a/LibOS/shim/src/sys/shim_socket.c
+++ b/LibOS/shim/src/sys/shim_socket.c
@@ -1154,7 +1154,7 @@ static int check_msghdr(struct msghdr* msg, bool is_recv) {
         return -EMSGSIZE;
     }
 
-    if (test_user_memory(msg->msg_iov, size, /*write=*/false)) {
+    if (!msg->msg_iov || test_user_memory(msg->msg_iov, size, /*write=*/false)) {
         return -EFAULT;
     }
 
--- a/LibOS/shim/src/sys/shim_uname.c
+++ b/LibOS/shim/src/sys/shim_uname.c
@@ -44,7 +44,7 @@ long shim_do_sethostname(char* name, int len) {
     if (len < 0 || (size_t)len >= sizeof(g_current_uname.nodename))
         return -EINVAL;
 
-    if (test_user_memory(name, len, /*write=*/false))
+    if (!name || test_user_memory(name, len, /*write=*/false))
         return -EFAULT;
 
     memcpy(&g_current_uname.nodename, name, len);
@@ -56,7 +56,7 @@ long shim_do_setdomainname(char* name, int len) {
     if (len < 0 || (size_t)len >= sizeof(g_current_uname.domainname))
         return -EINVAL;
 
-    if (test_user_memory(name, len, /*write=*/false))
+    if (!name || test_user_memory(name, len, /*write=*/false))
         return -EFAULT;
 
     memcpy(&g_current_uname.domainname, name, len);


stefanberger:
Finding those LTP-level bugs went a lot faster than anticipated since the pattern was always the same... there are still some left, though.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2102#-MRS3yzb8q9ehcoLz-ZY:-MRS3yzc7Grz8OwxUBSG:b-60wcl8) (no related file):*
This is actually a bug in `test_user_memory`, but this function is so hacky that I'm fine with this temporary solution (but it's not really correct, users can have memory mapped at zero sometimes and syscalls work correctly on it if it's actually mapped).

---

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

dimakuv:
Jenkins, test this please

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: :shipit: complete! all files reviewed, all discussions resolved

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, 1 unresolved discussion (waiting on @stefanberger)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2102#-MRS3yzb8q9ehcoLz-ZY:-MRWwpFyDm-4JF1z4jEP:b-4t7m38) (no related file):*

This is actually a bug in `test_user_memory`, but this function is so hacky that I'm fine with this temporary solution (but it's not really correct, users can have memory mapped at zero sometimes and syscalls work correctly on it if it's actually mapped).
</blockquote></details>

As far as `test_user_memory` is broken, it does handle `NULL` well (maybe not well, but as good as any other address). I'm not sure what went wrong exactly, but this is not the fix.

---

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2102#-MRS3yzb8q9ehcoLz-ZY:-MRX69U67N7mS4eUSFtP:bs7xgo8) (no related file):*

As far as `test_user_memory` is broken, it does handle `NULL` well (maybe not well, but as good as any other address). I'm not sure what went wrong exactly, but this is not the fix.
</blockquote></details>

There are tons of locations that look like this here:
```
long shim_do_read(int fd, void* buf, size_t count) {
    if (!buf || test_user_memory(buf, count, true))
        return -EFAULT;

long shim_do_write(int fd, const void* buf, size_t count) {
    if (!buf || test_user_memory((void*)buf, count, false))
        return -EFAULT;

long shim_do_pwrite64(int fd, char* buf, size_t count, loff_t pos) {
    if (!buf || test_user_memory(buf, count, false))
        return -EFAULT;

```

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, 1 unresolved discussion (waiting on @stefanberger)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2102#-MRS3yzb8q9ehcoLz-ZY:-MRX7tJU1iMeYue8hmVE:bu3cff5) (no related file):*

There are tons of locations that look like this here:
```
long shim_do_read(int fd, void* buf, size_t count) {
    if (!buf || test_user_memory(buf, count, true))
        return -EFAULT;

long shim_do_write(int fd, const void* buf, size_t count) {
    if (!buf || test_user_memory((void*)buf, count, false))
        return -EFAULT;

long shim_do_pwrite64(int fd, char* buf, size_t count, loff_t pos) {
    if (!buf || test_user_memory(buf, count, false))
        return -EFAULT;

```
</blockquote></details>

That doesn't make it correct. Graphene is full of places that need to go away ...

---

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2102#-MRS3yzb8q9ehcoLz-ZY:-MRXAK2z87SUfqV5HRI7:bnslw74) (no related file):*

That doesn't make it correct. Graphene is full of places that need to go away ...
</blockquote></details>

Then at least it would be justifiable that it follows an established pattern that can be cleaned up once test_user_memory() handles the NULL pointer correctly. Of course it should handle address 0x8 correctly as well. 
I suppose the fix would be to walk the VMas there? It seems to actually test memory by writing to it, which can end up re-writing bytes in Graphene code itself... ?

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, 1 unresolved discussion (waiting on @stefanberger)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2102#-MRS3yzb8q9ehcoLz-ZY:-MRXD4D9EjhU4NVCQmbd:b827hh1) (no related file):*

Then at least it would be justifiable that it follows an established pattern that can be cleaned up once test_user_memory() handles the NULL pointer correctly. Of course it should handle address 0x8 correctly as well. 
I suppose the fix would be to walk the VMas there? It seems to actually test memory by writing to it, which can end up re-writing bytes in Graphene code itself... ?
</blockquote></details>

`test_user_memory` handles the NULL pointer correctly right now, it's just that this function is one big undefined behavior.
I have the rewrite of it in my pipeline, but I'm waiting for the signals PR to land before it.

---

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2102#-MRS3yzb8q9ehcoLz-ZY:-MRzTYA46JX5IGC89j8F:b-ht2gt) (no related file):*

`test_user_memory` handles the NULL pointer correctly right now, it's just that this function is one big undefined behavior.
I have the rewrite of it in my pipeline, but I'm waiting for the signals PR to land before it.
</blockquote></details>

So we won't merge this but have to wait until the signal stuff, which I saw is using the same pattern as below in some cases, is merged in?

---

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2102#-MRS3yzb8q9ehcoLz-ZY:-MRzZfGS90BJRYwCGqhY:b-wx7uc9) (no related file):*

So we won't merge this but have to wait until the signal stuff, which I saw is using the same pattern as below in some cases, is merged in?
</blockquote></details>

What Borys means is that we don't understand why your patch would change anything, as `test_user_memory` seems to handle NULL pointers already. The fact that it's very hacky is a separate issue, but it seems that it should work in this case without the additional `if`s you added. Quoting him: "As far as test_user_memory is broken, it does handle NULL well (maybe not well, but as good as any other address). I'm not sure what went wrong exactly, but this is not the fix.".
So, what was broken exactly before this patch? And why did it change?

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2102#-MRS3yzb8q9ehcoLz-ZY:-MRzbCbR2sHtmEUhfAcW:b-d0mgh3) (no related file):*

What Borys means is that we don't understand why your patch would change anything, as `test_user_memory` seems to handle NULL pointers already. The fact that it's very hacky is a separate issue, but it seems that it should work in this case without the additional `if`s you added. Quoting him: "As far as test_user_memory is broken, it does handle NULL well (maybe not well, but as good as any other address). I'm not sure what went wrong exactly, but this is not the fix.".
So, what was broken exactly before this patch? And why did it change?
</blockquote></details>

1) We don't have to wait for signals PR, it's completely orthogonal.
2) The PR you mentioned is using a similar but a bit different pattern: `NULL` argument is valid there and means that this parameter is not used, hence checks like `if (ss && test_user_memory(ss, ...)`

---

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2102#-MRS3yzb8q9ehcoLz-ZY:-MRzoqw59uwNCc8fpKcC:bz79f1j) (no related file):*

1) We don't have to wait for signals PR, it's completely orthogonal.
2) The PR you mentioned is using a similar but a bit different pattern: `NULL` argument is valid there and means that this parameter is not used, hence checks like `if (ss && test_user_memory(ss, ...)`
</blockquote></details>

My current guess is that UBSAN jumps on writing to the NULL pointer:

```
UndefinedBehaviorSanitizer (UBSan) is a fast undefined behavior detector. UBSan modifies the program at compile-time to catch various kinds of undefined behavior during program execution, for example:

    Using misaligned or null pointer
    Signed integer overflow
    Conversion to, from, or between floating-point types which would overflow the destination
```

https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

---

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2102#-MRS3yzb8q9ehcoLz-ZY:-MRzsR9gBjkk6wdFg9ZG:bafztxq) (no related file):*

My current guess is that UBSAN jumps on writing to the NULL pointer:

```
UndefinedBehaviorSanitizer (UBSan) is a fast undefined behavior detector. UBSan modifies the program at compile-time to catch various kinds of undefined behavior during program execution, for example:

    Using misaligned or null pointer
    Signed integer overflow
    Conversion to, from, or between floating-point types which would overflow the destination
```

https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

</blockquote></details>

So then the correct fix is to disable UBSAN in the `test_user_memory()` function using `__attribute__((no_sanitize("undefined")))` until this part is rewritten (based on VMA walk I suppose).

---

stefanberger (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2102)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2102#-MRS3yzb8q9ehcoLz-ZY:-MRzv5Eo979d1ff9-zw-:b-o7q21n) (no related file):*

So then the correct fix is to disable UBSAN in the `test_user_memory()` function using `__attribute__((no_sanitize("undefined")))` until this part is rewritten (based on VMA walk I suppose).
</blockquote></details>

Let me close this PR (wrong title) and send another one just with this fix.

---

Issue #1520, opened 2020-05-19
Future direction of Graphene Secure Containers


vahldiek:
## Description of the problem

The first version of Graphene Secure Containers is worked on in PR #1430. This issue lists limitations and possible future features, and should serve as a discussion forum for others to highlight which features are a priority.

- List of trusted files does not include files in folders of symlinks
  - Implement os.walk in [finalize_manifests.py](https://github.com/oscarlab/graphene/blob/133fb6a21d0689b4665bd5a7637231bc51f3317a/Tools/gsc/finalize_manifests.py#L21) to follow symbolic links while avoiding cyclic dependencies
- Change test Docker image repositories to public repositories such as [python](https://hub.docker.com/_/python) or [nginx](https://hub.docker.com/_/nginx)
- Test and support protectFS
- Instead of relying on a list of trusted files, translate the Docker image to a protectFS image to avoid large amounts of metadata and startup time
- Support RPM-based distributions (different dependencies & yum instead of apt)
- Support for additional Docker `build` arguments such as `build_args` or `network_mode` in the configuration file
- Detect scripts as entrypoints and correctly change the entrypoint to the script interpreter
- Provide flag to switch off overestimation to allow more security conscious application of GSC
- Support local SGX driver header files instead of header files from repository
- Support using binary distribution of Graphene instead of compiling it in multi-stage Docker build (1st stage is obsolete with binary distribution of Graphene), e.g. files from local storage
- Support for specification of parent-child relationship instead of the implemented overestimation
- Support for non-ASCII characters in filenames of trusted files (mostly issue in Graphene manifest handling)
- Support to supply the signing file in the configuration, use it for signing and then destroy

dimakuv:
Some thoughts on this list, coming from January 2021 (after GSC rework):

##  GSC uses insecure environment variables

This is not even true. GSC doesn't do *anything* about environment variables (which is a bug). in Graphene, there must be one of the following manifest options: `loader.insecure__use_host_env = 1` or `loader.env.[ENVIRON] = "[VALUE]" / 
loader.env_src_file = "file:file_with_serialized_envs"` (see https://graphene.readthedocs.io/en/latest/manifest-syntax.html#environment-variables). However, none of these options are specified in the GSC template manifest, check https://github.com/oscarlab/graphene/blob/master/Tools/gsc/templates/entrypoint.manifest.template.

[...]

PR #2112, opened 2021-01-21
[GSC] Refactor GSC and make it work again

### LINKED ISSUE #2117: docker build of gsc does not take env into account
## Description of the problem
Sometimes one might need a custom docker host. (kubernetes docker for my case) However, `gsc build` command fails to build on the custom docker host pointed by the bash environment variables. Instead, it issues the build command on the default docker host which is listening on `unix://var/run/docker.sock`. 

I have analyzed some of the code on the repository. I have tested using commit `a37a85ba` but the latest commit `0a88f93d` doesn't seem to be much modified on this part. 

At `gsc.py:220`, the script issues the build command using function `build_docker_image`. 
Then, at `gsc.py:224` the `gsc.py` script checks whether the build succeeded using the `docker_socket` client object constructed at `gsc.py:201` using `docker.from_env()`.  However, inspecting the function `build_docker_image` in `gsc.py:162` the build was issued on the default docker environment using APIClient class object `docker_api`. 

When using kubernetes, the default docker daemon is not used, so this leads to a misleading behavior. 

## Steps to reproduce

I have tested using commit `a37a85ba` but the latest commit `0a88f93d` doesn't seem to be much modified on this part. I have used this commit because the `Tools/gsc/config.yaml.template` of the master branch uses this commit. 

I was trying to use gsc on docker of kubernetes. 

```
minikube start

eval $(minikube docker-env)

cd $GRAPHENE_DIR/Tools/gsc

./gsc build --insecure-args python test/ubuntu18.04-python3.manifest
```

## Expected results
output ending with follows. workaround applied.
```
Successfully built d0d55e66b30d
Successfully tagged gsc-python-unsigned:latest
Successfully graphenized docker image python into docker image gsc-python-unsigned
```

## Actual results
output ending with follows.
```
Successfully built ec9e5cceb893
Successfully tagged gsc-python-unsigned:latest
Failed to build graphenized image for python
```

## Additional information

I have found 2 workarounds for making this work. 

# Method 1
Just use method `docker_socket.images.build` instead of function `build_docker_image`. However, this workaround does not show the docker build command output, i.e. the docker build command runs silently. I figure the source tried to avoid this. Am I right?

# Method 2
Modify `build_docker_image` function. Currently the function newly constructs the APIClient object from socket `unix://var/run/docker.sock`. Instead, receive the backend of `docker_socket` object used in functions `gsc_build`, `gsc_sign_image`, `gsc_build_graphene` and others. We could access the backend of `docker_socket` by referencing `docker_socket.api`. Warning, [this](https://github.com/docker/docker-py/blob/ce2669e3edfe5d3215ba501cc9771fc0ffad680a/docker/client.py#L45) is not a documented feature of the package and could be changed, as far as I know. See `git diff` output. 

```
diff --git a/Tools/gsc/gsc.py b/Tools/gsc/gsc.py
index 565e7f47..d3e6b67e 100755
--- a/Tools/gsc/gsc.py
+++ b/Tools/gsc/gsc.py
@@ -159,8 +159,8 @@ def get_docker_image(docker_socket, image):
     except (docker.errors.ImageNotFound, docker.errors.APIError):
         return None
 
-def build_docker_image(path, name, dockerfile, **kwargs):
-    docker_api = docker.APIClient(base_url='unix://var/run/docker.sock')
+def build_docker_image(docker_api, path, name, dockerfile, **kwargs):
+    #docker_api = docker.APIClient(base_url='unix://var/run/docker.sock')
     # Docker build returns stream of json output
     stream = docker_api.build(path=path,
                               tag=name,
@@ -217,7 +217,7 @@ def gsc_build(args):
 
     buildargs_dict = extract_build_args(args)
 
-    build_docker_image(gsc_image_name(image), gsc_unsigned_image_name(image), 'Dockerfile.build',
+    build_docker_image(docker_socket.api, gsc_image_name(image), gsc_unsigned_image_name(image), 'Dockerfile.build',
                        rm=args.rm, nocache=args.no_cache, buildargs=buildargs_dict)
 
     # Check if docker build failed
@@ -269,7 +269,7 @@ def gsc_build_graphene(args):
     else:
         buildargs_dict = extract_build_args(args)
 
-        build_docker_image(image, image, 'Dockerfile.compile',
+        build_docker_image(docker_socket.api, image, image, 'Dockerfile.compile',
                        rm=args.rm, nocache=args.no_cache, buildargs=buildargs_dict)
 
         # Check if docker build failed
@@ -314,7 +314,7 @@ def gsc_sign_image(args):
     try:
         # We force the removal of intermediate Docker images to not leave the signing
         # key in a Docker container.
-        build_docker_image(gsc_image_name(image), gsc_image_name(image),
+        build_docker_image(docker_socket.api, gsc_image_name(image), gsc_image_name(image),
                            'Dockerfile.sign_manifests', forcerm=True)
 
     finally:
```

dimakuv:
## Description of the changes 

Commit "Introduce one, central manifest, zero-config children and constant MRENCLAVE" broke GSC. This commit fixes GSC (mainly adjusts it the single-manifest change in that commit). Also, significant internal refactoring is done (no user-visible changes).

Fixes #2117.

## How to test this PR? 

If someone re-enables the GSC pipeline in Jenkins, then it should pass. Alternatively, just do the tests manually.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: 0 of 20 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2112#-MR_pSW_23zu9Re9A7Sb:-MR_pSWaCKebwHH4VqZ1:b-hyj6ru) (no related file):*
Summoning @vahldiek for a quick review, just to have an ack from him.

---

vahldiek (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: 0 of 20 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2112#-MR_pSW_23zu9Re9A7Sb:-MR_sY5e0FakJX54h-Kb:b-1jtwzu) (no related file):*

Summoning @vahldiek for a quick review, just to have an ack from him.
</blockquote></details>

look good to me

---

dimakuv:
Jenkins, retest Jenkins-18.04 please (internet connectivity)

dimakuv:
Jenkins, retest Jenkins-18.04 please (internet connectivity)

dimakuv:
Jenkins, retest Jenkins please (internet connectivity)

dimakuv:
Jenkins, test Jenkins-SGX-gsc please

woju:
Jenkins, test Jenkins-SGX-gsc please

woju:
Jenkins, test Jenkins-SGX-gsc please

mkow (changes requested):
> Reviewed 18 of 20 files at r1, 1 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Documentation/manpages/gsc.rst, line 60 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS98D8S5J80sdbZr9CS:-MS98D8S5J80sdbZr9CT:br79vst) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L60)):*
> ```reStructuredText
> - `Intel SGX driver <https://github.com/intel/linux-sgx-driver>`__
> - `Intel SGX SDK <https://01.org/intel-software-guard-extensions/downloads>`__
> - `Graphene SGX Driver (kernel module) <https://github.com/oscarlab/graphene-sgx-driver>`__
> ```

we should drop this and ask to use a kernel with fsgsbase
___
*[Documentation/manpages/gsc.rst, line 70 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS98eHZDFz3p6WNyKNX:-MS98eHZDFz3p6WNyKNY:bchyvss) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L70)):*
> ```
>    You could also run commands as root.
> ```

I'd drop this sentence, the whole warning sounds really weird with it.
___
*[Documentation/manpages/gsc.rst, line 114 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS99J5w9A3S6sybZEo5:-MS99J5w9A3S6sybZEo6:bf461s5) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L114)):*
> ```reStructuredText
>    any arguments specified during :command:`docker run` are ignored.
> 
> .. option:: -nc
> ```

shouldn't this be `--`, not `-`?
___
*[Documentation/manpages/gsc.rst, line 149 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS99cdyCUPfrq0AaAtD:-MS99cdyCUPfrq0AaAtE:b8mqsp1) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L149)):*
> ```reStructuredText
> Docker image called ``gsc-<IMAGE-NAME>``. :command:`gsc sign-image` always
> removes intermediate Docker images, if successful or not, to ensure the removal
> of the signing key in intermediate Docker images.
> ```

`in intermediate Docker images` -> `in them` (repetition)
___
*[Documentation/manpages/gsc.rst, line 238 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS9ApzQE7oG5H1ZEw-t:-MS9ApzQE7oG5H1ZEw-u:bbwbk6k) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L238)):*
> ```
> Similarly to Docker, Graphene has a corresponding option (``libos.entrypoint``) [...]
> ```

Not sure if this whole sentence is actually needed, seems the users doesn't need to know about manifests?
___
*[Documentation/manpages/gsc.rst, line 269 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS9BklhBZwKYolDadt8:-MS9BklhBZwKYolDadt9:bs71oh1) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L269)):*
> ```reStructuredText
>    SGX token and starts the :program:`pal-Linux-SGX` loader. Note that the
>    generated image (``gsc-<image-name>-unsigned``) cannot successfully load an
>    Intel SGX enclave, since essential files and the signing of the enclave are
> ```

`signing of the enclave` -> `signature of the enclave`
___
*[Documentation/manpages/gsc.rst, line 281 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS9CCWPEwVWwpc91WjK:-MS9CCWPEwVWwpc91WjL:b-gfhohg) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L281)):*
> ```
> Note that prebuilt Graphene images will be provided for popular cloud-provider
> offerings or can be created via :command:`gsc build-graphene`.
> ```

`Note that` -> `In the future we plan to provide [...]`
___
*[Documentation/manpages/gsc.rst, line 434 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS9Egs7CRmnDsX89d-T:-MS9Egs7CRmnDsX89d-U:b7mdbho) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L434)):*
> ```reStructuredText
>    .. code-block:: sh
> 
>       docker run --device=/dev/gsgx --device=/dev/sgx/enclave \
> ```

` SGXDriver.Repository` default value is pointing to the old OOT (and basically most of this guide is pointing to it).
___
*[Tools/gsc/finalize_manifest.py, line 42 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSJJ3PTDq5kad2Wwx6z:-MSJJ3PTDq5kad2Wwx7-:b-707tol) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Tools/gsc/finalize_manifest.py#L42)):*
> ```Python
>             if exclude_re.match(filename) or not os.path.isfile(filename):
>                 continue
>             if not is_ascii(filename):
> ```

that's not enough, there are still names which will break the generation, e.g. `asdf"qwer`. Not sure about newlines, we'd need to verify with tomlc99.
___
*[Tools/gsc/gsc.py, line 254 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSOtHN62wWgKRUUCb6F:-MSOtHN62wWgKRUUCb6G:bnk2j3w) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Tools/gsc/gsc.py#L254)):*
> ```Python
> 
>     try:
>         # forcerm parameter forces removal of intermediate Docker images even after unsuccessful
> ```

please wrap `forcerm` with backticks, looked like a typo for me at first ;)
___
*[Tools/gsc/templates/entrypoint.manifest.template, line 20 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSOoPejDagTTcj4FPZM:-MSOoPek3rGwS4pa8tkV:b-e3w6cd) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Tools/gsc/templates/entrypoint.manifest.template#L20)):*
> ```
> # !! INSECURE !! Allow passing command-line arguments from the host without validation.
> # Most Docker images rely on runtime arguments and hence, a more general technique is required.
> # The issue is documented for future release of GSC in issue #1520.
> ```

Please link to the issue, the numbers may change if we migrate the Graphene repo or GSC.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSOuMBL7QN_8cwj2dW4:-MSOuMBL7QN_8cwj2dW5:b-10r070) (no related file):*
Waaait, what's that thing in the Jenkins log, of that successful run?

```
make[1]: Entering directory '/home/jenkins/workspace/graphene-sgx-gsc/Tools/gsc/test'
echo "Building base image ubuntu18.04-python3.dockerfile..."
Building base image ubuntu18.04-python3.dockerfile...
docker build --rm --no-cache -t ubuntu18.04-python3-a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f -f ubuntu18.04-python3.dockerfile ../../../Examples
free(): invalid pointer
SIGABRT: abort
PC=0x7fde76fa4fb7 m=0 sigcode=18446744073709551610
signal arrived during cgo execution

goroutine 1 [syscall, locked to thread]:
runtime.cgocall(0x4afd50, 0xc42005fcc0, 0xc42005fce8)
	/usr/lib/go-1.8/src/runtime/cgocall.go:131 +0xe2 fp=0xc42005fc90 sp=0xc42005fc50
github.com/docker/docker-credential-helpers/secretservice._Cfunc_free(0xa16270)
	github.com/docker/docker-credential-helpers/secretservice/_obj/_cgo_gotypes.go:111 +0x41 fp=0xc42005fcc0 sp=0xc42005fc90
github.com/docker/docker-credential-helpers/secretservice.Secretservice.List.func5(0xa16270)
	/build/golang-github-docker-docker-credential-helpers-cMhSy1/golang-github-docker-docker-credential-helpers-0.5.0/obj-x86_64-linux-gnu/src/github.com/docker/docker-credential-helpers/secretservice/secretservice_linux.go:96 +0x60 fp=0xc42005fcf8 sp=0xc42005fcc0
github.com/docker/docker-credential-helpers/secretservice.Secretservice.List(0x0, 0x756060, 0xc4200143d0)
	/build/golang-github-docker-docker-credential-helpers-cMhSy1/golang-github-docker-docker-credential-helpers-0.5.0/obj-x86_64-linux-gnu/src/github.com/docker/docker-credential-helpers/secretservice/secretservice_linux.go:97 +0x217 fp=0xc42005fda0 sp=0xc42005fcf8
github.com/docker/docker-credential-helpers/secretservice.(*Secretservice).List(0x77e548, 0xc42005fe88, 0x410022, 0xc420014330)
	<autogenerated>:4 +0x46 fp=0xc42005fde0 sp=0xc42005fda0
github.com/docker/docker-credential-helpers/credentials.List(0x756ba0, 0x77e548, 0x7560e0, 0xc42000e018, 0x0, 0x10)
	/build/golang-github-docker-docker-credential-helpers-cMhSy1/golang-github-docker-docker-credential-helpers-0.5.0/obj-x86_64-linux-gnu/src/github.com/docker/docker-credential-helpers/credentials/credentials.go:145 +0x3e fp=0xc42005fe68 sp=0xc42005fde0
github.com/docker/docker-credential-helpers/credentials.HandleCommand(0x756ba0, 0x77e548, 0x7ffe2c183f1c, 0x4, 0x7560a0, 0xc42000e010, 0x7560e0, 0xc42000e018, 0x40e398, 0x4d35c0)
	/build/golang-github-docker-docker-credential-helpers-cMhSy1/golang-github-docker-docker-credential-helpers-0.5.0/obj-x86_64-linux-gnu/src/github.com/docker/docker-credential-helpers/credentials/credentials.go:60 +0x16d fp=0xc42005fed8 sp=0xc42005fe68
github.com/docker/docker-credential-helpers/credentials.Serve(0x756ba0, 0x77e548)
	/build/golang-github-docker-docker-credential-helpers-cMhSy1/golang-github-docker-docker-credential-helpers-0.5.0/obj-x86_64-linux-gnu/src/github.com/docker/docker-credential-helpers/credentials/credentials.go:41 +0x1cb fp=0xc42005ff58 sp=0xc42005fed8
main.main()
	/build/golang-github-docker-docker-credential-helpers-cMhSy1/golang-github-docker-docker-credential-helpers-0.5.0/secretservice/cmd/main_linux.go:9 +0x4f fp=0xc42005ff88 sp=0xc42005ff58
runtime.main()
	/usr/lib/go-1.8/src/runtime/proc.go:185 +0x20a fp=0xc42005ffe0 sp=0xc42005ff88
runtime.goexit()
	/usr/lib/go-1.8/src/runtime/asm_amd64.s:2197 +0x1 fp=0xc42005ffe8 sp=0xc42005ffe0

goroutine 17 [syscall, locked to thread]:
runtime.goexit()
	/usr/lib/go-1.8/src/runtime/asm_amd64.s:2197 +0x1

rax    0x0
rbx    0x7ffe2c182b30
rcx    0x7fde76fa4fb7
rdx    0x0
rdi    0x2
rsi    0x7ffe2c1828c0
rbp    0x7ffe2c182c30
rsp    0x7ffe2c1828c0
r8     0x0
r9     0x7ffe2c1828c0
r10    0x8
r11    0x246
r12    0x7ffe2c182b30
r13    0x1000
r14    0x0
r15    0x30
rip    0x7fde76fa4fb7
rflags 0x246
cs     0x33
fs     0x0
gs     0x0
Sending build context to Docker daemon   1.54MB
```

See https://leeroy.cs.unc.edu/job/graphene-sgx-gsc/1021/consoleFull.
How could it succeed?

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSOuMBL7QN_8cwj2dW4:-MSSdq50-3dVoHbPZq22:bevkfmy) (no related file):*

Waaait, what's that thing in the Jenkins log, of that successful run?

```
make[1]: Entering directory '/home/jenkins/workspace/graphene-sgx-gsc/Tools/gsc/test'
echo "Building base image ubuntu18.04-python3.dockerfile..."
Building base image ubuntu18.04-python3.dockerfile...
docker build --rm --no-cache -t ubuntu18.04-python3-a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f -f ubuntu18.04-python3.dockerfile ../../../Examples
free(): invalid pointer
SIGABRT: abort
PC=0x7fde76fa4fb7 m=0 sigcode=18446744073709551610
signal arrived during cgo execution

goroutine 1 [syscall, locked to thread]:
runtime.cgocall(0x4afd50, 0xc42005fcc0, 0xc42005fce8)
	/usr/lib/go-1.8/src/runtime/cgocall.go:131 +0xe2 fp=0xc42005fc90 sp=0xc42005fc50
github.com/docker/docker-credential-helpers/secretservice._Cfunc_free(0xa16270)
	github.com/docker/docker-credential-helpers/secretservice/_obj/_cgo_gotypes.go:111 +0x41 fp=0xc42005fcc0 sp=0xc42005fc90
github.com/docker/docker-credential-helpers/secretservice.Secretservice.List.func5(0xa16270)
	/build/golang-github-docker-docker-credential-helpers-cMhSy1/golang-github-docker-docker-credential-helpers-0.5.0/obj-x86_64-linux-gnu/src/github.com/docker/docker-credential-helpers/secretservice/secretservice_linux.go:96 +0x60 fp=0xc42005fcf8 sp=0xc42005fcc0
github.com/docker/docker-credential-helpers/secretservice.Secretservice.List(0x0, 0x756060, 0xc4200143d0)
	/build/golang-github-docker-docker-credential-helpers-cMhSy1/golang-github-docker-docker-credential-helpers-0.5.0/obj-x86_64-linux-gnu/src/github.com/docker/docker-credential-helpers/secretservice/secretservice_linux.go:97 +0x217 fp=0xc42005fda0 sp=0xc42005fcf8
github.com/docker/docker-credential-helpers/secretservice.(*Secretservice).List(0x77e548, 0xc42005fe88, 0x410022, 0xc420014330)
	<autogenerated>:4 +0x46 fp=0xc42005fde0 sp=0xc42005fda0
github.com/docker/docker-credential-helpers/credentials.List(0x756ba0, 0x77e548, 0x7560e0, 0xc42000e018, 0x0, 0x10)
	/build/golang-github-docker-docker-credential-helpers-cMhSy1/golang-github-docker-docker-credential-helpers-0.5.0/obj-x86_64-linux-gnu/src/github.com/docker/docker-credential-helpers/credentials/credentials.go:145 +0x3e fp=0xc42005fe68 sp=0xc42005fde0
github.com/docker/docker-credential-helpers/credentials.HandleCommand(0x756ba0, 0x77e548, 0x7ffe2c183f1c, 0x4, 0x7560a0, 0xc42000e010, 0x7560e0, 0xc42000e018, 0x40e398, 0x4d35c0)
	/build/golang-github-docker-docker-credential-helpers-cMhSy1/golang-github-docker-docker-credential-helpers-0.5.0/obj-x86_64-linux-gnu/src/github.com/docker/docker-credential-helpers/credentials/credentials.go:60 +0x16d fp=0xc42005fed8 sp=0xc42005fe68
github.com/docker/docker-credential-helpers/credentials.Serve(0x756ba0, 0x77e548)
	/build/golang-github-docker-docker-credential-helpers-cMhSy1/golang-github-docker-docker-credential-helpers-0.5.0/obj-x86_64-linux-gnu/src/github.com/docker/docker-credential-helpers/credentials/credentials.go:41 +0x1cb fp=0xc42005ff58 sp=0xc42005fed8
main.main()
	/build/golang-github-docker-docker-credential-helpers-cMhSy1/golang-github-docker-docker-credential-helpers-0.5.0/secretservice/cmd/main_linux.go:9 +0x4f fp=0xc42005ff88 sp=0xc42005ff58
runtime.main()
	/usr/lib/go-1.8/src/runtime/proc.go:185 +0x20a fp=0xc42005ffe0 sp=0xc42005ff88
runtime.goexit()
	/usr/lib/go-1.8/src/runtime/asm_amd64.s:2197 +0x1 fp=0xc42005ffe8 sp=0xc42005ffe0

goroutine 17 [syscall, locked to thread]:
runtime.goexit()
	/usr/lib/go-1.8/src/runtime/asm_amd64.s:2197 +0x1

rax    0x0
rbx    0x7ffe2c182b30
rcx    0x7fde76fa4fb7
rdx    0x0
rdi    0x2
rsi    0x7ffe2c1828c0
rbp    0x7ffe2c182c30
rsp    0x7ffe2c1828c0
r8     0x0
r9     0x7ffe2c1828c0
r10    0x8
r11    0x246
r12    0x7ffe2c182b30
r13    0x1000
r14    0x0
r15    0x30
rip    0x7fde76fa4fb7
rflags 0x246
cs     0x33
fs     0x0
gs     0x0
Sending build context to Docker daemon   1.54MB
```

See https://leeroy.cs.unc.edu/job/graphene-sgx-gsc/1021/consoleFull.
How could it succeed?
</blockquote></details>

After reading through internet, this is a commonly seen benign bug in Docker. It was fixed in the mainline Docker but seems like not backported in Ubuntu 18.04. See these:
- https://github.com/docker/docker-credential-helpers/issues/103
- https://github.com/docker/docker-credential-helpers/pull/29 -- this is the fix
- https://github.com/docker/docker-credential-helpers/issues/103#issuecomment-479508473 -- the favorite way of fixing this.

So looks like we can ignore this benign error on Docker side. Or fix it via `dpkg -r --force-depends golang-docker-credential-helpers` on that Jenkins worker. @woju what would be your advise?

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @woju)

___
*[Documentation/manpages/gsc.rst, line 114 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS99J5w9A3S6sybZEo5:-MSSf5hz3R-0r3Q-Ro3x:b44fq18) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L114)):*

shouldn't this be `--`, not `-`?
</blockquote></details>

No, the option is either `-nc` or `--no-cache`. So this is correct.
___
*[Documentation/manpages/gsc.rst, line 434 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS9Egs7CRmnDsX89d-T:-MSSfHsi1vdXnx7bttQQ:b-9d6uyg) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L434)):*

` SGXDriver.Repository` default value is pointing to the old OOT (and basically most of this guide is pointing to it).
</blockquote></details>

Not sure what you're suggesting here. We have `SGXDriver.Repository` pointing to the old OOT driver because this is what is used in our Jenkins. Anyway, this is something that the user must change for himself.

Or do you mean to simply replace `/dev/sgx/enclave` here with `/dev/isgx`, for consistency in all GSC docs/files?
___
*[Tools/gsc/finalize_manifest.py, line 42 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSJJ3PTDq5kad2Wwx6z:-MSSfeyIEq8cU0vHXmKj:b3dur6a) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Tools/gsc/finalize_manifest.py#L42)):*

that's not enough, there are still names which will break the generation, e.g. `asdf"qwer`. Not sure about newlines, we'd need to verify with tomlc99.
</blockquote></details>

Yeah, I'm not sure what to do here? Should I remove this check completely and hope that the filenames are TOML-formatted? Should I white-list only latin alphanumerics plus a couple special signs?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[Documentation/manpages/gsc.rst, line 114 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS99J5w9A3S6sybZEo5:-MSSn0He3ELHlIRCys9T:b-rs2375) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L114)):*

No, the option is either `-nc` or `--no-cache`. So this is correct.
</blockquote></details>

I meant that usually you have `--` for long options and `-` for single-letter ones, so you can concat them (i.e. `-nc` is the same as `-n -c`). Seems confusing, as all other options used here seem to stick to this convention, except this one.
___
*[Documentation/manpages/gsc.rst, line 434 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS9Egs7CRmnDsX89d-T:-MSSnNidC8eS1nfHfjYW:b-4flevz) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L434)):*
> We have `SGXDriver.Repository` pointing to the old OOT driver because this is what is used in our Jenkins.

That's a bad argument for user-facing defaults ;) Users should not need to know about our Jenkins.

> Or do you mean to simply replace /dev/sgx/enclave here with /dev/isgx, for consistency in all GSC docs/files?

Whichever way, but make it consistent.
___
*[Tools/gsc/finalize_manifest.py, line 42 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSJJ3PTDq5kad2Wwx6z:-MSSnqUc2c6mhL2a6uLK:b24ijeq) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Tools/gsc/finalize_manifest.py#L42)):*

Yeah, I'm not sure what to do here? Should I remove this check completely and hope that the filenames are TOML-formatted? Should I white-list only latin alphanumerics plus a couple special signs?
</blockquote></details>

Maybe just escape `"` and `\`? Isn't this enough? (but please check with TOML specs/tomlc99 code, because I'm not sure about newlines and similar stuff)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: 16 of 20 files reviewed, 12 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[Documentation/manpages/gsc.rst, line 60 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS98D8S5J80sdbZr9CS:-MSWl40X3R8zYiI9D1nt:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L60)):*

we should drop this and ask to use a kernel with fsgsbase
</blockquote></details>

Done.
___
*[Documentation/manpages/gsc.rst, line 70 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS98eHZDFz3p6WNyKNX:-MSWl6gO4E5QLY8pWJIx:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L70)):*

> ```
>    You could also run commands as root.
> ```

I'd drop this sentence, the whole warning sounds really weird with it.
</blockquote></details>

Done.
___
*[Documentation/manpages/gsc.rst, line 114 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS99J5w9A3S6sybZEo5:-MSWlPB6EBpbFSQ51QAz:b-sa9pe1) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L114)):*

I meant that usually you have `--` for long options and `-` for single-letter ones, so you can concat them (i.e. `-nc` is the same as `-n -c`). Seems confusing, as all other options used here seem to stick to this convention, except this one.
</blockquote></details>

Done. Got it now, changed to `--`.
___
*[Documentation/manpages/gsc.rst, line 149 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS99cdyCUPfrq0AaAtD:-MSWl_5e5Xym0faR71lr:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L149)):*

`in intermediate Docker images` -> `in them` (repetition)
</blockquote></details>

Done.
___
*[Documentation/manpages/gsc.rst, line 238 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS9ApzQE7oG5H1ZEw-t:-MSWmBDu5gU8LqxFm9WN:b-puiu4v) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L238)):*

> ```
> Similarly to Docker, Graphene has a corresponding option (``libos.entrypoint``) [...]
> ```

Not sure if this whole sentence is actually needed, seems the users doesn't need to know about manifests?
</blockquote></details>

Done. I believe you added this :)

But yes, this whole section looks unnecessary. Users only use the original Docker image and (optionally) the extra-manifest (with things like `sgx.enclave_size`). They don't specify `libos.entrypoint` at all. So I removed the whole section.
___
*[Documentation/manpages/gsc.rst, line 269 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS9BklhBZwKYolDadt8:-MSWmaEv6CdhvUS_94VI:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L269)):*

`signing of the enclave` -> `signature of the enclave`
</blockquote></details>

Done.
___
*[Documentation/manpages/gsc.rst, line 281 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS9CCWPEwVWwpc91WjK:-MSWmolX61PW2qZArGU9:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L281)):*

> ```
> Note that prebuilt Graphene images will be provided for popular cloud-provider
> offerings or can be created via :command:`gsc build-graphene`.
> ```

`Note that` -> `In the future we plan to provide [...]`
</blockquote></details>

Done.
___
*[Documentation/manpages/gsc.rst, line 434 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS9Egs7CRmnDsX89d-T:-MSWmuK232qiQx2JwJnS:brz82hj) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L434)):*

> We have `SGXDriver.Repository` pointing to the old OOT driver because this is what is used in our Jenkins.

That's a bad argument for user-facing defaults ;) Users should not need to know about our Jenkins.

> Or do you mean to simply replace /dev/sgx/enclave here with /dev/isgx, for consistency in all GSC docs/files?

Whichever way, but make it consistent.
</blockquote></details>

Done. We're currently stuck with supporting multiple versions of the SGX driver. This is a sad reality our users should be aware of for at least a year or two.

I fixed to use old `isgx` in this example.
___
*[Tools/gsc/finalize_manifest.py, line 42 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSJJ3PTDq5kad2Wwx6z:-MSWq6818ulHHpfoizQK:b-klp4l1) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Tools/gsc/finalize_manifest.py#L42)):*

Maybe just escape `"` and `\`? Isn't this enough? (but please check with TOML specs/tomlc99 code, because I'm not sure about newlines and similar stuff)
</blockquote></details>

Done. I ended up using single-line basic strings in TOML. I only restrict newlines and carriage returns, and we must escape backslashes and quotes, but other stuff is supported (must be in UTF-8). See https://toml.io/en/v1.0.0#string
___
*[Tools/gsc/gsc.py, line 254 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSOtHN62wWgKRUUCb6F:-MSWqU9ZCdod0vveBeKo:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Tools/gsc/gsc.py#L254)):*

please wrap `forcerm` with backticks, looked like a typo for me at first ;)
</blockquote></details>

Done.
___
*[Tools/gsc/templates/entrypoint.manifest.template, line 20 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSOoPejDagTTcj4FPZM:-MSWqq8L92ZHIUFa88uO:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Tools/gsc/templates/entrypoint.manifest.template#L20)):*

Please link to the issue, the numbers may change if we migrate the Graphene repo or GSC.
</blockquote></details>

Done.

dimakuv:
Jenkins, retest Jenkins please (`sendfile08` LTP test failed, known issue)

mkow (changes requested):
> Reviewed 3 of 6 files at r4, 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[Documentation/manpages/gsc.rst, line 238 at r3](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS9ApzQE7oG5H1ZEw-t:-MS_AaCsC9zrFrAfZyqg:b-fr74hu) ([raw file](https://github.com/oscarlab/graphene/blob/a532c8a366b1d66b256c5d3c8e3ea56de7cbe27f/Documentation/manpages/gsc.rst#L238)):*
> Done. I believe you added this :)

Yes, I know :) But now I re-read the whole GSC documentation and noticed that the users don't really need to know our manifest syntax to use it, and moreover, this entry is automatically generated. So, it's not very useful to mention it here.
___
*[Tools/gsc/finalize_manifest.py, line 17 at r5](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS_C2Yy4_O7-9D3YDvl:-MS_C2Yy4_O7-9D3YDvm:b-v255h2) ([raw file](https://github.com/oscarlab/graphene/blob/dceefd684e0a2233affedbf9df672b4320f54ea1/Tools/gsc/finalize_manifest.py#L17)):*
> ```Python
> def is_utf8(string):
>     try:
>         string.encode('utf-8')
> ```

not `.decode()`? and `bytes`, not `string`

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[Tools/gsc/finalize_manifest.py, line 17 at r5](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS_C2Yy4_O7-9D3YDvl:-MSbVQe_EcyROFsyAB63:b-nuc5vg) ([raw file](https://github.com/oscarlab/graphene/blob/dceefd684e0a2233affedbf9df672b4320f54ea1/Tools/gsc/finalize_manifest.py#L17)):*

not `.decode()`? and `bytes`, not `string`
</blockquote></details>

No, my version looks correct. We pass the `str` object to this function -- which is the Unicode string in Python3. Then we `encode()` it to get the `bytes` object (these are the objects that are used to store to file).

Please see https://medium.com/better-programming/strings-unicode-and-bytes-in-python-3-everything-you-always-wanted-to-know-27dc02ff2686. Python3 `strings` vs `bytes` are a mess, but much better than Python2 :)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[Tools/gsc/finalize_manifest.py, line 17 at r5](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS_C2Yy4_O7-9D3YDvl:-MSboqKh3IDl96jHQdsr:bz7kn2i) ([raw file](https://github.com/oscarlab/graphene/blob/dceefd684e0a2233affedbf9df672b4320f54ea1/Tools/gsc/finalize_manifest.py#L17)):*

No, my version looks correct. We pass the `str` object to this function -- which is the Unicode string in Python3. Then we `encode()` it to get the `bytes` object (these are the objects that are used to store to file).

Please see https://medium.com/better-programming/strings-unicode-and-bytes-in-python-3-everything-you-always-wanted-to-know-27dc02ff2686. Python3 `strings` vs `bytes` are a mess, but much better than Python2 :)
</blockquote></details>

I understand bytes vs str in Python3, seems I got confused by the function name - `is_utf8` - UTF-8 is an encoding, so only a byte array can be "utf-8". What does this mean for a string?
Also, looking only at the implementation, how can something be non-encodable in UTF-8? I thought it can encode all unicode codepoints?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[Tools/gsc/finalize_manifest.py, line 17 at r5](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS_C2Yy4_O7-9D3YDvl:-MSbv52L1z_ruomFoWn3:b-rvq7b8) ([raw file](https://github.com/oscarlab/graphene/blob/dceefd684e0a2233affedbf9df672b4320f54ea1/Tools/gsc/finalize_manifest.py#L17)):*

I understand bytes vs str in Python3, seems I got confused by the function name - `is_utf8` - UTF-8 is an encoding, so only a byte array can be "utf-8". What does this mean for a string?
Also, looking only at the implementation, how can something be non-encodable in UTF-8? I thought it can encode all unicode codepoints?
</blockquote></details>

Yes, I think I didn't conceptualize it to myself previously. So the issue is like this:

1. Python3 uses utf-8, utf-16 or utf-32 to represent `string` objects internally (it chooses the fitting encoding at runtime, see https://www.python.org/dev/peps/pep-0393/).

2. TOML only accepts utf-8.

3. Some filenames in some Docker images surprisingly have non-utf-8 names (I got the error "cant parse this string" on some Python image, I think with `numpy`).

Thus, it is possible that a `filename` from `os.walk()` returns a utf-32 string which can't be represented in utf-8. So `string.encode('utf-8')` fails, and we skip outputting this filename to TOML.

Should I rename the function or add a comment?

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[Tools/gsc/finalize_manifest.py, line 17 at r5](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS_C2Yy4_O7-9D3YDvl:-MSc27AFB9bGGWqZiUMs:bfy8tea) ([raw file](https://github.com/oscarlab/graphene/blob/dceefd684e0a2233affedbf9df672b4320f54ea1/Tools/gsc/finalize_manifest.py#L17)):*

Yes, I think I didn't conceptualize it to myself previously. So the issue is like this:

1. Python3 uses utf-8, utf-16 or utf-32 to represent `string` objects internally (it chooses the fitting encoding at runtime, see https://www.python.org/dev/peps/pep-0393/).

2. TOML only accepts utf-8.

3. Some filenames in some Docker images surprisingly have non-utf-8 names (I got the error "cant parse this string" on some Python image, I think with `numpy`).

Thus, it is possible that a `filename` from `os.walk()` returns a utf-32 string which can't be represented in utf-8. So `string.encode('utf-8')` fails, and we skip outputting this filename to TOML.

Should I rename the function or add a comment?
</blockquote></details>

1. ? Why would this matter for anything? It's just the internal representation. Python strings AFAIK are conceptually just arrays of unicode codepoints.
2. Yes.
3. Filenames on Linux are not in UTF-8. They are arbitrary byte sequences without null bytes and slashes. I guess the bug is that GSC treats them as UTF-8 sequences?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[Tools/gsc/finalize_manifest.py, line 17 at r5](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS_C2Yy4_O7-9D3YDvl:-MSc7ME17rNyDZm_mkgG:b-noai46) ([raw file](https://github.com/oscarlab/graphene/blob/dceefd684e0a2233affedbf9df672b4320f54ea1/Tools/gsc/finalize_manifest.py#L17)):*

Yes, I think I didn't conceptualize it to myself previously. So the issue is like this:

1. Python3 uses utf-8, utf-16 or utf-32 to represent `string` objects internally (it chooses the fitting encoding at runtime, see https://www.python.org/dev/peps/pep-0393/).

2. TOML only accepts utf-8.

3. Some filenames in some Docker images surprisingly have non-utf-8 names (I got the error "cant parse this string" on some Python image, I think with `numpy`).

Thus, it is possible that a `filename` from `os.walk()` returns a utf-32 string which can't be represented in utf-8. So `string.encode('utf-8')` fails, and we skip outputting this filename to TOML.

Should I rename the function or add a comment?
</blockquote></details>

Sorry, I'm educating myself as I'm writing these comments. I was wrong in the previous comments.

https://www.b-list.org/weblog/2017/sep/05/how-python-does-unicode/ -- reading this document helped me clear the confusion between UTF-8 vs UTF-16 vs UTF-32 vs latin-1 vs UCS-2 and UCS-4.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[Tools/gsc/finalize_manifest.py, line 17 at r5](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS_C2Yy4_O7-9D3YDvl:-MSc8uBY2rwqHAB1FcS3:b-o4ehp0) ([raw file](https://github.com/oscarlab/graphene/blob/dceefd684e0a2233affedbf9df672b4320f54ea1/Tools/gsc/finalize_manifest.py#L17)):*

Sorry, I'm educating myself as I'm writing these comments. I was wrong in the previous comments.

https://www.b-list.org/weblog/2017/sep/05/how-python-does-unicode/ -- reading this document helped me clear the confusion between UTF-8 vs UTF-16 vs UTF-32 vs latin-1 vs UCS-2 and UCS-4.
</blockquote></details>

Let me debug what happens below. My head is spinning.

Ok, so my code basically does this, right:
```python
    for root, _, files in os.walk(root_dir, followlinks=False):
        for file in files:
            filename = os.path.join(root, file)
            filename.encode('utf-8') # this can fail!
```

We have the `string` Python object. We encode it into UTF-8. All strings must be representable in UTF-8, unless the string is not really a string but a sequence of non-UTF-8-formatted bytes...

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[Tools/gsc/finalize_manifest.py, line 17 at r5](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS_C2Yy4_O7-9D3YDvl:-MScB4mF5xN8R385XQGC:b-p96c09) ([raw file](https://github.com/oscarlab/graphene/blob/dceefd684e0a2233affedbf9df672b4320f54ea1/Tools/gsc/finalize_manifest.py#L17)):*

Let me debug what happens below. My head is spinning.

Ok, so my code basically does this, right:
```python
    for root, _, files in os.walk(root_dir, followlinks=False):
        for file in files:
            filename = os.path.join(root, file)
            filename.encode('utf-8') # this can fail!
```

We have the `string` Python object. We encode it into UTF-8. All strings must be representable in UTF-8, unless the string is not really a string but a sequence of non-UTF-8-formatted bytes...
</blockquote></details>

Wait, `os.walk` returns `bytes`, not `str`? I don't get what's happening in that listing :P

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[Tools/gsc/finalize_manifest.py, line 17 at r5](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS_C2Yy4_O7-9D3YDvl:-MScDeua6coRYgQYDRuX:b-vzrnr6) ([raw file](https://github.com/oscarlab/graphene/blob/dceefd684e0a2233affedbf9df672b4320f54ea1/Tools/gsc/finalize_manifest.py#L17)):*

Wait, `os.walk` returns `bytes`, not `str`? I don't get what's happening in that listing :P
</blockquote></details>

No, filenames have a type `str`. Just check this:
```python
for root, _, files in os.walk(".", followlinks=False):
    for file in files:
        print(f"{file} {type(file)}")
```

It prints things like:
```
.ubuntu18.04-bash.dockerfile.swp <class 'str'>
ubuntu18.04-pytorch.dockerfile <class 'str'>
.ubuntu18.04-numpy.dockerfile.swp <class 'str'>
```

(Same if I do `type(os.path.join(root, file))`)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[Tools/gsc/finalize_manifest.py, line 17 at r5](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS_C2Yy4_O7-9D3YDvl:-MScG0ZUBUExEnqc9OlR:b-oc0qrl) ([raw file](https://github.com/oscarlab/graphene/blob/dceefd684e0a2233affedbf9df672b4320f54ea1/Tools/gsc/finalize_manifest.py#L17)):*

No, filenames have a type `str`. Just check this:
```python
for root, _, files in os.walk(".", followlinks=False):
    for file in files:
        print(f"{file} {type(file)}")
```

It prints things like:
```
.ubuntu18.04-bash.dockerfile.swp <class 'str'>
ubuntu18.04-pytorch.dockerfile <class 'str'>
.ubuntu18.04-numpy.dockerfile.swp <class 'str'>
```

(Same if I do `type(os.path.join(root, file))`)
</blockquote></details>

Do `os.walk(b'.', ...)` and it will return bytes.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: 19 of 20 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[Tools/gsc/finalize_manifest.py, line 17 at r5](https://reviewable.io/reviews/oscarlab/graphene/2112#-MS_C2Yy4_O7-9D3YDvl:-MScY3ZjCxTAjGcDeaDY:bh38inh) ([raw file](https://github.com/oscarlab/graphene/blob/dceefd684e0a2233affedbf9df672b4320f54ea1/Tools/gsc/finalize_manifest.py#L17)):*

Do `os.walk(b'.', ...)` and it will return bytes.
</blockquote></details>

Done. I decided to have explicit `UTF-8` encoding everywhere to make it very clear that we (well, TOML) want this encoding and nothing else.
___
*[Tools/gsc/finalize_manifest.py, line 49 at r6](https://reviewable.io/reviews/oscarlab/graphene/2112#-MScYjuzCMjoud5eB7Ap:-MScYjuzCMjoud5eB7Aq:b-7ft9hp) ([raw file](https://github.com/oscarlab/graphene/blob/3a9613daf93b5fff53b8965219193965e19ef9c8/Tools/gsc/finalize_manifest.py#L49)):*
> ```Python
>                 # we append filenames as TOML strings which must be in UTF-8, ignore filenames
>                 # that cannot be represented in UTF-8 encoding
>                 print(f'\t[from inside Docker container] File {filename} is not in UTF-8!')
> ```

Currently I'm just printing the "bad" filename. I will change to loud fail after I analyze which files are non-utf-8.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: 19 of 21 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[Tools/gsc/finalize_manifest.py, line 49 at r6](https://reviewable.io/reviews/oscarlab/graphene/2112#-MScYjuzCMjoud5eB7Ap:-MScihdiCId4p10EWQ_v:b-42pu5h) ([raw file](https://github.com/oscarlab/graphene/blob/3a9613daf93b5fff53b8965219193965e19ef9c8/Tools/gsc/finalize_manifest.py#L49)):*

Currently I'm just printing the "bad" filename. I will change to loud fail after I analyze which files are non-utf-8.
</blockquote></details>

Done. Changed to a fail now.

dimakuv:
Jenkins, retest Jenkins-SGX-apps please (`./pal_loader redis-server --save '' --protected-mode no enclave EINIT failed - Invalid measurement`, some Makefile issue that Michal mentioned?)

mkow (dismissed):
> Reviewed 2 of 2 files at r7, 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @woju)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: 20 of 21 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSOuMBL7QN_8cwj2dW4:-MSg5-R-B-z_dKXlOHZk:brvd3zh) (no related file):*

After reading through internet, this is a commonly seen benign bug in Docker. It was fixed in the mainline Docker but seems like not backported in Ubuntu 18.04. See these:
- https://github.com/docker/docker-credential-helpers/issues/103
- https://github.com/docker/docker-credential-helpers/pull/29 -- this is the fix
- https://github.com/docker/docker-credential-helpers/issues/103#issuecomment-479508473 -- the favorite way of fixing this.

So looks like we can ignore this benign error on Docker side. Or fix it via `dpkg -r --force-depends golang-docker-credential-helpers` on that Jenkins worker. @woju what would be your advise?
</blockquote></details>

Should we do anything about it? @woju Ping again.

---

dimakuv:
Jenkins, retest Jenkins-Debug please (Apache web server seems to have taken more than 1 second to initialize due to a LOT of debug messages, and the client couldn't connect and panicked)

mkow (commented):
> Reviewed 1 of 1 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @woju)

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSOuMBL7QN_8cwj2dW4:-MShXle33Ie5onhmLToq:bxwwuek) (no related file):*

Should we do anything about it? @woju Ping again.
</blockquote></details>

If it's not critical and builds won't fail, I don't think we should do anything about this. Also I wonder why we didn't see this previously. Sorry for the delay.

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2112#-MSOuMBL7QN_8cwj2dW4:-MSivDPzCT_hohB087cZ:b2xir44) (no related file):*

If it's not critical and builds won't fail, I don't think we should do anything about this. Also I wonder why we didn't see this previously. Sorry for the delay.
</blockquote></details>

Ok, thanks for the input. I suggest to unblock this then.

We probably simply haven't noticed this before. This thing happens in the very beginning of the Jenkins job, and is buried in the several-MBs log.

---

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2112)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2147, opened 2021-02-08
[LibOS] Error out on anonymous shared mappings in mmap syscall

### LINKED ISSUE #2146: [LibOS,PAL] MAP_SHARED flag is not propagated to PAL
## Description of the problem

During `shim_do_mmap()`, there could be a combination `mmap(..., MAP_SHARED, ...)` which is not possible on Linux-SGX PAL. However, the Linux-SGX PAL layer ignores this flag (which is transformed into `!PAL_PROT_WRITECOPY` via `LINUX_PROT_TO_PAL()`).

It seems that the simple fix is to check that `prot` doesn't contain `PAL_PROT_WRITECOPY` inside Linux-SGX's `_DkVirtualMemoryAlloc()` and fail with some PAL error.

It looks like the same situation happens in Linux PAL: the mapping is created as `MAP_PRIVATE` (never `MAP_SHARED`) always.

mkow:
## Description of the changes 

LibOS can't implement such a combination of flags because PAL doesn't support shared mappings. I'm not sure about the non-anonymous shared case, as it currently works on Linux PAL, but only by an accident? (at least it seems...)

Fixes #2146.

## How to test this PR? 

Check if `mmap(NULL, 4096, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_SHARED, -1, 0)` fails. And wait for CI to see if this breaks any of our examples.

### DIFF
--- a/LibOS/shim/src/sys/shim_mmap.c
+++ b/LibOS/shim/src/sys/shim_mmap.c
@@ -77,6 +77,8 @@ void* shim_do_mmap(void* addr, size_t length, int prot, int flags, int fd, off_t
     if (flags & MAP_ANONYMOUS) {
         switch (flags & MAP_TYPE) {
             case MAP_SHARED:
+                /* PAL API doesn't support anonymous shared memory. */
+                return (void*)-ENOSYS;
             case MAP_PRIVATE:
                 break;
             default:


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2147)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/src/sys/shim_mmap.c, line 80 at r1](https://reviewable.io/reviews/oscarlab/graphene/2147#-MT2gtOk1_nutXwP3i6I:-MT2gtOl65V2MSlLxz8K:b-chd4v0) ([raw file](https://github.com/oscarlab/graphene/blob/5b67047d4f3bee5f5d55d5f50a4d000f86843d38/LibOS/shim/src/sys/shim_mmap.c#L80)):*
> ```C
>         switch (flags & MAP_TYPE) {
>             case MAP_SHARED:
>                 /* PAL API doesn't support anonymous shared memory. */
> ```

It does, see `PAL_PROT_WRITECOPY`.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2147)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/sys/shim_mmap.c, line 80 at r1](https://reviewable.io/reviews/oscarlab/graphene/2147#-MT2gtOk1_nutXwP3i6I:-MT2kh7k8tUcsctApKlS:bte1gxs) ([raw file](https://github.com/oscarlab/graphene/blob/5b67047d4f3bee5f5d55d5f50a4d000f86843d38/LibOS/shim/src/sys/shim_mmap.c#L80)):*

It does, see `PAL_PROT_WRITECOPY`.
</blockquote></details>

Except `PAL_PROT_WRITECOPY` is broken by design and probably unfixable...

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2147)* status: 2 of 5 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "DO NOT MERGE" found in commit messages' one-liners (waiting on @mkow)

___
*[.ci/lib/stage-clean-check.jenkinsfile, line 24 at r2](https://reviewable.io/reviews/oscarlab/graphene/2147#-MT5O2VV0jxKAEUj4rle:-MT5O2VW0XlecbvF2FHc:b-6ezuwq) ([raw file](https://github.com/oscarlab/graphene/blob/36f90a6c8968f54131d2e5a071a27ef65a5482e9/.ci/lib/stage-clean-check.jenkinsfile#L24)):*
> ```
>         make -C Examples/redis distclean
>         make -C Examples/lighttpd distclean
>         #make -C Examples/nginx distclean
> ```

OMG. Graphene had a silent `MAP_SHARED -> MAP_PRIVATE` error in Nginx and Apache all these years? And despite this, both Nginx and Apache worked? This is sad.

Upon further inspection in Nginx, I see `src/os/unix/ngx_shmem.c: ngx_shm_alloc()` which is used in several places in Nginx. I don't see a way to disable shared memory (we can only switch between 3 implementations: `NGX_HAVE_MAP_ANON`, `NGX_HAVE_MAP_DEVZERO` and `NGX_HAVE_SYSVSHM`).

We are doomed :( At least for Nginx, I don't see how we can simply error out on `mmap(MAP_ANON | MAP_SHARED)`. There is no config option in Nginx to disable shared memory.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2147)* status: 2 of 5 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "DO NOT MERGE" found in commit messages' one-liners (waiting on @dimakuv)

___
*[.ci/lib/stage-clean-check.jenkinsfile, line 24 at r2](https://reviewable.io/reviews/oscarlab/graphene/2147#-MT5O2VV0jxKAEUj4rle:-MT5qbXP3Ku6_FNkZycR:b-9337oo) ([raw file](https://github.com/oscarlab/graphene/blob/36f90a6c8968f54131d2e5a071a27ef65a5482e9/.ci/lib/stage-clean-check.jenkinsfile#L24)):*

OMG. Graphene had a silent `MAP_SHARED -> MAP_PRIVATE` error in Nginx and Apache all these years? And despite this, both Nginx and Apache worked? This is sad.

Upon further inspection in Nginx, I see `src/os/unix/ngx_shmem.c: ngx_shm_alloc()` which is used in several places in Nginx. I don't see a way to disable shared memory (we can only switch between 3 implementations: `NGX_HAVE_MAP_ANON`, `NGX_HAVE_MAP_DEVZERO` and `NGX_HAVE_SYSVSHM`).

We are doomed :( At least for Nginx, I don't see how we can simply error out on `mmap(MAP_ANON | MAP_SHARED)`. There is no config option in Nginx to disable shared memory.
</blockquote></details>

Uh, that's sad...
Do you know if the shared memory was actually used after forking? How could it work if we aren't really allocating it with `MAP_SHARED`?

If this mapping isn't really used then we could push down the error to the first shared usage by returning a normal mapping but marking it as "remove-on-fork" internally and then remove it after fork (and reserve this region in bookkeeping). This way fork+execve would still work, but fork+using the memory in the child would error out. Cons: requires SGX2 and EDMM.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2147)* status: 2 of 5 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "DO NOT MERGE" found in commit messages' one-liners (waiting on @mkow)

___
*[.ci/lib/stage-clean-check.jenkinsfile, line 24 at r2](https://reviewable.io/reviews/oscarlab/graphene/2147#-MT5O2VV0jxKAEUj4rle:-MT7GezmBwALVeRFxTzH:b-31bf1t) ([raw file](https://github.com/oscarlab/graphene/blob/36f90a6c8968f54131d2e5a071a27ef65a5482e9/.ci/lib/stage-clean-check.jenkinsfile#L24)):*

Uh, that's sad...
Do you know if the shared memory was actually used after forking? How could it work if we aren't really allocating it with `MAP_SHARED`?

If this mapping isn't really used then we could push down the error to the first shared usage by returning a normal mapping but marking it as "remove-on-fork" internally and then remove it after fork (and reserve this region in bookkeeping). This way fork+execve would still work, but fork+using the memory in the child would error out. Cons: requires SGX2 and EDMM.
</blockquote></details>

From what I understood, Nginx pre-allocates a shared memory region (of a predefined size) for future use by any Nginx modules at runtime. I.e., it's like a scratch space for Nginx processes. So yes, at least in our simple flows, it seems to be unused, so we could use your trick of "push the error down".

I didn't look into Apache code.

boryspoplawski:
Jenkins, retest this please (just making jenkins load for other tests, unrelated to this PR)

PR #2155, opened 2021-02-11
fix the deadlock of _DkSystemTimeQuery()

ying2liu:
_DkSystemTimeQuery() can be interrupted by async signal when it holds spin lock,
signal handler can recursively issue _DkSystemTimeQuery() with tsc_lock held,
thus it can result in deadlock(self recursive lock).
Solve such deadlock by introducing sequential number for data integrity for
fast path and use spinlock_trylock() instead spinlock_lock() for slow path to
update start_usec, start_tsc.

Change-Id: Ia922676890cfaaa1ca9a32b74130f74bf540f719
Signed-off-by: Isaku Yamahata <saku.yamahata@intel.com>
Signed-off-by: Ying Liu <ying2.liu@intel.com>

## Description of the changes 

_DkSystemTimeQuery() can be interrupted by async signal when it holds spin lock,
signal handler can recursively issue _DkSystemTimeQuery() with tsc_lock held,
thus it can result in deadlock(self recursive lock).
Solve such deadlock by introducing sequential number for data integrity for
fast path and use spinlock_trylock() instead spinlock_lock() for slow path to
update start_usec, start_tsc.

## How to test this PR? 
This commit also includes a unit test case.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2155)

### DIFF
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -73,6 +73,7 @@ c_executables = \
 	sigaction_per_process \
 	sigaltstack \
 	sighandler_reset \
+	sighandler_deadlock \
 	sighandler_sigpipe \
 	signal_multithread \
 	sigprocmask_pending \
--- a/LibOS/shim/test/regression/sighandler_deadlock.c
+++ b/LibOS/shim/test/regression/sighandler_deadlock.c
@@ -1,0 +1,72 @@
+#define _XOPEN_SOURCE 700
+#include <errno.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/time.h>
+#include <unistd.h>
+
+static int received = 0;
+#define SIGNAL     (SIGUSR1)
+#define NUM_SIGNAL (20000)
+static struct timeval *p_timeval = NULL;
+static void handler(int sig) {
+    if (sig == SIGNAL) {
+        received++;
+        int ret = gettimeofday(p_timeval, NULL);
+        if (ret != 0) {
+            fprintf(stderr, "handler received the signal, receive = %d ret %d, errno %d/%s \n", received, ret,
+                   errno, strerror(errno));
+            return;
+        }
+        printf("handler received the signal, receive = %d\n", received);
+    }
+}
+
+int main() {
+    int count = 0;
+
+    pid_t pid2 = fork();
+    if (pid2 < 0)
+        fprintf(stderr, "failed to create child process\n");
+    else if (pid2 == 0) {  // child
+        p_timeval = malloc(sizeof(*p_timeval));
+        struct sigaction new_action;
+        new_action.sa_handler = handler;
+        sigemptyset(&new_action.sa_mask);
+        new_action.sa_flags = 0;
+        int sigret = sigaction(SIGNAL, &new_action, NULL);
+        if (sigret < 0) {
+            fprintf(stderr, "sigaction failed, sigret = %d\n", sigret);
+            return 1;
+        }
+        struct timeval start;
+        int last_received = 0;
+        while (received < NUM_SIGNAL) {
+            if (received != last_received) {
+                printf("SIGNAL received = %d.\n", received);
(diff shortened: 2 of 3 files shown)


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2155)* status: 0 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @ying2liu)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2155#-MTEKPu93pQd9Iejj4wq:-MTEKPuC7HCOeZ0A53Ps:bfiy9vp) (no related file):*
> _DkSystemTimeQuery() can be interrupted by async signal when it holds spin lock,
> signal handler can recursively issue _DkSystemTimeQuery() with tsc_lock held,
> thus it can result in deadlock(self recursive lock).

Could you specify where exactly a signal handler can recursively issue `_DkSystemTimeQuery()`? Currently signal handlers should be reentrant...

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2155)* status: 0 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @ying2liu)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2155#-MTIz6eP--t9pY7cWczM:-MTIz6eP--t9pY7cWczN:b-79lzk2) (no related file):*
Please check #2156 it might solve the issues you are having.

---

ying2liu:
Yes, signal handler is reentrant, but it is an async signal and might be called in a thread just within _DkSystemTimeQuery and holding the lock.
This would cause recursive locking and dead lock if the handler calls gettimeofdata/_DkSystemTimeQuery. This commit tries to solve this problem because
lots of signal hander calls gettimeofday for bookkeeping.

boryspoplawski (commented):
For the future: pleas respond directly in reviewable or at least github, responding via emials completely breaks formatting.
> Yes, signal handler is reentrant, but it is an async signal and might be called in a thread just within _DkSystemTimeQuery and holding the lock. This would cause recursive locking and dead lock if the handler calls gettimeofdata/_DkSystemTimeQuery. This commit tries to solve this problem because lots of signal hander calls gettimeofday for bookkeeping.
 
This is not true. User-level signal handlers are only called when returning from a syscall / interrupt (host-level signal) - they cannot hold a Graphene-level lock. Host-level signal handlers should be reentrant, hence they do not use locks at all (this means that they do not call functions holding any locks, such as `DkSystemTimeQuery`).

Keep in mind that recently a bug was discovered (#2156) which could cause host-level singal to be nested, if the first one appeared while the code was in host-level vdso. That's why I asked you to verify it it fixes issues you are having, as this might have been the underlying issue to your deadlocks.

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2155#-:-MTME_nK28Rio_TI6THZ:b7o3fuv)* status: 0 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @ying2liu)

ying2liu:
I will run some tests using #2156.

Thanks

Ying

ying2liu:
I ran 80000 tests and didn’t see the deadlock. It seems that #2156 has fixed the deadlock of _DkSystemTimeQuery().

Thanks

Ying

From: Liu, Ying2
Sent: Friday, February 12, 2021 11:08 AM
To: oscarlab/graphene <reply@reply.github.com>; oscarlab/graphene <graphene@noreply.github.com>
Cc: Mention <mention@noreply.github.com>
Subject: RE: [oscarlab/graphene] fix the deadlock of _DkSystemTimeQuery() (#2155)

I will run some tests using #2156.

Thanks

Ying

PR #2156, opened 2021-02-12
[Pal/Linux] Change `ADDR_IN_PAL` to `ADDR_IN_PAL_OR_VDSO`

boryspoplawski:
## Description of the changes 
When an async exception triggers, LibOS needs to know whether it
happened while inside PAL code. Since Linux-PAL sometimes call VDSO
functions, they need to be accounted for as well.

Closes #2155.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)

### DIFF
--- a/Pal/src/Makefile
+++ b/Pal/src/Makefile
@@ -114,6 +114,14 @@ ifneq ($(pal_lib),)
 $(pal_lib): $(addprefix $(OBJ_DIR)/,$(objs)) \
 	    $(host_lib) $(graphene_lib) $(pal_lib_deps)
 	$(call cmd,ld_so_o)
+ifeq ($(PAL_HOST),Linux)
+	@if readelf -a $@ | grep -q NEEDED; then              \
+		echo Error: $@ should not be dynamically linked ; \
+		mv $@ $@.dynamic;                                 \
+		exit 1;                                           \
+	fi
+endif
+
 
 $(runtime_lib): $(pal_lib)
 	$(call cmd,ln_sfr)
--- a/Pal/src/host/Linux/db_exception.c
+++ b/Pal/src/host/Linux/db_exception.c
@@ -121,7 +121,7 @@ static void handle_sync_signal(int signum, siginfo_t* info, struct ucontext* uc)
     assert(event > 0);
 
     uintptr_t rip = ucontext_get_ip(uc);
-    if (!ADDR_IN_PAL(rip)) {
+    if (!ADDR_IN_PAL_OR_VDSO(rip)) {
         /* exception happened in application or LibOS code, normal benign case */
         perform_signal_handling(event, /*is_in_pal=*/false, (PAL_NUM)info->si_addr, uc);
         return;
@@ -141,11 +141,12 @@ static void handle_sync_signal(int signum, siginfo_t* info, struct ucontext* uc)
             break;
     }
 
-    printf("*** Unexpected %s occurred inside PAL (PID = %ld, TID = %ld, RIP = +0x%08lx)! ***\n",
-           name, INLINE_SYSCALL(getpid, 0), INLINE_SYSCALL(gettid, 0), rip - (uintptr_t)TEXT_START);
+    bool in_vdso = is_in_vdso(rip);
+    printf("*** Unexpected %s occurred inside %s (PID = %ld, TID = %ld, RIP = +0x%08lx)! ***\n",
+           name, in_vdso ? "VDSO" : "PAL", INLINE_SYSCALL(getpid, 0), INLINE_SYSCALL(gettid, 0),
+           rip - (in_vdso ? get_vdso_start() : (uintptr_t)TEXT_START));
 
     _DkProcessExit(1);
-    return;
 }
 
 static void handle_async_signal(int signum, siginfo_t* info, struct ucontext* uc) {
@@ -155,7 +156,7 @@ static void handle_async_signal(int signum, siginfo_t* info, struct ucontext* uc
     assert(event > 0);
 
     uintptr_t rip = ucontext_get_ip(uc);
-    perform_signal_handling(event, ADDR_IN_PAL(rip), /*addr=*/0, uc);
+    perform_signal_handling(event, ADDR_IN_PAL_OR_VDSO(rip), /*addr=*/0, uc);
 }
 
 /* TODO: remove this function. It's not an exception handling, it's just returning an error from
--- a/Pal/src/host/Linux/db_rtld.c
+++ b/Pal/src/host/Linux/db_rtld.c
@@ -10,13 +10,26 @@
  */
 
 #include "api.h"
+#include "cpu.h"
 #include "debug_map.h"
 #include "elf-arch.h"
 #include "elf/elf.h"
 #include "pal.h"
 #include "pal_debug.h"
+#include "pal_internal.h"
 #include "pal_linux.h"
 #include "pal_rtld.h"
+
+static uintptr_t vdso_start = 0;
+static uintptr_t vdso_end = 0;
+
+uintptr_t get_vdso_start(void) {
+    return vdso_start;
+}
+
+bool is_in_vdso(uintptr_t addr) {
+    return (vdso_start || vdso_end) && vdso_start <= addr && addr < vdso_end;
+}
 
 void _DkDebugMapAdd(const char* name, void* addr) {
     int ret = debug_map_add(name, addr);
@@ -58,16 +71,28 @@ void setup_vdso_map(ElfW(Addr) addr) {
 
     ElfW(Addr) load_offset = 0;
     const ElfW(Phdr) * ph;
+    unsigned long pt_loads_count = 0;
     for (ph = vdso_map.l_phdr; ph < &vdso_map.l_phdr[vdso_map.l_phnum]; ph++)
         switch (ph->p_type) {
             case PT_LOAD:
                 load_offset = addr + (ElfW(Addr))ph->p_offset - (ElfW(Addr))ph->p_vaddr;
+                vdso_start = (uintptr_t)addr;
+                vdso_end = ALIGN_UP(vdso_start + (size_t)ph->p_memsz, PAGE_SIZE);
+                pt_loads_count++;
                 break;
             case PT_DYNAMIC:
                 vdso_map.l_real_ld = vdso_map.l_ld = (void*)addr + ph->p_offset;
                 vdso_map.l_ldnum = ph->p_memsz / sizeof(ElfW(Dyn));
                 break;
         }
+
+    if (pt_loads_count != 1) {
+        log_warning("The VDSO has %lu PT_LOAD segments, but only 1 was expected.\n",
+                    pt_loads_count);
+        vdso_start = 0;
+        vdso_end = 0;
+        return;
+    }
 
     ElfW(Dyn) local_dyn[4];
     int ndyn = 0;
--- a/Pal/src/host/Linux/pal_linux.h
+++ b/Pal/src/host/Linux/pal_linux.h
@@ -8,6 +8,8 @@
 #include <asm/stat.h>
 #include <linux/mman.h>
 #include <sigset.h>
+#include <stdbool.h>
+#include <stdint.h>
 #include <sys/syscall.h>
 #include <sys/types.h>
 #include <unistd.h>
@@ -103,6 +105,9 @@ int clone(int (*__fn)(void* __arg), void* __child_stack, int __flags, const void
 /* PAL main function */
 noreturn void pal_linux_main(void* initial_rsp, void* fini_callback);
 
+uintptr_t get_vdso_start(void);
+bool is_in_vdso(uintptr_t addr);
+
 struct link_map;
 void setup_pal_map(struct link_map* map);
 void setup_vdso_map(ElfW(Addr) addr);
@@ -133,8 +138,8 @@ extern char __text_start, __text_end, __data_start, __data_end;
 #define DATA_START ((void*)(&__text_start))
 #define DATA_END   ((void*)(&__text_end))
 
-#define ADDR_IN_PAL(addr) \
-        ((void*)(addr) > TEXT_START && (void*)(addr) < TEXT_END)
+#define ADDR_IN_PAL_OR_VDSO(addr) \
+        (((void*)(addr) > TEXT_START && (void*)(addr) < TEXT_END) || is_in_vdso(addr))
 
 typedef struct pal_tcb_linux {
     PAL_TCB common;


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: 0 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFMy2Es5gkGA-IBIr:-MTKFMy2Es5gkGA-IBIs:b5knxqz) (no related file):*
Linux-SGX could do the same, right? It doesn't do it now (afaik), but it can surely do it in the future.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTKFWAt-J-ljoI5gh6C:b-mu36hg) (no related file):*
Why not:
1. Replace the naive `ADDR_IN_PAL()` macro with a proper function `is_in_pal()`
2. This function is initialized with results of parsing `/proc/self/maps` to find all maps with permission `x`
3. For each PAL, a common PAL initialization routine should parse `/proc/self/maps` right-before jumping into LibOS (so we don't have any LibOS-related VMAs yet) and mark all `x`-permission VMAs as "in PAL".

---

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: 0 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFMy2Es5gkGA-IBIr:-MTLUF5c4SD7NRcERssf:b2szx65) (no related file):*

Linux-SGX could do the same, right? It doesn't do it now (afaik), but it can surely do it in the future.
</blockquote></details>

Linux-SGX needs to do a ocall anyway and we are only interested in trusted part (which doesn't and never will use VDSO, at least direclty).

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTLUef8A7xEuIygO_AD:b-2wbvan) (no related file):*

Why not:
1. Replace the naive `ADDR_IN_PAL()` macro with a proper function `is_in_pal()`
2. This function is initialized with results of parsing `/proc/self/maps` to find all maps with permission `x`
3. For each PAL, a common PAL initialization routine should parse `/proc/self/maps` right-before jumping into LibOS (so we don't have any LibOS-related VMAs yet) and mark all `x`-permission VMAs as "in PAL".
</blockquote></details>

1) Possible.
2,3) This would make no sense for some PALs (like Linux-SGX, we care about the trusted part only).
We could do that only for specific PALs (here: Linux). Should I?
cc @mkow 

---

boryspoplawski:
Jenkins, retest this please (jenkins internal failure)

boryspoplawski:
Jenkins, retest this please (testing new nodes)

boryspoplawski:
Jenkins, retest this please (testing new nodes)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: 0 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTT5fum9YV0nDaZOJlz:bnou47q) (no related file):*

1) Possible.
2,3) This would make no sense for some PALs (like Linux-SGX, we care about the trusted part only).
We could do that only for specific PALs (here: Linux). Should I?
cc @mkow 
</blockquote></details>

I don't like 2 and 3, they look complicated and brittle (e.g. what about libraries which are loaded lazily? you'll silently miss them and reintroduce this hard-to-reproduce and hard-to-pinpoint bug).

What about the idea we discussed on priv, about having the LibOS address range and just checking if something is outside it? Would be trivial and correct. The only problem with it, as Borys mentioned, is that we might not want to have a separate address range for LibOS in the future, but I think this requires a major redesign and we can think about this problem when doing this redesign.

---

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: 0 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTX67TQAG0Iwie84Lop:bk0pgth) (no related file):*
> but I think this requires a major redesign and we can think about this problem when doing this redesign.
As far as I remember, LibOS part (vma bookkeeping) should be ready for such change, so doesn't sound like a major redesign.

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: 0 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTY70KbA9sM0Rpn_G8n:b-2it4vd) (no related file):*

> but I think this requires a major redesign and we can think about this problem when doing this redesign.
As far as I remember, LibOS part (vma bookkeeping) should be ready for such change, so doesn't sound like a major redesign.
</blockquote></details>

But how do you imagine solving the problem we have here without the address space separation? For me your current approach looks like something which will blow up eventually and again lead to hours of debugging...

---

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: 0 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTYIz258EejzHV4vyGM:b3tjj96) (no related file):*

But how do you imagine solving the problem we have here without the address space separation? For me your current approach looks like something which will blow up eventually and again lead to hours of debugging...
</blockquote></details>

As long as we build PAL statically nothing should blow up. We could also add a comment in makefile, where we set `-static` for PAL, about this function.

---

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTYQVu-2y2RtC9uBjIs:b-sdbnhk) (no related file):*

As long as we build PAL statically nothing should blow up. We could also add a comment in makefile, where we set `-static` for PAL, about this function.
</blockquote></details>

Ok, please add such a comment then.

---
___
*[Pal/src/host/Linux/db_rtld.c, line 76 at r1](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTYREvL9aVqMjcGk1Lm:-MTYREvL9aVqMjcGk1Ln:b15mujl) ([raw file](https://github.com/oscarlab/graphene/blob/55f517f05fe1a11e8ab326ec020dc4030fef47b7/Pal/src/host/Linux/db_rtld.c#L76)):*
> ```C
>         switch (ph->p_type) {
>             case PT_LOAD:
>                 /* Assumes that VDSO has only one PT_LOAD segment. */
> ```

we should actually assert this... but I'm not blocking, this issues was already here before

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFMy2Es5gkGA-IBIr:-MTZhKuz8BvjIxF5DqKk:b-dkjlih) (no related file):*

Linux-SGX needs to do a ocall anyway and we are only interested in trusted part (which doesn't and never will use VDSO, at least direclty).
</blockquote></details>

Ok, this is true.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTZhSii-Ecdca9DxjBi:bci3vfe) (no related file):*

Ok, please add such a comment then.
</blockquote></details>

LGTM, my bad. I didn't realize this issue doesn't exist in Linux-SGX PAL.

---

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: 2 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTg0A0a9k2f2cDtR-DN:b-b8r3bw) (no related file):*

LGTM, my bad. I didn't realize this issue doesn't exist in Linux-SGX PAL.
</blockquote></details>

I couldn't find a place to put it. I wanted to add an assert on `readelf -a libpal.so | fgrep -q NEEDED` being empty, but that would be PAL-Linux specific check, but `libpal.so` is a generic makefile target...

---
___
*[Pal/src/host/Linux/db_rtld.c, line 76 at r1](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTYREvL9aVqMjcGk1Lm:-MTg3_Vd3EdPZ3l9St9s:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/55f517f05fe1a11e8ab326ec020dc4030fef47b7/Pal/src/host/Linux/db_rtld.c#L76)):*

we should actually assert this... but I'm not blocking, this issues was already here before
</blockquote></details>

Done.

boryspoplawski:
Jenkins, retest Jenkins-SGX-apps please (apache build failed with `util_expr_eval.c:(.text+0x112b): undefined reference to `ap_expr_yylex_init'`, broken disc?)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTgtdkoB816HdAYb9kR:bowmvua) (no related file):*

I couldn't find a place to put it. I wanted to add an assert on `readelf -a libpal.so | fgrep -q NEEDED` being empty, but that would be PAL-Linux specific check, but `libpal.so` is a generic makefile target...
</blockquote></details>

If this is a shared file between all PALs and one of them has some requirements, then I'm fine with asserting them for that common file. Why not?

---
___
*[Pal/src/host/Linux/db_rtld.c, line 89 at r2](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTgtwnw1EBKKSxl3gS4:-MTgtwnx0wEg4oPMZ1eZ:b8orhwi) ([raw file](https://github.com/oscarlab/graphene/blob/9d6b9feef79a63904ac6cc27569e24af1feebabf/Pal/src/host/Linux/db_rtld.c#L89)):*
> ```C
> 
>     if (pt_loads_count != 1) {
>         /* This code assumes that VDSO has exactly one PT_LOAD segment. */
> ```

Shouldn't we print an error and fail here? Silently continuing will:
a) make the bug you fixed here reappear
b) we may never learn that some systems has such a problematic vdso

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTgv16r1OcVf2FSO3Wj:bhij0b9) (no related file):*

If this is a shared file between all PALs and one of them has some requirements, then I'm fine with asserting them for that common file. Why not?
</blockquote></details>

This file is not shared by all PALs, it just has the very same build recipe.

---
___
*[Pal/src/host/Linux/db_rtld.c, line 89 at r2](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTgtwnw1EBKKSxl3gS4:-MTgv9DL9uubRf8miJU_:bvrub8y) ([raw file](https://github.com/oscarlab/graphene/blob/9d6b9feef79a63904ac6cc27569e24af1feebabf/Pal/src/host/Linux/db_rtld.c#L89)):*

Shouldn't we print an error and fail here? Silently continuing will:
a) make the bug you fixed here reappear
b) we may never learn that some systems has such a problematic vdso
</blockquote></details>

1) this function does not return errors and callers do not handle them.
2) no it will not. If this code doesn't set vdso address, Graphene will not use it (we skip setting on errors).

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTgwPsy7fAg-1k8Z0Oq:b-m9ylcm) (no related file):*

This file is not shared by all PALs, it just has the very same build recipe.
</blockquote></details>

Ah, then maybe just an `if` on `$(SGX)`?

---
___
*[Pal/src/host/Linux/db_rtld.c, line 89 at r2](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTgtwnw1EBKKSxl3gS4:-MTgwZGB1wRiP-vMRYoo:b-8wcn4j) ([raw file](https://github.com/oscarlab/graphene/blob/9d6b9feef79a63904ac6cc27569e24af1feebabf/Pal/src/host/Linux/db_rtld.c#L89)):*

1) this function does not return errors and callers do not handle them.
2) no it will not. If this code doesn't set vdso address, Graphene will not use it (we skip setting on errors).
</blockquote></details>

Ah, so we aren't using vsdo at all in this case? Then it's fine, but maybe just log a warning then?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTKFWAt-J-ljoI5gh6B:-MTkcu-Y6dHNEVwIYFYM:b-2j2qhz) (no related file):*

Ah, then maybe just an `if` on `$(SGX)`?
</blockquote></details>

Rather on `Pal-Linux` as it's that-Pal-specific.
Done.

---
___
*[Pal/src/host/Linux/db_rtld.c, line 89 at r2](https://reviewable.io/reviews/oscarlab/graphene/2156#-MTgtwnw1EBKKSxl3gS4:-MTkbXqDDfyE65u93WG9:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/9d6b9feef79a63904ac6cc27569e24af1feebabf/Pal/src/host/Linux/db_rtld.c#L89)):*

Ah, so we aren't using vsdo at all in this case? Then it's fine, but maybe just log a warning then?
</blockquote></details>

Done.

mkow (approved):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (approved):
> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2156)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2163, opened 2021-02-16
Fix race condition in example's Makefiles

mkow:
## Description of the changes 

I hate Make. We should really get rid of it finally...

It turns out that Make in versions that doesn't support the `&:` operator ("Rules with Grouped Targets") silently ignores it and calls the recipe for each target separately, without even a warning. When using parallel builds (`-j`) the result is a race, because a few command invocations write to the same set of target files at the same time.

Additionally, I ported the Redis example to Ubuntu 20.04 to test the fix.

## How to test this PR? 

CI + build the examples manually with `-j8` and check the build log.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2163)

### DIFF
--- a/Examples/blender/Makefile
+++ b/Examples/blender/Makefile
@@ -59,7 +59,12 @@ blender.manifest: blender.manifest.template trusted-libs | $(RUN_DIR)
 	     $<; \
     cat trusted-libs) > $@
 
-blender.sig blender.manifest.sgx &: $(BLENDER_DIR)/blender blender.manifest \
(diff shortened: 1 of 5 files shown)


dimakuv (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2163)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

mkow:
Jenkins, retest Jenkins-Debug please (`apps.LTP.sendfile08` failed, known issue)

PR #2168, opened 2021-02-17
[Pal/Linux-SGX] Fix stack unwinding on enclave exit

### LINKED ISSUE #2160: GDB integration breaks on SGX when stack is split twice
CC @boryspoplawski 

## Description of the problem

GDB backtrace breaks when we're inside LibOS _and_ inside enclave ocall.

## Steps to reproduce

```sh
# Build Graphene and tests
cd graphene
SGX=1 DEBUG=1 make 
cd LibOS/shim/test/regression
SGX=1 DEBUG=1 make 
SGX=1 DEBUG=1 make sgx-tokens

# Run program under GDB
SGX=1 GDB=1 ./pal_loader ./debug
...
# Set breakpoint outside of enclave and run
(gdb) break sgx_ocall_write
Breakpoint 2 at 0x2844: file sgx_enclave.c, line 134.
(gdb) run
...
Thread 1 "pal-Linux-SGX" hit Breakpoint 2, sgx_ocall_write (pms=0x7ffe16c7fce0) at sgx_enclave.c:134
134	static long sgx_ocall_write(void* pms) {
```

## Expected results

After `backtrace` command, we should see an unbroken backtrace up to `_start`.

## Actual results

GDB stops backtrace at LibOS entry point:

```sh
(gdb) backtrace
#0  sgx_ocall_write (pms=0x7ffe16c7fce0) at sgx_enclave.c:134
#1  0x000055beb30d6f13 in sgx_raise () at sgx_entry.S:144
#2  0x000000000b6495b6 in __morestack () at enclave_entry.S:652
#3  0x000000000b63beb9 in sgx_exitless_ocall (code=7, ms=0x7ffe16c7fd12) at enclave_ocalls.c:55
#4  0x000000000b63cdcc in ocall_write (fd=1, buf=0x589c2a0, count=6) at enclave_ocalls.c:499
#5  0x000000000b6270ae in dev_write (handle=0xb6909f0 <g_mem_pool+29040>, offset=0, size=6, buffer=0x589c2a0) at db_devices.c:110
#6  0x000000000b624a6d in _DkStreamWrite (handle=0xb6909f0 <g_mem_pool+29040>, offset=0, count=6, buf=0x589c2a0, addr=0x0, addrlen=0)
    at db_streams.c:308
#7  0x000000000b624b08 in DkStreamWrite (handle=0xb6909f0 <g_mem_pool+29040>, offset=0, count=6, buffer=0x589c2a0, dest=0x0)
    at db_streams.c:325
#8  0x000000000b5a1fc9 in chroot_write (hdl=0xb5291c8, buf=0x589c2a0, count=6) at fs/chroot/fs.c:662
#9  0x000000000b5ce95e in do_handle_write (hdl=0xb5291c8, buf=0x589c2a0, count=6) at sys/shim_open.c:73
#10 0x000000000b5ce9e7 in shim_do_write (fd=1, buf=0x589c2a0, count=6) at sys/shim_open.c:84
#11 0x000000000b57ff03 in shim_emulate_syscall (context=0xb521f38) at shim_syscalls.c:37
#12 0x000000000b580099 in syscalldb () at syscallas-x86_64.S:173
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
```

## Additional information

Verified on gdb 8.2 (Ubuntu 18.04) and 9.2 (Ubuntu 20.04).

This is actually tested, but SGX=1 DEBUG=1 configuration is not present on Jenkins.

The problem seems to be that there are two `__morestack` symbols in the backtrace. It works as long as there is only one `__morestack` symbol.

[Relevant file in GDB](https://github.com/bminor/binutils-gdb/blob/master/gdb/frame.c). I guess GDB tries to disambiguate the symbols somehow?...

pwmarcz:
## Description of the changes 

A previous change in LibOS started storing fake syscall return address in R14. This broke stack unwinding when the stack contained both LibOS syscall and SGX OCALL, because we had no CFI directives for recovering register values except for RBP, RSP and RIP, and the value of R14 was lost.
    
This change adds proper CFI to enclave exit code, by making sure callee-saved registers can be recovered.

Fixes #2160.

## How to test this PR? 

This fixes the debug regression test under SGX (we didn't notice it break because the configuration is not on Jenkins yet). 

You can run it using `SGX=1 .../graphene/Scripts/run-pytest -k TC_50_GDB` in `LibOS/shim/test/regression`. Make sure to compile with `DEBUG=1`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_entry.S
+++ b/Pal/src/host/Linux-SGX/enclave_entry.S
@@ -510,6 +510,7 @@ sgx_ocall:
 	#   xregs
 	# (padding)
 	# --- stack may be non-contiguous as we may switch the stack to signal stack
+	# previous RBX
 	# previous RBP
 	# previous RIP: pushed by callq
 
@@ -519,6 +520,8 @@ sgx_ocall:
 	movq %rsp, %rbp
 	.cfi_offset %rbp, -16
 	.cfi_def_cfa_register %rbp
+	pushq %rbx
+	.cfi_offset %rbx, -24
 
 	CHECK_IF_SIGNAL_STACK_IS_USED %rsp, .Lon_signal_stack_ocall, .Lout_of_signal_stack_ocall
 
@@ -593,7 +596,6 @@ sgx_ocall:
 
 #ifdef DEBUG
 	# Push %rip of some code inside __morestack() on untrusted stack.
-	# At sgx_entry(), GDB deduces saved_rip by looking at CFA-8 = %rsp.
 	leaq .Lfor_cfa_debug_info(%rip), %r8
 	pushq %r8
 #endif
@@ -611,27 +613,7 @@ sgx_ocall:
 	#     %rdi, %rsi: (optional) arguments to untrusted code.
 .Lclear_and_eexit:
 
-#ifdef DEBUG
-	# Enclave and untrusted stacks are split (segmented). GDB refuses to
-	# unwind such stacks because it looks like stack frames "jump" back
-	# and forth. Luckily, GDB special-cases stack frames for a function
-	# with hardcoded name "__morestack". Declare this dummy function
-	# to make GDB happy.
-
-	.global __morestack
-	.type __morestack, @function
-__morestack:
-
 	.cfi_startproc
-
-	# Parse trusted stack frame from (saved) RBP.
-	.cfi_def_cfa %rbp, 0
-	.cfi_offset %rip, 8
-	.cfi_offset %rbp, 0
-#else
-	.cfi_startproc
-#endif
-
 	# Clear "extended" state (FPU aka x87, SSE, AVX, ...).
 
 	# g_pal_sec.enclave_attributes.xfrm will always be zero before
@@ -653,22 +635,21 @@ __morestack:
 	# %rsi, %rdi are arguments to the untrusted code
 
 #ifdef DEBUG
-.Lfor_cfa_debug_info:
-	# Leave %rbp pointing to OCALL function on trusted stack.
+	# Keep callee-saved registers in order to recover stack later (see __morestack() below).
 #else
-	# In non-debug mode, clear %rbp to not leak trusted stack address.
+	# In non-debug mode, clear these registers to prevent information leaks.
 	xorq %rbp, %rbp
+	xorq %r12, %r12
+	xorq %r13, %r13
+	xorq %r14, %r14
+	xorq %r15, %r15
 #endif
 
 	# %rsp points to untrusted stack
 	xorq %r8, %r8
 	xorq %r9, %r9
 	xorq %r10, %r10
-	xorq %r11, %r11
-	xorq %r12, %r12
-	xorq %r13, %r13
-	xorq %r14, %r14
-	subq %r15, %r15 # use sub to set flags to a fixed value
+	subq %r11, %r11  # use sub to set flags to a fixed value
 
 	movq $EEXIT, %rax
 	ENCLU
@@ -873,3 +854,39 @@ restore_xregs:
 	popq %r11
 	jmp __restore_xregs
 	.cfi_endproc
+
+#ifdef DEBUG
+	# CFI "trampoline" to make GDB happy. GDB normally does not handle switching stack in the
+	# middle of backtrace (which is what happens when we exit the enclave), unless the function
+	# doing it is called __morestack.
+	#
+	# To make GDB backtrace work, we make sure that the first function outside of enclave
+	# (sgx_entry) has a return address on stack, pointing inside __morestack. We will not actually
+	# return to this function (sgx_entry performs EENTER to go back to enclave), but GDB will make a
+	# stack frame for it.
+	#
+	# The function contains CFI directives to make sure that all callee-saved registers can be
+	# recovered. They should reflect the situation during EEXIT in code above.
+
+	.global __morestack
+	.type __morestack, @function
+__morestack:
+	.cfi_startproc
+
+	# Callee-saved registers:
+
+	# RIP, RSP: deduced from current RBP (which was not cleared in debug mode)
+	.cfi_def_cfa %rbp, 16
+
+	# RBP, RBX: saved on stack (at the beginning of sgx_ocall)
+	.cfi_offset %rbp, -16
+	.cfi_offset %rbx, -24
+
+	# R12, R13, R14, R15: not changed (not cleared in debug mode)
+
+	nop
+.Lfor_cfa_debug_info:
+	nop
+
+	.cfi_endproc
+#endif


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Pal/src/host/Linux-SGX/enclave_entry.S, line 858 at r1](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTlh-IH8Ck6aOKUERaJ:-MTlh-IH8Ck6aOKUERaK:b-wy7urs) ([raw file](https://github.com/oscarlab/graphene/blob/4763cc3128cf202356848316486a3f04557c3230/Pal/src/host/Linux-SGX/enclave_entry.S#L858)):*
> ```
> 	# Inserting a fake `__morestack` function is unfortunately the only way to keep GDB happy when
> 	# switching stacks. Here, an additional complication is that the registers are saved on an
> 	# signal stack, which might be different than application stack. As a result we have to insert
> ```

`an signal stack` -> `a signal stack`

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTlob_t3_SWvlHuf2Rp:-MTlob_u0ASuC_YwJtlR:b-inq8zv) (no related file):*
LibOS does not use `r8`, I do not understand the commit message...
Why do we need two "fake frames"? I'm quite lost in all of this tbh.
Also: not clearing callee-saved registers (ofc in debug builds only) seems to be enough - in LibOS we intentionally use r14, which is callee saved, to keep fake return address.
Check out https://github.com/boryspoplawski/graphene/commits/borys/backtrace_test - I was able to get a clear backtrace spanning untrusted PAL <-> trusted PAL <-> LibOS <-> user program
Does it miss some edge case?

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTlob_t3_SWvlHuf2Rp:-MTmnstb0Nsj6py-AK_R:b8a623m) (no related file):*
>LibOS does not use r8, I do not understand the commit message...

My mistake. LibOS uses R14, and that's what we need to preserve.

>Why do we need two "fake frames"? I'm quite lost in all of this tbh.

Because there is signal stack involved: `sgx_ocall` switches to signal stack, saves the registers there, and only then exits enclave. If we want CFI to take the saved registers into account, we need a stack frame that is on the signal stack. Unfortunately that means we jump stacks not once, but twice, so we need two `__morestack` frames.

To be honest I have no idea why `sgx_ocall` switches to the signal stack. If we could get rid of that, it would simplify things. But your idea sounds even better.

>Also: not clearing callee-saved registers (ofc in debug builds only) seems to be enough - in LibOS we intentionally use r14, which is callee saved, to keep fake return address.

That's good news! And much simpler than trying to reach the saved registers.

I agree that callee-saved registers should be enough, we are being called like a normal function.

>Does it miss some edge case?

Looking at that code, it looks like we can account for all callee-saved registers except for RBX. It can be saved next to RBP, like so:

```diff
 sgx_ocall:
     pushq %rbp
+    pushq %rbx
     movq %rsp, %rbp
```

Does that sound good? Technically we don't need RBX for the stack trace, but it makes the CFI "correct", and I guess it could be useful when examining upper frames in GDB. But I'm not sure if we should bother.

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: 0 of 1 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTlob_t3_SWvlHuf2Rp:-MTp4sXO6-JeG580iJD2:bdhtorm) (no related file):*

>LibOS does not use r8, I do not understand the commit message...

My mistake. LibOS uses R14, and that's what we need to preserve.

>Why do we need two "fake frames"? I'm quite lost in all of this tbh.

Because there is signal stack involved: `sgx_ocall` switches to signal stack, saves the registers there, and only then exits enclave. If we want CFI to take the saved registers into account, we need a stack frame that is on the signal stack. Unfortunately that means we jump stacks not once, but twice, so we need two `__morestack` frames.

To be honest I have no idea why `sgx_ocall` switches to the signal stack. If we could get rid of that, it would simplify things. But your idea sounds even better.

>Also: not clearing callee-saved registers (ofc in debug builds only) seems to be enough - in LibOS we intentionally use r14, which is callee saved, to keep fake return address.

That's good news! And much simpler than trying to reach the saved registers.

I agree that callee-saved registers should be enough, we are being called like a normal function.

>Does it miss some edge case?

Looking at that code, it looks like we can account for all callee-saved registers except for RBX. It can be saved next to RBP, like so:

```diff
 sgx_ocall:
     pushq %rbp
+    pushq %rbx
     movq %rsp, %rbp
```

Does that sound good? Technically we don't need RBX for the stack trace, but it makes the CFI "correct", and I guess it could be useful when examining upper frames in GDB. But I'm not sure if we should bother.
</blockquote></details>

Updated. The new version is much simpler - apart from not clearing callee-saved registers, I mostly moved `__morestack` out of the way to make the code less confusing, and did my best to comment it. I also added RBX, as described above.

---
___
*[Pal/src/host/Linux-SGX/enclave_entry.S, line 858 at r1](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTlh-IH8Ck6aOKUERaJ:-MTp5Fvy4KXTnw0cfuAp:b-3ex806) ([raw file](https://github.com/oscarlab/graphene/blob/4763cc3128cf202356848316486a3f04557c3230/Pal/src/host/Linux-SGX/enclave_entry.S#L858)):*

`an signal stack` -> `a signal stack`
</blockquote></details>

Fixed - the new version doesn't use or mention the signal stack at all.

dimakuv (changes requested):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTlob_t3_SWvlHuf2Rp:-MTpb8Pp4WRNM5YrldTB:b-agezz7) (no related file):*

Updated. The new version is much simpler - apart from not clearing callee-saved registers, I mostly moved `__morestack` out of the way to make the code less confusing, and did my best to comment it. I also added RBX, as described above.
</blockquote></details>

Thanks for this PR! I would like to keep `rbx` visible in the stack trace, so I am for pushing it to the stack as you implemented.

Thankfully I looked at this PR after the first version was thrown away :)

---
___
*[Pal/src/host/Linux-SGX/enclave_entry.S, line 638 at r2](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTpbcXQ21OopX1X40Ap:-MTpbcXQ21OopX1X40Aq:b-g9henj) ([raw file](https://github.com/oscarlab/graphene/blob/f451cf56ef99a8ae4cb2e5c74a1ce7dc252d6b8c/Pal/src/host/Linux-SGX/enclave_entry.S#L638)):*
> ```
> 
> #ifdef DEBUG
> 	# Keep callee-saved registers in order to recover stack later (see __morestack() below).
> ```

Do we want to add a warning somewhere in the documentation that `DEBUG=1` Graphene leaks registers on OCALLs? @mkow @boryspoplawski @yamahata . I can imagine people building Graphene in DEBUG mode and using this in production...

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski, @mkow, and @pwmarcz)

___
*[Pal/src/host/Linux-SGX/enclave_entry.S, line 638 at r2](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTpbcXQ21OopX1X40Ap:-MTpbv2J3Cc-UwfQtny2:b-q5rsqo) ([raw file](https://github.com/oscarlab/graphene/blob/f451cf56ef99a8ae4cb2e5c74a1ce7dc252d6b8c/Pal/src/host/Linux-SGX/enclave_entry.S#L638)):*

Do we want to add a warning somewhere in the documentation that `DEBUG=1` Graphene leaks registers on OCALLs? @mkow @boryspoplawski @yamahata . I can imagine people building Graphene in DEBUG mode and using this in production...
</blockquote></details>

Not in the docs, but maybe a big fat warning when Graphene-SGX starts.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @boryspoplawski, @mkow, and @yamahata)

___
*[Pal/src/host/Linux-SGX/enclave_entry.S, line 638 at r2](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTpbcXQ21OopX1X40Ap:-MTpekuG6DYOy5G4npL3:b-dfylhq) ([raw file](https://github.com/oscarlab/graphene/blob/f451cf56ef99a8ae4cb2e5c74a1ce7dc252d6b8c/Pal/src/host/Linux-SGX/enclave_entry.S#L638)):*

Not in the docs, but maybe a big fat warning when Graphene-SGX starts.
</blockquote></details>

Actually, what do you think about saving R12, R13, R14, R15 on stack in the same way  as RBX, and clearing them here? This way, we would directly expose only RBP (same as before), and the other registers would be accessible only if it is a debug enclave and we can access enclave memory.

A warning under `#ifdef DEBUG` might still be a good idea.

boryspoplawski (changes requested):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @mkow, @pwmarcz, and @yamahata)

___
*[Pal/src/host/Linux-SGX/enclave_entry.S, line 638 at r2](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTpbcXQ21OopX1X40Ap:-MTpgPuF5otPfExm6m7N:b-e3vge6) ([raw file](https://github.com/oscarlab/graphene/blob/f451cf56ef99a8ae4cb2e5c74a1ce7dc252d6b8c/Pal/src/host/Linux-SGX/enclave_entry.S#L638)):*

Actually, what do you think about saving R12, R13, R14, R15 on stack in the same way  as RBX, and clearing them here? This way, we would directly expose only RBP (same as before), and the other registers would be accessible only if it is a debug enclave and we can access enclave memory.

A warning under `#ifdef DEBUG` might still be a good idea.
</blockquote></details>

1) I see no difference in exposing all of them vs just rbp, so imo this could can stay as it is
2) I remember talking about it with @mkow some time ago and afair we've reached the conclusion not to bother about it atm, but add such a warning after Graphene reaches ready-for-productio state.
___
*[Pal/src/host/Linux-SGX/enclave_entry.S, line 885 at r2](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTphh8kEjbZV8_1TmdH:-MTphh8kEjbZV8_1TmdI:bn4qcz1) ([raw file](https://github.com/oscarlab/graphene/blob/f451cf56ef99a8ae4cb2e5c74a1ce7dc252d6b8c/Pal/src/host/Linux-SGX/enclave_entry.S#L885)):*
> ```
> 	.cfi_offset %rbx, -24
> 
> 	# RDI, RSI: not changed (function parameters to sgx_ocall)
> ```

These two are caller-saved and as such cannot be used meaningfully in cfi (we can drop this comment)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: 0 of 1 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTlob_t3_SWvlHuf2Rp:-MTpnyyU7Rbj6mMf8lEX:b-mg7tlc) (no related file):*

Thanks for this PR! I would like to keep `rbx` visible in the stack trace, so I am for pushing it to the stack as you implemented.

Thankfully I looked at this PR after the first version was thrown away :)
</blockquote></details>

Yeah, I'm happy about this too :) I learned more about CFI, though.

---
___
*[Pal/src/host/Linux-SGX/enclave_entry.S, line 638 at r2](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTpbcXQ21OopX1X40Ap:-MTpnJMEBAQacrY4SMtz:b-g4zdah) ([raw file](https://github.com/oscarlab/graphene/blob/f451cf56ef99a8ae4cb2e5c74a1ce7dc252d6b8c/Pal/src/host/Linux-SGX/enclave_entry.S#L638)):*

1) I see no difference in exposing all of them vs just rbp, so imo this could can stay as it is
2) I remember talking about it with @mkow some time ago and afair we've reached the conclusion not to bother about it atm, but add such a warning after Graphene reaches ready-for-productio state.
</blockquote></details>

Okay, I'll keep the registers as they are right now.
___
*[Pal/src/host/Linux-SGX/enclave_entry.S, line 885 at r2](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTphh8kEjbZV8_1TmdH:-MTpnfQi3E5-7ed6_uKQ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f451cf56ef99a8ae4cb2e5c74a1ce7dc252d6b8c/Pal/src/host/Linux-SGX/enclave_entry.S#L885)):*

These two are caller-saved and as such cannot be used meaningfully in cfi (we can drop this comment)
</blockquote></details>

Done.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @yamahata)

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/enclave_entry.S, line 638 at r2](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTpbcXQ21OopX1X40Ap:-MTtKThM0rOGHuq6eH7S:bm4tas5) ([raw file](https://github.com/oscarlab/graphene/blob/f451cf56ef99a8ae4cb2e5c74a1ce7dc252d6b8c/Pal/src/host/Linux-SGX/enclave_entry.S#L638)):*

Okay, I'll keep the registers as they are right now.
</blockquote></details>

OK, fair.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux-SGX/enclave_entry.S, line 638 at r2](https://reviewable.io/reviews/oscarlab/graphene/2168#-MTpbcXQ21OopX1X40Ap:-MTvRWmn0T8FTDDh_HVa:b-c7sad6) ([raw file](https://github.com/oscarlab/graphene/blob/f451cf56ef99a8ae4cb2e5c74a1ce7dc252d6b8c/Pal/src/host/Linux-SGX/enclave_entry.S#L638)):*

OK, fair.
</blockquote></details>

Yup, I think a warning about debug builds is a good idea, now that we have a proper logging subsystem :)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

boryspoplawski:
Jenkins, retest Jenkins-SGX-gsc (please jenkins disc full)

boryspoplawski:
Jenkins, retest Jenkins-SGX-gsc please (jenkins disc full)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2168)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #570, opened 2019-03-29
[PAL/SGX] file_attrquery hangs if called on a named pipe


omeg:
SGX's `file_attrquery` implementation tries to open the given file and then calls `fstat`, in contrast to Linux-PAL which just calls `stat`. This causes unintended consequences on special files like named pipes (`open` blocks by default in such case until the other end is connected).

dimakuv:
The reason why SGX PAL does *not* perform `stat()` is because Graphene-SGX tries to use the minimum set of syscalls to the underlying host OS. SGX PAL only has `ocall_fstat()`. Introducing an new `ocall_stat()` would increase the attack surface.

Do you think there is a way to special-case named pipes so that SGX PAL still uses the combination of `open() + fstat()`?

omeg:
I'm not sure it's possible since even if opening the pipe succeeds it can have unintended side effects. Say, process A and process B use this pipe for communication. Process A opens the pipe which blocks until process B also opens it. If some unrelated process running in Graphene tries to open the pipe, the A-B communication is possibly disrupted.

Also, simple repro code:

```
#include <stdio.h>
#include <sys/stat.h>

int main(int argc, char *argv[])
{
    struct stat s;
    int rv = stat(argv[1], &s);
    printf("test: stat = %d\n", rv);
    return rv;
}
```

Create a pipe by `mkfifo test_pipe` in the shell, then run the above program with the pipe name as argument.

dimakuv:
Ok, I understand the issue now.

This is a peculiar behavior of `open()`. See a related SO question and comments here: https://stackoverflow.com/questions/5782279/why-does-a-read-only-open-of-a-named-pipe-block .

A simple (and hopefully sufficient) workaround for this particular issue would be to call `ocall_open()` with `O_NONBLOCK` inside `file_attrquery()`. We don't even need to special-case named pipes because `O_NONBLOCK` will be silently ignored by the host Linux for normal kinds of files.

pwmarcz:
Seems to be fixed by 13bd8c6e85f2f803d19f29079cf8fea5b5c5d5f7.

dimakuv:
Pawel will look at it again, but just in case, I'm assigning this bug Priority 1 (if this bug is still there).

pwmarcz:
Okay... looks like right now we don't support host-level named pipes (only emulated ones, created with `mknod` from inside Graphene).

[...]

PR #2169, opened 2021-02-18
[Pal/Linux-SGX] Add OCALL profiling modes

pwmarcz:
Based on #2168, I'll rebase once that is merged.

@dimakuv You might be interested in this :)

## Description of the changes 

This change extends the SGX profiler to handle OCALLs, not only asynchronous enclave exit. Depending on settings, we either record enclave state during OCALL, or the outer OCALL handler being executed.

Example output (mode = "ocall_outer"):

```
# Overhead  Command  Shared Object  Symbol                        
# ........  .......  .............  ..............................
#
    28.91%  pal-sgx  pal-sgx        [.] sgx_ocall_open
    27.61%  pal-sgx  pal-sgx        [.] sgx_ocall_close
    22.41%  pal-sgx  pal-sgx        [.] sgx_ocall_fstat
    17.03%  pal-sgx  pal-sgx        [.] sgx_ocall_getdents
     1.13%  pal-sgx  pal-sgx        [.] sgx_ocall_mmap_untrusted
     0.90%  pal-sgx  pal-sgx        [.] sgx_ocall_cpuid
     0.70%  pal-sgx  pal-sgx        [.] sgx_ocall_munmap_untrusted
     0.53%  pal-sgx  pal-sgx        [.] sgx_ocall_write
     0.25%  pal-sgx  pal-sgx        [.] sgx_ocall_debug_map_add
     0.13%  pal-sgx  pal-sgx        [.] sgx_ocall_pread
     0.05%  pal-sgx  pal-sgx        [.] sgx_ocall_futex
     0.05%  pal-sgx  pal-sgx        [.] sgx_ocall_gettime
     0.05%  pal-sgx  pal-sgx        [.] sgx_ocall_poll
     0.05%  pal-sgx  pal-sgx        [.] sgx_ocall_sleep
     0.05%  pal-sgx  pal-sgx        [.] sgx_ocall_socketpair
     0.03%  pal-sgx  pal-sgx        [.] sgx_ocall_clone_thread
     0.03%  pal-sgx  pal-sgx        [.] sgx_ocall_exit
     0.03%  pal-sgx  pal-sgx        [.] sgx_ocall_listen
     0.03%  pal-sgx  pal-sgx        [.] sgx_ocall_read
     0.03%  pal-sgx  pal-sgx        [.] sgx_ocall_send
     0.03%  pal-sgx  pal-sgx        [.] sgx_ocall_shutdown
```

Example output (mode = "ocall_inner"):

```
#
    99.05%  pal-sgx  libpal.so      [.] sgx_exitless_ocall
            |
            ---sgx_exitless_ocall
               |          
               |--28.92%--ocall_open
               |          |          
               |          |--22.01%--file_attrquery
               |          |          
               |          |--5.40%--dir_open
               |          |          
               |           --1.48%--file_open
               |          
               |--27.61%--ocall_close
               |          |          
               |          |--20.93%--file_attrquery
               |          |          
               |          |--5.40%--dir_close
               |          |          
               |          |--0.73%--file_close
               |          |          
               |           --0.53%--file_open
               |          
               |--22.41%--ocall_fstat
               |          |          
               |          |--20.93%--file_attrquery
               |          |          
               |           --1.48%--file_open
               |          
               |--17.04%--ocall_getdents
               |          dir_read
               |          
               |--1.13%--ocall_mmap_untrusted
               |          |          
               |           --0.90%--load_trusted_file
               |          
               |--0.70%--ocall_munmap_untrusted
               |          file_close
               |          
                --0.53%--ocall_write
                          _DkDebugLog
```

Note that these percentages are all ocall counts, not time spent.

## How to test this PR? 

Follow the instructions in documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)

### DIFF
--- a/Documentation/devel/performance.rst
+++ b/Documentation/devel/performance.rst
@@ -533,6 +533,33 @@ measuring the value of instruction pointer on each asynchronous enclave exit
 as page faults. While we attempt to measure time (and not only count
 occurences), the results might be inaccurate.
 
+.. _sgx-profile-ocall:
+
+OCALL profiling
+"""""""""""""""
+
+It's also possible to discover what OCALLs are being executed, which should help
+attribute the EEXIT numbers given by ``sgx.enable_stats``. There are two ways to
+do that:
+
+* Use ``sgx.profile.mode = "ocall_inner"`` and ``sgx.profile.with_stack =
+  1``. This will give you a report on what enclave code is causing the OCALLs
+  (best viewed with ``perf report --no-children``).
+
+  The ``with_stack`` option is important: without it, the report will only show
+  the last function before enclave exit, which is usually the same regardless of
+  which OCALL we're executing.
+
+* Use ``sgx.profile.mode = "ocall_outer"``. This will give you a report on what
+  outer PAL code is handling the OCALLs (``sgx_ocall_open``, ``sgx_ocall_write``
+  etc.)
+
+**Warning**: The report for OCALL modes should be interpreted in term of *number
+of OCALLs*, not time spent in them. The profiler records a sample every time an
+OCALL is executed, and ``perf report`` displays percentages based on the number
+of samples.
+
+
 Other useful tools for profiling
 --------------------------------
 
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -551,11 +551,30 @@ collect samples and save them to ``sgx-perf-<PID>.data``.
 The saved files can be viewed with the ``perf`` tool, e.g. ``perf report -i
 sgx-perf.data``.
 
-See :doc:`devel/performance` for more information.
+See :ref:`sgx-profile` for more information.
 
 *Note:* this option is insecure and cannot be used with production enclaves
 (``sgx.debug = 0``). If the production enclave is started with this option set,
 Graphene will fail initialization of the enclave.
+
+::
+
+    sgx.profile.mode = ["aex"|"ocall_inner"|"ocall_outer"]
+    (Default: "aex")
+
+Specifies what events to record:
+
+* ``aex``: Records enclave state during asynchronous enclave exit (AEX). Use
+  this to check where the CPU time is spent in the enclave.
+
+* ``ocall_inner``: Records enclave state during OCALL.
+
+* ``ocall_outer``: Records the outer OCALL function, i.e. what OCALL handlers
+  are going to be executed. Does not include stack information (cannot be used
+  with ``sgx.profile.with_stack = 1``).
+
+See also :ref:`sgx-profile-ocall` for more detailed advice regarding the OCALL
+modes.
 
 ::
 
@@ -577,3 +596,6 @@ lower overhead.
 
 Note that the accuracy is limited by how often the process is interrupted by
 Linux scheduler: the effective maximum is 250 samples per second.
+
+**Note**: This option applies only to ``aex`` mode. In the ``ocall_*`` modes,
+currently all samples are taken.
--- a/Pal/src/host/Linux-SGX/Makefile
+++ b/Pal/src/host/Linux-SGX/Makefile
@@ -78,6 +78,7 @@ urts-objs = \
 	sgx_platform.o \
 	sgx_process.o \
 	sgx_profile.o \
+	sgx_profile_glibc.o \
 	sgx_gdb_info.o \
 	sgx_thread.o \
 	quote/aesm.pb-c.o \
--- a/Pal/src/host/Linux-SGX/enclave_entry.S
+++ b/Pal/src/host/Linux-SGX/enclave_entry.S
@@ -631,13 +631,16 @@ sgx_ocall:
 	# %rax is argument to EEXIT
 	# %rbx is argument to EEXIT
 	# %rcx is set to AEP by EEXIT
-	xorq %rdx, %rdx
 	# %rsi, %rdi are arguments to the untrusted code
 
 #ifdef DEBUG
+	# Store pointer to context in RDX, for the SGX profiler.
+	movq %gs:SGX_PRE_OCALL_STACK, %rdx
+
 	# Keep callee-saved registers in order to recover stack later (see __morestack() below).
 #else
 	# In non-debug mode, clear these registers to prevent information leaks.
+	xorq %rdx, %rdx
 	xorq %rbp, %rbp
 	xorq %r12, %r12
 	xorq %r13, %r13
--- a/Pal/src/host/Linux-SGX/sgx_entry.S
+++ b/Pal/src/host/Linux-SGX/sgx_entry.S
@@ -84,9 +84,9 @@ async_exit_pointer:
 	.cfi_def_cfa_register %rbp
 	andq $~0xF, %rsp
 
-	# Call sgx_profile_sample with %rdi = TCS
+	# Call sgx_profile_sample_aex with %rdi = TCS
 	movq %rbx, %rdi
-	call sgx_profile_sample
+	call sgx_profile_sample_aex
 
 	# Restore stack
 	movq %rbp, %rsp
@@ -139,7 +139,26 @@ sgx_raise:
 	movq %rsp, %rbp
 	.cfi_offset %rbp, -16
 	.cfi_def_cfa_register %rbp
+
+#if DEBUG
+    # Adjust stack and save RDI
+	subq $8, %rsp
 	andq $~0xF, %rsp  # Required by System V AMD64 ABI.
+	movq %rdi, -8(%rbp)
+
+	# Call sgx_profile_sample_ocall_outer with RBX (ocall handler)
+	movq %rbx, %rdi
+	call sgx_profile_sample_ocall_outer
+
+	# Call sgx_profile_sample_ocall_inner with RDX (pointer to in-enclave context)
+	movq %rdx, %rdi
+	call sgx_profile_sample_ocall_inner
+
+	# Restore RDI
+	movq -8(%rbp), %rdi
+#else
+	andq $~0xF, %rsp  # Required by System V AMD64 ABI.
+#endif
 
 	callq *%rbx
 
--- a/Pal/src/host/Linux-SGX/sgx_internal.h
+++ b/Pal/src/host/Linux-SGX/sgx_internal.h
@@ -73,6 +73,7 @@ struct pal_enclave {
 #ifdef DEBUG
     /* profiling */
     bool profile_enable;
+    int profile_mode;
     char profile_filename[64];
     bool profile_with_stack;
     int profile_frequency;
@@ -164,6 +165,12 @@ void update_debugger(void);
 #define SGX_PROFILE_DEFAULT_FREQUENCY 50
 #define SGX_PROFILE_MAX_FREQUENCY 250
 
+enum {
+    SGX_PROFILE_MODE_AEX = 1,
+    SGX_PROFILE_MODE_OCALL_INNER = 2,
+    SGX_PROFILE_MODE_OCALL_OUTER = 3,
+};
+
 /* Filenames for saved data */
 #define SGX_PROFILE_FILENAME "sgx-perf.data"
 #define SGX_PROFILE_FILENAME_WITH_PID "sgx-perf-%d.data"
@@ -174,11 +181,20 @@ int sgx_profile_init(void);
 /* Finalize and close file */
 void sgx_profile_finish(void);
 
-/* Record a sample */
-void sgx_profile_sample(void* tcs);
+/* Record a sample during AEX */
+void sgx_profile_sample_aex(void* tcs);
+
+/* Record a sample during OCALL (inner state) */
+void sgx_profile_sample_ocall_inner(void* enclave_gpr);
+
+/* Record a sample during OCALL (function to be executed) */
+void sgx_profile_sample_ocall_outer(void* ocall_func);
 
 /* Record a new mapped ELF */
 void sgx_profile_report_elf(const char* filename, void* addr);
+
+/* Record all ELFs from outer PAL */
+void sgx_profile_report_urts_elfs(void);
 #endif
 
 /* perf.data output (sgx_perf_data.h) */
--- a/Pal/src/host/Linux-SGX/sgx_main.c
+++ b/Pal/src/host/Linux-SGX/sgx_main.c
@@ -534,6 +534,9 @@ static int initialize_enclave(struct pal_enclave* enclave, const char* manifest_
 
     debug_map_add(enclave->libpal_uri + URI_PREFIX_FILE_LEN, (void*)pal_area->addr);
     sgx_profile_report_elf(enclave->libpal_uri + URI_PREFIX_FILE_LEN, (void*)pal_area->addr);
+
+    /* Report outer PAL maps to profiler, so that we can record samples pointing to outer PAL. */
+    sgx_profile_report_urts_elfs();
 #endif
 
     ret = 0;
@@ -765,6 +768,29 @@ static int parse_loader_config(char* manifest, struct pal_enclave* enclave_info)
         goto out;
     }
 
+    char* profile_mode_str = NULL;
+    ret = toml_string_in(manifest_root, "sgx.profile.mode", &profile_mode_str);
+    if (ret < 0) {
+        urts_log_error("Cannot parse 'sgx.profile.mode' "
+                       "(the value must be \"aex\", \"ocall_inner\" or \"ocall_outer\")\n");
+        ret = -EINVAL;
+        goto out;
+    }
+    if (!profile_mode_str) {
+        enclave_info->profile_mode = SGX_PROFILE_MODE_AEX;
+    } else if (!strcmp(profile_mode_str, "aex")) {
+        enclave_info->profile_mode = SGX_PROFILE_MODE_AEX;
+    } else if (!strcmp(profile_mode_str, "ocall_inner")) {
+        enclave_info->profile_mode = SGX_PROFILE_MODE_OCALL_INNER;
+    } else if (!strcmp(profile_mode_str, "ocall_outer")) {
+        enclave_info->profile_mode = SGX_PROFILE_MODE_OCALL_OUTER;
+    } else {
+        urts_log_error("Invalid 'sgx.profile.mode' "
+                       "(the value must be \"aex\", \"ocall_inner\" or \"ocall_outer\")\n");
+        ret = -EINVAL;
+        goto out;
+    }
+
     int64_t profile_with_stack;
     ret = toml_int_in(manifest_root, "sgx.profile.with_stack", /*defaultval=*/0,
                       &profile_with_stack);
@@ -774,6 +800,15 @@ static int parse_loader_config(char* manifest, struct pal_enclave* enclave_info)
         goto out;
     }
     enclave_info->profile_with_stack = profile_with_stack;
+
+    if (enclave_info->profile_with_stack &&
+        enclave_info->profile_mode == SGX_PROFILE_MODE_OCALL_OUTER) {
+
+        urts_log_error("Invalid 'sgx.profile.mode' and 'sgx.profile.with_stack' combination "
+                       "(\"ocall_outer\" mode cannot be used with stack)\n");
+        ret = -EINVAL;
+        goto out;
+    }
 
     int64_t profile_frequency;
     ret = toml_int_in(manifest_root, "sgx.profile.frequency", SGX_PROFILE_DEFAULT_FREQUENCY,
--- a/Pal/src/host/Linux-SGX/sgx_profile.c
+++ b/Pal/src/host/Linux-SGX/sgx_profile.c
@@ -18,6 +18,7 @@
 #include "cpu.h"
 #include "elf-x86_64.h"
 #include "elf/elf.h"
+#include "linux_utils.h"
 #include "sgx_internal.h"
 #include "sgx_log.h"
 #include "sgx_tls.h"
@@ -33,6 +34,7 @@ static spinlock_t g_perf_data_lock = INIT_SPINLOCK_UNLOCKED;
 static struct perf_data* g_perf_data = NULL;
 
 static bool g_profile_enabled = false;
+static int g_profile_mode;
 static uint64_t g_profile_period;
 static int g_mem_fd = -1;
 
@@ -101,6 +103,7 @@ int sgx_profile_init(void) {
     assert(!g_perf_data);
 
     g_profile_period = NSEC_IN_SEC / g_pal_enclave.profile_frequency;
+    g_profile_mode = g_pal_enclave.profile_mode;
 
     ret = INLINE_SYSCALL(open, 3, "/proc/self/mem", O_RDONLY | O_LARGEFILE, 0);
     if (IS_ERR(ret)) {
@@ -172,18 +175,15 @@ void sgx_profile_finish(void) {
     g_profile_enabled = false;
 }
 
-static void sample_simple(void* tcs, pid_t pid, pid_t tid) {
+static void sample_simple(uint64_t rip) {
     int ret;
-    sgx_pal_gpr_t gpr;
 
-    ret = get_sgx_gpr(&gpr, tcs);
-    if (IS_ERR(ret)) {
-        urts_log_error("error reading GPR: %d\n", ret);
-        return;
-    }
+    // Report all events as the same PID so that they are grouped in report.
+    pid_t pid = g_pal_enclave.pal_sec.pid;
+    pid_t tid = pid;
 
     spinlock_lock(&g_perf_data_lock);
-    ret = pd_event_sample_simple(g_perf_data, gpr.rip, pid, tid, g_profile_period);
+    ret = pd_event_sample_simple(g_perf_data, rip, pid, tid, g_profile_period);
     spinlock_unlock(&g_perf_data_lock);
 
     if (IS_ERR(ret)) {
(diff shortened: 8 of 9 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 562 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtO7lvAbIDBRrDp-GM:-MTtO7lvAbIDBRrDp-GN:b-3lk497) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Documentation/manifest-syntax.rst#L562)):*
> ```reStructuredText
> ::
> 
>     sgx.profile.mode = ["aex"|"ocall_inner"|"ocall_outer"]
> ```

Please add that `aex` is the default.
___
*[Documentation/devel/performance.rst, line 555 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtPH0r760OxwdBFHML:-MTtPH0r760OxwdBFHMM:b-afh1sx) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Documentation/devel/performance.rst#L555)):*
> ```reStructuredText
> * Use ``sgx.profile.mode = "ocall_outer"``. This will give you a report on what
>   outer PAL code is handling the OCALLs (``sgx_ocall_open``, ``sgx_ocall_write``
>   etc.)
> ```

This mode looks a bit useless, since it's more or less the same info we can get via `strace` (outer OCALL implementations are very straight-forward and typically issue one host syscall). But still nice to have :)
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 288 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtUMrc7ng_N_VyXpBd:-MTtUMrd2m3sZ8worpka:buku4ei) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Pal/src/host/Linux-SGX/sgx_profile.c#L288)):*
> ```C
>     ret = debug_read_all(&gpr, enclave_gpr, sizeof(gpr));
>     if (IS_ERR(ret)) {
>         urts_log_error("sgx_profile_sample_eexit: error reading GPR: %d\n", ret);
> ```

Old function name?
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 303 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtVJB_3XXdHE6Sk9_Q:-MTtVJB_3XXdHE6Sk9_R:b-cvkk6o) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Pal/src/host/Linux-SGX/sgx_profile.c#L303)):*
> ```C
>         return;
> 
>     assert(!g_pal_enclave.profile_with_stack);
> ```

Maybe also assert that `ocall_func` is not NULL?
___
*[Pal/src/host/Linux-SGX/sgx_profile_glibc.c, line 37 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtSVF5C7ughd4r0yVD:-MTtSVF5C7ughd4r0yVE:b-x5sw9s) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Pal/src/host/Linux-SGX/sgx_profile_glibc.c#L37)):*
> ```C
> 
> void sgx_profile_report_urts_elfs(void) {
>     dl_iterate_phdr(&callback, NULL);
> ```

Was this feature needed for our untrusted PAL? I thought we only use Glibc and Protobuf libraries as dependencies. Did you want to collect the information on these outside-PAL libs? But you also disallow the stack with `ocall_outer`, so I'm confused what's the point of even looking at the outside libs.

Update: after reading the whole PR, it looks like you're mainly doing it to collect the ELF info on the untrusted PAL executable itself?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @pwmarcz)

___
*[Documentation/devel/performance.rst, line 555 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtPH0r760OxwdBFHML:-MTtlzcXCIWn8oSs19xA:bdp4p5) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Documentation/devel/performance.rst#L555)):*

This mode looks a bit useless, since it's more or less the same info we can get via `strace` (outer OCALL implementations are very straight-forward and typically issue one host syscall). But still nice to have :)
</blockquote></details>

Let's see:

Pros:
* It might not be exactly the same as `strace`, and differences could be helpful in hunting down problems (don't we have any non-trivial parts in the URTS except for ocalls?)
* Relatively cheap to implement
* Lower overhead than `ocall_inner` (we don't read enclave memory or record 8KB of stack sample every time), and possibly lower overhead than strace (since it happens in process, not though ptrace). I haven't experimented with it though.

Cons:
* Some added implementation complexity (adding `dl_iterate_phdr`)
* Added complexity for the user (you need to understand and decide between the two ocall modes)
* The same information, i.e. which ocall, is in `ocall_inner`. However, it's hidden at the 2nd level of stack, so you need to gather stack traces as well.

So, on balance, I'm slightly leaning towards keeping this mode. But if any of the above convinces you, I can remove it.

Also, maybe different names would help? `ocall` and `ocall_urts`?
___
*[Documentation/devel/performance.rst, line 557 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtojAG9qsTd-s4ItT2:-MTtojAG9qsTd-s4ItT3:b-1zsk4q) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Documentation/devel/performance.rst#L557)):*
> ```reStructuredText
>   etc.)
> 
> **Warning**: The report for OCALL modes should be interpreted in term of *number
> ```

One thing I'm worried about is lack of rate limiting, which means we will record every OCALL, increasing the overhead.

I don't want to do what the AEX code does (not take a sample if we took one recently). I'm afraid this will bias the results: for example, if two OCALLs are always performed one after another, we will never record the second one. For the same reason, I don't think it's a good idea to record every Nth sample. So far my only idea is something like `if rand() < threshold`. 

But then again, I'm not even sure if rate limiting is necessary, the overhead should be that high since we already execute some code (and a syscall) on every OCALL. So for now I'd rather keep it simple and see if it's useful without rate limiting.
___
*[Pal/src/host/Linux-SGX/sgx_profile_glibc.c, line 37 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtSVF5C7ughd4r0yVD:-MTtnQnkEnRa3MYSAg-H:b-8jqirf) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Pal/src/host/Linux-SGX/sgx_profile_glibc.c#L37)):*
>Update: after reading the whole PR, it looks like you're mainly doing it to collect the ELF info on the untrusted PAL executable itself?

Yes, that's what it's for. Information about glibc/protobuf/any other libraries is not necessary for now, although might be useful if we want to record something more.

If there is a simpler way to check the load address of main executable, I'll gladly change it. I figured out that since we're loaded by glibc (ld.so), it might be easiest to ask glibc itself. But as we can see, it's pretty inconvenient because of the header issue.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @pwmarcz)

___
*[Documentation/devel/performance.rst, line 555 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtPH0r760OxwdBFHML:-MTubq293K24qje2NzMJ:b6d0f72) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Documentation/devel/performance.rst#L555)):*

Let's see:

Pros:
* It might not be exactly the same as `strace`, and differences could be helpful in hunting down problems (don't we have any non-trivial parts in the URTS except for ocalls?)
* Relatively cheap to implement
* Lower overhead than `ocall_inner` (we don't read enclave memory or record 8KB of stack sample every time), and possibly lower overhead than strace (since it happens in process, not though ptrace). I haven't experimented with it though.

Cons:
* Some added implementation complexity (adding `dl_iterate_phdr`)
* Added complexity for the user (you need to understand and decide between the two ocall modes)
* The same information, i.e. which ocall, is in `ocall_inner`. However, it's hidden at the 2nd level of stack, so you need to gather stack traces as well.

So, on balance, I'm slightly leaning towards keeping this mode. But if any of the above convinces you, I can remove it.

Also, maybe different names would help? `ocall` and `ocall_urts`?
</blockquote></details>

I'm fine with current names and I am for keeping this mode.
___
*[Documentation/devel/performance.rst, line 557 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtojAG9qsTd-s4ItT2:-MTuc1QlDUwXi6k9Olsx:b8vkgkh) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Documentation/devel/performance.rst#L557)):*

One thing I'm worried about is lack of rate limiting, which means we will record every OCALL, increasing the overhead.

I don't want to do what the AEX code does (not take a sample if we took one recently). I'm afraid this will bias the results: for example, if two OCALLs are always performed one after another, we will never record the second one. For the same reason, I don't think it's a good idea to record every Nth sample. So far my only idea is something like `if rand() < threshold`. 

But then again, I'm not even sure if rate limiting is necessary, the overhead should be that high since we already execute some code (and a syscall) on every OCALL. So for now I'd rather keep it simple and see if it's useful without rate limiting.
</blockquote></details>

+1 for not adding rate limiting. It's like `strace -c` right now, which is totally fine.
___
*[Pal/src/host/Linux-SGX/sgx_profile_glibc.c, line 37 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtSVF5C7ughd4r0yVD:-MTucMJb1o26turTgKuE:b-iwcxpc) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Pal/src/host/Linux-SGX/sgx_profile_glibc.c#L37)):*

>Update: after reading the whole PR, it looks like you're mainly doing it to collect the ELF info on the untrusted PAL executable itself?

Yes, that's what it's for. Information about glibc/protobuf/any other libraries is not necessary for now, although might be useful if we want to record something more.

If there is a simpler way to check the load address of main executable, I'll gladly change it. I figured out that since we're loaded by glibc (ld.so), it might be easiest to ask glibc itself. But as we can see, it's pretty inconvenient because of the header issue.
</blockquote></details>

I'm fine with this solution. If anything, we can change this later.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: 7 of 9 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 562 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtO7lvAbIDBRrDp-GM:-MTuq0rqDOct7niHKdKv:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Documentation/manifest-syntax.rst#L562)):*

Please add that `aex` is the default.
</blockquote></details>

Done.
___
*[Documentation/devel/performance.rst, line 555 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtPH0r760OxwdBFHML:-MTuq1yc7681yt-zBzQR:b5iwv2w) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Documentation/devel/performance.rst#L555)):*

I'm fine with current names and I am for keeping this mode.
</blockquote></details>

Okay, then no changes here.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 288 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtUMrc7ng_N_VyXpBd:-MTur2mmBlkT3aKNgnWr:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Pal/src/host/Linux-SGX/sgx_profile.c#L288)):*

Old function name?
</blockquote></details>

Fixed.
___
*[Pal/src/host/Linux-SGX/sgx_profile.c, line 303 at r1](https://reviewable.io/reviews/oscarlab/graphene/2169#-MTtVJB_3XXdHE6Sk9_Q:-MTuqLJD6wDg4TDS06QU:b-ijlim) ([raw file](https://github.com/oscarlab/graphene/blob/a57ba07ea0939114380eb15cb2edab5cf2e98871/Pal/src/host/Linux-SGX/sgx_profile.c#L303)):*

Maybe also assert that `ocall_func` is not NULL?
</blockquote></details>

There's no harm here if it's NULL (we'll just log a 0 to the perf data file), but indeed that should never happen, as we're about to jump to `ocall_func`. I added an assertion.

mkow (changes requested):
> Reviewed 6 of 9 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2169#-MU0Q0pP0LDZQ9h_eJb6:-MU0Q0pP0LDZQ9h_eJb7:b-b9rwgl) (no related file):*
FYI: I played with this profiler briefly, and seems to be really useful for pinpointing reasons of high numbers of OCALLs. Example from Blender:
```
-   97.89%  pal-sgx  libpal.so      [.] sgx_exitless_ocall
   - sgx_exitless_ocall
      - 21.00% ocall_open
         + 14.25% file_attrquery
         + 5.02% file_open
         + 1.72% dir_open
```

This clearly shows an issue with our attrquery always requiring to open a file. It's already known to us (the reason is AFAIR PAL API design, which says it should be stateless), but I think it's a nice demo for this new feature :)
Also, this particular problem is probably something which would be good to fix when rewriting the FS subsystem.

---
___
*[Pal/src/host/Linux-SGX/sgx_profile_glibc.c, line 19 at r2](https://reviewable.io/reviews/oscarlab/graphene/2169#-MU0TmNt4GcjE8CWYA0W:-MU0TmNt4GcjE8CWYA0X:bfcgbjq) ([raw file](https://github.com/oscarlab/graphene/blob/7abefcb6dbc95ebb7b310aa9a19368ef5f93ca0b/Pal/src/host/Linux-SGX/sgx_profile_glibc.c#L19)):*
> ```C
> #include <stdlib.h>
> #include <string.h>
> #include <link.h>
> ```

nitpick: please sort :)
___
*[Pal/src/host/Linux-SGX/sgx_profile_glibc.c, line 30 at r2](https://reviewable.io/reviews/oscarlab/graphene/2169#-MU0Tv7dEFPnjPRbA26Q:-MU0Tv7dEFPnjPRbA26R:b-wb2zo7) ([raw file](https://github.com/oscarlab/graphene/blob/7abefcb6dbc95ebb7b310aa9a19368ef5f93ca0b/Pal/src/host/Linux-SGX/sgx_profile_glibc.c#L30)):*
> ```C
> static int callback(struct dl_phdr_info *info,
>                     __attribute__((unused)) size_t size,
>                     __attribute__((unused)) void *data) {
> ```

please move asterisks to the left (x2)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2169#-MU0Q0pP0LDZQ9h_eJb6:-MU0cGAX1KcH9XsiSBra:b9n8dn6) (no related file):*

FYI: I played with this profiler briefly, and seems to be really useful for pinpointing reasons of high numbers of OCALLs. Example from Blender:
```
-   97.89%  pal-sgx  libpal.so      [.] sgx_exitless_ocall
   - sgx_exitless_ocall
      - 21.00% ocall_open
         + 14.25% file_attrquery
         + 5.02% file_open
         + 1.72% dir_open
```

This clearly shows an issue with our attrquery always requiring to open a file. It's already known to us (the reason is AFAIR PAL API design, which says it should be stateless), but I think it's a nice demo for this new feature :)
Also, this particular problem is probably something which would be good to fix when rewriting the FS subsystem.
</blockquote></details>

Isn't it because we wanted PAL to support a minimal number of different syscalls, and so we don't do `stat` because we can do `open` + `fstat`? 

I don't necessarily agree, especially if it matters for performance, but that sounds like a reason.

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2169#-MU0Q0pP0LDZQ9h_eJb6:-MU0mS6XCxZNZkWYWK3a:bsewkqx) (no related file):*

Isn't it because we wanted PAL to support a minimal number of different syscalls, and so we don't do `stat` because we can do `open` + `fstat`? 

I don't necessarily agree, especially if it matters for performance, but that sounds like a reason.
</blockquote></details>

Hmm, actually, I just checked and we have `file_attrquerybyhdl`, but AFAIR some PAL APIs still had to use the handle-less versions for some reason (and that was the known issue I mentioned, but I don't remember the details). And we already have `ocall_fstat` :)

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2169#-MU0Q0pP0LDZQ9h_eJb6:-MU0nNfxDzVmXwPZQ6PM:bswv83c) (no related file):*

Hmm, actually, I just checked and we have `file_attrquerybyhdl`, but AFAIR some PAL APIs still had to use the handle-less versions for some reason (and that was the known issue I mentioned, but I don't remember the details). And we already have `ocall_fstat` :)
</blockquote></details>

Ah, sorry, misread your part about stat/fstat, yes, I think that was the reason to do open+fstat. I think we should just somehow get rid of the handle-less `attrquery`.

---

yamahata (approved):
> Reviewed 6 of 9 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @pwmarcz)

dimakuv (dismissed):
> Reviewed 2 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2169#-MU0Q0pP0LDZQ9h_eJb6:-MU7v7LdBzK0PwVj7Ew7:b-tmyxc0) (no related file):*

Ah, sorry, misread your part about stat/fstat, yes, I think that was the reason to do open+fstat. I think we should just somehow get rid of the handle-less `attrquery`.
</blockquote></details>

https://github.com/oscarlab/graphene/issues/570 -- Indeed, `ocall_stat` was never implemented because we can emulate the same via `ocall_open + ocall_fstat`. The reason is to minimize the number of OCALLs (to decrease attack surface).

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: 8 of 9 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2169#-MU0Q0pP0LDZQ9h_eJb6:-MU878fcE-wdfnG4r3DI:b-nqlpyb) (no related file):*

https://github.com/oscarlab/graphene/issues/570 -- Indeed, `ocall_stat` was never implemented because we can emulate the same via `ocall_open + ocall_fstat`. The reason is to minimize the number of OCALLs (to decrease attack surface).
</blockquote></details>

There's still room for improvement here. For example our `shim_do_getdents` ends up calling `DkStreamAttributesQuery` for every file. There is an OCALL for `getdents`, which PAL uses for listing the directory, but it uses only file names.

---
___
*[Pal/src/host/Linux-SGX/sgx_profile_glibc.c, line 19 at r2](https://reviewable.io/reviews/oscarlab/graphene/2169#-MU0TmNt4GcjE8CWYA0W:-MU88yhk4W1I3QQSxWSM:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7abefcb6dbc95ebb7b310aa9a19368ef5f93ca0b/Pal/src/host/Linux-SGX/sgx_profile_glibc.c#L19)):*

nitpick: please sort :)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_profile_glibc.c, line 30 at r2](https://reviewable.io/reviews/oscarlab/graphene/2169#-MU0Tv7dEFPnjPRbA26Q:-MU890TZ2WiiuC8a3lid:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7abefcb6dbc95ebb7b310aa9a19368ef5f93ca0b/Pal/src/host/Linux-SGX/sgx_profile_glibc.c#L30)):*

please move asterisks to the left (x2)
</blockquote></details>

Done.

dimakuv (approved):
> Reviewed 1 of 3 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2169#-MU0Q0pP0LDZQ9h_eJb6:-MU8C20F-Y9sb7LRSnQW:b-jxcynd) (no related file):*

There's still room for improvement here. For example our `shim_do_getdents` ends up calling `DkStreamAttributesQuery` for every file. There is an OCALL for `getdents`, which PAL uses for listing the directory, but it uses only file names.
</blockquote></details>

Yes, in terms of performance there is a lot of room for improvement... I'm not against adding `ocall_stat` if this lies on a critical path for FS-intensive workloads, but I think we currently just need to remove `DkStreamAttributesQuery()` from the critical paths (which should be possible).

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2169#-MU0Q0pP0LDZQ9h_eJb6:-MU8D_aQB34_meoNDFU2:b-r7dggi) (no related file):*

Yes, in terms of performance there is a lot of room for improvement... I'm not against adding `ocall_stat` if this lies on a critical path for FS-intensive workloads, but I think we currently just need to remove `DkStreamAttributesQuery()` from the critical paths (which should be possible).
</blockquote></details>

That's what I meant - we already use `getdents`, so if we just stop throwing away the results, a lot of this `open + fstat` can go away.

---

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2169)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2177, opened 2021-02-20
[LibOS] Migrate logging to the new API

mkow:
## Description of the changes 

PAL was already converted, only LibOS was left, so that's it.

After this PR the only thing left is code in Pal/lib, but this one is tricky - the code there is also used without the rest of Graphene, and also shared between PAL and LibOS. So, that's left for another PR.

Update: And `pal_printf`, I think we only updated untrusted PAL/Linux-SGX so far, but not the trusted part.

Also, fixed a bug in the `_log` macro which got triggered after converting some of the callsites.

## How to test this PR? 

Try different log_levels.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)

### DIFF
--- a/LibOS/shim/include/shim_checkpoint.h
+++ b/LibOS/shim/include/shim_checkpoint.h
@@ -296,7 +296,7 @@ struct shim_cp_map_entry* get_cp_map_entry(void* map, void* addr, bool create);
 
 #if DEBUG_RESUME == 1
 #define DEBUG_RS(fmt, ...) \
-    debug("GET %s(0x%08lx): " fmt "\n", CP_FUNC_NAME, entry->cp_val, ##__VA_ARGS__)
+    log_debug("GET %s(0x%08lx): " fmt "\n", CP_FUNC_NAME, entry->cp_val, ##__VA_ARGS__)
 #else
 #define DEBUG_RS(...) do {} while (0)
 #endif
@@ -313,7 +313,7 @@ struct shim_cp_map_entry* get_cp_map_entry(void* map, void* addr, bool create);
             if (ret < 0)                                                   \
                 goto out;                                                  \
                                                                            \
-            debug("complete checkpointing data\n");                        \
+            log_debug("complete checkpointing data\n");                    \
         out:                                                               \
             destroy_cp_map((store)->cp_map);                               \
         } while (0);                                                       \
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -46,11 +46,9 @@ void debug_puts(const char* str);
 void debug_putch(int ch);
 void debug_vprintf(const char* fmt, va_list ap) __attribute__((format(printf, 1, 0)));
 
-#define _log(level, fmt...)                          \
-    do {                                             \
-        if ((level) <= g_log_level)                  \
-            debug_printf(fmt);                       \
-    } while (0)
+// TODO(mkow): We should make it cross-object-inlinable, ideally by enabling LTO, less ideally by
+// pasting it here and making `inline`, but our current linker scripts prevent both.
+void _log(int level, const char* fmt, ...) __attribute__((format(printf, 2, 3)));
 
 #define log_error(fmt...)    _log(PAL_LOG_ERROR, fmt)
 #define log_warning(fmt...)  _log(PAL_LOG_WARNING, fmt)
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -64,16 +64,16 @@ void _log(int level, const char* fmt, ...) __attribute__((format(printf, 2, 3)))
 #define DEBUG_BREAK_ON_FAILURE() do {} while (0)
 #endif
 
-#define BUG()                                       \
-    do {                                            \
-        warn("BUG() " __FILE__ ":%d\n", __LINE__);  \
-        DEBUG_BREAK_ON_FAILURE();                   \
-        die_or_inf_loop();                          \
+#define BUG()                                           \
+    do {                                                \
+        log_error("BUG() " __FILE__ ":%d\n", __LINE__); \
+        DEBUG_BREAK_ON_FAILURE();                       \
+        die_or_inf_loop();                              \
     } while (0)
 
-#define DEBUG_HERE()                                         \
-    do {                                                     \
-        debug("%s (" __FILE__ ":%d)\n", __func__, __LINE__); \
+#define DEBUG_HERE()                                             \
+    do {                                                         \
+        log_debug("%s (" __FILE__ ":%d)\n", __func__, __LINE__); \
     } while (0)
 
 /*!
@@ -322,7 +322,7 @@ static inline int __ref_dec(REFTYPE* ref) {
     do {
         _c = __atomic_load_n(&ref->counter, __ATOMIC_SEQ_CST);
         if (!_c) {
-            debug("Fail: Trying to drop reference count below 0\n");
+            log_error("Fail: Trying to drop reference count below 0\n");
             BUG();
             return 0;
         }
--- a/LibOS/shim/include/shim_lock.h
+++ b/LibOS/shim/include/shim_lock.h
@@ -40,23 +40,12 @@ static inline void destroy_lock(struct shim_lock* l) {
     clear_lock(l);
 }
 
-#ifdef DEBUG
-#define lock(l) __lock(l, __FILE__, __LINE__)
-static void __lock(struct shim_lock* l, const char* file, int line) {
-#else
 static void lock(struct shim_lock* l) {
-#endif
     if (!lock_enabled) {
         return;
     }
-    /* TODO: This whole if should be just an assert. Change it once we are sure that it does not
-     * trigger (previous code allowed for this case). Same in unlock below. */
-    if (!l->lock) {
-#ifdef DEBUG
-        debug("Trying to lock an uninitialized lock at %s:%d!\n", file, line);
-#endif // DEBUG
-        __abort();
-    }
+
+    assert(l->lock);
 
     while (!DkSynchronizationObjectWait(l->lock, NO_TIMEOUT))
         /* nop */;
@@ -64,22 +53,12 @@ static void lock(struct shim_lock* l) {
     l->owner = get_cur_tid();
 }
 
-#ifdef DEBUG
-#define unlock(l) __unlock(l, __FILE__, __LINE__)
-static inline void __unlock(struct shim_lock* l, const char* file, int line) {
-#else
 static inline void unlock(struct shim_lock* l) {
-#endif
     if (!lock_enabled) {
         return;
     }
-    if (!l->lock) {
-#ifdef DEBUG
-        debug("Trying to unlock an uninitialized lock at %s:%d!\n", file, line);
-#endif // DEBUG
-        __abort();
-    }
 
+    assert(l->lock);
     l->owner = 0;
     DkMutexRelease(l->lock);
 }
@@ -102,11 +81,11 @@ extern struct shim_lock __master_lock;
 #define MASTER_LOCK()                                          \
     do {                                                       \
         lock(&__master_lock);                                  \
-        pal_printf("master lock " __FILE__ ":%d\n", __LINE__); \
+        log_debug("master lock " __FILE__ ":%d\n", __LINE__);  \
     } while (0)
 #define MASTER_UNLOCK()                                          \
     do {                                                         \
-        pal_printf("master unlock " __FILE__ ":%d\n", __LINE__); \
+        log_debug("master unlock " __FILE__ ":%d\n", __LINE__);  \
         unlock(&__master_lock);                                  \
     } while (0)
 #else
--- a/LibOS/shim/src/bookkeep/shim_handle.c
+++ b/LibOS/shim/src/bookkeep/shim_handle.c
@@ -398,7 +398,7 @@ void get_handle(struct shim_handle* hdl) {
 #ifdef DEBUG_REF
     int ref_count = REF_INC(hdl->ref_count);
 
-    debug("get handle %p(%s) (ref_count = %d)\n", hdl, __handle_name(hdl), ref_count);
+    log_debug("get handle %p(%s) (ref_count = %d)\n", hdl, __handle_name(hdl), ref_count);
 #else
     REF_INC(hdl->ref_count);
 #endif
@@ -414,7 +414,7 @@ void put_handle(struct shim_handle* hdl) {
     int ref_count = REF_DEC(hdl->ref_count);
 
 #ifdef DEBUG_REF
-    debug("put handle %p(%s) (ref_count = %d)\n", hdl, __handle_name(hdl), ref_count);
+    log_debug("put handle %p(%s) (ref_count = %d)\n", hdl, __handle_name(hdl), ref_count);
 #endif
 
     if (!ref_count) {
@@ -458,7 +458,7 @@ void put_handle(struct shim_handle* hdl) {
 
         if (hdl->pal_handle) {
 #ifdef DEBUG_REF
-            debug("handle %p closes PAL handle %p\n", hdl, hdl->pal_handle);
+            log_debug("handle %p closes PAL handle %p\n", hdl, hdl->pal_handle);
 #endif
             DkObjectClose(hdl->pal_handle);
             hdl->pal_handle = NULL;
--- a/LibOS/shim/src/bookkeep/shim_process.c
+++ b/LibOS/shim/src/bookkeep/shim_process.c
@@ -48,7 +48,7 @@ int init_process(void) {
     struct shim_dentry* dent = NULL;
     int ret = path_lookupat(NULL, "/", 0, &dent, NULL);
     if (ret < 0) {
-        debug("Could not set up dentry for \"/\", something is seriously broken.\n");
+        log_error("Could not set up dentry for \"/\", something is seriously broken.\n");
         return ret;
     }
     g_process.root = dent;
@@ -147,7 +147,7 @@ static bool mark_child_exited(child_cmp_t child_cmp, unsigned long arg, IDTYPE c
         fill_siginfo_code_and_status(&info, signal, exit_code);
         int x = kill_current_proc(&info);
         if (x < 0) {
-            debug("Sending child death signal failed: %d!\n", x);
+            log_error("Sending child death signal failed: %d!\n", x);
         }
     }
 
--- a/LibOS/shim/src/bookkeep/shim_signal.c
+++ b/LibOS/shim/src/bookkeep/shim_signal.c
@@ -61,7 +61,7 @@ void thread_sigaction_reset_on_execve(void) {
 }
 
 static noreturn void sighandler_kill(int sig) {
-    debug("killed by signal %d\n", sig & ~__WCOREDUMP_BIT);
+    log_debug("killed by signal %d\n", sig & ~__WCOREDUMP_BIT);
     process_exit(0, sig);
 }
 
@@ -311,12 +311,12 @@ static noreturn void internal_fault(const char* errstr, PAL_NUM addr, PAL_CONTEX
     PAL_NUM ip = pal_context_get_ip(context);
 
     if (context_is_libos(context))
-        warn("%s at 0x%08lx (IP = +0x%lx, VMID = %u, TID = %u)\n", errstr, addr,
-             (void*)ip - (void*)&__load_address, g_process_ipc_info.vmid,
-             is_internal_tid(tid) ? 0 : tid);
+        log_error("%s at 0x%08lx (IP = +0x%lx, VMID = %u, TID = %u)\n", errstr, addr,
+                  (void*)ip - (void*)&__load_address, g_process_ipc_info.vmid,
+                  is_internal_tid(tid) ? 0 : tid);
     else
-        warn("%s at 0x%08lx (IP = 0x%08lx, VMID = %u, TID = %u)\n", errstr, addr,
-             context ? ip : 0, g_process_ipc_info.vmid, is_internal_tid(tid) ? 0 : tid);
+        log_error("%s at 0x%08lx (IP = 0x%08lx, VMID = %u, TID = %u)\n", errstr, addr,
+                  context ? ip : 0, g_process_ipc_info.vmid, is_internal_tid(tid) ? 0 : tid);
 
     DEBUG_BREAK_ON_FAILURE();
     DkProcessExit(1);
@@ -330,7 +330,7 @@ static void arithmetic_error_upcall(bool is_in_pal, PAL_NUM addr, PAL_CONTEXT* c
     if (is_internal_tid(get_cur_tid()) || context_is_libos(context)) {
         internal_fault("Internal arithmetic fault", addr, context);
     } else {
-        debug("arithmetic fault at 0x%08lx\n", pal_context_get_ip(context));
+        log_debug("arithmetic fault at 0x%08lx\n", pal_context_get_ip(context));
         siginfo_t info = {
             .si_signo = SIGFPE,
             .si_code = FPE_INTDIV,
@@ -361,7 +361,7 @@ static void memfault_upcall(bool is_in_pal, PAL_NUM addr, PAL_CONTEXT* context) 
         internal_fault("Internal memory fault", addr, context);
     }
 
-    debug("memory fault at 0x%08lx (IP = 0x%08lx)\n", addr, pal_context_get_ip(context));
+    log_debug("memory fault at 0x%08lx (IP = 0x%08lx)\n", addr, pal_context_get_ip(context));
 
     siginfo_t info = {
         .si_addr = (void*)addr,
@@ -590,7 +590,7 @@ static void illegal_upcall(bool is_in_pal, PAL_NUM addr, PAL_CONTEXT* context) {
     /* Emulate syscall instruction, which is prohibited in Linux-SGX PAL and raises a SIGILL. */
     if (!maybe_emulate_syscall(context)) {
         void* rip = (void*)pal_context_get_ip(context);
-        debug("Illegal instruction during app execution at %p; delivering to app\n", rip);
+        log_debug("Illegal instruction during app execution at %p; delivering to app\n", rip);
         siginfo_t info = {
             .si_signo = SIGILL,
             .si_code = ILL_ILLOPC,
@@ -641,7 +641,7 @@ int init_signal_handling(void) {
     int ret = toml_int_in(g_manifest_root, "sys.enable_sigterm_injection", /*defaultval=*/0,
                           &allow_injection);
     if (ret < 0 || (allow_injection != 0 && allow_injection != 1)) {
-        debug("Cannot parse 'sys.enable_sigterm_injection' (the value must be 0 or 1)\n");
+        log_error("Cannot parse 'sys.enable_sigterm_injection' (the value must be 0 or 1)\n");
         return -EINVAL;
     }
     g_inject_host_signal_enabled = !!allow_injection;
@@ -650,7 +650,7 @@ int init_signal_handling(void) {
     ret = toml_int_in(g_manifest_root, "libos.check_invalid_pointers",
                       /*defaultval=*/1, &check_invalid_ptrs_int);
     if (ret < 0 || (check_invalid_ptrs_int != 0 && check_invalid_ptrs_int != 1)) {
-        debug("Cannot parse 'libos.check_invalid_pointers' (the value must be 0 or 1)\n");
+        log_error("Cannot parse 'libos.check_invalid_pointers' (the value must be 0 or 1)\n");
         return -EINVAL;
     }
     g_check_invalid_ptrs = !!check_invalid_ptrs_int;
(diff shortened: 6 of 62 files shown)


pwmarcz (commented):
> Reviewed 62 of 62 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU0_vpq7Ux40brfFDuD:-MU0_vpr7LcrWEH7jA87:bhw3q0j) (no related file):*
I haven't tried to run it, but is the `debug` output acceptable now, or should we demote some more noisy messages to `trace`? For instance IPC?

---
___
*[LibOS/shim/include/shim_internal.h, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU0TxGh5e-7V0cBfehT:-MU0TxGh5e-7V0cBfehU:bcdda3p) ([raw file](https://github.com/oscarlab/graphene/blob/2d4725908203b675507f041855927e9936ff1dbd/LibOS/shim/include/shim_internal.h#L49)):*
> ```objc
> void debug_vprintf(const char* fmt, va_list ap) __attribute__((format(printf, 1, 0)));
> 
> void _log(int level, const char* fmt, ...) __attribute__((format(printf, 2, 3)));
> ```

Why not inline? An inline function will preserve the side effects, but still might be optimized if we end up not needing to log something: for instance, `log_debug("%p", foo->bar)` can potentially omit dereferencing `foo` if the function is inline.
___
*[LibOS/shim/src/shim_malloc.c, line 97 at r1](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU0ZsWd47XSBFlrf1-W:-MU0ZsWd47XSBFlrf1-X:bh34ixm) ([raw file](https://github.com/oscarlab/graphene/blob/2d4725908203b675507f041855927e9936ff1dbd/LibOS/shim/src/shim_malloc.c#L97)):*
> ```C
>          * condition and must terminate the current process.
>          */
>         log_error("******** Out-of-memory in library OS ********\n");
> ```

Do we still want the `******`?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: 61 of 62 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU0_vpq7Ux40brfFDuD:-MU0nwRMB86rUoSbo0ng:b-6hpuii) (no related file):*

I haven't tried to run it, but is the `debug` output acceptable now, or should we demote some more noisy messages to `trace`? For instance IPC?
</blockquote></details>

Hard to say, it was still a bit noisy (not terribly), but that's mostly because it's totally random which events are logged and which aren't. Each subsystem is different in this terms. I think the log consisted mostly of thread creation/exit events and "async event installed [...]" entries.

What we should do is probably look at the whole codebase and remove all low-quality/noise/debugging-leftovers messages and add some important ones.

---
___
*[LibOS/shim/include/shim_internal.h, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU0TxGh5e-7V0cBfehT:-MU0om4OCcDyqhCPK4e7:b-an8zt0) ([raw file](https://github.com/oscarlab/graphene/blob/2d4725908203b675507f041855927e9936ff1dbd/LibOS/shim/include/shim_internal.h#L49)):*

Why not inline? An inline function will preserve the side effects, but still might be optimized if we end up not needing to log something: for instance, `log_debug("%p", foo->bar)` can potentially omit dereferencing `foo` if the function is inline.
</blockquote></details>

Normal functions can also be optimized, `inline` is just a hint to compiler's heuristic. I'm not convinced it should ever be used - in the end, the compiler should know better where and what to inline.
___
*[LibOS/shim/src/shim_malloc.c, line 97 at r1](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU0ZsWd47XSBFlrf1-W:-MU0pCRU8mywokDCLcps:b3unloe) ([raw file](https://github.com/oscarlab/graphene/blob/2d4725908203b675507f041855927e9936ff1dbd/LibOS/shim/src/shim_malloc.c#L97)):*

Do we still want the `******`?
</blockquote></details>

Nah, fixed.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: 61 of 62 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[LibOS/shim/include/shim_internal.h, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU0TxGh5e-7V0cBfehT:-MU2clCG3CqwwE8N32V9:bw3z1dr) ([raw file](https://github.com/oscarlab/graphene/blob/2d4725908203b675507f041855927e9936ff1dbd/LibOS/shim/include/shim_internal.h#L49)):*

Normal functions can also be optimized, `inline` is just a hint to compiler's heuristic. I'm not convinced it should ever be used - in the end, the compiler should know better where and what to inline.
</blockquote></details>

Functions in the same object file yes, but I think GCC will not optimize functions across compilation units unless enable link-time optimization (`-flto`). Should we enable it for Graphene?

(Or maybe what I know is outdated, and LTO is on by default, but I didn't find anything to confirm that).

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: 61 of 62 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_internal.h, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU0TxGh5e-7V0cBfehT:-MU4Y96XEGUDUqxgyckJ:b-wu73ad) ([raw file](https://github.com/oscarlab/graphene/blob/2d4725908203b675507f041855927e9936ff1dbd/LibOS/shim/include/shim_internal.h#L49)):*

Functions in the same object file yes, but I think GCC will not optimize functions across compilation units unless enable link-time optimization (`-flto`). Should we enable it for Graphene?

(Or maybe what I know is outdated, and LTO is on by default, but I didn't find anything to confirm that).
</blockquote></details>

Yup, but this has (AFAIK) nothing to do with `inline` specifier.

And yes, we should enable LTO, but I think it may be non-trivial, considering our custom linking scripts. I'm not sure if anyone has tried it already with Graphene.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: 61 of 62 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[LibOS/shim/include/shim_internal.h, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU0TxGh5e-7V0cBfehT:-MU4Zeuw1PzVLDv2TZKc:b-sfbsn1) ([raw file](https://github.com/oscarlab/graphene/blob/2d4725908203b675507f041855927e9936ff1dbd/LibOS/shim/include/shim_internal.h#L49)):*
> Yup, but this has (AFAIK) nothing to do with inline specifier.

Not directly, but: if we don't have LTO, and if we want this function to be inlined, it needs to be defined here in the header file. And `inline` is the only way to do that, because otherwise the linker will complain about multiple definitions across `.o` files. (`inline` is used not to ensure the function is inlined, but to ensure we're generating a weak symbol).

So, since we don't compile with LTO right now, I understand the choice is:
* define `_log` as inline in the header file, and it will get inlined by compiler when appropriate,
* define `_log` in a separate `.c` file, and it will never get inlined (except possibly in other functions in that file).

Do I have it right?

(Now that I think of it, we also tend to abuse `static inline`, which causes compiler to emit separate copies if the functions are not inlined...)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: 61 of 62 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[LibOS/shim/include/shim_internal.h, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU0TxGh5e-7V0cBfehT:-MU4dPi7CaeFo1SUjaOm:bjk39t1) ([raw file](https://github.com/oscarlab/graphene/blob/2d4725908203b675507f041855927e9936ff1dbd/LibOS/shim/include/shim_internal.h#L49)):*
>(Now that I think of it, we also tend to abuse static inline, which causes compiler to emit separate copies if the functions are not inlined...)

That's unrelated, but... ouch:

```
$ nm Runtime/libsysdb.so | grep ref_inc
000000000001723f t __ref_inc
000000000002023c t __ref_inc
000000000002df2e t __ref_inc
00000000000418fb t __ref_inc
0000000000043533 t __ref_inc
000000000004716f t __ref_inc
000000000005dea1 t __ref_inc
```

`__ref_inc` is `static inline`. And changing it to `inline` causes linker errors about undefined symbols. Either I don't understand something or our build script is broken and `static inline` was a workaround for that :)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: 61 of 62 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_internal.h, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU0TxGh5e-7V0cBfehT:-MU4h6g20FXSNWdNgukG:bqyv6wo) ([raw file](https://github.com/oscarlab/graphene/blob/2d4725908203b675507f041855927e9936ff1dbd/LibOS/shim/include/shim_internal.h#L49)):*

>(Now that I think of it, we also tend to abuse static inline, which causes compiler to emit separate copies if the functions are not inlined...)

That's unrelated, but... ouch:

```
$ nm Runtime/libsysdb.so | grep ref_inc
000000000001723f t __ref_inc
000000000002023c t __ref_inc
000000000002df2e t __ref_inc
00000000000418fb t __ref_inc
0000000000043533 t __ref_inc
000000000004716f t __ref_inc
000000000005dea1 t __ref_inc
```

`__ref_inc` is `static inline`. And changing it to `inline` causes linker errors about undefined symbols. Either I don't understand something or our build script is broken and `static inline` was a workaround for that :)
</blockquote></details>

Okay, my conclusion is: I think `_log` should be possible to be inlined by compiler, and I would recommend it to be defined `inline` in the header... once our linking is fixed. If you agree, maybe add a TODO comment? I'm resolving the discussion anyway.

(Usiing `static inline` right now would work pretty badly since we would generate separate `_log` function for each file). 

I ticketed the linking issue as https://github.com/oscarlab/graphene/issues/2179.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: 59 of 62 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_internal.h, line 49 at r1](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU0TxGh5e-7V0cBfehT:-MU6HdF45gXAQRSiA2f6:bh715yw) ([raw file](https://github.com/oscarlab/graphene/blob/2d4725908203b675507f041855927e9936ff1dbd/LibOS/shim/include/shim_internal.h#L49)):*

Okay, my conclusion is: I think `_log` should be possible to be inlined by compiler, and I would recommend it to be defined `inline` in the header... once our linking is fixed. If you agree, maybe add a TODO comment? I'm resolving the discussion anyway.

(Usiing `static inline` right now would work pretty badly since we would generate separate `_log` function for each file). 

I ticketed the linking issue as https://github.com/oscarlab/graphene/issues/2179.
</blockquote></details>

(for others: at this point we had some discussions on priv, so I'll skip replying to older points)

Hmm, I'd rather say that the ideal way of resolving this would be to leave this code as-is and enable LTO, as it doesn't require us to manually mark all cross-object-inlinable functions. The question here is how hard would it be to enable LTO and whether it will be possible in near future.

Anyways, a TODO added.

yamahata (dismissed):
> Reviewed 58 of 62 files at r1, 1 of 1 files at r2, 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

dimakuv (changes requested):
> Reviewed 58 of 62 files at r1, 1 of 1 files at r2, 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/src/fs/sys/cache_info.c, line 47 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7qwZ34F1yQ-WeybLs:-MU7qwZ34F1yQ-WeybLt:bnk71o7) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/fs/sys/cache_info.c#L47)):*
> ```C
>         cache_filebuf = pal_control.topo_info.core_topology[cpunum].cache[idx].physical_line_partition;
>     } else {
>         log_debug("Unrecognized file %s\n", name);
> ```

For all these sysfs files, I'd prefer `log_warning` (I can imagine a situation when some app goes bonkers because a sysfs file is not found, and the end user shouldn't scroll through MBs of `debug` logs to find this). But no hard preference.
___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 602 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7qlo6C72ZHTUHjC-s:-MU7qlo6C72ZHTUHjC-t:b-okhvmt) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/ipc/shim_ipc_helper.c#L602)):*
> ```C
>                     ret = send_response_ipc_message(port, msg->src, ret, msg->seq);
>                     if (ret < 0) {
>                         log_debug("Sending IPC_MSG_RESP msg on port %p (handle %p) to %u failed\n",
> ```

I'd prefer `log_error`. This is a Graphene-internal error which should never happen in normal circumstances.
___
*[LibOS/shim/src/sys/shim_clone.c, line 253 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7rv_KApaXFmTfHEpk:-MU7rv_KApaXFmTfHEpl:b-p90i5x) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_clone.c#L253)):*
> ```C
> 
>     if (flags & ~supported_flags) {
>         log_warning("clone called with unsupported flags argument.\n");
> ```

Shouldn't it be `log_error`?

Here and below, my assumption is:
- `warning` is used for situations "weird syscall arguments are given, it is application's fault OR Graphene will continue but with caveats"
- `error` is used for situations "unsupported arguments are given, Graphene will fail unexpectedly for application"
___
*[LibOS/shim/src/sys/shim_clone.c, line 273 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7rxgzDQrJC-8jNVK6:-MU7rxgzDQrJC-8jNVK7:b-ea14kv) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_clone.c#L273)):*
> ```C
>     if (flags & CLONE_VM) {
>         if (!((flags & CLONE_THREAD) || (flags & CLONE_VFORK))) {
>             log_warning("CLONE_VM without either CLONE_THREAD or CLONE_VFORK is unsupported\n");
> ```

Shouldn't it be `log_error`?
___
*[LibOS/shim/src/sys/shim_futex.c, line 799 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7sDBu-ZEFA6n34swO:-MU7sDBu-ZEFA6n34swP:b-81wohb) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_futex.c#L799)):*
> ```C
>         case FUTEX_CMP_REQUEUE_PI:
>         case FUTEX_WAIT_REQUEUE_PI:
>             log_warning("PI futexes are not yet supported!\n");
> ```

`log_error`?
___
*[LibOS/shim/src/sys/shim_mmap.c, line 255 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7sNoEC79BVGNOs94x:-MU7sNoEC79BVGNOs94y:bw9gsbk) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_mmap.c#L255)):*
> ```C
>             }
>         } else {
>             log_warning("Memory that was about to be mprotected was unmapped, your program is "
> ```

`log_error`? Even though it is application's fault, sounds like something to complain about very loudly.
___
*[LibOS/shim/src/sys/shim_sleep.c, line 54 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7sbjl6kuY_YiCc3zo:-MU7sbjl6kuY_YiCc3zp:b-81wohb) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_sleep.c#L54)):*
> ```C
> 
>     if (flags) {
>         log_warning("Graphene's clock_nanosleep does not support non-zero flags (0x%x)\n", flags);
> ```

`log_error`?
___
*[LibOS/shim/src/sys/shim_socket.c, line 484 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7slsj9hZRBwo05khZ:-MU7slsj9hZRBwo05kh_:b-81wohb) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_socket.c#L484)):*
> ```C
>     if (!pal_hdl) {
>         ret = (PAL_NATIVE_ERRNO() == PAL_ERROR_STREAMEXIST) ? -EADDRINUSE : -PAL_ERRNO();
>         log_warning("bind: invalid handle returned\n");
> ```

`log_error`?
___
*[LibOS/shim/src/sys/shim_socket.c, line 1509 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7sykF9hdQHn5TuOyF:-MU7sykF9hdQHn5TuOyG:b-81wohb) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_socket.c#L1509)):*
> ```C
>     /* TODO(donporter): timeout properly. For now, explicitly return an error. */
>     if (timeout) {
>         log_warning("recvmmsg(): timeout parameter unsupported.\n");
> ```

`log_error`?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/src/sys/shim_clone.c, line 253 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7rv_KApaXFmTfHEpk:-MU84v2B2sf9Nghx-cFH:b-rwebsa) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_clone.c#L253)):*

Shouldn't it be `log_error`?

Here and below, my assumption is:
- `warning` is used for situations "weird syscall arguments are given, it is application's fault OR Graphene will continue but with caveats"
- `error` is used for situations "unsupported arguments are given, Graphene will fail unexpectedly for application"
</blockquote></details>

In the [documentation about log levels](https://graphene.readthedocs.io/en/latest/manifest-syntax.html#log-level), I think both of these situations fall into `warning`, and `error` is reserved for fatal errors / problems with Graphene itself.

My thinking behind that was that an application could (possibly frequently) request something unsupported, but continue anyway (for instance, fall back to an older mechanism). Because `error` is on by default, so the user will get a lot of noise on stderr.

(As an example, we don't produce `error` on unsupported syscalls, and probably shouldn't?)

So if you want to treat some of these as `error`, I would advise to at least test them on our examples. And it might be good to update the documentation.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/src/sys/shim_clone.c, line 253 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7rv_KApaXFmTfHEpk:-MU887W84-8fG8pkQL2I:b2nwzrj) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_clone.c#L253)):*

In the [documentation about log levels](https://graphene.readthedocs.io/en/latest/manifest-syntax.html#log-level), I think both of these situations fall into `warning`, and `error` is reserved for fatal errors / problems with Graphene itself.

My thinking behind that was that an application could (possibly frequently) request something unsupported, but continue anyway (for instance, fall back to an older mechanism). Because `error` is on by default, so the user will get a lot of noise on stderr.

(As an example, we don't produce `error` on unsupported syscalls, and probably shouldn't?)

So if you want to treat some of these as `error`, I would advise to at least test them on our examples. And it might be good to update the documentation.
</blockquote></details>

Thanks, Pawel, that's true. I'll resolve my comments on this.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: 60 of 62 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @pwmarcz, and @yamahata)

___
*[LibOS/shim/src/fs/sys/cache_info.c, line 47 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7qwZ34F1yQ-WeybLs:-MU8crCK9R_M96qGPqF4:bv12iub) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/fs/sys/cache_info.c#L47)):*

For all these sysfs files, I'd prefer `log_warning` (I can imagine a situation when some app goes bonkers because a sysfs file is not found, and the end user shouldn't scroll through MBs of `debug` logs to find this). But no hard preference.
</blockquote></details>

I thought about it, but it's actually a completely valid case which just returns `ENOENT`. And some apps do try to open such files, e.g. `lscpu` - for indexed entries (like `cache/index<number>`) it just loops until ENOENT. So, it'd rather keep it at debug level.
___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 602 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7qlo6C72ZHTUHjC-s:-MU8dosS5iGiB96_iI77:bwlwmf2) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/ipc/shim_ipc_helper.c#L602)):*

I'd prefer `log_error`. This is a Graphene-internal error which should never happen in normal circumstances.
</blockquote></details>

Right, I overlooked this one.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 255 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7sNoEC79BVGNOs94x:-MU8f-Pf7RLurhTNcWUk:b18ho2g) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_mmap.c#L255)):*

`log_error`? Even though it is application's fault, sounds like something to complain about very loudly.
</blockquote></details>

I thought about this one, but it's not really a fatal error, the application can continue afterwards (with its buggy behavior). OTOH, it's rare and definitely something which the user should see.
@pwmarcz What do you think?
___
*[LibOS/shim/src/sys/shim_socket.c, line 484 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7slsj9hZRBwo05khZ:-MU8g5T_22-3bbAtxCjG:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_socket.c#L484)):*

`log_error`?
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/fs/sys/cache_info.c, line 47 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7qwZ34F1yQ-WeybLs:-MU8lBNz1o8TnIZYUjIK:b2nhktp) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/fs/sys/cache_info.c#L47)):*

I thought about it, but it's actually a completely valid case which just returns `ENOENT`. And some apps do try to open such files, e.g. `lscpu` - for indexed entries (like `cache/index<number>`) it just loops until ENOENT. So, it'd rather keep it at debug level.
</blockquote></details>

OK.
___
*[LibOS/shim/src/sys/shim_mmap.c, line 255 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7sNoEC79BVGNOs94x:-MU8lKdwDEZ7WMLnK8sr:blz3wp9) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_mmap.c#L255)):*

I thought about this one, but it's not really a fatal error, the application can continue afterwards (with its buggy behavior). OTOH, it's rare and definitely something which the user should see.
@pwmarcz What do you think?
</blockquote></details>

OK, resolving from my side.

yamahata (approved):
> Reviewed 1 of 2 files at r4, 5 of 5 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

dimakuv (approved):
> Reviewed 5 of 5 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: :shipit: complete! all files reviewed, all discussions resolved

pwmarcz (commented):
> Reviewed 3 of 5 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2177)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[LibOS/shim/src/sys/shim_mmap.c, line 255 at r3](https://reviewable.io/reviews/oscarlab/graphene/2177#-MU7sNoEC79BVGNOs94x:-MUAIV0SEasuN-GnbkLF:bap5hn) ([raw file](https://github.com/oscarlab/graphene/blob/9f98e7300343b817ca94c738539a304fe9f74d94/LibOS/shim/src/sys/shim_mmap.c#L255)):*

OK, resolving from my side.
</blockquote></details>

I think that by our definition it's `warning` - as I understand, it's recoverable, it doesn't indicate a problem in Graphene, and it's not a security issue.

PR #2191, opened 2021-02-25
[Pal] Fix {READ,WRITE}_ONCE macros

### LINKED ISSUE #2172: Pre-heat feature not working in master
## Description of the problem
Preheat enclave feature in the master is not showing any performance improvements and is a regression compared to the first patch (https://github.com/mkow/graphene/commit/089375d450ce5198406ffe6cc234c975c88223af) where we saw massive performance improvements.

Replacing the code in the master with the patch code, again shows the same performance improvements. 

## Steps to reproduce

Run the 3DRESUNET 1 instance test with GSGX with preheat enabled, disabled and replacing the code from patch:

average_time | std | max | min | epoch | compute type | threads | instances
-- | -- | -- | -- | -- | -- | -- | --
4.47 | 3.44 | 14.77 | 3.23 | 10 | preheat_enabled_manifest (latest master) | 20 | 1
4.43 | 3.44 | 14.74 | 3.23 | 10 | preheat_disabled_manifest (latest master) | 20 | 1
3.4 | 0.45 | 4.75 | 3.24 | 10 | preheat enabled by applying the patch | 20 | 1

## Expected results

Preheat feature should show performance improvements on apps like 3Dresunet

mkow:
## Description of the changes 

As it turns out, our READ_ONCE macro isn't strong enough to convince GCC to always emit a read.

The interesting part is that we [ported that macro from Linux v5.6](https://elixir.bootlin.com/linux/v5.6/source/include/linux/compiler.h#L182), where it seems to work correctly only for sizes 1, 2, 4 and 8 (these special cases were if-ed out there), and not for other sizes, which it also tries to support. For a demo see https://godbolt.org/z/1Gzq9c.

This implementation in Linux was recently [refactored and simplified](https://github.com/torvalds/linux/commit/a5460b5e5fb82656807840d40d3deaecad094044), because Linux dropped support for old, buggy GCC, which required such workarounds. It seems that this particular bug was fixed during the refactoring, without even noticing that it existed.

So far so good, but when I tried to use [the new implementation](https://elixir.bootlin.com/linux/v5.11.1/source/include/asm-generic/rwonce.h#L44), it turned out to be still broken in some cases. But this time it seems to be a bug in GCC - we [reported it](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99258) and GCC developers confirmed that it's a bug.

The bug was noticed because `sgx.preheat_enclave` didn't work on some configurations.

After the changes some type mismatches started to get detected by the compiler, so I also had to fix them.

Fixes #2172.

## How to test this PR? 

Check if `sgx.preheat_enclave` works on your configuration.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)

### DIFF
--- a/Pal/include/lib/api.h
+++ b/Pal/include/lib/api.h
@@ -204,24 +204,12 @@ void* calloc(size_t nmemb, size_t size);
 
 #define COMPILER_BARRIER() ({ __asm__ __volatile__("" ::: "memory"); })
 
-/* Idea taken from: https://elixir.bootlin.com/linux/v5.6/source/include/linux/compiler.h */
-#define READ_ONCE(x)                            \
-    ({                                          \
-        __typeof__(x) _y;                       \
-        COMPILER_BARRIER();                     \
-        __builtin_memcpy(&_y, &(x), sizeof(x)); \
-        COMPILER_BARRIER();                     \
-        _y;                                     \
-    })
+/* We need this artificial assignment in READ_ONCE because of a GCC bug:
+ * https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99258
+ */
+#define READ_ONCE(x) ({ __typeof__(x) y = *(volatile __typeof__(x)*)&(x); y;})
 
-#define WRITE_ONCE(x, y)                        \
-    ({                                          \
-        __typeof__(x) _y = (__typeof__(x))(y);  \
-        COMPILER_BARRIER();                     \
-        __builtin_memcpy(&(x), &_y, sizeof(x)); \
-        COMPILER_BARRIER();                     \
-        _y;                                     \
-    })
+#define WRITE_ONCE(x, y) do { *(volatile __typeof__(x)*)&(x) = (y); } while (0)
 
 /* Libc printf functions. stdio.h/stdarg.h. */
 void fprintfmt(int (*_fputch)(void*, int, void*), void* f, void* putdat, const char* fmt, ...)
--- a/Pal/include/lib/spinlock.h
+++ b/Pal/include/lib/spinlock.h
@@ -24,7 +24,7 @@
 #endif // IN_SHIM
 
 typedef struct {
-    int lock;
+    uint32_t lock;
 #ifdef DEBUG_SPINLOCKS_SHIM
     unsigned int owner;
 #endif // DEBUG_SPINLOCKS_SHIM
@@ -94,7 +94,7 @@ static inline int spinlock_trylock(spinlock_t* lock) {
  * \brief Acquire spinlock.
  */
 static inline void spinlock_lock(spinlock_t* lock) {
-    int val;
+    uint32_t val;
 
     /* First check if lock is already free. */
     if (__atomic_exchange_n(&lock->lock, SPINLOCK_LOCKED, __ATOMIC_ACQUIRE) == SPINLOCK_UNLOCKED) {
@@ -122,7 +122,7 @@ out:
  * \return            0 if acquiring the lock succeeded, 1 if timed out.
  */
 static inline int spinlock_lock_timeout(spinlock_t* lock, unsigned long iterations) {
-    int val;
+    uint32_t val;
 
     /* First check if lock is already free. */
     if (__atomic_exchange_n(&lock->lock, SPINLOCK_LOCKED, __ATOMIC_ACQUIRE) == SPINLOCK_UNLOCKED) {
@@ -157,7 +157,7 @@ out_success:
  * of `*lock` are written into `*expected`. If `desired` is written into `*lock` then true is
  * returned.
  */
-static inline int spinlock_cmpxchg(spinlock_t* lock, int* expected, int desired) {
+static inline int spinlock_cmpxchg(spinlock_t* lock, uint32_t* expected, uint32_t desired) {
     static_assert(SAME_TYPE(&lock->lock, expected), "spinlock is not implemented as int*");
     return __atomic_compare_exchange_n(&lock->lock, expected, desired, /*weak=*/false,
                                        __ATOMIC_ACQUIRE, __ATOMIC_RELAXED);
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -709,7 +709,7 @@ noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char*
     }
     if (preheat_enclave == 1) {
         for (uint8_t* i = g_pal_sec.heap_min; i < (uint8_t*)g_pal_sec.heap_max; i += g_page_size)
-            READ_ONCE(*i);
+            READ_ONCE(*(size_t*)i);
     }
 
     ret = toml_sizestring_in(g_pal_state.manifest_root, "loader.pal_internal_mem_size",
--- a/Pal/src/host/Linux-SGX/db_mutex.c
+++ b/Pal/src/host/Linux-SGX/db_mutex.c
@@ -31,12 +31,12 @@ int _DkMutexCreate(PAL_HANDLE* handle, int initialCount) {
     PAL_HANDLE mut = malloc(HANDLE_SIZE(mutex));
     SET_HANDLE_TYPE(mut, mutex);
     __atomic_store_n(&mut->mutex.mut.nwaiters.counter, 0, __ATOMIC_SEQ_CST);
-    mut->mutex.mut.locked = malloc_untrusted(sizeof(uint32_t));
+    mut->mutex.mut.locked = malloc_untrusted(sizeof(*mut->mutex.mut.locked));
     if (!mut->mutex.mut.locked) {
         free(mut);
         return -PAL_ERROR_NOMEM;
     }
-    *handle                  = mut;
+    *handle = mut;
     __atomic_store_n(mut->mutex.mut.locked, initialCount, __ATOMIC_SEQ_CST);
     return 0;
 }
--- a/Pal/src/host/Linux-SGX/enclave_ocalls.c
+++ b/Pal/src/host/Linux-SGX/enclave_ocalls.c
@@ -95,7 +95,7 @@ static long sgx_exitless_ocall(uint64_t code, void* ms) {
     if (timedout) {
         /* OCALL takes a lot of time, so fallback to waiting on a futex; at this point we exit
          * enclave to perform syscall; this code is based on Mutex 2 from Futexes are Tricky */
-        int c = SPINLOCK_UNLOCKED;
+        uint32_t c = SPINLOCK_UNLOCKED;
 
         /* at this point can be a subtle data race: RPC thread is only now done with OCALL and
          * moved lock in UNLOCKED state; in this racey case, lock = UNLOCKED = 0 and we do not
--- a/Pal/src/host/Linux/pal_host.h
+++ b/Pal/src/host/Linux/pal_host.h
@@ -146,7 +146,7 @@ typedef struct pal_handle {
             PAL_BOL isnotification;
         } event;
     };
-} * PAL_HANDLE;
+}* PAL_HANDLE;
 
 #define RFD(n)   (1 << (MAX_FDS * 0 + (n)))
 #define WFD(n)   (1 << (MAX_FDS * 1 + (n)))


mkow:
Jenkins, retest Jenkins-SGX-apps please (internal Jenkins issues)

mkow:
Jenkins, retest Jenkins-SGX please (internal Jenkins issues)

boryspoplawski (changes requested):
> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[Pal/include/lib/api.h, line 210 at r1](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUPQW0pB8Ptn_bmuLdk:-MUPQW0rEwhfdpOxrEX6:bkrg8gb) ([raw file](https://github.com/oscarlab/graphene/blob/77f3747e18a0f94ce7b0d534abc88b41d2f97157/Pal/include/lib/api.h#L210)):*
> ```objc
>  * https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99258
>  */
> #define READ_ONCE(x) ({ __typeof__(x) y = *(const volatile __typeof__(x)*)&(x); y;})
> ```

Why the `const`? Doesn't seem to be needed.
___
*[Pal/include/lib/api.h, line 212 at r1](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUPQ_0m7IX0zzkFXCbO:-MUPQ_0m7IX0zzkFXCbP:b-awp65u) ([raw file](https://github.com/oscarlab/graphene/blob/77f3747e18a0f94ce7b0d534abc88b41d2f97157/Pal/include/lib/api.h#L212)):*
> ```objc
> #define READ_ONCE(x) ({ __typeof__(x) y = *(const volatile __typeof__(x)*)&(x); y;})
> 
> #define WRITE_ONCE(x, y) do { *(volatile __typeof__(x)*)&(x) = (y); } while (0)
> ```

Why  use`do {} while (0)` here and a statement expression above?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[Pal/src/host/Linux/pal_host.h, line 24 at r1](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUPSGoB2-bI8J94ix0f:-MUPSGoB2-bI8J94ix0g:b1wewom) ([raw file](https://github.com/oscarlab/graphene/blob/9e0ac10856cdb666172d640776024704384e7ced/Pal/src/host/Linux/pal_host.h#L24)):*
> ```objc
>  */
> typedef struct mutex_handle {
>     uint32_t locked;
> ```

Should be `uin32_t` here and everywhere else.
My mistake, sorry (suggested this change in private chat).

mkow (commented):
> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: 0 of 5 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[Pal/include/lib/api.h, line 210 at r1](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUPQW0pB8Ptn_bmuLdk:-MUPXcqQB3otz0Yla4v-:b-wi5pvm) ([raw file](https://github.com/oscarlab/graphene/blob/77f3747e18a0f94ce7b0d534abc88b41d2f97157/Pal/include/lib/api.h#L210)):*

Why the `const`? Doesn't seem to be needed.
</blockquote></details>

Hmm, seems it's actually not needed, removed.
___
*[Pal/include/lib/api.h, line 212 at r1](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUPQ_0m7IX0zzkFXCbO:-MUPXqSXEGutY8SXeyqP:b-lha0xa) ([raw file](https://github.com/oscarlab/graphene/blob/77f3747e18a0f94ce7b0d534abc88b41d2f97157/Pal/include/lib/api.h#L212)):*

Why  use`do {} while (0)` here and a statement expression above?
</blockquote></details>

I wanted to prevent anyone assuming the value of `WRITE_ONCE(x, y)` and using it inside some larger expression.
___
*[Pal/src/host/Linux/pal_host.h, line 24 at r1](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUPSGoB2-bI8J94ix0f:-MUPZkkJ2IaYtssxI-B-:b-w4bvp9) ([raw file](https://github.com/oscarlab/graphene/blob/9e0ac10856cdb666172d640776024704384e7ced/Pal/src/host/Linux/pal_host.h#L24)):*

Should be `uin32_t` here and everywhere else.
My mistake, sorry (suggested this change in private chat).
</blockquote></details>

Heh, ok. Fixed :)

dimakuv (approved):
> Reviewed 5 of 9 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

mkow (commented):
FYI: I rebased it onto the current master to get the fixes for the new CI nodes (#2192).

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191#-:-MUQTBVc-R0YR7EVfKuy:b-pnz0uv)* status: 2 of 5 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @dimakuv)

boryspoplawski (changes requested):
> Reviewed 2 of 9 files at r2, 3 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[Pal/include/lib/api.h, line 212 at r1](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUPQ_0m7IX0zzkFXCbO:-MURAgHD0Stn9uNQmDag:b-tdzo6s) ([raw file](https://github.com/oscarlab/graphene/blob/77f3747e18a0f94ce7b0d534abc88b41d2f97157/Pal/include/lib/api.h#L212)):*

I wanted to prevent anyone assuming the value of `WRITE_ONCE(x, y)` and using it inside some larger expression.
</blockquote></details>

What's wrong it that? I would assume the value is the newly written one (`y`, or `x` after assignment). I.e. I would have `WRITE_ONCE` have the same semantics as `=` in C.

jinengandhi-intel (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MURtuL5Ed4HCuNO-bW-:bpi1agy) (no related file):*
Attached are the performance comparison results from Pre-heat feature fix (PR2191 using READ_ONCE) and applying the WRITE_ONCE/MEMSET on the same PR.

There seems to be a clear difference in terms of the max time as well as the std. deviation values for the READ_ONCE vs. WRITE_ONCE. 

The runs were done on an ICX server running 5.9 rc8 intel next kernel.



---

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @jinengandhi-intel and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUSCoW-8tVF4InjMSV4:b-hh493r) (no related file):*
<details><summary><i>Previously, jinengandhi-intel wrote…</i></summary><blockquote>

Attached are the performance comparison results from Pre-heat feature fix (PR2191 using READ_ONCE) and applying the WRITE_ONCE/MEMSET on the same PR.

There seems to be a clear difference in terms of the max time as well as the std. deviation values for the READ_ONCE vs. WRITE_ONCE. 

The runs were done on an ICX server running 5.9 rc8 intel next kernel.



</blockquote></details>

Thanks @jinengandhi-intel , so the new `READ_ONCE` seems to work fine!

---

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @jinengandhi-intel and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUTWB6MEudTV7EHjKmr:bv0b4iq) (no related file):*

Thanks @jinengandhi-intel , so the new `READ_ONCE` seems to work fine!
</blockquote></details>

But why are `READ_ONCE` and `WRITE_ONCE` timings so much different? In the current version they are actually the same, they should translate to only copy only...

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski and @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUT_h1S-ImUCiUgNJUB:bku7ggp) (no related file):*

But why are `READ_ONCE` and `WRITE_ONCE` timings so much different? In the current version they are actually the same, they should translate to only copy only...
</blockquote></details>

Hmm, writing shouldn't be faster than reading, or rather, in our case, they should be identical (because the time includes only working on the final effect, not the operation itself). I'll take a look at this.

btw. @jinengandhi-intel Please don't paste images, it's not possible to copy the data from them. Markdown has a syntax for tables, please see https://guides.github.com/features/mastering-markdown/.

---
___
*[Pal/include/lib/api.h, line 212 at r1](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUPQ_0m7IX0zzkFXCbO:-MUTZyPZAcHliXONuK--:b-8y01ak) ([raw file](https://github.com/oscarlab/graphene/blob/77f3747e18a0f94ce7b0d534abc88b41d2f97157/Pal/include/lib/api.h#L212)):*

What's wrong it that? I would assume the value is the newly written one (`y`, or `x` after assignment). I.e. I would have `WRITE_ONCE` have the same semantics as `=` in C.
</blockquote></details>

a) I don't want anyone to write monsters like `WRITE_ONCE(a, b) = c`, I prefer to explicitly forbid such constructions. I don't think anything like this would ever pass our review :)
b) It's not clear if the returned lvalue is volatile or not.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion (waiting on @jinengandhi-intel)

___
*[Pal/include/lib/api.h, line 212 at r1](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUPQ_0m7IX0zzkFXCbO:-MUTb8VD6coxF3NcyeW9:bk0c7bm) ([raw file](https://github.com/oscarlab/graphene/blob/77f3747e18a0f94ce7b0d534abc88b41d2f97157/Pal/include/lib/api.h#L212)):*

a) I don't want anyone to write monsters like `WRITE_ONCE(a, b) = c`, I prefer to explicitly forbid such constructions. I don't think anything like this would ever pass our review :)
b) It's not clear if the returned lvalue is volatile or not.
</blockquote></details>

1) Changing to statement expression wouldn't allow that (it would still not compile), but it would enable `c = WRITE_ONCE(a, b)`. Which might be useful, or not, who knows.
2) Imo it should be clear it's not (i.e. it doesn't have to be, but might be). `WRITE_ONCE(c, WRITE_ONCE(a, b))` would be the thing, if you want to have the same semantics on both writes. But I would not expect such thing to be used anywhere.

To sum up: I'm unblocking, as I don't see any usecase for it, but I don't thing that explicitly blocking it is needed.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: 5 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUUcw7CELEC8qHyKeuH:bftw9bh) (no related file):*

Hmm, writing shouldn't be faster than reading, or rather, in our case, they should be identical (because the time includes only working on the final effect, not the operation itself). I'll take a look at this.

btw. @jinengandhi-intel Please don't paste images, it's not possible to copy the data from them. Markdown has a syntax for tables, please see https://guides.github.com/features/mastering-markdown/.
</blockquote></details>

I can't make any sense of these results, but I managed to reproduce them - WRITE_ONCE seems to be slightly faster than READ_ONCE. What's even more troubling, is that the total runtime (i.e. counting the preheating) with WRITE_ONCE is better than READ_ONCE, which makes no sense to me. Anyone has an idea for an explanation of this?

So, I changed the code to use WRITE_ONCE, but I don't know why it's faster.

My results:

| Config              | Total time | Time per epoch |
| :------------------ | ---------: | -------------: |
| no zeroing          |   254.362s |         15.58s |
| memset              |   360.170s |         14.85s |
| READ_ONCE           |   334.066s |         15.02s |
| WRITE_ONCE          |   327.260s |         14.38s |

---
___
*[Pal/include/lib/api.h, line 212 at r1](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUPQ_0m7IX0zzkFXCbO:-MUUcc2WA2F9j6KoiPlc:b-hgk7u9) ([raw file](https://github.com/oscarlab/graphene/blob/77f3747e18a0f94ce7b0d534abc88b41d2f97157/Pal/include/lib/api.h#L212)):*

1) Changing to statement expression wouldn't allow that (it would still not compile), but it would enable `c = WRITE_ONCE(a, b)`. Which might be useful, or not, who knows.
2) Imo it should be clear it's not (i.e. it doesn't have to be, but might be). `WRITE_ONCE(c, WRITE_ONCE(a, b))` would be the thing, if you want to have the same semantics on both writes. But I would not expect such thing to be used anywhere.

To sum up: I'm unblocking, as I don't see any usecase for it, but I don't thing that explicitly blocking it is needed.
</blockquote></details>

I definitely prefer `WRITE_ONCE(a, b); WRITE_ONCE(c, b)` to `WRITE_ONCE(c, WRITE_ONCE(a, b))`.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUVKhey9MW_0UvJ9Mn7:bf8sg7k) (no related file):*

I can't make any sense of these results, but I managed to reproduce them - WRITE_ONCE seems to be slightly faster than READ_ONCE. What's even more troubling, is that the total runtime (i.e. counting the preheating) with WRITE_ONCE is better than READ_ONCE, which makes no sense to me. Anyone has an idea for an explanation of this?

So, I changed the code to use WRITE_ONCE, but I don't know why it's faster.

My results:

| Config              | Total time | Time per epoch |
| :------------------ | ---------: | -------------: |
| no zeroing          |   254.362s |         15.58s |
| memset              |   360.170s |         14.85s |
| READ_ONCE           |   334.066s |         15.02s |
| WRITE_ONCE          |   327.260s |         14.38s |
</blockquote></details>

Did you check what code does the compiler output in both version? What's the actual difference? They both do one write, so there should be very similar...

---
___
*[Pal/include/lib/api.h, line 212 at r1](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUPQ_0m7IX0zzkFXCbO:-MUVKvg50VU-Q1YcybsA:be2hrvq) ([raw file](https://github.com/oscarlab/graphene/blob/77f3747e18a0f94ce7b0d534abc88b41d2f97157/Pal/include/lib/api.h#L212)):*

I definitely prefer `WRITE_ONCE(a, b); WRITE_ONCE(c, b)` to `WRITE_ONCE(c, WRITE_ONCE(a, b))`.
</blockquote></details>

So do I, I never said it's what we want to use, it was just an answer to your point about return value being volatile or not.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: 5 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski and @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUVVH8265FjH638wvKD:bew4yww) (no related file):*
> They both do one write

I guess you meant "access", not "write".

I verified the disasm, looks ok (write just writes 0, read just reads and writes to stack). The additional write to the stack on read may have caused a slight overhead in the total time, but it should be negligible... (as this line will be always in cache).

I also noticed that I previously built Graphene in debug mode, so I did another round of tests in release. Additionally, I switched to writing whole machine words, which is faster than a single byte.

Updated results: (each entry is an average of 3 runs, each 5 epochs)

Debug:

| Config              | Total time | Time per epoch |
| :------------------ | ---------: | -------------: |
| no zeroing          |    254.36s |         15.58s |
| memset              |    360.17s |         14.85s |
| READ_ONCE           |    334.06s |         15.02s |
| WRITE_ONCE          |    327.26s |         14.38s |
| READ_ONCE long      |    324.33s |         13.12s |
| WRITE_ONCE long     |    322.54s |         12.39s |

Release:

| Config              | Total time | Time per epoch |
| :------------------ | ---------: | -------------: |
| no zeroing          |    184.09s |         14.78s |
| memset              |    300.70s |         16.70s |
| READ_ONCE byte      |    272.66s |         15.91s |
| WRITE_ONCE byte     |    269.32s |         15.85s |
| READ_ONCE size_t    |    261.43s |         14.06s |
| WRITE_ONCE size_t   |    256.97s |         13.21s |

The variance was quite huge and the results may not be conclusive (with release, memset seems to give no improvement, and it wasn't optimized out, so I'd blame variance).

@jinengandhi-intel could you verify if WRITE_ONCE with size_t also gives the best results for you?

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUYlPE275rLwth_ndrY:b-dei4vc) (no related file):*

> They both do one write

I guess you meant "access", not "write".

I verified the disasm, looks ok (write just writes 0, read just reads and writes to stack). The additional write to the stack on read may have caused a slight overhead in the total time, but it should be negligible... (as this line will be always in cache).

I also noticed that I previously built Graphene in debug mode, so I did another round of tests in release. Additionally, I switched to writing whole machine words, which is faster than a single byte.

Updated results: (each entry is an average of 3 runs, each 5 epochs)

Debug:

| Config              | Total time | Time per epoch |
| :------------------ | ---------: | -------------: |
| no zeroing          |    254.36s |         15.58s |
| memset              |    360.17s |         14.85s |
| READ_ONCE           |    334.06s |         15.02s |
| WRITE_ONCE          |    327.26s |         14.38s |
| READ_ONCE long      |    324.33s |         13.12s |
| WRITE_ONCE long     |    322.54s |         12.39s |

Release:

| Config              | Total time | Time per epoch |
| :------------------ | ---------: | -------------: |
| no zeroing          |    184.09s |         14.78s |
| memset              |    300.70s |         16.70s |
| READ_ONCE byte      |    272.66s |         15.91s |
| WRITE_ONCE byte     |    269.32s |         15.85s |
| READ_ONCE size_t    |    261.43s |         14.06s |
| WRITE_ONCE size_t   |    256.97s |         13.21s |

The variance was quite huge and the results may not be conclusive (with release, memset seems to give no improvement, and it wasn't optimized out, so I'd blame variance).

@jinengandhi-intel could you verify if WRITE_ONCE with size_t also gives the best results for you?
</blockquote></details>

Could you run the test with `READ_ONCE` temporarily overwritten as:
```
static inline size_t read_once_tmp(volatile size_t* x) {
    return *x;
}
```
(just in the preheating)? As you pointed out the additional write to stack in `READ_ONCE` should be negligible, but for some reason we see a speed-up on `WRITE_ONCE`...

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski and @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUYzJ7w5XYKh9TZmRsl:b-r6qj5r) (no related file):*

Could you run the test with `READ_ONCE` temporarily overwritten as:
```
static inline size_t read_once_tmp(volatile size_t* x) {
    return *x;
}
```
(just in the preheating)? As you pointed out the additional write to stack in `READ_ONCE` should be negligible, but for some reason we see a speed-up on `WRITE_ONCE`...
</blockquote></details>

I got 257.98s total and 13.03s per epoch. But I have very low confidence in these values, the variance is huge (see Jinen's table), so maybe what we see is just randomness.
I don't have a good idea how to stabilize the results without spending significant amount of time on analysis of this particular workload...

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski and @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUZ1iJ18l0L2_CANIFn:bi4daac) (no related file):*

I got 257.98s total and 13.03s per epoch. But I have very low confidence in these values, the variance is huge (see Jinen's table), so maybe what we see is just randomness.
I don't have a good idea how to stabilize the results without spending significant amount of time on analysis of this particular workload...
</blockquote></details>

I checked the assembly, and both reading versions actually produce identical loops on release builds. The original check I did for READ_ONCE was on a debug build, sorry, my mistake :)

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUZWFnCCGhfMGomSsT3:b-cl8peb) (no related file):*

I checked the assembly, and both reading versions actually produce identical loops on release builds. The original check I did for READ_ONCE was on a debug build, sorry, my mistake :)
</blockquote></details>

Since the results were inconclusive and the last run of read was the fastest, I propose changing the preheating back to `READ_ONCE`.

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: 5 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @boryspoplawski and @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUZkVHYAdDQEncjLp_z:bbczm3z) (no related file):*

Since the results were inconclusive and the last run of read was the fastest, I propose changing the preheating back to `READ_ONCE`.
</blockquote></details>

Makes sense, done. Let's keep it this way until we have stronger data for the unintuitive variant (i.e. WRITE_ONCE).

---

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @jinengandhi-intel)

jinengandhi-intel (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "squash! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUfxmpL1wT5CNkTRj_O:b-iz9o37) (no related file):*

Makes sense, done. Let's keep it this way until we have stronger data for the unintuitive variant (i.e. WRITE_ONCE).
</blockquote></details>

@mkow Please find the results for READ_ONCE and WRITE_ONCE size_t. 

From the std.dev and the max times, I am seeing that the WRITE_ONCE numbers are much more consistent than READ_ONCE. The lowest numbers for READ_ONCE do match the WRITE_ONCE numbers but as I said it is very inconsistent and most of the time has an higher overhead than the WRITE_ONCE.

Data | Pre-heat fix (PR 2191 using READ_ONCE size_t) | WRITE_ONCE size_t
------------ | ------------- | -------------
Run1
Avg time|5.05 |4.42
Max time |8.53 |6.39
Min time |3.74 |3.67
Std. Dev |1.79 |1.06
 |  | 
Run2
Avg time | 4.35 |4.45
Max time|6.82|6.85
Min time|3.7|3.7
Std. Dev|1.24|1.22
 |  | 
Run3
Avg time|4.64|4.41
Max time|7.83|6.63
Min time|3.7|3.68
Std. Dev|1.61|1.14
 |  | 
Avg. time for 3 runs|4.68|4.42667

---

dimakuv (approved):
> Reviewed 3 of 4 files at r3, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion, "squash! " found in commit messages' one-liners (waiting on @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUhAz4E8joCwFy0YBhh:bh93o0o) (no related file):*
<details><summary><i>Previously, jinengandhi-intel wrote…</i></summary><blockquote>

@mkow Please find the results for READ_ONCE and WRITE_ONCE size_t. 

From the std.dev and the max times, I am seeing that the WRITE_ONCE numbers are much more consistent than READ_ONCE. The lowest numbers for READ_ONCE do match the WRITE_ONCE numbers but as I said it is very inconsistent and most of the time has an higher overhead than the WRITE_ONCE.

Data | Pre-heat fix (PR 2191 using READ_ONCE size_t) | WRITE_ONCE size_t
------------ | ------------- | -------------
Run1
Avg time|5.05 |4.42
Max time |8.53 |6.39
Min time |3.74 |3.67
Std. Dev |1.79 |1.06
 |  | 
Run2
Avg time | 4.35 |4.45
Max time|6.82|6.85
Min time|3.7|3.7
Std. Dev|1.24|1.22
 |  | 
Run3
Avg time|4.64|4.41
Max time|7.83|6.63
Min time|3.7|3.68
Std. Dev|1.61|1.14
 |  | 
Avg. time for 3 runs|4.68|4.42667
</blockquote></details>

I'm fine with both READ_ONCE and WRITE_ONCE. It looks like the difference is quite significant (WRITE_ONCE is better most of the time), but it is hard to explain (and may be a quirk of this particular CPU model). I would stick with the current intuitive code of using READ_ONCE.

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion, "squash! " found in commit messages' one-liners (waiting on @jinengandhi-intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MURtuL5Ed4HCuNO-bVz:-MUhJg9m6y1FrTfnGCg4:b4gpyzr) (no related file):*

I'm fine with both READ_ONCE and WRITE_ONCE. It looks like the difference is quite significant (WRITE_ONCE is better most of the time), but it is hard to explain (and may be a quirk of this particular CPU model). I would stick with the current intuitive code of using READ_ONCE.
</blockquote></details>

@jinengandhi-intel I thought the same, but then did this test again and got READ_ONCE significantly faster than WRITE_ONCE (see my reply to Borys). For now I'd merge the current version, as it has better theoretical foundations, and switch to WRITE_ONCE if more extensive tests (e.g. 50 or 100 runs) would show that it's indeed faster. For now the variance is so high that it may be just randomness.

---

mkow:
Jenkins, retest Jenkins-SGX-18.04 please (glibc build failed with `sed: -e expression #1, char 97: unknown option to `s'`, weird, but unrelated)

mkow:
Jenkins, retest Jenkins-SGX-18.04-apps please (landed on a Jenkins node with broken SGX, I disabled it and pinged @woju about this)

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUhjq6x1gzMLKvJpysQ:-MUhjq6zDyayqyXJ5Jdo:b-hzvrfv) (no related file):*
2nd commit has `Write a whole (...)` - it should be `Read a whole (...)`

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2191#-MUhjq6x1gzMLKvJpysQ:-MUhkckN7fNqnxFesEbh:bxkgfz6) (no related file):*

2nd commit has `Write a whole (...)` - it should be `Read a whole (...)`
</blockquote></details>

Oops, nice catch, fixed.

---

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

mkow:
Jenkins, retest Jenkins-SGX-18.04 please (glibc build failed (again) with sed: -e expression #1, char 97: unknown option to s'`, I have no idea why is this happening)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2191)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2207, opened 2021-03-04
[Pal/Linux-SGX] Fallback to calculating TSC freq from Processor freq

dimakuv:
## Description of the changes 

Some Intel CPUs return ECX == 0 on CPUID 15H (TSC and Nominal Core Crystal Clock Information Leaf). Previously, Linux-SGX PAL considered this as "no support for invariant TSC" and fell back to issuing `ocall_gettime()`, which hurts SGX performance. In reality, Linux and other software devised a workaround by querying CPUID 16H (Processor Frequency Information Leaf) and calculating the TSC frequency via Processor Base Frequency. This PR adds this logic to Graphene and thus improves SGX performance.

For more context, see:
- https://github.com/torvalds/linux/blob/v5.11/arch/x86/kernel/tsc.c#L621
- https://bugzilla.kernel.org/show_bug.cgi?id=197299
- https://edk2.groups.io/g/devel/message/45750
- https://github.com/acidanthera/bugtracker/issues/448

## How to test this PR? 

Performance improvement on some Intel CPUs. We've seen significant performance boost with this PR (Graphene-SGX overhead decreases from 20% to 5%).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_exception.c
+++ b/Pal/src/host/Linux-SGX/db_exception.c
@@ -114,6 +114,9 @@ static void save_pal_context(PAL_CONTEXT* ctx, sgx_cpu_context_t* uc,
 static void emulate_rdtsc_and_print_warning(sgx_cpu_context_t* uc) {
     static int first = 0;
     if (__atomic_exchange_n(&first, 1, __ATOMIC_RELAXED) == 0) {
+        /* if we end up emulating RDTSC/RDTSCP instruction, we cannot use invariant TSC */
+        extern uint64_t g_tsc_hz;
+        g_tsc_hz = 0;
         log_warning("Warning: all RDTSC/RDTSCP instructions are emulated (imprecisely) via "
                     "gettime() syscall.\n");
     }
--- a/Pal/src/host/Linux-SGX/db_main-x86_64.c
+++ b/Pal/src/host/Linux-SGX/db_main-x86_64.c
@@ -31,6 +31,7 @@
 
 #define CPUID_LEAF_INVARIANT_TSC 0x80000007
 #define CPUID_LEAF_TSC_FREQ 0x15
+#define CPUID_LEAF_PROC_FREQ 0x16
 
 static const char* const g_cpu_flags[] = {
     "fpu",    // "x87 FPU on chip"
@@ -95,24 +96,40 @@ static double get_bogomips(void) {
 
 bool is_tsc_usable(void) {
     uint32_t words[PAL_CPUID_WORD_NUM];
-
     _DkCpuIdRetrieve(CPUID_LEAF_INVARIANT_TSC, 0, words);
     return words[PAL_CPUID_WORD_EDX] & 1 << 8;
 }
 
+/* return TSC frequency or 0 if invariant TSC is not supported */
 uint64_t get_tsc_hz(void) {
     uint32_t words[PAL_CPUID_WORD_NUM];
-    uint64_t crys_hz;
 
     _DkCpuIdRetrieve(CPUID_LEAF_TSC_FREQ, 0, words);
-    if (words[PAL_CPUID_WORD_EBX] > 0 && words[PAL_CPUID_WORD_EAX] > 0) {
-        /* nominal frequency of the core crystal clock in kHz */
-        crys_hz = words[PAL_CPUID_WORD_ECX];
-        if (crys_hz > 0) {
-            return crys_hz * words[PAL_CPUID_WORD_EBX] / words[PAL_CPUID_WORD_EAX];
-        }
+    if (!words[PAL_CPUID_WORD_EAX] || !words[PAL_CPUID_WORD_EBX]) {
+        /* TSC/core crystal clock ratio is not enumerated, can't use RDTSC for accurate time */
+        return 0;
     }
-    return 0;
+
+    if (words[PAL_CPUID_WORD_ECX] > 0) {
+        /* calculate TSC frequency as core crystal clock frequency (EAX) * EBX / EAX; cast to 64-bit
+         * first to prevent integer overflow */
+        uint64_t ecx_hz = words[PAL_CPUID_WORD_ECX];
+        return ecx_hz * words[PAL_CPUID_WORD_EBX] / words[PAL_CPUID_WORD_EAX];
+    }
+
+    /* some Intel CPUs do not report nominal frequency of crystal clock, let's calculate it
+     * based on Processor Frequency Information Leaf (CPUID 16H); this leaf always exists if
+     * TSC Frequency Leaf exists; logic is taken from Linux 5.11's arch/x86/kernel/tsc.c */
+    _DkCpuIdRetrieve(CPUID_LEAF_PROC_FREQ, 0, words);
+    if (!words[PAL_CPUID_WORD_EAX]) {
+        /* processor base frequency (in MHz) is not enumerated, can't calculate frequency */
+        return 0;
+    }
+
+    /* processor base frequency is in MHz but we need to return TSC frequency in Hz; cast to 64-bit
+     * first to prevent integer overflow */
+    uint64_t base_frequency_mhz = words[PAL_CPUID_WORD_EAX];
+    return base_frequency_mhz * 1000000;
 }
 
 int _DkGetCPUInfo(PAL_CPU_INFO* ci) {
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -625,7 +625,6 @@ noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char*
     init_enclave_key();
 
     init_cpuid();
-    init_tsc();
 
     /* now we can add a link map for PAL itself */
     setup_pal_map(&g_pal_map);
@@ -658,6 +657,14 @@ noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char*
     g_linux_state.process_id = g_pal_sec.pid;
 
     SET_ENCLAVE_TLS(ready_for_exceptions, 1UL);
+
+    /* initialize "Invariant TSC" HW feature for fast and accurate gettime and immediately probe
+     * RDTSC instruction inside SGX enclave (via dummy get_tsc) -- it is possible that
+     * the CPU supports invariant TSC but doesn't support executing RDTSC inside SGX enclave, in
+     * this case the SIGILL exception is generated and leads to emulate_rdtsc_and_print_warning()
+     * which unsets invariant TSC, and we end up falling back to the slower ocall_gettime() */
+    init_tsc();
+    (void)get_tsc(); /* must be after `ready_for_exceptions=1` since it may generate SIGILL */
 
     /* Now that enclave memory is set up, parse and store host topology info into g_pal_sec struct */
     ret = parse_host_topo_info(&sec_info);
--- a/Pal/src/host/Linux-SGX/db_misc.c
+++ b/Pal/src/host/Linux-SGX/db_misc.c
@@ -25,7 +25,7 @@
 
 #define TSC_REFINE_INIT_TIMEOUT_USECS 10000000
 
-static uint64_t g_tsc_hz = 0;
+uint64_t g_tsc_hz = 0; /* TSC frequency for fast and accurate time ("invariant TSC" HW feature) */
 static uint64_t g_start_tsc = 0;
 static uint64_t g_start_usec = 0;
 static PAL_LOCK g_tsc_lock = LOCK_INIT;


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: 0 of 4 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUxXaIs5-fu3MP-T-Wu:-MUxXaIs5-fu3MP-T-Wv:b-uemaat) (no related file):*
It took me a while to realize that my initial versions of this PR were wrong. So I'm explaining below, for other reviewers and for future reference.

It is important to understand that there are two independent HW features at play here:
1. Invariant TSC (means that executing the RDTSC instruction and adjusting it to the CPU-specific TSC frequency results in correct estimation of elapsed time in microseconds)
2. RDTSC instruction enabled in SGX enclaves (older Intel CPUs sigill on RDTSC, newer Intel CPUs allow RDTSC)

These two features can exist in all four combinations:
1. Invariant TSC -- no, RDTSC in SGX -- no: trivial, must fall back to slow `ocall_gettime()`
2. Invariant TSC -- no, RDTSC in SGX -- yes: using RDTSC inside the SGX enclave doesn't give an accurate estimate of elapsed time, no sense in using it for gettime, so fall back to slow `ocall_gettime()`
3. Invariant TSC -- yes, RDTSC in SGX -- no: tricky, we can use RDTSC for accurate estimate of elapsed time, but **only outside of SGX enclaves**, so we can't do anything but fall back to slow `ocall_gettime()`
4. Invariant TSC -- yes, RDTSC in SGX -- yes: trivial, use RDTSC with adjustment for gettime (gives performance boost!)

Another interesting aspect is that there is a CPUID bit to check whether Invariant TSC exists (and we use it in Graphene-SGX). But there is no CPUID bit to check (or indeed any other indication) whether RDTSC is allowed in SGX enclaves. So for RDTSC check, we have no other way but to probe the RDTSC instruction -- if the instruction results in SIGILL, we know that we have variant 3, and we should fall back to slow `ocall_gettime()`.

---
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 128 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUxZWc39Qyx8-aF6EzA:-MUxZWc39Qyx8-aF6EzB:b-ye8p91) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_main-x86_64.c#L128)):*
> ```C
> 
>     /* processor base frequency is in MHz but we need to return TSC frequency in Hz */
>     return words[PAL_CPUID_WORD_EAX] * 1000000;
> ```

FYI: Linux code does some acrobatics with adjusting the `crystal_khz` variable but all that arithmetic results in negating each other. See https://github.com/torvalds/linux/blob/master/arch/x86/kernel/tsc.c#L668 and https://github.com/torvalds/linux/blob/master/arch/x86/kernel/tsc.c#L692.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski and @dimakuv)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 669 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUyV2hZE3hAGYa1nrIz:-MUyV2h_6d-LbIHZM13y:b99qzav) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_main.c#L669)):*
> ```C
> 
>     uint64_t dummy_time;
>     ret = _DkSystemTimeQuery(&dummy_time);
> ```

Could we move this check into `is_tsc_usable`? Then we could also remove zeroing `g_tsc_hz` in `emulate_rdtsc...`

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 669 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUyV2hZE3hAGYa1nrIz:-MUyXGd92mmABko-mKJq:b5aqc7s) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_main.c#L669)):*

Could we move this check into `is_tsc_usable`? Then we could also remove zeroing `g_tsc_hz` in `emulate_rdtsc...`
</blockquote></details>

I can move the check, but how can we remove the `g_tsc_hz` unsetting from `emulate_rdtsc...`? If we remove that code, we need a way to tell our normal context (where `init_tsc()` happens) that there was a SIGILL exception leading to `emulate_rdtsc...` execution. Is there a way to detect this in Graphene currently?

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 669 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUyV2hZE3hAGYa1nrIz:-MUy_ZpT3BBupKctRW5V:bczop7j) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_main.c#L669)):*

I can move the check, but how can we remove the `g_tsc_hz` unsetting from `emulate_rdtsc...`? If we remove that code, we need a way to tell our normal context (where `init_tsc()` happens) that there was a SIGILL exception leading to `emulate_rdtsc...` execution. Is there a way to detect this in Graphene currently?
</blockquote></details>

If `is_tsc_usable` returns `false`, then `get_tsc_hz` is never called and hence the `g_tcs_hz` is always `0`.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 669 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUyV2hZE3hAGYa1nrIz:-MUybF0J1qKQ-wFUlAv9:bu63926) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_main.c#L669)):*

If `is_tsc_usable` returns `false`, then `get_tsc_hz` is never called and hence the `g_tcs_hz` is always `0`.
</blockquote></details>

But what if `is_tsc_usable == true` but RDTSC-in-SGX is not supported by this CPU? I still don't understand how you suggest to handle this case (we need to set `g_tcs_hz = 0` in this case).

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[Pal/src/host/Linux-SGX/db_main.c, line 669 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUyV2hZE3hAGYa1nrIz:-MUydcQ-92wR9JUO0QU-:b-pgkqmm) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_main.c#L669)):*

But what if `is_tsc_usable == true` but RDTSC-in-SGX is not supported by this CPU? I still don't understand how you suggest to handle this case (we need to set `g_tcs_hz = 0` in this case).
</blockquote></details>

Ah, sorry, for some reason I thought we get that rdtsc is not supported in enclave from the return value. Ignore this comment.

yamahata (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: all files reviewed, 2 unresolved discussions (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/db_exception.c, line 119 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUzabLwDvRbxvg2zcOR:-MUzabLwDvRbxvg2zcOS:bhg6ukr) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_exception.c#L119)):*
> ```C
>         /* if we end up emulating RDTSC/RDTSCP instruction, we cannot use invariant TSC */
>         extern uint64_t g_tsc_hz;
>         __atomic_store_n(&g_tsc_hz, 0, __ATOMIC_SEQ_CST);
> ```

non blocking: plan g_tsc_hz = 0.
Now we know we have single thread when issuing first rdtsc instruction.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 669 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUzaP_52-2Rgv8jxbfb:-MUzaP_52-2Rgv8jxbfc:blua5il) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_main.c#L669)):*
> ```C
> 
>     uint64_t dummy_time;
>     ret = _DkSystemTimeQuery(&dummy_time);
> ```

just get_tsc() for single rdtsc intruction.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: all files reviewed, 4 unresolved discussions (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUxXaIs5-fu3MP-T-Wu:-MUzfUroCm0vVJlSNzmL:buwt4c) (no related file):*

It took me a while to realize that my initial versions of this PR were wrong. So I'm explaining below, for other reviewers and for future reference.

It is important to understand that there are two independent HW features at play here:
1. Invariant TSC (means that executing the RDTSC instruction and adjusting it to the CPU-specific TSC frequency results in correct estimation of elapsed time in microseconds)
2. RDTSC instruction enabled in SGX enclaves (older Intel CPUs sigill on RDTSC, newer Intel CPUs allow RDTSC)

These two features can exist in all four combinations:
1. Invariant TSC -- no, RDTSC in SGX -- no: trivial, must fall back to slow `ocall_gettime()`
2. Invariant TSC -- no, RDTSC in SGX -- yes: using RDTSC inside the SGX enclave doesn't give an accurate estimate of elapsed time, no sense in using it for gettime, so fall back to slow `ocall_gettime()`
3. Invariant TSC -- yes, RDTSC in SGX -- no: tricky, we can use RDTSC for accurate estimate of elapsed time, but **only outside of SGX enclaves**, so we can't do anything but fall back to slow `ocall_gettime()`
4. Invariant TSC -- yes, RDTSC in SGX -- yes: trivial, use RDTSC with adjustment for gettime (gives performance boost!)

Another interesting aspect is that there is a CPUID bit to check whether Invariant TSC exists (and we use it in Graphene-SGX). But there is no CPUID bit to check (or indeed any other indication) whether RDTSC is allowed in SGX enclaves. So for RDTSC check, we have no other way but to probe the RDTSC instruction -- if the instruction results in SIGILL, we know that we have variant 3, and we should fall back to slow `ocall_gettime()`.
</blockquote></details>

So, this PR actually fixes two issues:
* We previously assumed that Invariant TSC implies RDTSC in SGX, which isn't true.
* Adds a fallback calculations if `Nominal Core Crystal Clock Information Leaf` is not available.

Do I understand it correctly?

---
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 109 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUzgD6eCwhMas78qtVZ:-MUzgD6eCwhMas78qtV_:bykva2v) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_main-x86_64.c#L109)):*
> ```
> no TSC/core crystal clock ratio is not enumerated
> ```

Something went wrong here I think :)
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 128 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUzgXah9rQoYN05PXRI:-MUzgXah9rQoYN05PXRJ:bdxd2do) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_main-x86_64.c#L128)):*
> ```C
> 
>     /* processor base frequency is in MHz but we need to return TSC frequency in Hz */
>     return words[PAL_CPUID_WORD_EAX] * 1000000;
> ```

Won't this just overflow before being extended into `uint64_t` and returned? Seems we need to test things also on 5 GHz CPUs :)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: 1 of 4 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUxXaIs5-fu3MP-T-Wu:-MV0IRLa2VaaMCp7y-1F:b-63ilrs) (no related file):*

So, this PR actually fixes two issues:
* We previously assumed that Invariant TSC implies RDTSC in SGX, which isn't true.
* Adds a fallback calculations if `Nominal Core Crystal Clock Information Leaf` is not available.

Do I understand it correctly?
</blockquote></details>

Yes, this is a correct. Should I augment the commit message? If you want it during final rebase, please give a suggestion how to change the commit message.

---
___
*[Pal/src/host/Linux-SGX/db_exception.c, line 119 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUzabLwDvRbxvg2zcOR:-MV0J0a62rg1AYfvfOES:baalanl) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_exception.c#L119)):*

non blocking: plan g_tsc_hz = 0.
Now we know we have single thread when issuing first rdtsc instruction.
</blockquote></details>

Done, this is true.
___
*[Pal/src/host/Linux-SGX/db_main.c, line 669 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUzaP_52-2Rgv8jxbfb:-MV0K9JB5vWRTL6b9G5N:b-h369k) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_main.c#L669)):*

just get_tsc() for single rdtsc intruction.
</blockquote></details>

Done, good point
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 109 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUzgD6eCwhMas78qtVZ:-MV0KHL93q1E4QCOjYuM:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_main-x86_64.c#L109)):*

> ```
> no TSC/core crystal clock ratio is not enumerated
> ```

Something went wrong here I think :)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_main-x86_64.c, line 128 at r1](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUzgXah9rQoYN05PXRI:-MV0LYor01cPKDAHxIKm:b-shw9i0) ([raw file](https://github.com/oscarlab/graphene/blob/205f05de335b2f2641a6d65a54cf831e1a617a53/Pal/src/host/Linux-SGX/db_main-x86_64.c#L128)):*

Won't this just overflow before being extended into `uint64_t` and returned? Seems we need to test things also on 5 GHz CPUs :)
</blockquote></details>

Done. Oops, that would be a fail.

boryspoplawski (dismissed):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

yamahata (approved):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (approved):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUxXaIs5-fu3MP-T-Wu:-MV3gVbE5-kyn1WdoNYl:b2j7rjf) (no related file):*

Yes, this is a correct. Should I augment the commit message? If you want it during final rebase, please give a suggestion how to change the commit message.
</blockquote></details>

I'm not sure, if I see correctly, this whole description misses the first point entirely. Maybe just add it there?

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2207#-MUxXaIs5-fu3MP-T-Wu:-MVFteQsANu3HKijKi58:b-896fix) (no related file):*

I'm not sure, if I see correctly, this whole description misses the first point entirely. Maybe just add it there?
</blockquote></details>

Done.

---

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2207)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2210, opened 2021-03-05
[Pal/Linux-SGX] sgx_sign.py: Use proper TOML parser for manifests

### LINKED ISSUE #2205: Add recursive directory scan to trusted files
## Description of the problem

It should be possible to add whole directories using `sgx.trusted_files` syntax. Right now all files need to be listed manually, which is very PITA and causes problems if the precise list varies between platforms.

Resolving this will greatly simplify our manifests and make them more portable (e.g. between Ubuntu versions).

mkow:
## Description of the changes 

This PR is a preparation for another one, which will resolve #2205.

The changes are:
* Changing sgx_sign.py to use a proper TOML parsing library, not some handcrafted string shuffling.
* Fixing a small bug in trusted files handling, where a missing/non-matching `sgx.trusted_checksum` entry resulted in treating the file as allowed.

## How to test this PR? 

CI should be enough. For the `sgx.trusted_checksum` issue you need to modify `sgx_sign.py` to make it generate invalid entries.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)


woju (changes requested):
> Reviewed 10 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: 10 of 11 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[.pylintrc, line 326 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3WynCE0wqX1spYobD:-MV3WynCE0wqX1spYobE:b-mrpjjn) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/.pylintrc#L326)):*
> ```
> good-names=
>     e,
>     f,
> ```

Please no. If you use `f` in `with open() as f:`, it's also OK to use `... as file:`. `e` in `except ... as e` can also be `exc`.
___
*[python/graphenelibos/sgx_sign.py, line 122 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3U3po666e8mF8N8ML:-MV3U3ppE_wSPZsmmTSZ:bz9wy4n) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/python/graphenelibos/sgx_sign.py#L122)):*
> ```Python
>         attributes.add('FLAG_MODE64BIT')
> 
>     for opt, flag in manifest_options:
> ```

If `manifest_options` remained a `dict` (right now it's a `set` is this intended?), you could do `for opt, flag in manifest_options.items()`. Both are OK.
___
*[python/graphenelibos/sgx_sign.py, line 145 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3Uuk96PK8lRkA0wTb:-MV3Uuk96PK8lRkA0wTc:b5esh5t) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/python/graphenelibos/sgx_sign.py#L145)):*
> ```Python
> def resolve_uri(uri, check_exist=True):
>     if not uri.startswith('file:'):
>         raise RuntimeError('Unsupported URI type: ' + uri)
> ```

Note to self: after retiring Ubuntu 16.04, convert this to f-strings.
___
*[python/graphenelibos/sgx_sign.py, line 660 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3Es5V7Awwn4OfdgGP:-MV3Es5W9jCJF963zFdQ:bnd846p) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/python/graphenelibos/sgx_sign.py#L660)):*
> ```Python
> 
>     manifest.setdefault('sgx', {})
>     sgx = manifest['sgx']
> ```

`setdefault()` returns whatever is the result, so `sgx = manifest.setdefault(...)` is enough. Same below. 
___
*[python/graphenelibos/sgx_sign.py, line 798 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3W4KeCi5q-b3450Xm:-MV3W4KeCi5q-b3450Xn:b-xs0g4r) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/python/graphenelibos/sgx_sign.py#L798)):*
> ```Python
>         manifest = read_manifest(manifest_path)
>     except toml.TomlDecodeError as e:
>         print('Parsing %s as TOML failed: %s' % (manifest_path, e), file=stderr)
> ```

`{}`.format(), plz. Or defer this for 1 more month and convert to f-strings after retiring Ubuntu 16.04.

yamahata (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[python/graphenelibos/sgx_sign.py, line 176 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV4FocJAGObR7Z4F6fS:-MV4FocJAGObR7Z4F6fT:bdzprwu) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/python/graphenelibos/sgx_sign.py#L176)):*
> ```
> _hash
> ```

nitpick: typically hash_ (trailing undercore) is used to avoid name clash with builtin name.
non-blocking.
___
*[python/graphenelibos/sgx_sign.py, line 726 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV4GPuZ9RoKVzIr_AHM:-MV4GPuZ9RoKVzIr_AHN:b75mxki) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/python/graphenelibos/sgx_sign.py#L726)):*
> ```
> _hash
> ```

ditto.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: 2 of 11 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju and @yamahata)

___
*[.pylintrc, line 326 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3WynCE0wqX1spYobD:-MV8AmY96zkQpBhZf41n:bxsd6d4) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/.pylintrc#L326)):*

Please no. If you use `f` in `with open() as f:`, it's also OK to use `... as file:`. `e` in `except ... as e` can also be `exc`.
</blockquote></details>

Honestly, why? `with open() as f:` is quite a popular idiom with very short `with` blocks, same as `i` in loops and `e` for exceptions.
This rule seems much too picky to me.
___
*[python/graphenelibos/sgx_sign.py, line 122 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3U3po666e8mF8N8ML:-MV8AOlC0QBr6rMPqnEZ:b-ptl766) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/python/graphenelibos/sgx_sign.py#L122)):*

If `manifest_options` remained a `dict` (right now it's a `set` is this intended?), you could do `for opt, flag in manifest_options.items()`. Both are OK.
</blockquote></details>

Oh, sorry, that was supposed to be just a list (list seems more proper here to me, as it just lists arguments for consecutive iterations).
___
*[python/graphenelibos/sgx_sign.py, line 176 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV4FocJAGObR7Z4F6fS:-MV89UKLCLbcoYrKr4U2:bohdn7b) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/python/graphenelibos/sgx_sign.py#L176)):*

> ```
> _hash
> ```

nitpick: typically hash_ (trailing undercore) is used to avoid name clash with builtin name.
non-blocking.
</blockquote></details>

oh, right, I misremembered which side to put it, thanks!
___
*[python/graphenelibos/sgx_sign.py, line 660 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3Es5V7Awwn4OfdgGP:-MV8BU_FE919mu5IS3v_:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/python/graphenelibos/sgx_sign.py#L660)):*

`setdefault()` returns whatever is the result, so `sgx = manifest.setdefault(...)` is enough. Same below. 
</blockquote></details>

Done.
___
*[python/graphenelibos/sgx_sign.py, line 798 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3W4KeCi5q-b3450Xm:-MV8B_7m-zNeXqHLoj0s:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/python/graphenelibos/sgx_sign.py#L798)):*

`{}`.format(), plz. Or defer this for 1 more month and convert to f-strings after retiring Ubuntu 16.04.
</blockquote></details>

Done.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: 1 of 12 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju and @yamahata)

___
*[Pal/lib/string/toml_utils.c, line 20 at r2](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVD6bVh-wINReP7A3fD:-MVD6bVi8v-do2BbY1PC:b-gamta1) ([raw file](https://github.com/oscarlab/graphene/blob/d6f756bbb3a0e9464943689522bb6b94106e88b0/Pal/lib/string/toml_utils.c#L20)):*
> ```C
> 
> /* Searches for a dotted-key (e.g. "fs.mount.lib1.type") from `root`; returns NULL if value for
>  * such key is not found. Double quotes are respected, same as in TOML. */
> ```

FYI: I made this change because it turned out that we didn't handle dots in keys correctly (e.g. in keys like `sgx.trusted_checksum."/lib64/ld-linux-x86-64.so.2"`) and some tests failed because of this.

mkow:
Jenkins, retest Jenkins-Debug-18.04 please (sendfile08 failed, known issue)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: 1 of 15 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju and @yamahata)

___
*[LibOS/shim/src/fs/shim_fs.c, line 269 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVEDReBAfToHkEhIvF_:-MVEDReC2JWDWZXTWi-s:b7wdyz1) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/LibOS/shim/src/fs/shim_fs.c#L269)):*
> ```C
>         return 0;
> 
>     /* *** Warning: A _very_ ugly hack below (hopefully only temporary) ***
> ```

I'm sorry.
Maybe you will have better ideas how to workaround this, but I couldn't figure out any...

dimakuv (changes requested):
> Reviewed 1 of 11 files at r1, 10 of 11 files at r2, 3 of 4 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[LibOS/shim/src/fs/shim_fs.c, line 285 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGF8k18wrrP3EvDbBq:-MVGF8k18wrrP3EvDbBr:b-rze6oz) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/LibOS/shim/src/fs/shim_fs.c#L285)):*
> ```C
>      * fast.
>      *
>      * Corresponding issue: https://github.com/oscarlab/graphene/issues/2214.
> ```

But why do we suddenly need to do this (other than being correct)? Everything worked before, even though it is technically wrong to rely on undocumented feature of the TOML parser to preserve the order of entries on the manifest file. Why did you need this hack in this PR?
___
*[Pal/lib/string/toml_utils.c, line 20 at r2](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVD6bVh-wINReP7A3fD:-MVGHOyo9c1Qf6HW-dok:b-94gj7g) ([raw file](https://github.com/oscarlab/graphene/blob/d6f756bbb3a0e9464943689522bb6b94106e88b0/Pal/lib/string/toml_utils.c#L20)):*

FYI: I made this change because it turned out that we didn't handle dots in keys correctly (e.g. in keys like `sgx.trusted_checksum."/lib64/ld-linux-x86-64.so.2"`) and some tests failed because of this.
</blockquote></details>

Thanks, my bad, I didn't take into account this case.

But your code still doesn't allow a key with an escaped `"` somewhere in the middle. Which is fine, but maybe add a comment that it is not completely TOML-compatible?
___
*[Pal/lib/string/toml_utils.c, line 27 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGGd4s1DaRbQX7gfAS:-MVGGd4s1DaRbQX7gfAT:bl2lrjz) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/Pal/lib/string/toml_utils.c#L27)):*
> ```C
>         return NULL;
> 
>     toml_raw_t res = NULL;
> ```

If you want to rename `raw` -> `res`, then please do the same in other functions in this file.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 726 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGHsPOAfEoK5XaGK6w:-MVGHsPOAfEoK5XaGK6x:b-2zr1kj) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/Pal/src/host/Linux-SGX/enclave_framework.c#L726)):*
> ```C
> 
>     /* read sgx.trusted_checksum.<key> entry from manifest */
>     char* fullkey = alloc_concat3("sgx.trusted_checksum.\"", -1, key, -1, "\"", -1);
> ```

We have a special case for `sgx.trusted_checksum."blabla"` entries now. Do you want to keep it special, or do we want to update all other TOML keys like this in our C code?
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 733 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGIF2ACwVOm_A7pcb8:-MVGIF2ACwVOm_A7pcb9:b-z3j83b) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/Pal/src/host/Linux-SGX/enclave_framework.c#L733)):*
> ```C
>     ret = toml_string_in(g_pal_state.manifest_root, fullkey, &trusted_checksum);
>     if (ret < 0) {
>         log_error("Cannot parse \'%s\'\n", fullkey);
> ```

We don't need escaping of `'`.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 738 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGIHVh3QEx5jwuTLat:-MVGIHVh3QEx5jwuTLau:b-z3j83b) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/Pal/src/host/Linux-SGX/enclave_framework.c#L738)):*
> ```C
>     }
>     if (!trusted_checksum) {
>         log_error("Missing \'%s\' entry\n", fullkey);
> ```

We don't need escaping of `'`.
___
*[python/graphenelibos/sgx_sign.py, line 118 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGJPSx7QdPROodELFi:-MVGJPSy0T1HsgxQU4ql:bg90ljh) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/python/graphenelibos/sgx_sign.py#L118)):*
> ```Python
>     ]
> 
>     attributes = {'XFRM_LEGACY'}
> ```

Can you add a comment that this attribute set (which means "SSE supported") always needed to be set for SGX?
___
*[python/graphenelibos/sgx_sign.py, line 148 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGJnyQDsLLk_jLIsfo:-MVGJnyRBptENE7h3nYp:b-9wo4l0) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/python/graphenelibos/sgx_sign.py#L148)):*
> ```Python
>     path = Path(uri[len('file:'):])
>     if check_exist and not path.exists():
>         raise Exception('Cannot resolve ' + uri + ' or the file does not exist.')
> ```

What is the difference between `RuntimeError` and `Exception`? Why do you have two different ones in here?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: 13 of 15 files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[LibOS/shim/src/fs/shim_fs.c, line 285 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGF8k18wrrP3EvDbBq:-MVGP-tKCJ7FjbCufemF:bnqphat) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/LibOS/shim/src/fs/shim_fs.c#L285)):*

But why do we suddenly need to do this (other than being correct)? Everything worked before, even though it is technically wrong to rely on undocumented feature of the TOML parser to preserve the order of entries on the manifest file. Why did you need this hack in this PR?
</blockquote></details>

It worked before by an accident and when I touched sgx_sign.py it started failing in CI.

Previously it worked because of two things:
* tomlc99 seems to preserve the order from the file (but I think this is just a side effect of their implementation, not a documented feature),
* old sgx_sign.py was very limited and only able to append data to an already existing manifest.

I broke the second point by using `toml` library in Python, which by design doesn't preserve the order (which is perfectly correct).
___
*[Pal/lib/string/toml_utils.c, line 20 at r2](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVD6bVh-wINReP7A3fD:-MVGQwJzArN-3hlytU8T:bgmjnv8) ([raw file](https://github.com/oscarlab/graphene/blob/d6f756bbb3a0e9464943689522bb6b94106e88b0/Pal/lib/string/toml_utils.c#L20)):*

Thanks, my bad, I didn't take into account this case.

But your code still doesn't allow a key with an escaped `"` somewhere in the middle. Which is fine, but maybe add a comment that it is not completely TOML-compatible?
</blockquote></details>

Wait, can you use `"` in the middle of the key? This doesn't work for me, at least in Python:
```
In [2]: toml.loads('a.b"c"d = 1')                                                                                                            
TomlDecodeError: Found invalid character in key name: '"'. Try quoting the key name. (line 1 column 4 char 3)
```
___
*[Pal/lib/string/toml_utils.c, line 27 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGGd4s1DaRbQX7gfAS:-MVGRdkn7OUIUDPxaX_P:b-4krivu) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/Pal/lib/string/toml_utils.c#L27)):*

If you want to rename `raw` -> `res`, then please do the same in other functions in this file.
</blockquote></details>

Hmm, I think it makes less sense in others, which are very short and `raw` is directly assigned from `toml_raw_*`. Here I wanted to emphasize that this variable holds the final return value, which isn't obvious if it's just "raw", but the whole control flow depends on it being NULL and just does simple `goto out` on failure (without setting any variables to denote the error).
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 726 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGHsPOAfEoK5XaGK6w:-MVGSVMI3XOr05Fqptm7:b-fdrnpp) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/Pal/src/host/Linux-SGX/enclave_framework.c#L726)):*

We have a special case for `sgx.trusted_checksum."blabla"` entries now. Do you want to keep it special, or do we want to update all other TOML keys like this in our C code?
</blockquote></details>

I'd leave it as is, this problem will disappear once we migrate to a proper TOML (the key names won't be those artifical IDs then).
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 733 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGIF2ACwVOm_A7pcb8:-MVGSrEw0OqKyHafISoV:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/Pal/src/host/Linux-SGX/enclave_framework.c#L733)):*

We don't need escaping of `'`.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 738 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGIHVh3QEx5jwuTLat:-MVGSraCDIW-RyWT0sDv:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/Pal/src/host/Linux-SGX/enclave_framework.c#L738)):*

We don't need escaping of `'`.
</blockquote></details>

Done.
___
*[python/graphenelibos/sgx_sign.py, line 118 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGJPSx7QdPROodELFi:-MVGTDVR2of8oyc0oKcv:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/python/graphenelibos/sgx_sign.py#L118)):*

Can you add a comment that this attribute set (which means "SSE supported") always needed to be set for SGX?
</blockquote></details>

Done.
___
*[python/graphenelibos/sgx_sign.py, line 148 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGJnyQDsLLk_jLIsfo:-MVGTLhVDQD7LQFEEpQX:b-6a2qik) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/python/graphenelibos/sgx_sign.py#L148)):*

What is the difference between `RuntimeError` and `Exception`? Why do you have two different ones in here?
</blockquote></details>

Ah, I didn't notice we use another one here.
@woju Which one should I use, assuming that writing a custom exception for this is an overkill? (we'd need to refactor quite a lot of code here to properly report errors and handle them in `main()`, and this PR is already large).

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[LibOS/shim/src/fs/shim_fs.c, line 285 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGF8k18wrrP3EvDbBq:-MVGY6fG5NHpPk1jimUa:b-i0cano) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/LibOS/shim/src/fs/shim_fs.c#L285)):*

It worked before by an accident and when I touched sgx_sign.py it started failing in CI.

Previously it worked because of two things:
* tomlc99 seems to preserve the order from the file (but I think this is just a side effect of their implementation, not a documented feature),
* old sgx_sign.py was very limited and only able to append data to an already existing manifest.

I broke the second point by using `toml` library in Python, which by design doesn't preserve the order (which is perfectly correct).
</blockquote></details>

OK, I understand now. You use `toml.dump()` in Python, which shuffles TOML entries in the final `blabla.manifest.sgx`.
___
*[Pal/lib/string/toml_utils.c, line 20 at r2](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVD6bVh-wINReP7A3fD:-MVGYZCI3WZUehplzPB6:b-3wo0k) ([raw file](https://github.com/oscarlab/graphene/blob/d6f756bbb3a0e9464943689522bb6b94106e88b0/Pal/lib/string/toml_utils.c#L20)):*

Wait, can you use `"` in the middle of the key? This doesn't work for me, at least in Python:
```
In [2]: toml.loads('a.b"c"d = 1')                                                                                                            
TomlDecodeError: Found invalid character in key name: '"'. Try quoting the key name. (line 1 column 4 char 3)
```
</blockquote></details>

I meant something like `a.b."escaped \" works"`.
___
*[Pal/lib/string/toml_utils.c, line 27 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGGd4s1DaRbQX7gfAS:-MVGYhW103UTLoaojCMT:b32w0x) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/Pal/lib/string/toml_utils.c#L27)):*

Hmm, I think it makes less sense in others, which are very short and `raw` is directly assigned from `toml_raw_*`. Here I wanted to emphasize that this variable holds the final return value, which isn't obvious if it's just "raw", but the whole control flow depends on it being NULL and just does simple `goto out` on failure (without setting any variables to denote the error).
</blockquote></details>

OK

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[LibOS/shim/src/fs/shim_fs.c, line 285 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGF8k18wrrP3EvDbBq:-MVG_Dg6EjMr3E8TM618:btih660) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/LibOS/shim/src/fs/shim_fs.c#L285)):*

OK, I understand now. You use `toml.dump()` in Python, which shuffles TOML entries in the final `blabla.manifest.sgx`.
</blockquote></details>

Yes, exactly.
___
*[Pal/lib/string/toml_utils.c, line 20 at r2](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVD6bVh-wINReP7A3fD:-MVG_Nl25aGLL3iRUx47:b-z9ty36) ([raw file](https://github.com/oscarlab/graphene/blob/d6f756bbb3a0e9464943689522bb6b94106e88b0/Pal/lib/string/toml_utils.c#L20)):*

I meant something like `a.b."escaped \" works"`.
</blockquote></details>

Nah, I considered this case, but checked it in Python and it seems it's not allowed?
```
In [4]: toml.loads(r'''a.b\"cd = 1''')                                                                                                       
TomlDecodeError: Found invalid character in key name: '\'. Try quoting the key name. (line 1 column 4 char 3)
```
(and no, quoting doesn't help)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[Pal/lib/string/toml_utils.c, line 20 at r2](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVD6bVh-wINReP7A3fD:-MVGfnKh0Z2-5C4cM_FN:b-xpuyrb) ([raw file](https://github.com/oscarlab/graphene/blob/d6f756bbb3a0e9464943689522bb6b94106e88b0/Pal/lib/string/toml_utils.c#L20)):*

Nah, I considered this case, but checked it in Python and it seems it's not allowed?
```
In [4]: toml.loads(r'''a.b\"cd = 1''')                                                                                                       
TomlDecodeError: Found invalid character in key name: '\'. Try quoting the key name. (line 1 column 4 char 3)
```
(and no, quoting doesn't help)
</blockquote></details>

Surprising but Ok, resolving.

boryspoplawski (changes requested):
> Reviewed 1 of 11 files at r1, 8 of 11 files at r2, 3 of 4 files at r3, 1 of 1 files at r4, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[LibOS/shim/src/fs/shim_fs.c, line 288 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVIeDqn2ZLFbg4YJ6w-:-MVIeDqn2ZLFbg4YJ6w0:b6n1rsm) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/LibOS/shim/src/fs/shim_fs.c#L288)):*
> ```C
>      */
>     const char** keys = malloc(mounts_cnt * sizeof(*keys));
>     size_t* path_len = malloc(mounts_cnt * sizeof(*path_len));
> ```

Please use plural form
___
*[LibOS/shim/src/fs/shim_fs.c, line 318 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVIgcB6E2inCWUlata6:-MVIgcB75DLks77J-NYs:b-lwqil0) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/LibOS/shim/src/fs/shim_fs.c#L318)):*
> ```C
>         }
>     }
> out:
> ```

Since we do not do a proper cleanup on error paths in this function, please remove gotos (just return)
___
*[Pal/lib/string/toml_utils.c, line 20 at r2](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVD6bVh-wINReP7A3fD:-MVIiS9z2mhVppy41WRk:bhvnrw8) ([raw file](https://github.com/oscarlab/graphene/blob/d6f756bbb3a0e9464943689522bb6b94106e88b0/Pal/lib/string/toml_utils.c#L20)):*

Surprising but Ok, resolving.
</blockquote></details>

Python is weird, `toml.loads('"b\" = 1')` is apparently correct TOML
___
*[Pal/lib/string/toml_utils.c, line 39 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVIlh7fBOCNPkn0ecth:-MVIlh7fBOCNPkn0ecti:b-n20fm3) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/Pal/lib/string/toml_utils.c#L39)):*
> ```C
>             subkey++;
>             subkey_end = find_next_char(subkey, '"');
>             if (*subkey_end != '"'  || (*(subkey_end+1) != '.' && *(subkey_end+1) != '\0'))
> ```

Why not `subkey_end[1]`?
___
*[python/graphenelibos/sgx_sign.py, line 148 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGJnyQDsLLk_jLIsfo:-MVJUPf56JYc-wcIHQhF:biu7uoz) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/python/graphenelibos/sgx_sign.py#L148)):*

Ah, I didn't notice we use another one here.
@woju Which one should I use, assuming that writing a custom exception for this is an overkill? (we'd need to refactor quite a lot of code here to properly report errors and handle them in `main()`, and this PR is already large).
</blockquote></details>

Imo `FileNotFoundError` is what we need here.`RuntimeError` is generic bag for everything that you don't have specific exception for and `Exception` is base class for most of the exceptions and raising it directly makes no sense to me.
___
*[python/graphenelibos/sgx_sign.py, line 15 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJRL1oDf7LcUb0FVdh:-MVJRL1p0AHyktHMKa0m:bodyc8c) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L15)):*
> ```Python
> import subprocess
> from sys import stderr
> ```

Why the empty line?
___
*[python/graphenelibos/sgx_sign.py, line 84 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJRp8r9B9yg4cTwbkf:-MVJRp8r9B9yg4cTwbkg:bugky08) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L84)):*
> ```Python
> 
> 
> # Loading Enclave Attributes
> ```

Maybe move this comment after next function? Looks weird now.
___
*[python/graphenelibos/sgx_sign.py, line 661 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJXVHu6DGpPf-kjBzu:-MVJXVHu6DGpPf-kjBzv:b-yt7w65) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L661)):*
> ```Python
> 
>     sgx = manifest.setdefault('sgx', {})
>     sgx.setdefault('trusted_files', {})
> ```

Please move these values to a separate dict of defaults and iterate over that dict here.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: 13 of 15 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @woju)

___
*[LibOS/shim/src/fs/shim_fs.c, line 288 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVIeDqn2ZLFbg4YJ6w-:-MVJmyrU822ZHmBgoRm6:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/LibOS/shim/src/fs/shim_fs.c#L288)):*

Please use plural form
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/shim_fs.c, line 318 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVIgcB6E2inCWUlata6:-MVJn0Nt0b4maj2cgnF7:bh2uaon) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/LibOS/shim/src/fs/shim_fs.c#L318)):*

Since we do not do a proper cleanup on error paths in this function, please remove gotos (just return)
</blockquote></details>

What do you mean? That we don't undo mounts?
___
*[Pal/lib/string/toml_utils.c, line 39 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVIlh7fBOCNPkn0ecth:-MVJnKPz1xaxwtn4Tbk9:bxd1j9b) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/Pal/lib/string/toml_utils.c#L39)):*

Why not `subkey_end[1]`?
</blockquote></details>

I thought about it, and decided to keep both derefs in the same form (because we're checking a character at the iterator and then a following one, so the checks are "coupled" logically). `subkey_end[0]` would look quite odd as a dereference of an iterator, this syntax it's usually used for arrays (i.e. their beginning pointers), not iterators inside them.
___
*[python/graphenelibos/sgx_sign.py, line 148 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGJnyQDsLLk_jLIsfo:-MVJp5Qu4IsVlvN9xMt1:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/python/graphenelibos/sgx_sign.py#L148)):*

Imo `FileNotFoundError` is what we need here.`RuntimeError` is generic bag for everything that you don't have specific exception for and `Exception` is base class for most of the exceptions and raising it directly makes no sense to me.
</blockquote></details>

Done.
___
*[python/graphenelibos/sgx_sign.py, line 15 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJRL1oDf7LcUb0FVdh:-MVJo_H7DM9Cbkele_l8:b-60asmt) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L15)):*

Why the empty line?
</blockquote></details>

That's a pep8 requirement, `toml` is a third party library, it's not a built-in.
___
*[python/graphenelibos/sgx_sign.py, line 84 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJRp8r9B9yg4cTwbkf:-MVJqTMY1ceFZEwBVpUb:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L84)):*

Maybe move this comment after next function? Looks weird now.
</blockquote></details>

Done.
___
*[python/graphenelibos/sgx_sign.py, line 661 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJXVHu6DGpPf-kjBzu:-MVJpXH56rVa4uR7iSa9:bj1ujti) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L661)):*

Please move these values to a separate dict of defaults and iterate over that dict here.
</blockquote></details>

How exactly? For the loader part below this would be an overkill (and I don't want to have a recursive implementation to use just one dict, this would be an even larger overkill) and the having one set of values initialized differently than the other makes this simple logic more convoluted.
I'd rather just leave it as is.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[LibOS/shim/src/fs/shim_fs.c, line 318 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVIgcB6E2inCWUlata6:-MVK07RH3ITwTJWdou5E:bkr0gpm) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/LibOS/shim/src/fs/shim_fs.c#L318)):*

What do you mean? That we don't undo mounts?
</blockquote></details>

Yes and I thought you don't free `mount_path` on error, but the only error nearby is when it's not created, my bad. In such case I'm resolving this.
___
*[Pal/lib/string/toml_utils.c, line 39 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVIlh7fBOCNPkn0ecth:-MVK2fs04h8Je_iT6nwd:bh02te7) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/Pal/lib/string/toml_utils.c#L39)):*

I thought about it, and decided to keep both derefs in the same form (because we're checking a character at the iterator and then a following one, so the checks are "coupled" logically). `subkey_end[0]` would look quite odd as a dereference of an iterator, this syntax it's usually used for arrays (i.e. their beginning pointers), not iterators inside them.
</blockquote></details>

If you want to treat this as an iterator (which imo is a big overstatement, we are in C land) then you can go further and say this is a slice. Then `subkey_end[0]` will make perfect sense.
In any case I dislike the current version, deref of pointer + offset is hard to read imo.
(btw, there are no spaces around `+`)
___
*[python/graphenelibos/sgx_sign.py, line 661 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJXVHu6DGpPf-kjBzu:-MVK3dYb7uSEOqtWb3Ml:b-mp7pu0) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L661)):*

How exactly? For the loader part below this would be an overkill (and I don't want to have a recursive implementation to use just one dict, this would be an even larger overkill) and the having one set of values initialized differently than the other makes this simple logic more convoluted.
I'd rather just leave it as is.
</blockquote></details>

I just wanted to have:
```
SGX_DEFAULT_VALUES_OR_SOME_BETTER_NAME = {
    'trusted_files': {},
    'enclave_size': DEFAULT_ENCLAVE_SIZE,
    ...
}

for k, v in dict.items():
    sgx.setdefault(k, v)
```
Current version looks plain ugly and harder to maintain. Imo default values should be separated and even taken outside of the scope of this function.
Since we have only one entry (and a very simple one, not really assigning a meaningful value) in `loader` we could leave it as is, but if there were more I would move them too.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: 14 of 15 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @woju)

___
*[Pal/lib/string/toml_utils.c, line 39 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVIlh7fBOCNPkn0ecth:-MVLgQ_493nwaTFbO_-C:b-8mrenx) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/Pal/lib/string/toml_utils.c#L39)):*

If you want to treat this as an iterator (which imo is a big overstatement, we are in C land) then you can go further and say this is a slice. Then `subkey_end[0]` will make perfect sense.
In any case I dislike the current version, deref of pointer + offset is hard to read imo.
(btw, there are no spaces around `+`)
</blockquote></details>

I didn't mean "iterator" in C++ sense, just a variable iterating over an array.
And I dislike the `subkey_end[0]` version :)
Let's have @dimakuv vote and decide :)
___
*[python/graphenelibos/sgx_sign.py, line 661 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJXVHu6DGpPf-kjBzu:-MVLhL8C1CTfRcJQZ8W5:b12fldc) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L661)):*

I just wanted to have:
```
SGX_DEFAULT_VALUES_OR_SOME_BETTER_NAME = {
    'trusted_files': {},
    'enclave_size': DEFAULT_ENCLAVE_SIZE,
    ...
}

for k, v in dict.items():
    sgx.setdefault(k, v)
```
Current version looks plain ugly and harder to maintain. Imo default values should be separated and even taken outside of the scope of this function.
Since we have only one entry (and a very simple one, not really assigning a meaningful value) in `loader` we could leave it as is, but if there were more I would move them too.
</blockquote></details>

But then you have both categories set in a different way, which is IMO worse of the two.
Let's ask @dimakuv what he prefers ;)

dimakuv (dismissed):
> Reviewed 1 of 2 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @woju)

___
*[Pal/lib/string/toml_utils.c, line 39 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVIlh7fBOCNPkn0ecth:-MVLx90P2afACs2I7FcB:b2dlc32) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/Pal/lib/string/toml_utils.c#L39)):*

I didn't mean "iterator" in C++ sense, just a variable iterating over an array.
And I dislike the `subkey_end[0]` version :)
Let's have @dimakuv vote and decide :)
</blockquote></details>

I'm siding with Borys on this one.
___
*[python/graphenelibos/sgx_sign.py, line 661 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJXVHu6DGpPf-kjBzu:-MVLxvMn4hcOy0_8v-uz:bfeky5p) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L661)):*

But then you have both categories set in a different way, which is IMO worse of the two.
Let's ask @dimakuv what he prefers ;)
</blockquote></details>

I'm fine with the current (Michal's) version.

To be honest, this should go away. We should have a better handling of defaults -- used both in C code and in Python code. Right now we have duplication of these defaults.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: 14 of 15 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @woju)

___
*[Pal/lib/string/toml_utils.c, line 39 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVIlh7fBOCNPkn0ecth:-MVLyo9F71_K7eF3Im1Y:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/Pal/lib/string/toml_utils.c#L39)):*

I'm siding with Borys on this one.
</blockquote></details>

Done.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: 14 of 15 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @woju)

___
*[python/graphenelibos/sgx_sign.py, line 661 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJXVHu6DGpPf-kjBzu:-MVLyqbvCUN1MUwskvRG:b-c5fet8) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L661)):*

I'm fine with the current (Michal's) version.

To be honest, this should go away. We should have a better handling of defaults -- used both in C code and in Python code. Right now we have duplication of these defaults.
</blockquote></details>

Why do we need defaults in C code? I thought we always emit these entries (since we have defaults here).

boryspoplawski (approved):
> Reviewed 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

dimakuv (approved):
> Reviewed 1 of 2 files at r6, 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[python/graphenelibos/sgx_sign.py, line 661 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJXVHu6DGpPf-kjBzu:-MVM3g0X9tsWj8_PJ-d3:b28yulm) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L661)):*

Why do we need defaults in C code? I thought we always emit these entries (since we have defaults here).
</blockquote></details>

No, we don't emit all entries in this Python script. E.g. `sgx.log_level` or `sgx.preheat_enclave`. That's my point -- we don't emit everything in the Python script, so some manifest entries are implicit (and thus default to whatever C code has as default). But this is not for this PR -- here Michal simply re-shuffled the entries already used before him.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[python/graphenelibos/sgx_sign.py, line 661 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJXVHu6DGpPf-kjBzu:-MVMNy2nAe2cxtIwIwbR:b-kj6fnu) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L661)):*

No, we don't emit all entries in this Python script. E.g. `sgx.log_level` or `sgx.preheat_enclave`. That's my point -- we don't emit everything in the Python script, so some manifest entries are implicit (and thus default to whatever C code has as default). But this is not for this PR -- here Michal simply re-shuffled the entries already used before him.
</blockquote></details>

So imo we should decide on one place: either handle defaults in C or always emit everything here.
Ofc this is not to be done in this PR.

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[.pylintrc, line 326 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3WynCE0wqX1spYobD:-MVMYFbQ1gP18es7tZ-x:b7loaw2) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/.pylintrc#L326)):*

Honestly, why? `with open() as f:` is quite a popular idiom with very short `with` blocks, same as `i` in loops and `e` for exceptions.
This rule seems much too picky to me.
</blockquote></details>

Not to give other people reason to use 1-letter variables for random things. I know `f` for `file` is OK, because it doesn't affect readability, but it's not OK for `fizzbuzz` or anything else. The problem is, `pylint` can't know in advance how the variable is used, so the option is to either have `f` allowed everywhere or prohibited everywhere.
___
*[LibOS/shim/src/fs/shim_fs.c, line 269 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVEDReBAfToHkEhIvF_:-MVMYzcl8qoFkEVReDnj:b-8jv1de) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/LibOS/shim/src/fs/shim_fs.c#L269)):*

I'm sorry.
Maybe you will have better ideas how to workaround this, but I couldn't figure out any...
</blockquote></details>

I don't see what's wrong with sorting by length (other than CPU time).
___
*[python/graphenelibos/sgx_sign.py, line 122 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3U3po666e8mF8N8ML:-MVMZBcL84BwErmotJ35:brxp5g7) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/python/graphenelibos/sgx_sign.py#L122)):*

Oh, sorry, that was supposed to be just a list (list seems more proper here to me, as it just lists arguments for consecutive iterations).
</blockquote></details>

Ah, OK, this also works, as long as there are no duplicate keys.
___
*[python/graphenelibos/sgx_sign.py, line 148 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGJnyQDsLLk_jLIsfo:-MVMZRbU9fOm1XehOnx2:b-m1nqky) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/python/graphenelibos/sgx_sign.py#L148)):*

Done.
</blockquote></details>

`FileNotFoundError` is a child of `OSError`, which is generally a wrapper for `errno`. So `FileNotFoundError` is wrong. It's better to have a generic `Exception()`. Or maybe you even can skip this check entirely and instead rely on a real `FileNotFoundError` raised when someone actually attempts to `open()` the file.

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[.pylintrc, line 326 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3WynCE0wqX1spYobD:-MVMd-603Pmnh7RFOcVo:b-iz9ly0) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/.pylintrc#L326)):*

Not to give other people reason to use 1-letter variables for random things. I know `f` for `file` is OK, because it doesn't affect readability, but it's not OK for `fizzbuzz` or anything else. The problem is, `pylint` can't know in advance how the variable is used, so the option is to either have `f` allowed everywhere or prohibited everywhere.
</blockquote></details>

Following this line of thought, `i` is also wrong and should be removed.
___
*[python/graphenelibos/sgx_sign.py, line 148 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGJnyQDsLLk_jLIsfo:-MVMdEE46c15XQq4UGyf:bqvczz) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/python/graphenelibos/sgx_sign.py#L148)):*

`FileNotFoundError` is a child of `OSError`, which is generally a wrapper for `errno`. So `FileNotFoundError` is wrong. It's better to have a generic `Exception()`. Or maybe you even can skip this check entirely and instead rely on a real `FileNotFoundError` raised when someone actually attempts to `open()` the file.
</blockquote></details>

Imo its good, this is exactly what we want to return here: file was not found (exactly because OS could not find it).

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

___
*[python/graphenelibos/sgx_sign.py, line 148 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGJnyQDsLLk_jLIsfo:-MVMgqxk-reUV0IkP4T7:b-t9p4uh) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/python/graphenelibos/sgx_sign.py#L148)):*

No, this isn't a fault from syscall (well, indirectly it is from `stat()`, but you know). Also this misses expectations from anyone who'll want to wrap this code in try/except (I guess s/he'll try to catch for the case of missing manifest file, not something wrong inside the manifest)
</blockquote></details>

Well, it is exactly a fault from a syscall ;)
In what way having a generic `Exception` is better? It strips you from all means of handling this error in a meaningful way. This is a case of a missing file that this manifest depends on, so it probably should be treated in a similar way.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

yamahata (dismissed):
> Reviewed 8 of 11 files at r2, 2 of 4 files at r3, 1 of 2 files at r5, 2 of 2 files at r6, 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow and @woju)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[.pylintrc, line 326 at r1](https://reviewable.io/reviews/oscarlab/graphene/2210#-MV3WynCE0wqX1spYobD:-MVOhDsRAIH8OWkaCIXS:b-nwyy8u) ([raw file](https://github.com/oscarlab/graphene/blob/f5df50aba5cf50aac6480ae47cfd3c4d24b24d03/.pylintrc#L326)):*

No, `i`, `j`, `k` are common for iteration and there are no good replacements. But `file` or `fd` instead of `f` are reasonable. 
</blockquote></details>

I'd say this rule tries to enforce something which makes sense to enforce only via reviews. Anyways, as this is not too important in this PR, I reverted this change.
___
*[LibOS/shim/src/fs/shim_fs.c, line 269 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVEDReBAfToHkEhIvF_:-MVOi3YFE7hmY766d7N-:b-9wx3j9) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/LibOS/shim/src/fs/shim_fs.c#L269)):*

I don't see what's wrong with sorting by length (other than CPU time).
</blockquote></details>

The hackiness here is that what we actually need here is topological sort (mounting parents, then children). But path length is a pretty good approximation, if we stripped double-dots then I may be even correct. Anyways, this is just temporary.
___
*[python/graphenelibos/sgx_sign.py, line 661 at r5](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVJXVHu6DGpPf-kjBzu:-MVOifLxCk_XiRClpWyn:b-1sl2gu) ([raw file](https://github.com/oscarlab/graphene/blob/3517c79cb64bd1ef2ad6c840459941ec3ddb4066/python/graphenelibos/sgx_sign.py#L661)):*

So imo we should decide on one place: either handle defaults in C or always emit everything here.
Ofc this is not to be done in this PR.
</blockquote></details>

Here it's a bit different, the default set here have also another role (not just setting defaults) - ensuring that we can safely do `manifest['sgx']['trusted_files']`, etc., without a chain of `if`s or try-except blocks.

dimakuv (approved):
> Reviewed 2 of 2 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @woju)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: 13 of 14 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[python/graphenelibos/sgx_sign.py, line 148 at r4](https://reviewable.io/reviews/oscarlab/graphene/2210#-MVGJnyQDsLLk_jLIsfo:-MVR7tE-E1a97QYw04GU:b-d3othq) ([raw file](https://github.com/oscarlab/graphene/blob/31559c332f126e16290d279121153160cc535d2d/python/graphenelibos/sgx_sign.py#L148)):*

Well, it is exactly a fault from a syscall ;)
In what way having a generic `Exception` is better? It strips you from all means of handling this error in a meaningful way. This is a case of a missing file that this manifest depends on, so it probably should be treated in a similar way.
</blockquote></details>

Ok, I introduced a custom exception for this, please take a look.

woju (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (approved):
> Reviewed 2 of 2 files at r11.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2210)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

PR #2222, opened 2021-03-09
[Pal/{lib,Linux-SGX}] Add seqlocks and use them in SGX's _DkSystemTimeQuery()

dimakuv:
## Description of the changes 

For gettime emulation, Linux-SGX PAL uses RDTSC instruction if its execution is allowed inside the SGX enclave and if the CPU supports Invariant TSC. Previously, this code path used a global spinlock to synchronize read/write accesses to shared-time-variables (threads very frequently read these shared variables and very rarely update them).

Using a spinlock led to significant performance overhead due to contention between readers (on some systems, around 140 simultaneous reader threads). This commit introduces seqlocks to allow fast lock-free readers, and replaces spinlock with a seqlock in `_DkSystemTimeQuery()`.

## How to test this PR? 

New test to sanity-check `gettimeofday()` emulation on Linux-SGX was added to LibOS regression suite.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)

### DIFF
--- a/LibOS/shim/test/regression/.gitignore
+++ b/LibOS/shim/test/regression/.gitignore
@@ -61,6 +61,7 @@
 /getdents
 /getsockname
 /getsockopt
+/gettimeofday
 /groups
 /helloworld
 /host_root_fs
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -40,6 +40,7 @@ c_executables = \
 	getdents \
 	getsockname \
 	getsockopt \
+	gettimeofday \
 	groups \
 	helloworld \
 	host_root_fs \
@@ -149,6 +150,7 @@ CFLAGS-spinlock += -I$(PALDIR)/../include/lib -I$(PALDIR)/../include/arch/$(ARCH
 CFLAGS-sigaction_per_process += -pthread
 CFLAGS-signal_multithread += -pthread
 CFLAGS-pthread_set_get_affinity += -pthread
+CFLAGS-gettimeofday += -pthread
 
 CFLAGS-attestation += -I$(PALDIR)/../lib/crypto/mbedtls/crypto/include \
                       -I$(PALDIR)/host/Linux-SGX \
--- a/LibOS/shim/test/regression/gettimeofday.c
+++ b/LibOS/shim/test/regression/gettimeofday.c
@@ -1,0 +1,62 @@
+#include <err.h>
+#include <errno.h>
+#include <pthread.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/time.h>
+#include <unistd.h>
+
+#define MAX_TIME_DIFF_SEC 20
+#define THREAD_NUM 4
+#define ITERATIONS 1000000
+
+struct timeval base_tv;
+
+static void* foo(void* arg) {
+    int ret;
+
+    for (size_t i = 0; i < ITERATIONS; i++) {
+        struct timeval tv;
+        ret = gettimeofday(&tv, NULL);
+        if (ret < 0)
+            err(1, "thread gettimeofday");
+
+        if (tv.tv_sec < base_tv.tv_sec || tv.tv_sec - base_tv.tv_sec > MAX_TIME_DIFF_SEC)
+            errx(1, "Retrieved time is more than 20 seconds away from base time");
+    }
+
+    return NULL;
+}
+
+int main(int argc, char** argv) {
+    int ret;
+
+    ret = gettimeofday(&base_tv, NULL);
+    if (ret < 0)
+        err(1, "base gettimeofday");
+
+    printf("Starting time: %lu sec, %lu usec\n", base_tv.tv_sec, base_tv.tv_usec);
+
+    pthread_t thread[THREAD_NUM];
+    for (size_t i = 0; i < THREAD_NUM; i++) {
+        pthread_create(&thread[i], NULL, foo, NULL);
+    }
+    for (size_t i = 0; i < THREAD_NUM; i++) {
+        pthread_join(thread[i], NULL);
+    }
+
+    struct timeval end_tv;
+    ret = gettimeofday(&end_tv, NULL);
+    if (ret < 0)
+        err(1, "base gettimeofday");
+
+    uint64_t sec_diff  = end_tv.tv_sec - base_tv.tv_sec -
+                         (end_tv.tv_usec < base_tv.tv_usec ? 1 : 0);
+    uint64_t usec_diff = end_tv.tv_usec > base_tv.tv_usec ? end_tv.tv_usec - base_tv.tv_usec
+                                                          : base_tv.tv_usec - end_tv.tv_usec;
+    printf("Finish time: %lu sec, %lu usec (passed: %lu sec, %lu usec)\n", end_tv.tv_sec,
+           end_tv.tv_usec, sec_diff, usec_diff);
+    puts("TEST OK");
+    return 0;
+}
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -528,6 +528,10 @@ class TC_30_Syscall(RegressionTestCase):
         stdout, _ = self.run_binary(['pthread_set_get_affinity', '1000'])
         self.assertIn('TEST OK', stdout)
 
+    def test_103_gettimeofday(self):
+        stdout, _ = self.run_binary(['gettimeofday'])
+        self.assertIn('TEST OK', stdout)
+
 @unittest.skipUnless(HAS_SGX,
     'This test is only meaningful on SGX PAL because only SGX catches raw '
     'syscalls and redirects to Graphene\'s LibOS. If we will add seccomp to '
--- a/Pal/include/lib/seqlock.h
+++ b/Pal/include/lib/seqlock.h
@@ -1,0 +1,89 @@
+/* SPDX-License-Identifier: LGPL-3.0-or-later */
+/* Copyright (C) 2021 Intel Labs */
+
+/* NOTE: This seqlock implementation follows the spinlock one; see spinlock.h for details.
+ *       This seqlock implementation is based on the Linux one; see include/linux/seqlock.h.
+ *
+ * Reader/writer consistent mechanism with fast lock-free readers and without starving writers.
+ * Seqlocks allow readers to race with writers, but readers detect when a race occurs and retry the
+ * critical section (i.e., reads are speculative). Writers synchronize their accesses with other
+ * writers via a spinlock, but never wait for readers.
+ *
+ * Seqlocks can be only used when readers do not mutate any shared state and do not use pointers.
+ * For example, seqlocks are a perfect fit for time management routines. Typical usage for readers:
+ *     do {
+ *         seq = read_seqbegin(&foo);
+ *         ...
+ *     } while (read_seqretry(&foo, seq));
+ */
+
+#ifndef _SEQLOCK_H
+#define _SEQLOCK_H
+
+#include "api.h"
+#include "cpu.h"
+#include "spinlock.h"
+
+typedef struct {
+    uint32_t sequence;
+    spinlock_t lock;
+} seqlock_t;
+
+/*!
+ * \brief Initialize seqlock with *static* storage duration.
+ */
+#define INIT_SEQLOCK_UNLOCKED { .sequence = 0, .lock = INIT_SPINLOCK_UNLOCKED }
+
+
+/*!
+ * \brief Initialize seqlock with *dynamic* storage duration.
+ */
+static inline void seqlock_init(seqlock_t* sl) {
+    __atomic_store_n(&sl->sequence, 0, __ATOMIC_RELAXED);
+    spinlock_init(&sl->lock);
+}
+
+/*!
+ * \brief Start a reader-side critical section (no locking).
+ */
+static inline uint32_t read_seqbegin(seqlock_t* sl) {
+    uint32_t start;
+    while (true) {
+        start = __atomic_load_n(&sl->sequence, __ATOMIC_RELAXED);
+        if ((start & 1) == 0) {
+            /* there is no writer in the middle of an update, data is currently consistent */
+            break;
+        }
+        CPU_RELAX();
+    }
+    RMB();
+    return start;
+}
+
+/*!
+ * \brief Returns true if reader must retry the critical section.
+ */
+static inline bool read_seqretry(seqlock_t* sl, uint32_t start) {
+    RMB();
+    return start != __atomic_load_n(&sl->sequence, __ATOMIC_RELAXED);
+}
+
+/*!
+ * \brief Start a writer-side critical section (acquires spinlock).
+ */
+static inline void write_seqbegin(seqlock_t* sl) {
+    spinlock_lock(&sl->lock);
+    __atomic_add_fetch(&sl->sequence, 1, __ATOMIC_RELAXED);
+    WMB();
+}
+
+/*!
+ * \brief End a writer-side critical section (releases spinlock).
+ */
+static inline void write_seqend(seqlock_t* sl) {
+    WMB();
+    __atomic_add_fetch(&sl->sequence, 1, __ATOMIC_RELAXED);
+    spinlock_unlock(&sl->lock);
+}
+
+#endif // _SEQLOCK_H
--- a/Pal/src/host/Linux-SGX/db_misc.c
+++ b/Pal/src/host/Linux-SGX/db_misc.c
@@ -11,6 +11,7 @@
 
 #include "api.h"
 #include "cpu.h"
+#include "gsgx.h"
 #include "linux_utils.h"
 #include "pal.h"
 #include "pal_debug.h"
@@ -20,7 +21,7 @@
 #include "pal_linux.h"
 #include "pal_linux_defs.h"
 #include "pal_security.h"
-#include "gsgx.h"
+#include "seqlock.h"
 #include "sgx_api.h"
 #include "sgx_attest.h"
 #include "toml.h"
@@ -30,7 +31,7 @@
 uint64_t g_tsc_hz = 0; /* TSC frequency for fast and accurate time ("invariant TSC" HW feature) */
 static uint64_t g_start_tsc = 0;
 static uint64_t g_start_usec = 0;
-static PAL_LOCK g_tsc_lock = LOCK_INIT;
+static seqlock_t g_tsc_lock = INIT_SEQLOCK_UNLOCKED;
 
 /**
  * Initialize the data structures used for date/time emulation using TSC
@@ -43,54 +44,67 @@ void init_tsc(void) {
 
 /* TODO: result comes from the untrusted host, introduce some schielding */
 int _DkSystemTimeQuery(uint64_t* out_usec) {
-    uint64_t usec = 0;
-    uint64_t tsc_usec = 0, tsc_cyc1, tsc_cyc2, tsc_cyc, tsc_diff;
     int ret;
 
-    if (g_tsc_hz > 0) {
-        _DkInternalLock(&g_tsc_lock);
-        if (g_start_tsc > 0 && g_start_usec > 0) {
-            /* calculate the TSC-based time */
-            tsc_diff = get_tsc() - g_start_tsc;
-            if (tsc_diff < INT64_MAX / 1000000) {
-                tsc_usec = g_start_usec + (tsc_diff * 1000000 / g_tsc_hz);
-                if (tsc_usec < g_start_usec + TSC_REFINE_INIT_TIMEOUT_USECS) {
-                    /* no need to refine yet */
-                    usec = tsc_usec;
-                }
-            }
-        }
-        _DkInternalUnlock(&g_tsc_lock);
-
-        if (!usec) {
-            /* refresh the baseline usec and TSC to contain the drift */
-            tsc_cyc1 = get_tsc();
-            ret = ocall_gettime(&usec);
-            tsc_cyc2 = get_tsc();
-            if (!ret) {
-                /* the ocall_gettime() is a time consuming operation.   *
-                 * it includes EENTER and EEXIT instructions, our best  *
-                 * estimation is the timestamp obtained in the middle   *
-                 * time point, therefore, the tsc_cyc as baseline will  *
-                 * be calibrated precisely in this way.                 */
-                tsc_cyc = ((tsc_cyc2 - tsc_cyc1) / 2) + tsc_cyc1;
-                _DkInternalLock(&g_tsc_lock);
-                /* refresh the baseline data if no other thread updated g_start_tsc */
-                if (g_start_tsc < tsc_cyc) {
-                    g_start_usec = usec;
-                    g_start_tsc = tsc_cyc;
-                }
-                _DkInternalUnlock(&g_tsc_lock);
-            } else {
-                return -PAL_ERROR_DENIED;
-            }
-        }
-    } else {
-        /* fallback to the slow ocall */
-        ret = ocall_gettime(&usec);
-        if (ret < 0)
-            return -PAL_ERROR_DENIED;
+    if (!g_tsc_hz) {
+        /* RDTSC is not allowed or no Invariant TSC feature -- fallback to the slow ocall */
+        return ocall_gettime(out_usec);
     }
+
+    uint32_t seq;
+    uint64_t start_tsc;
+    uint64_t start_usec;
+    do {
+        seq = read_seqbegin(&g_tsc_lock);
+        start_tsc  = g_start_tsc;
+        start_usec = g_start_usec;
+    } while (read_seqretry(&g_tsc_lock, seq));
+
+    uint64_t usec = 0;
+    if (start_tsc > 0 && start_usec > 0) {
+        /* baseline TSC/usec pair was initialized, can calculate time via RDTSC (but should be
+         * careful with integer overflow during calculations) */
+        uint64_t diff_tsc = get_tsc() - start_tsc;
+        if (diff_tsc < UINT64_MAX / 1000000) {
+            uint64_t diff_usec = diff_tsc * 1000000 / g_tsc_hz;
+            if (diff_usec < TSC_REFINE_INIT_TIMEOUT_USECS) {
+                /* less than TSC_REFINE_INIT_TIMEOUT_USECS passed from the previous update of
+                 * TSC/usec pair (time drift is contained), use the RDTSC-calculated time */
+                usec = start_usec + diff_usec;
+                if (usec < start_usec)
+                    return -PAL_ERROR_OVERFLOW;
+            }
+        }
+    }
+
+    if (usec) {
+        *out_usec = usec;
+        return 0;
+    }
+
+    /* if we are here, either the baseline TSC/usec pair was not yet initialized or too much time
+     * passed since the previous TSC/usec update, so let's refresh them to contain the time drift */
+    uint64_t tsc_cyc1 = get_tsc();
+    ret = ocall_gettime(&usec);
+    if (ret < 0)
+        return -PAL_ERROR_DENIED;
+    uint64_t tsc_cyc2 = get_tsc();
+
+    /* we need to match the OCALL-obtained timestamp (`usec`) with the RDTSC-obtained number of
+     * cycles (`tsc_cyc`); since OCALL is a time-consuming operation, we estimate `tsc_cyc` as a
+     * mid-point between the RDTSC values obtained right-before and right-after the OCALL. */
+    uint64_t tsc_cyc = tsc_cyc1 + (tsc_cyc2 - tsc_cyc1) / 2;
+    if (tsc_cyc < tsc_cyc1)
+        return -PAL_ERROR_OVERFLOW;
+
+    /* refresh the baseline data if no other thread updated g_start_tsc */
+    write_seqbegin(&g_tsc_lock);
+    if (g_start_tsc < tsc_cyc) {
+        g_start_tsc  = tsc_cyc;
+        g_start_usec = usec;
+    }
+    write_seqend(&g_tsc_lock);
+
     *out_usec = usec;
     return 0;
 }


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: 0 of 6 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners

___
*[Pal/include/lib/seqlock.h, line 59 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVLu2D8-BkZMHgt50vJ:-MVLu2D8-BkZMHgt50vK:behkeyk) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/include/lib/seqlock.h#L59)):*
> ```objc
>         CPU_RELAX();
>     }
>     RMB();
> ```

FYI: Interestingly, https://elixir.bootlin.com/linux/v3.19.8/source/include/linux/seqlock.h (Linux 3.19) didn't have this `RMB` but https://elixir.bootlin.com/linux/v5.11.4/source/include/linux/seqlock.h has it. So I added too.

boryspoplawski (changes requested):
> Reviewed 5 of 6 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @yamahata)

___
*[LibOS/shim/test/regression/gettimeofday.c, line 14 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVM-sgD8FTtKVcUocH6:-MVM-sgD8FTtKVcUocH7:b3vkhm6) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/LibOS/shim/test/regression/gettimeofday.c#L14)):*
> ```C
> 
> struct timeval base_tv;
> struct timeval tv[THREAD_NUM];
> ```

Why this is global array? You only use one entry at a time and only as a local variable.
___
*[LibOS/shim/test/regression/gettimeofday.c, line 25 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVM-HP82j7T1Hj5YSB8:-MVM-HP96o0yHkQrHGiK:b-bmej7i) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/LibOS/shim/test/regression/gettimeofday.c#L25)):*
> ```C
>             err(1, "thread gettimeofday");
> 
>         if (tv[idx].tv_sec < base_tv.tv_sec || tv[idx].tv_sec - base_tv.tv_sec > 20)
> ```

Please introduce a define for this constant value.
___
*[LibOS/shim/test/regression/gettimeofday.c, line 57 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVM-fhuABm5fqcywMV6:-MVM-fhuABm5fqcywMV7:bov84zg) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/LibOS/shim/test/regression/gettimeofday.c#L57)):*
> ```C
>                                                           : base_tv.tv_usec - end_tv.tv_usec;
>     printf("Finish time: %lu sec, %lu usec (passed: %lu sec, %lu usec)\n", end_tv.tv_sec,
>            end_tv.tv_usec, end_tv.tv_sec - base_tv.tv_sec, usec_diff);
> ```

We should subtract 1 from 3rd param if `end_tv.tv_usec < base_tv.tv_usec`.
___
*[Pal/include/lib/seqlock.h, line 23 at r1](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVLq0tk2QGWQToV-kpT:-MVM07G38eR10VsONo10:b-8ef97z) ([raw file](https://github.com/oscarlab/graphene/blob/ec42fcf811716c0f35beb3abca3c85b9ae249a7c/Pal/include/lib/seqlock.h#L23)):*

@boryspoplawski I haven't added the debugging stuff that you had in spinlocks. Do we need those?
</blockquote></details>

Nah, we can add them later if needed.
___
*[Pal/include/lib/seqlock.h, line 66 at r1](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVLq7jGCU1qE2cE5K3S:-MVM9NeM2RguDqGFFLO3:bo7q3bm) ([raw file](https://github.com/oscarlab/graphene/blob/ec42fcf811716c0f35beb3abca3c85b9ae249a7c/Pal/include/lib/seqlock.h#L66)):*

@boryspoplawski @yamahata I used `RMB` and `WMB` (read and write memory barriers) macros instead of relying on GCC atomics' memory ordering. I wasn't sure whether I can and should use `__ATOMIC_*` keywords instead of the explicit RMB/WMB. Please advise.
</blockquote></details>

`__ATOMIC_RELAXED` gives no synchronization, so we need these barrier.
I think we could use `ACQUIRE/RELEASE` in other places, to have the same (or actually stronger) guarantees as now, but this (`seqretry`) case is special and we cannot use any of those. In such case I prefer to have barriers everywhere.
___
*[Pal/include/lib/seqlock.h, line 59 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVLu2D8-BkZMHgt50vJ:-MVM275L3eslmuU5_svV:b-xs73uw) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/include/lib/seqlock.h#L59)):*

FYI: Interestingly, https://elixir.bootlin.com/linux/v3.19.8/source/include/linux/seqlock.h (Linux 3.19) didn't have this `RMB` but https://elixir.bootlin.com/linux/v5.11.4/source/include/linux/seqlock.h has it. So I added too.
</blockquote></details>

Both versions you've linked have the barrier (and it's essential for this to work correctly).
___
*[Pal/include/lib/seqlock.h, line 74 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMHce16V26gE6pWWWX:-MVMHce16V26gE6pWWWY:bg064w7) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/include/lib/seqlock.h#L74)):*
> ```objc
>  * \brief Start a writer-side critical section (acquires spinlock).
>  */
> static inline void write_seqlock(seqlock_t* sl) {
> ```

Could you rename this to `write_seqbegin` to match `read` version?
___
*[Pal/include/lib/seqlock.h, line 76 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMEviZ6X0lh5lgcu89:-MVMEviZ6X0lh5lgcu8A:b-8d2cp5) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/include/lib/seqlock.h#L76)):*
> ```objc
> static inline void write_seqlock(seqlock_t* sl) {
>     spinlock_lock(&sl->lock);
>     __atomic_add_fetch(&sl->sequence, 1, __ATOMIC_RELAXED);
> ```

Fun fact: kernel uses a plain `++` here (https://elixir.bootlin.com/linux/v5.11.4/source/include/linux/seqlock.h#L473) which I believe is incorrect. Nothing stops the compiler from doing something like:
```
x = load(seq);
*seq = 5;
x += 1;
*seq = x;
WMB();
```
and in such case it's clear that a reader can see an invalid value. As far as this case is unrealistic, there could be some others like: stores and reads get shredded into byte-by-byte read/write.
Otoh kernel uses `READ_ONCE` in readers, which might be enough (I think it prevents shredding)...
___
*[Pal/include/lib/seqlock.h, line 83 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMHhN44kLk1N2cvelD:-MVMHhN5AIYCq5DIkzlw:blhx4oh) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/include/lib/seqlock.h#L83)):*
> ```objc
>  * \brief End a writer-side critical section (releases spinlock).
>  */
> static inline void write_sequnlock(seqlock_t* sl) {
> ```

ditto (name)
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 54 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMJV5J5l9Vfwq3glgj:-MVMJV5J5l9Vfwq3glgk:b-p2j61c) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/src/host/Linux-SGX/db_misc.c#L54)):*
> ```C
>             return -PAL_ERROR_DENIED;
> 
>         *out_usec = usec;
> ```

Do we need the additional parameter? Why not just pass `out_usec` to `ocall_gettime`?
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 58 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMJQ5_3HZhEZdvZ8vB:-MVMJQ5_3HZhEZdvZ8vC:bw4slw6) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/src/host/Linux-SGX/db_misc.c#L58)):*
> ```C
>     }
> 
>     uint32_t seq;
> ```

Please move this inside `do {} while`.
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 65 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMKJiL8bRMGDP9YmRk:-MVMKJiL8bRMGDP9YmRl:b-sb0oik) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/src/host/Linux-SGX/db_misc.c#L65)):*
> ```C
>              * careful with integer overflow during calculations) */
>             uint64_t tsc_diff = get_tsc() - g_start_tsc;
>             if (tsc_diff < INT64_MAX / 1000000) {
> ```

Shouldn't this be `UINT64_MAX`?
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 66 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMK1lC7jZN7KVf6Zz_:-MVMK1lC7jZN7KVf6Zza:b-5xq8qk) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/src/host/Linux-SGX/db_misc.c#L66)):*
> ```C
>             uint64_t tsc_diff = get_tsc() - g_start_tsc;
>             if (tsc_diff < INT64_MAX / 1000000) {
>                 uint64_t tsc_usec = g_start_usec + tsc_diff * 1000000 / g_tsc_hz;
> ```

Don't we need `add_overflow` check here?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/test/regression/gettimeofday.c, line 14 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVM-sgD8FTtKVcUocH6:-MVN83yp1G9f6d2EKkh6:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/LibOS/shim/test/regression/gettimeofday.c#L14)):*

Why this is global array? You only use one entry at a time and only as a local variable.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/gettimeofday.c, line 25 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVM-HP82j7T1Hj5YSB8:-MVN8NV46R43PgGTQgIE:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/LibOS/shim/test/regression/gettimeofday.c#L25)):*

Please introduce a define for this constant value.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/gettimeofday.c, line 57 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVM-fhuABm5fqcywMV6:-MVNFWyNE7IkV9pFn2_M:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/LibOS/shim/test/regression/gettimeofday.c#L57)):*

We should subtract 1 from 3rd param if `end_tv.tv_usec < base_tv.tv_usec`.
</blockquote></details>

Done.
___
*[Pal/include/lib/seqlock.h, line 66 at r1](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVLq7jGCU1qE2cE5K3S:-MVNS5etBlX3Kmk89cmx:bnjom5k) ([raw file](https://github.com/oscarlab/graphene/blob/ec42fcf811716c0f35beb3abca3c85b9ae249a7c/Pal/include/lib/seqlock.h#L66)):*

`__ATOMIC_RELAXED` gives no synchronization, so we need these barrier.
I think we could use `ACQUIRE/RELEASE` in other places, to have the same (or actually stronger) guarantees as now, but this (`seqretry`) case is special and we cannot use any of those. In such case I prefer to have barriers everywhere.
</blockquote></details>

Ok, then I hope I added all the barriers correctly?
___
*[Pal/include/lib/seqlock.h, line 59 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVLu2D8-BkZMHgt50vJ:-MVNRzBwEVdKe5hs56rt:b7dv4i) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/include/lib/seqlock.h#L59)):*

Both versions you've linked have the barrier (and it's essential for this to work correctly).
</blockquote></details>

True, my eyes deceived me on 3.19.
___
*[Pal/include/lib/seqlock.h, line 74 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMHce16V26gE6pWWWX:-MVNSMbI8r1CK3R0ptMj:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/include/lib/seqlock.h#L74)):*

Could you rename this to `write_seqbegin` to match `read` version?
</blockquote></details>

Done.
___
*[Pal/include/lib/seqlock.h, line 83 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMHhN44kLk1N2cvelD:-MVNScaX2UjtlzMsSMtd:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/include/lib/seqlock.h#L83)):*

ditto (name)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 54 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMJV5J5l9Vfwq3glgj:-MVNGhM30AflxacDjS3R:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/src/host/Linux-SGX/db_misc.c#L54)):*

Do we need the additional parameter? Why not just pass `out_usec` to `ocall_gettime`?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 58 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMJQ5_3HZhEZdvZ8vB:-MVNJvxp50jJS-HI-i7o:b-fyg9q7) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/src/host/Linux-SGX/db_misc.c#L58)):*

Please move this inside `do {} while`.
</blockquote></details>

This doesn't work?
```
db_misc.c: In function ‘_DkSystemTimeQuery’:
db_misc.c:54:18: error: unused variable ‘seq’ [-Werror=unused-variable]
db_misc.c:70:41: error: ‘seq’ undeclared (first use in this function); did you mean ‘usec’?
```

Keeping as-is for now, maybe I'm being stupid but I don't think `seq` will still exist in `while (<condition>)`.
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 65 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMKJiL8bRMGDP9YmRk:-MVNLNdI5nLDFVy-nP2I:bwlq46u) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/src/host/Linux-SGX/db_misc.c#L65)):*

Shouldn't this be `UINT64_MAX`?
</blockquote></details>

Done. I guess at some point we had `int64_t` types here...
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 66 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMK1lC7jZN7KVf6Zz_:-MVNQsCv46TOxDcjf4Ls:b-actlme) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/src/host/Linux-SGX/db_misc.c#L66)):*

Don't we need `add_overflow` check here?
</blockquote></details>

Done. I really don't know what to do if time overflows and wraps around... For now, I'm doing a post-check (since we have `uint`) and return an OVERFLOW error.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, @yamahata, and @ying2liu)

___
*[Pal/include/lib/seqlock.h, line 66 at r1](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVLq7jGCU1qE2cE5K3S:-MVNh7Me3v59SeEMPbXI:b-de04pm) ([raw file](https://github.com/oscarlab/graphene/blob/ec42fcf811716c0f35beb3abca3c85b9ae249a7c/Pal/include/lib/seqlock.h#L66)):*

Ok, then I hope I added all the barriers correctly?
</blockquote></details>

Hopefully :)
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 58 at r2](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVMJQ5_3HZhEZdvZ8vB:-MVNk57A9BBNdOEnDUwB:bx73zw0) ([raw file](https://github.com/oscarlab/graphene/blob/7a77aa7a431bf589733a2323e3b29dd862b8842c/Pal/src/host/Linux-SGX/db_misc.c#L58)):*

This doesn't work?
```
db_misc.c: In function ‘_DkSystemTimeQuery’:
db_misc.c:54:18: error: unused variable ‘seq’ [-Werror=unused-variable]
db_misc.c:70:41: error: ‘seq’ undeclared (first use in this function); did you mean ‘usec’?
```

Keeping as-is for now, maybe I'm being stupid but I don't think `seq` will still exist in `while (<condition>)`.
</blockquote></details>

Sorry, apparently it does not work.
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 66 at r3](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVNhkMeBE_HeAbBOFGf:-MVNhkMeBE_HeAbBOFGg:b-q5xbnx) ([raw file](https://github.com/oscarlab/graphene/blob/95499a4edd4fb180555bfd989a9eb119ca67cd48/Pal/src/host/Linux-SGX/db_misc.c#L66)):*
> ```C
>                      * TSC/usec pair (time drift is contained), use the RDTSC-calculated time */
>                     usec = g_start_usec + diff_usec;
>                     if (usec < g_start_usec)
> ```

This case is tricky, but since we are inside seqlock, the value can be random garbage, unless we get out and confirm we passed the critical section successfully. Here we need to save the value, get out of the loop and only then check if it is correct (does not overflow).
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 73 at r3](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVNlkQVBl1BwGp3f-yI:-MVNlkQVBl1BwGp3f-yJ:be7co5y) ([raw file](https://github.com/oscarlab/graphene/blob/95499a4edd4fb180555bfd989a9eb119ca67cd48/Pal/src/host/Linux-SGX/db_misc.c#L73)):*
> ```C
>     } while (read_seqretry(&g_tsc_lock, seq));
> 
>     if (usec) {
> ```

If one loop fails (due to parallel writer), and the second due to int overflow, we will still exit successfully here.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: 5 of 6 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, @yamahata, and @ying2liu)

___
*[Pal/src/host/Linux-SGX/db_misc.c, line 66 at r3](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVNhkMeBE_HeAbBOFGf:-MVQ97FxBlLGRqefhtB4:b-1m9zjr) ([raw file](https://github.com/oscarlab/graphene/blob/95499a4edd4fb180555bfd989a9eb119ca67cd48/Pal/src/host/Linux-SGX/db_misc.c#L66)):*

This case is tricky, but since we are inside seqlock, the value can be random garbage, unless we get out and confirm we passed the critical section successfully. Here we need to save the value, get out of the loop and only then check if it is correct (does not overflow).
</blockquote></details>

Done. Good catch. I decided to move all these calculations outside of the critical section -- we only need to read a consistent set of global vars in the critical section, and the rest can be done outside of it.
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 73 at r3](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVNlkQVBl1BwGp3f-yI:-MVQ9NHVCZucN9AbhOF4:br2il1u) ([raw file](https://github.com/oscarlab/graphene/blob/95499a4edd4fb180555bfd989a9eb119ca67cd48/Pal/src/host/Linux-SGX/db_misc.c#L73)):*

If one loop fails (due to parallel writer), and the second due to int overflow, we will still exit successfully here.
</blockquote></details>

I don't understand your scenario. Are you meaning two parallel threads?

yamahata (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: 5 of 6 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, @yamahata, and @ying2liu)

___
*[Pal/src/host/Linux-SGX/db_misc.c, line 58 at r4](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVRqTXyD_KMoXmPGd0l:-MVRqTXyD_KMoXmPGd0m:b-xqi8gg) ([raw file](https://github.com/oscarlab/graphene/blob/44e965224f888ee1d8fc98f3559c60168d8df60d/Pal/src/host/Linux-SGX/db_misc.c#L58)):*
> ```
>         start_tsc  = READ_ONCE(g_start_tsc);
>         start_usec = READ_ONCE(g_start_usec);
> ```

Because The integrity is guaranteed by seqlock, READ_ONCE is overkill.
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 68 at r4](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVRqeDT94TAcOi0oUTh:-MVRqeDT94TAcOi0oUTi:blbcsfq) ([raw file](https://github.com/oscarlab/graphene/blob/44e965224f888ee1d8fc98f3559c60168d8df60d/Pal/src/host/Linux-SGX/db_misc.c#L68)):*
> ```
>         if (diff_tsc < UINT64_MAX / 1000000) {
>             uint64_t diff_usec = diff_tsc * 1000000 / g_tsc_hz;
>             if (diff_usec < TSC_REFINE_INIT_TIMEOUT_USECS) {
> ```

These two if clause can be done in single if clause without UINT64_MAX.
But it's a matter of preference. Non-blocking

yamahata (commented):
> Reviewed 3 of 6 files at r1, 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: 5 of 6 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, @yamahata, and @ying2liu)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow, @yamahata, and @ying2liu)

___
*[Pal/src/host/Linux-SGX/db_misc.c, line 73 at r3](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVNlkQVBl1BwGp3f-yI:-MVRu4QYCQgOrun93SCu:b-fh64py) ([raw file](https://github.com/oscarlab/graphene/blob/95499a4edd4fb180555bfd989a9eb119ca67cd48/Pal/src/host/Linux-SGX/db_misc.c#L73)):*

I don't understand your scenario. Are you meaning two parallel threads?
</blockquote></details>

No, I meant first loop iteration vs second iteration. 2nd could fail this `if (diff_tsc < UINT64_MAX / 1000000)`.
Anyway, with the new version the issue is not present anymore.
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 58 at r4](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVRqTXyD_KMoXmPGd0l:-MVRtE-NDXnuqxDfAW04:b-ocirl) ([raw file](https://github.com/oscarlab/graphene/blob/44e965224f888ee1d8fc98f3559c60168d8df60d/Pal/src/host/Linux-SGX/db_misc.c#L58)):*

> ```
>         start_tsc  = READ_ONCE(g_start_tsc);
>         start_usec = READ_ONCE(g_start_usec);
> ```

Because The integrity is guaranteed by seqlock, READ_ONCE is overkill.

</blockquote></details>

+1 we do not need them. Any reason why you added them now?
___
*[Pal/src/host/Linux-SGX/db_misc.c, line 68 at r4](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVRqeDT94TAcOi0oUTh:-MVRupre2Gc-HK85AoCM:b-rngmim) ([raw file](https://github.com/oscarlab/graphene/blob/44e965224f888ee1d8fc98f3559c60168d8df60d/Pal/src/host/Linux-SGX/db_misc.c#L68)):*

> ```
>         if (diff_tsc < UINT64_MAX / 1000000) {
>             uint64_t diff_usec = diff_tsc * 1000000 / g_tsc_hz;
>             if (diff_usec < TSC_REFINE_INIT_TIMEOUT_USECS) {
> ```

These two if clause can be done in single if clause without UINT64_MAX.
But it's a matter of preference. Non-blocking
</blockquote></details>

Hmm, how exactly would we do that? How to make sure we do not overflow?
We could use `__builtin_mul_overflow`, but that's different approach entirely.

yamahata (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, @yamahata, and @ying2liu)

___
*[Pal/src/host/Linux-SGX/db_misc.c, line 68 at r4](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVRqeDT94TAcOi0oUTh:-MVRx3Tg6mMnU9EMIAAR:b-ef6ol) ([raw file](https://github.com/oscarlab/graphene/blob/44e965224f888ee1d8fc98f3559c60168d8df60d/Pal/src/host/Linux-SGX/db_misc.c#L68)):*

Hmm, how exactly would we do that? How to make sure we do not overflow?
We could use `__builtin_mul_overflow`, but that's different approach entirely.
</blockquote></details>

builtin is ugly. Let's keep this as is. I scratch my comment here.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: 5 of 6 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @yamahata)

___
*[Pal/src/host/Linux-SGX/db_misc.c, line 58 at r4](https://reviewable.io/reviews/oscarlab/graphene/2222#-MVRqTXyD_KMoXmPGd0l:-MVRxqQg7y2upFOgbjqa:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/44e965224f888ee1d8fc98f3559c60168d8df60d/Pal/src/host/Linux-SGX/db_misc.c#L58)):*

I wanted to make it explicit, but yes, no reason other than "in your face readability". Will change to normal assignment.
</blockquote></details>

Done.

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2222)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv:
Jenkins, retest Jenkins-Debug please (Jenkins log was removed due to other jobs in the meantime, don't know the reason for failing)

PR #2224, opened 2021-03-10
Cleanup shim_rtld

pwmarcz:
This is an incremental refactoring. I'm planning more (mostly, split the huge `__map_elf_object` function), but I don't want to submit it all at once, so I hope I can get these out of the way first.

## Description of the changes 

See commit messages for more details. The most important change is that we don't run the ELF loader on an already-mapped code, which necessitated many special cases (`OBJ_INTERNAL`, `OBJ_VDSO`, `OBJ_MAPPED`).

## How to test this PR? 

The current tests should be enough.

## Planned next steps

Feedback appreciated.

* **Split `__map_elf_object` into smaller parts**, and clean it up some more (weird variables, `goto`, vestigial `RELOCATE` macro, etc.)
* Maybe get rid of `search_map_by_{name,handle}`? We want to load the ELF and then execute it, there should be no need to decouple these steps.
* Simplify checkpointing: I think the only thing we do is migrate data to another process (for which the list of libraries is empty).
* Apply the lessons learned here to PAL (I wouldn't go as far as extracting a common part, the use cases are different)

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)


mkow (changes requested):
> Reviewed 2 of 2 files at r1, 3 of 3 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 317 at r3](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVSQuJo1e5jks23I8sd:-MVSQuJp0izcA7tQQacb:bof8xjy) ([raw file](https://github.com/oscarlab/graphene/blob/33740d82c3ade1a8e52622b728239a65b417f1d0/LibOS/shim/src/shim_rtld.c#L317)):*
> ```C
> 
>         if (l->l_phdr == 0 && (ElfW(Off))c->mapoff <= ehdr->e_phoff &&
>             ((size_t)(c->mapend - c->mapstart + c->mapoff) >= phdr_size))
> ```

could you indent this line one level more? just to make it more obvious that it's still a part of the condition
___
*[LibOS/shim/src/shim_rtld.c, line 381 at r3](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVSR4UeEZJL2FOIuV7n:-MVSR4UeEZJL2FOIuV7o:b-quwuvy) ([raw file](https://github.com/oscarlab/graphene/blob/33740d82c3ade1a8e52622b728239a65b417f1d0/LibOS/shim/src/shim_rtld.c#L381)):*
> ```C
>         /* The program header is not contained in any of the segments. We have to allocate memory
>            ourself.
>         */
> ```

Please use the following style:
```
/* asdf
 * 11234
 */
```

boryspoplawski (changes requested):
> Reviewed 2 of 2 files at r1, 3 of 3 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow and @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 120 at r2](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVSRW4_DLW4r5VTQvKm:-MVSRW4_DLW4r5VTQvKn:b-pqxp8k) ([raw file](https://github.com/oscarlab/graphene/blob/65503626d9e906b70b82a083272dd98382666267/LibOS/shim/src/shim_rtld.c#L120)):*
> ```C
> static struct link_map* __map_elf_object(struct shim_handle* file, const void* fbp, size_t fbp_len) {
>     ElfW(Phdr)* new_phdr = NULL;
>     if (!(file && file->fs && file->fs->fs_ops))
> ```

Was there any reason the old code allowed for `file` to be `NULL`?
___
*[LibOS/shim/src/shim_rtld.c, line 317 at r3](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVSQuJo1e5jks23I8sd:-MVSYjSn4RUQWYW2BcZE:b-74tm21) ([raw file](https://github.com/oscarlab/graphene/blob/33740d82c3ade1a8e52622b728239a65b417f1d0/LibOS/shim/src/shim_rtld.c#L317)):*

could you indent this line one level more? just to make it more obvious that it's still a part of the condition
</blockquote></details>

Don't we also have convention of moving operator to the new line?
___
*[LibOS/shim/src/shim_rtld.c, line 381 at r3](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVSR4UeEZJL2FOIuV7n:-MVSZB5MBj9Ka8dio97b:b-i8z6ov) ([raw file](https://github.com/oscarlab/graphene/blob/33740d82c3ade1a8e52622b728239a65b417f1d0/LibOS/shim/src/shim_rtld.c#L381)):*

Please use the following style:
```
/* asdf
 * 11234
 */
```
</blockquote></details>

Actually please use either:
```
/* aaaaaaaaaaa
 * bbbbbbbbbbbbbbbbbbb */
```
or
```
/*
 * aaaaaaaaaaaaaaa
 * bbbbbbbbbbbbbbbbbbbb
 */
```
___
*[LibOS/shim/src/shim_rtld.c, line 559 at r3](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVS_7E34pwNI-DNsq9h:-MVS_7E4ChWldT27A7uj:bmqrtn1) ([raw file](https://github.com/oscarlab/graphene/blob/33740d82c3ade1a8e52622b728239a65b417f1d0/LibOS/shim/src/shim_rtld.c#L559)):*
> ```C
>     if ((size_t)ret < size)
>         return -EINVAL;
>     return 0;
> ```

The old version rewind handle to offset `0`, is that not needed anymore? (TBH I do not understand how this mess works as a whole).

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 317 at r3](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVSQuJo1e5jks23I8sd:-MVSgXB537-4N-5V0Trn:b-szr3xu) ([raw file](https://github.com/oscarlab/graphene/blob/33740d82c3ade1a8e52622b728239a65b417f1d0/LibOS/shim/src/shim_rtld.c#L317)):*

Don't we also have convention of moving operator to the new line?
</blockquote></details>

Officially not, but +1 to doing this from my side.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/src/shim_rtld.c, line 120 at r2](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVSRW4_DLW4r5VTQvKm:-MVSjXEA4FT5E45BofdH:b-luv50y) ([raw file](https://github.com/oscarlab/graphene/blob/65503626d9e906b70b82a083272dd98382666267/LibOS/shim/src/shim_rtld.c#L120)):*

Was there any reason the old code allowed for `file` to be `NULL`?
</blockquote></details>

Yes. `file` was null when we called this function on LibOS object itself, and on the vDSO page, both of which were already mapped. The code was supposed to read the program header directly from memory, and there were many `if`s disabling mmaps in these cases.

I believe this stopped being necessary when I removed dynamic linking: before, we needed to load symbols from LibOS (and also do a second pass of "Graphene-aware" dynamic linking for the already-loaded code).
___
*[LibOS/shim/src/shim_rtld.c, line 381 at r3](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVSR4UeEZJL2FOIuV7n:-MVSlVQb-l7k8Wdj6FKQ:b-ajrf8x) ([raw file](https://github.com/oscarlab/graphene/blob/33740d82c3ade1a8e52622b728239a65b417f1d0/LibOS/shim/src/shim_rtld.c#L381)):*

Actually please use either:
```
/* aaaaaaaaaaa
 * bbbbbbbbbbbbbbbbbbb */
```
or
```
/*
 * aaaaaaaaaaaaaaa
 * bbbbbbbbbbbbbbbbbbbb
 */
```
</blockquote></details>

ACK, will do (and also `ourself` -> `ourselves`)
___
*[LibOS/shim/src/shim_rtld.c, line 559 at r3](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVS_7E34pwNI-DNsq9h:-MVSl_MU6wU1K4RUuYYA:bldyxvp) ([raw file](https://github.com/oscarlab/graphene/blob/33740d82c3ade1a8e52622b728239a65b417f1d0/LibOS/shim/src/shim_rtld.c#L559)):*

The old version rewind handle to offset `0`, is that not needed anymore? (TBH I do not understand how this mess works as a whole).
</blockquote></details>

AFAIK everything that touches this handle does an explicit seek, and the rewind is not necessary. But I will double-check.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/src/shim_rtld.c, line 120 at r2](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVSRW4_DLW4r5VTQvKm:-MVSncY_AQuBqt02u_1O:b-wsvasg) ([raw file](https://github.com/oscarlab/graphene/blob/65503626d9e906b70b82a083272dd98382666267/LibOS/shim/src/shim_rtld.c#L120)):*

Yes. `file` was null when we called this function on LibOS object itself, and on the vDSO page, both of which were already mapped. The code was supposed to read the program header directly from memory, and there were many `if`s disabling mmaps in these cases.

I believe this stopped being necessary when I removed dynamic linking: before, we needed to load symbols from LibOS (and also do a second pass of "Graphene-aware" dynamic linking for the already-loaded code).
</blockquote></details>

Thanks for the explanation!

dimakuv (changes requested):
> Reviewed 1 of 2 files at r1, 2 of 3 files at r2, 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVHZ-xCJbwSCBHOhSO:-MVVHZ-xCJbwSCBHOhSP:b-90lae1) (no related file):*
> Maybe get rid of search_map_by_{name,handle}? We want to load the ELF and then execute it, there should be no need to decouple these steps.

Don't you need search functions for things like `init_brk_from_executable()`, `execute_elf_object()` and checkpoint func? On the other hand, you can just have a single global variable that points to the executable info, instead of searching through the link_map.

---
___
*[LibOS/shim/src/shim_rtld.c, line 61 at r5](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVGN1XCVOstnFumg1o:-MVVGN1XCVOstnFumg1p:b-n7asds) ([raw file](https://github.com/oscarlab/graphene/blob/2f757200621e536ca73ea755a0a58686eca8816b/LibOS/shim/src/shim_rtld.c#L61)):*
> ```C
>     bool l_phdr_allocated;
>     ElfW(Addr) l_entry;        /* Entry point location.  */
>     ElfW(Half) l_phnum;        /* Number of program header entries.  */
> ```

You have strange alignments of comments here. Please fix.
___
*[LibOS/shim/src/shim_rtld.c, line 527 at r5](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVE9G4DsMdjX2wD_Sn:-MVVE9G5CYrHyL7WydP6:bccdzqs) ([raw file](https://github.com/oscarlab/graphene/blob/2f757200621e536ca73ea755a0a58686eca8816b/LibOS/shim/src/shim_rtld.c#L527)):*
> ```C
> 
> static int __load_elf_header(struct shim_handle* file, ElfW(Ehdr)* ehdr) {
>     int ret = read_file_fragment(file, ehdr, 0, sizeof(*ehdr));
> ```

Could you add a named argument here: `0` -> `/*offset=*/0`
___
*[LibOS/shim/src/shim_rtld.c, line 541 at r5](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVEELa8MB8H1YGqIJH:-MVVEELa8MB8H1YGqIJI:bccdzqs) ([raw file](https://github.com/oscarlab/graphene/blob/2f757200621e536ca73ea755a0a58686eca8816b/LibOS/shim/src/shim_rtld.c#L541)):*
> ```C
>     ElfW(Ehdr) ehdr;
> 
>     int ret = read_file_fragment(file, &ehdr, 0, sizeof(ehdr));
> ```

Could you add a named argument here: `0` -> `/*offset=*/0`
___
*[LibOS/shim/src/elf/shim_rtld.c, line 42 at r5](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVEdRE5mz8tF-fYWlf:-MVVEdRE5mz8tF-fYWlg:b-hrmfoc) ([raw file](https://github.com/oscarlab/graphene/blob/cea083122941c52b8e9b21dfd327364ffbc3f0cb/LibOS/shim/src/elf/shim_rtld.c#L42)):*
> ```C
>     OBJECT_MAPPED   = 2,
>     OBJECT_VDSO     = 3,
> };
> ```

For my own understanding: now we only have the case of `OBJECT_LOAD` because:
- `OBJECT_INTERNAL` was LibOS shared lib (`libsysdb.so`) which is not loaded by this loader anyway
- `OBJECT_VDSO` was our vDSO, but it is specially-loaded (mmapped and initialized) by this loader in `vdso_map_init()` -- and there is never a need for applications to search for vDSO (I am unsure about this -- is vDSO completely transparent to the applications?)
- `OBJECT_MAPPED` was a hack to allow double-adding of objects (e.g., they were somehow double-added during a checkpoint restore), now there is no need for this after your recent changes

Is this all correct?
___
*[LibOS/shim/src/elf/shim_rtld.c, line 131 at r5](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVGldk1wYpi0mZ62wA:-MVVGldl45kumcAQc-5T:bq07j3n) ([raw file](https://github.com/oscarlab/graphene/blob/cea083122941c52b8e9b21dfd327364ffbc3f0cb/LibOS/shim/src/elf/shim_rtld.c#L131)):*
> ```C
>    opened on FD */
> static struct link_map* __map_elf_object(struct shim_handle* file, const void* fbp, size_t fbp_len,
>                                          void* addr, int type) {
> ```

For my own understanding: we're getting rid of `addr` because it was previously used to hint that the object is already loaded in memory (e.g., for vDSO and for libsysdb)?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: 4 of 5 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVHZ-xCJbwSCBHOhSO:-MVVzEKd6RG0xZ4Cp9jn:bffeuj3) (no related file):*
>On the other hand, you can just have a single global variable that points to the executable info, instead of searching through the link_map.

I think that makes sense. I'd rather not pretend that this is a generic loader for all kinds of ELF binaries: these functions already kind of assume there is a specific executable binary (and optionally, an interpreter binary).

As for checkpointing: is there actually a case where we want to *replace* an existing link_map record? I thought we only transfer a list of of records to a newly created process.

---
___
*[LibOS/shim/src/shim_rtld.c, line 317 at r3](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVSQuJo1e5jks23I8sd:-MVW8LEZEFPU0Gg6tI-X:b-80i6qo) ([raw file](https://github.com/oscarlab/graphene/blob/33740d82c3ade1a8e52622b728239a65b417f1d0/LibOS/shim/src/shim_rtld.c#L317)):*

Officially not, but +1 to doing this from my side.
</blockquote></details>

Both done, and I added braces so that it reads better (because of the comment after `if`).
___
*[LibOS/shim/src/shim_rtld.c, line 381 at r3](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVSR4UeEZJL2FOIuV7n:-MVW8jnB8Zxhgzvwfw2T:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/33740d82c3ade1a8e52622b728239a65b417f1d0/LibOS/shim/src/shim_rtld.c#L381)):*

ACK, will do (and also `ourself` -> `ourselves`)
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_rtld.c, line 559 at r3](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVS_7E34pwNI-DNsq9h:-MVW9tl72HLHpuOMxmlm:b-a6iu8g) ([raw file](https://github.com/oscarlab/graphene/blob/33740d82c3ade1a8e52622b728239a65b417f1d0/LibOS/shim/src/shim_rtld.c#L559)):*

AFAIK everything that touches this handle does an explicit seek, and the rewind is not necessary. But I will double-check.
</blockquote></details>

Okay, as far as I can tell the only case where this handle is visible outside is `g_process.exec`, and we don't do any other read operations on it. 

In any case, I don't think it's reasonable to expect a specific offset for this handle, and it's better not to rewind so that we don't mask bugs when somebody tries reading from it.
___
*[LibOS/shim/src/shim_rtld.c, line 61 at r5](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVGN1XCVOstnFumg1o:-MVW1bHm5LTgrRk7HejP:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/2f757200621e536ca73ea755a0a58686eca8816b/LibOS/shim/src/shim_rtld.c#L61)):*

You have strange alignments of comments here. Please fix.
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_rtld.c, line 527 at r5](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVE9G4DsMdjX2wD_Sn:-MVWAj3C32S1fdsLLpF6:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/2f757200621e536ca73ea755a0a58686eca8816b/LibOS/shim/src/shim_rtld.c#L527)):*

Could you add a named argument here: `0` -> `/*offset=*/0`
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_rtld.c, line 541 at r5](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVEELa8MB8H1YGqIJH:-MVWAjZdC_qQwbdo4-Hk:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/2f757200621e536ca73ea755a0a58686eca8816b/LibOS/shim/src/shim_rtld.c#L541)):*

Could you add a named argument here: `0` -> `/*offset=*/0`
</blockquote></details>

Done.
___
*[LibOS/shim/src/elf/shim_rtld.c, line 42 at r5](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVEdRE5mz8tF-fYWlf:-MVVzePQABK1oy62kUXm:b-37xm60) ([raw file](https://github.com/oscarlab/graphene/blob/cea083122941c52b8e9b21dfd327364ffbc3f0cb/LibOS/shim/src/elf/shim_rtld.c#L42)):*
Mostly. In short, there is no need to keep track of all these objects anymore. That job is already done by glibc/ld.so (except for LibOS itself, which is "kernel space" as far as the application is concerned).

>`OBJECT_INTERNAL` was LibOS shared lib (libsysdb.so) which is not loaded by this loader anyway

Yes. We did need to read it for the shared symbols (`syscalldb` etc.), but don't need to anymore.

>`OBJECT_VDSO` was our vDSO, but it is specially-loaded (mmapped and initialized) by this loader in vdso_map_init() -- and there is never a need for applications to search for vDSO (I am unsure about this -- is vDSO completely transparent to the applications?)

The applications do use the vDSO, but we don't need to keep track of it here. The vDSO address is provided in the aux vector when running the application (under `AT_SYSINFO_EHDR`). Then `ld.so` picks it up and loads it to *its own* link map, so that it's available to glibc (and to user applications using `dlopen()`).

>`OBJECT_MAPPED` was a hack to allow double-adding of objects (e.g., they were somehow double-added during a checkpoint restore), now there is no need for this after your recent changes

I don't think that happened during checkpoint restore, but we did add objects that were already loaded by `ld.so` (via `register_library`). This was in order to perform the dynamic linking again.
___
*[LibOS/shim/src/elf/shim_rtld.c, line 131 at r5](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVGldk1wYpi0mZ62wA:-MVW1tOA4CLf1oSJXAif:bb77ufg) ([raw file](https://github.com/oscarlab/graphene/blob/cea083122941c52b8e9b21dfd327364ffbc3f0cb/LibOS/shim/src/elf/shim_rtld.c#L131)):*

For my own understanding: we're getting rid of `addr` because it was previously used to hint that the object is already loaded in memory (e.g., for vDSO and for libsysdb)?
</blockquote></details>

Yes.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: all files reviewed, 4 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVHZ-xCJbwSCBHOhSO:-MVW_6iSBw9lHuS4Ai1b:b-ysudxq) (no related file):*

>On the other hand, you can just have a single global variable that points to the executable info, instead of searching through the link_map.

I think that makes sense. I'd rather not pretend that this is a generic loader for all kinds of ELF binaries: these functions already kind of assume there is a specific executable binary (and optionally, an interpreter binary).

As for checkpointing: is there actually a case where we want to *replace* an existing link_map record? I thought we only transfer a list of of records to a newly created process.
</blockquote></details>

Regarding checkpointing: to be honest, I don't know. I haven't looked deeply into checkpoint + loaded objects.

---
___
*[LibOS/shim/src/elf/shim_rtld.c, line 42 at r5](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVVEdRE5mz8tF-fYWlf:-MVW_MPgA9F4Qs2R6Q8B:btif8r7) ([raw file](https://github.com/oscarlab/graphene/blob/cea083122941c52b8e9b21dfd327364ffbc3f0cb/LibOS/shim/src/elf/shim_rtld.c#L42)):*
> Then ld.so picks it up and loads it to its own link map, so that it's available to glibc (and to user applications using dlopen()).

Sweet, this is exactly what I wanted to know.

yamahata (dismissed):
> Reviewed 1 of 2 files at r1, 2 of 3 files at r2, 1 of 2 files at r4, 1 of 1 files at r5, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: all files reviewed, 4 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/src/shim_rtld.c, line 559 at r3](https://reviewable.io/reviews/oscarlab/graphene/2224#-MVS_7E34pwNI-DNsq9h:-MVXNgi52Pa1C1kDsUw4:b-dxcbts) ([raw file](https://github.com/oscarlab/graphene/blob/33740d82c3ade1a8e52622b728239a65b417f1d0/LibOS/shim/src/shim_rtld.c#L559)):*

Okay, as far as I can tell the only case where this handle is visible outside is `g_process.exec`, and we don't do any other read operations on it. 

In any case, I don't think it's reasonable to expect a specific offset for this handle, and it's better not to rewind so that we don't mask bugs when somebody tries reading from it.
</blockquote></details>

Makes sense.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

yamahata (approved):
> Reviewed 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: :shipit: complete! all files reviewed, all discussions resolved

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2224)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2225, opened 2021-03-10
 [LibOS] Connect to IPC leader at process startup

boryspoplawski:
## Description of the changes 
Previously Graphene made IPC connections on demand and there was a hidden assumption that the main process is an IPC leader that never exits. Although there was some code to find new IPC leader in case the old one dies, it was broken and only partialy used.
This commit causes Graphene processes to connect to the IPC leader on startup and disconnect of the leader is treated as a fatal error in the child process.

## How to test this PR? 
Added a new LibOS regression test, which checks if we have a working IPC connection even if our parent (intermediate process) dies.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)

### DIFF
--- a/LibOS/shim/include/shim_ipc.h
+++ b/LibOS/shim/include/shim_ipc.h
@@ -44,11 +44,8 @@ enum {
 enum {
     IPC_MSG_RESP = 0,
     IPC_MSG_CHILDEXIT,
-    IPC_MSG_FINDNS,
-    IPC_MSG_TELLNS,
     IPC_MSG_LEASE,
     IPC_MSG_OFFER,
-    IPC_MSG_RENEW,
     IPC_MSG_SUBLEASE,
     IPC_MSG_QUERY,
     IPC_MSG_QUERYALL,
@@ -87,7 +84,6 @@ struct shim_ipc_info {
 
 struct shim_process_ipc_info {
     IDTYPE vmid;
-    struct shim_lock lock;
     struct shim_ipc_info* self;
     struct shim_ipc_info* parent;
     struct shim_ipc_info* ns;
@@ -137,7 +133,6 @@ int add_ipc_subrange(IDTYPE idx, IDTYPE owner, const char* uri);
 IDTYPE allocate_ipc_id(IDTYPE min, IDTYPE max);
 void release_ipc_id(IDTYPE idx);
 
-int connect_ns(IDTYPE* vmid, struct shim_ipc_port** portptr);
 int connect_owner(IDTYPE idx, struct shim_ipc_port** portptr, IDTYPE* owner);
 
 /* sysv namespace */
@@ -176,20 +171,6 @@ struct shim_ipc_cld_exit {
 int ipc_cld_exit_send(unsigned int exitcode, unsigned int term_signal);
 int ipc_cld_exit_callback(struct shim_ipc_msg* msg, struct shim_ipc_port* port);
 
-/* FINDNS: find namespace leader (its IPC port) */
-int ipc_findns_send(bool block);
-int ipc_findns_callback(struct shim_ipc_msg* msg, struct shim_ipc_port* port);
-
-/* TELLNS: tell namespace leader (its IPC port) */
-struct shim_ipc_tellns {
-    IDTYPE vmid;
-    char uri[1];
-} __attribute__((packed));
-
-int ipc_tellns_send(struct shim_ipc_port* port, IDTYPE dest, struct shim_ipc_info* leader,
-                    unsigned long seq);
-int ipc_tellns_callback(struct shim_ipc_msg* msg, struct shim_ipc_port* port);
-
 /* LEASE: lease a range of IDs */
 struct shim_ipc_lease {
     char uri[1];
@@ -207,15 +188,6 @@ struct shim_ipc_offer {
 int ipc_offer_send(struct shim_ipc_port* port, IDTYPE dest, IDTYPE base, IDTYPE size,
                    unsigned long seq);
 int ipc_offer_callback(struct shim_ipc_msg* msg, struct shim_ipc_port* port);
-
-/* RENEW: renew lease of a range of IDs */
-struct shim_ipc_renew {
-    IDTYPE base;
-    IDTYPE size;
-} __attribute__((packed));
-
-int ipc_renew_send(IDTYPE base, IDTYPE size);
-int ipc_renew_callback(struct shim_ipc_msg* msg, struct shim_ipc_port* port);
 
 /* SUBLEASE: lease a range of IDs */
 struct shim_ipc_sublease {
@@ -413,7 +385,6 @@ struct shim_process_ipc_info* create_process_ipc_info(void);
 void free_process_ipc_info(struct shim_process_ipc_info* process);
 
 struct shim_ipc_info* create_ipc_info_and_port(bool use_vmid_as_port_name);
-int get_ipc_info_cur_process(struct shim_ipc_info** pinfo);
 
 void add_ipc_port_by_id(IDTYPE vmid, PAL_HANDLE hdl, IDTYPE type, port_fini fini,
                         struct shim_ipc_port** portptr);
@@ -456,8 +427,6 @@ int send_response_ipc_message(struct shim_ipc_port* port, IDTYPE dest, int ret, 
 void ipc_port_with_child_fini(struct shim_ipc_port* port, IDTYPE vmid);
 
 struct shim_thread* terminate_ipc_helper(void);
-
-int prepare_ipc_leader(void);
 
 int init_ipc_ports(void);
 int init_ns_ranges(void);
--- a/LibOS/shim/src/ipc/shim_ipc.c
+++ b/LibOS/shim/src/ipc/shim_ipc.c
@@ -45,7 +45,6 @@ int init_ipc(void) {
     int ret = 0;
 
     if (!create_lock(&ipc_info_lock)
-        || !create_lock(&g_process_ipc_info.lock)
         || !create_lock(&ipc_info_mgr_lock)) {
         return -ENOMEM;
     }
@@ -168,8 +167,6 @@ struct shim_process_ipc_info* create_process_ipc_info(void) {
     if (!new_process_ipc_info)
         return NULL;
 
-    lock(&g_process_ipc_info.lock);
-
     /* current process must have been initialized with info on its own IPC info */
     assert(g_process_ipc_info.self);
 
@@ -182,20 +179,17 @@ struct shim_process_ipc_info* create_process_ipc_info(void) {
     if (!new_process_ipc_info->parent)
         goto fail;
 
-    /* new process inherits the same namespace leader */
-    if (g_process_ipc_info.ns) {
-        new_process_ipc_info->ns = create_ipc_info(g_process_ipc_info.ns->vmid,
-                                                   qstrgetstr(&g_process_ipc_info.ns->uri),
-                                                   g_process_ipc_info.ns->uri.len);
-        if (!new_process_ipc_info->ns)
-            goto fail;
-    }
+    assert(g_process_ipc_info.ns);
+    assert(!qstrempty(&g_process_ipc_info.ns->uri));
+    new_process_ipc_info->ns = create_ipc_info(g_process_ipc_info.ns->vmid,
+                                               qstrgetstr(&g_process_ipc_info.ns->uri),
+                                               g_process_ipc_info.ns->uri.len);
+    if (!new_process_ipc_info->ns)
+        goto fail;
 
-    unlock(&g_process_ipc_info.lock);
     return new_process_ipc_info;
 
 fail:
-    unlock(&g_process_ipc_info.lock);
     free_process_ipc_info(new_process_ipc_info);
     return NULL;
 }
@@ -334,8 +328,6 @@ out:
 }
 
 struct shim_ipc_info* create_ipc_info_and_port(bool use_vmid_as_port_name) {
-    assert(locked(&g_process_ipc_info.lock));
-
     struct shim_ipc_info* info = create_ipc_info(g_process_ipc_info.vmid, NULL, 0);
     if (!info)
         return NULL;
@@ -356,24 +348,6 @@ struct shim_ipc_info* create_ipc_info_and_port(bool use_vmid_as_port_name) {
     }
 
     return info;
-}
-
-int get_ipc_info_cur_process(struct shim_ipc_info** info) {
-    lock(&g_process_ipc_info.lock);
-
-    if (!g_process_ipc_info.self) {
-        g_process_ipc_info.self = create_ipc_info_and_port(/*use_vmid_as_port_name=*/true);
-        if (!g_process_ipc_info.self) {
-            unlock(&g_process_ipc_info.lock);
-            return -EACCES;
-        }
-    }
-
-    get_ipc_info(g_process_ipc_info.self);
-    *info = g_process_ipc_info.self;
-
-    unlock(&g_process_ipc_info.lock);
-    return 0;
 }
 
 BEGIN_CP_FUNC(ipc_info) {
@@ -465,8 +439,6 @@ BEGIN_RS_FUNC(process_ipc_info) {
         get_ipc_info(process_ipc_info->ns);
 
     g_process_ipc_info = *process_ipc_info;
-    // this lock will be created in init_ipc
-    clear_lock(&g_process_ipc_info.lock);
 
     DEBUG_RS("vmid=%u,uri=%s,parent=%u(%s)", process_ipc_info->vmid,
              process_ipc_info->self ? qstrgetstr(&process_ipc_info->self->uri) : "",
--- a/LibOS/shim/src/ipc/shim_ipc_helper.c
+++ b/LibOS/shim/src/ipc/shim_ipc_helper.c
@@ -48,11 +48,8 @@ typedef int (*ipc_callback)(struct shim_ipc_msg* msg, struct shim_ipc_port* port
 static ipc_callback ipc_callbacks[] = {
     [IPC_MSG_RESP]          = &ipc_resp_callback,
     [IPC_MSG_CHILDEXIT]     = &ipc_cld_exit_callback,
-    [IPC_MSG_FINDNS]        = &ipc_findns_callback,
-    [IPC_MSG_TELLNS]        = &ipc_tellns_callback,
     [IPC_MSG_LEASE]         = &ipc_lease_callback,
     [IPC_MSG_OFFER]         = &ipc_offer_callback,
-    [IPC_MSG_RENEW]         = &ipc_renew_callback,
     [IPC_MSG_SUBLEASE]      = &ipc_sublease_callback,
     [IPC_MSG_QUERY]         = &ipc_query_callback,
     [IPC_MSG_QUERYALL]      = &ipc_queryall_callback,
@@ -74,67 +71,68 @@ static ipc_callback ipc_callbacks[] = {
 };
 
 static int init_self_ipc_port(void) {
-    lock(&g_process_ipc_info.lock);
-
     assert(!g_process_ipc_info.self);
     /* very first process or clone/fork case: create IPC port from scratch */
     g_process_ipc_info.self = create_ipc_info_and_port(/*use_vmid_as_port_name=*/true);
     if (!g_process_ipc_info.self) {
-        unlock(&g_process_ipc_info.lock);
-        return -EACCES;
+        return -ENOMEM;
     }
 
-    unlock(&g_process_ipc_info.lock);
     return 0;
 }
 
 static int init_parent_ipc_port(void) {
-    if (!PAL_CB(parent_process) || !g_process_ipc_info.parent) {
+    if (!PAL_CB(parent_process)) {
         /* no parent process, no sense in creating parent IPC port */
         return 0;
     }
 
-    lock(&g_process_ipc_info.lock);
     assert(g_process_ipc_info.parent && g_process_ipc_info.parent->vmid);
 
     add_ipc_port_by_id(g_process_ipc_info.parent->vmid, PAL_CB(parent_process),
                        IPC_PORT_CONNECTION | IPC_PORT_DIRECTPARENT, /*fini=*/NULL,
                        &g_process_ipc_info.parent->port);
 
-    unlock(&g_process_ipc_info.lock);
     return 0;
+}
+
+static void ipc_leader_disconnect_callback(struct shim_ipc_port* port, IDTYPE vmid) {
+    __UNUSED(port);
+    __UNUSED(vmid);
+
+    log_error("IPC leader died\n");
+    DkProcessExit(1);
 }
 
 static int init_ns_ipc_port(void) {
     if (!g_process_ipc_info.ns) {
-        /* no NS info from parent process, no sense in creating NS IPC port */
+        assert(!g_process_ipc_info.parent);
+        assert(!PAL_CB(parent_process));
+
+        /* We are the very first Graphene process, hence also IPC leader. */
+        assert(g_process_ipc_info.self);
+        g_process_ipc_info.ns = create_ipc_info(g_process_ipc_info.self->vmid,
+                                                qstrgetstr(&g_process_ipc_info.self->uri),
+                                                g_process_ipc_info.self->uri.len);
+        if (!g_process_ipc_info.ns) {
+            return -ENOMEM;
+        }
+        assert(!g_process_ipc_info.ns->port);
         return 0;
     }
 
-    if (g_process_ipc_info.ns->port) {
-        return 0;
-    }
-
-    if (qstrempty(&g_process_ipc_info.ns->uri)) {
-        /* there is no connection to NS leader via PAL handle and there is no URI to find NS leader:
-         * do not create NS IPC port now, it will be created on-demand during NS leader lookup */
-        return 0;
-    }
-
-    lock(&g_process_ipc_info.lock);
+    assert(!g_process_ipc_info.ns->port);
+    assert(!qstrempty(&g_process_ipc_info.ns->uri));
 
     log_debug("Reconnecting IPC port %s\n", qstrgetstr(&g_process_ipc_info.ns->uri));
     PAL_HANDLE handle = NULL;
     int ret = DkStreamOpen(qstrgetstr(&g_process_ipc_info.ns->uri), 0, 0, 0, 0, &handle);
     if (ret < 0) {
-        unlock(&g_process_ipc_info.lock);
         return pal_to_unix_errno(ret);
     }
 
-    add_ipc_port_by_id(g_process_ipc_info.ns->vmid, handle, IPC_PORT_CONNECTION, /*fini=*/NULL,
-                       &g_process_ipc_info.ns->port);
-
-    unlock(&g_process_ipc_info.lock);
+    add_ipc_port_by_id(g_process_ipc_info.ns->vmid, handle, IPC_PORT_CONNECTION,
+                       ipc_leader_disconnect_callback, &g_process_ipc_info.ns->port);
     return 0;
 }
 
--- a/LibOS/shim/src/ipc/shim_ipc_pid.c
+++ b/LibOS/shim/src/ipc/shim_ipc_pid.c
@@ -23,16 +23,8 @@ int init_ns_pid(void) {
      * initialized), so we should have only one thread, whose tid is equal to the process pid. */
     assert(cur_thread->tid == g_process.pid);
 
-    struct shim_ipc_info* info = NULL;
-    int ret = get_ipc_info_cur_process(&info);
-    if (ret < 0) {
-        return ret;
-    }
-
-    ret = add_ipc_subrange(cur_thread->tid, info->vmid, qstrgetstr(&info->uri));
-
-    put_ipc_info(info);
-    return ret;
+    return add_ipc_subrange(cur_thread->tid, g_process_ipc_info.vmid,
+                            qstrgetstr(&g_process_ipc_info.self->uri));
 }
 
 // TODO: for KILL_THREAD we don't know which process has a thread with given id
--- a/LibOS/shim/src/ipc/shim_ipc_ranges.c
+++ b/LibOS/shim/src/ipc/shim_ipc_ranges.c
@@ -72,17 +72,6 @@ static int nowned = 0;
 static int noffered = 0;
 static int nsubed = 0;
 
-DEFINE_LIST(ns_query);
-struct ns_query {
-    IDTYPE dest;
-    unsigned long seq;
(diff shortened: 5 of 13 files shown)


boryspoplawski:
Jenkins, retest Jenkins-Debug please

boryspoplawski:
Jenkins, retest Jenkins-Debug please (cmon do it)

boryspoplawski:
Jenkins, retest Jenkins-Debug please

boryspoplawski:
Jenkins, retest this please

boryspoplawski:
Jenkins, retest Jenkins-Debug please

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 0 of 13 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 115 at r1](https://reviewable.io/reviews/oscarlab/graphene/2225#-MVuq4MYB8u0GhmXVZWR:-MVuq4MYB8u0GhmXVZWS:b-yg28vz) ([raw file](https://github.com/oscarlab/graphene/blob/cbc27b289fb6b3e42c070a3b3cf6959f32ad8308/LibOS/shim/src/ipc/shim_ipc_helper.c#L115)):*
> ```C
>         assert(g_process_ipc_info.self);
>         get_ipc_info(g_process_ipc_info.self);
>         g_process_ipc_info.ns = g_process_ipc_info.self;
> ```

I don't think that's right: `self` has a `pipesrv` type handle, which I think is actually a listening UNIX socket. So trying to send a message to `g_process_info.ns->port` fails if you're a process leader: PAL complains about wrong handle type (and LibOS, confusingly, interprets that as "port was removed during sending").

So, if we set `ns` here, I think we want it to be a connected handle (same as it is for child processes). Something like this worked for me:

```c
        /* We are the very first Graphene process, hence also IPC leader. */
        assert(g_process_ipc_info.self);
        g_process_ipc_info.ns = create_ipc_info(g_process_ipc_info.self->vmid,
                                                qstrgetstr(&g_process_ipc_info.self->uri),
                                                g_process_ipc_info.self->uri.len);
    }

    assert(!g_process_ipc_info.ns->port);
    ...
```

(I need it for the "client-server" sync: the IPC helper thread of the main process is my "server", and I want all processes to talk to it, including the main process.)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 0 of 13 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 115 at r1](https://reviewable.io/reviews/oscarlab/graphene/2225#-MVuq4MYB8u0GhmXVZWR:-MVwJtLCAdBAXvsX6JJN:b-g90vmu) ([raw file](https://github.com/oscarlab/graphene/blob/cbc27b289fb6b3e42c070a3b3cf6959f32ad8308/LibOS/shim/src/ipc/shim_ipc_helper.c#L115)):*

I don't think that's right: `self` has a `pipesrv` type handle, which I think is actually a listening UNIX socket. So trying to send a message to `g_process_info.ns->port` fails if you're a process leader: PAL complains about wrong handle type (and LibOS, confusingly, interprets that as "port was removed during sending").

So, if we set `ns` here, I think we want it to be a connected handle (same as it is for child processes). Something like this worked for me:

```c
        /* We are the very first Graphene process, hence also IPC leader. */
        assert(g_process_ipc_info.self);
        g_process_ipc_info.ns = create_ipc_info(g_process_ipc_info.self->vmid,
                                                qstrgetstr(&g_process_ipc_info.self->uri),
                                                g_process_ipc_info.self->uri.len);
    }

    assert(!g_process_ipc_info.ns->port);
    ...
```

(I need it for the "client-server" sync: the IPC helper thread of the main process is my "server", and I want all processes to talk to it, including the main process.)
</blockquote></details>

This is the main process, which can skip the part of communicating via pipes. All the places have an if on whether we are the IPC leader, this sounds like some are missing (this is how the old code worked and I did not change that part).

Besides I'm removing `shim_ipc_info` completely (to be done yet).

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 0 of 13 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 115 at r1](https://reviewable.io/reviews/oscarlab/graphene/2225#-MVuq4MYB8u0GhmXVZWR:-MVwNuP33Xp1H7-Kjqq4:b-6ij4rd) ([raw file](https://github.com/oscarlab/graphene/blob/cbc27b289fb6b3e42c070a3b3cf6959f32ad8308/LibOS/shim/src/ipc/shim_ipc_helper.c#L115)):*
Then why set `g_process_ipc_info.ns` at all here, if using it doesn't work in the main process? Leaving it at null seems like the more straightforward / less confusing thing.

>This is the main process, which can skip the part of communicating via pipes. All the places have an if on whether we are the IPC leader, this sounds like some are missing (this is how the old code worked and I did not change that part).

I know that's how the current code works, it doesn't sound good for my subsystem though. I have a more complicated communication pattern than single IPC send or a single round trip, and special-casing the main process would mean maintaining separate synchronous and asynchronous code paths for all interactions.

(I'll think about it some more, but most likely I'd like to start with "loopback" IPC, and possibly think about special-casing later on.)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 0 of 13 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 115 at r1](https://reviewable.io/reviews/oscarlab/graphene/2225#-MVuq4MYB8u0GhmXVZWR:-MVwpV2k5iSWYGOk3JQs:bc7l130) ([raw file](https://github.com/oscarlab/graphene/blob/cbc27b289fb6b3e42c070a3b3cf6959f32ad8308/LibOS/shim/src/ipc/shim_ipc_helper.c#L115)):*
> Then why set g_process_ipc_info.ns at all here, if using it doesn't work in the main process? Leaving it at null seems like the more straightforward / less confusing thing.

Because currently code checks for `g_process_ipc_info.ns->vmid == my_vmid`.
> I have a more complicated communication pattern than single IPC send or a single round trip, and special-casing the main process would mean maintaining separate synchronous and asynchronous code paths for all interactions.

What else do you need beside msg+response primitive? IPC was supposed to provide just that...
Also I was thinking more of a deliver this message directly to the IPC helper instead of passing it via pipes, as a special case.

Keep in mind that this time I was hoping for a graduate rewrite of the IPC, not a 1-2 big PRs changing everything.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 0 of 13 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 115 at r1](https://reviewable.io/reviews/oscarlab/graphene/2225#-MVuq4MYB8u0GhmXVZWR:-MVwr3oH4heohqEszFHZ:bpqrsac) ([raw file](https://github.com/oscarlab/graphene/blob/cbc27b289fb6b3e42c070a3b3cf6959f32ad8308/LibOS/shim/src/ipc/shim_ipc_helper.c#L115)):*
> Because currently code checks for g_process_ipc_info.ns->vmid == my_vmid.

Okay. I don't like that the `ns->pal_handle` is there but I guess it's going to be removed later.

>What else do you need beside msg+response primitive? IPC was supposed to provide just that...

Short version: Message and a response, but the server either responds immediately, or after some more negotiation with other clients.

Long version: https://github.com/pwmarcz/fs-demo#example-dentry-cache

>Also I was thinking more of a deliver this message directly to the IPC helper instead of passing it via pipes, as a special case.

That would solve my problem, yes.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 0 of 13 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 115 at r1](https://reviewable.io/reviews/oscarlab/graphene/2225#-MVuq4MYB8u0GhmXVZWR:-MVxA5Sm7iWB-1RpCGkw:b-tpqeiy) ([raw file](https://github.com/oscarlab/graphene/blob/cbc27b289fb6b3e42c070a3b3cf6959f32ad8308/LibOS/shim/src/ipc/shim_ipc_helper.c#L115)):*
I've changed the code to use a new ipc info for `ns` - it currently does not matter and should behave exactly as the old version (but it does not have port now).
> Short version: Message and a response, but the server either responds immediately, or after some more negotiation with other clients.

This should not matter from the IPC point of view.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 0 of 13 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 115 at r1](https://reviewable.io/reviews/oscarlab/graphene/2225#-MVuq4MYB8u0GhmXVZWR:-MW4wuQo9-gv3LhUbKtG:b-wf70u4) ([raw file](https://github.com/oscarlab/graphene/blob/cbc27b289fb6b3e42c070a3b3cf6959f32ad8308/LibOS/shim/src/ipc/shim_ipc_helper.c#L115)):*
Okay, I do see a problem with sending a message to self: if it's sent *from* the IPC helper to itself, it might block on a full pipe. So I guess I'll be working on a way to queue messages for the IPC helper thread.

>This should not matter from the IPC point of view.

My point is: in this case, it's pretty hard to replace a "loopback" IPC message send with a function call, because a function call has different semantics (it's executed immediately, not queued). I'll try writing some more code and see what comes up.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 0 of 13 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 115 at r1](https://reviewable.io/reviews/oscarlab/graphene/2225#-MVuq4MYB8u0GhmXVZWR:-MW9p82JEzLNXmgdVnKQ:bu48ejy) ([raw file](https://github.com/oscarlab/graphene/blob/cbc27b289fb6b3e42c070a3b3cf6959f32ad8308/LibOS/shim/src/ipc/shim_ipc_helper.c#L115)):*

Okay, I do see a problem with sending a message to self: if it's sent *from* the IPC helper to itself, it might block on a full pipe. So I guess I'll be working on a way to queue messages for the IPC helper thread.

>This should not matter from the IPC point of view.

My point is: in this case, it's pretty hard to replace a "loopback" IPC message send with a function call, because a function call has different semantics (it's executed immediately, not queued). I'll try writing some more code and see what comes up.
</blockquote></details>

I'm not sure why it matters to you: cannot you just assume that IPC send passes message to the IPC leader, ignoring how is it done?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 0 of 13 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 115 at r1](https://reviewable.io/reviews/oscarlab/graphene/2225#-MVuq4MYB8u0GhmXVZWR:-MW9uOfvAYid5Kd7KV04:bz3egwx) ([raw file](https://github.com/oscarlab/graphene/blob/cbc27b289fb6b3e42c070a3b3cf6959f32ad8308/LibOS/shim/src/ipc/shim_ipc_helper.c#L115)):*

I'm not sure why it matters to you: cannot you just assume that IPC send passes message to the IPC leader, ignoring how is it done?
</blockquote></details>

Sure, that's what I want! If we have a "send IPC message" operation that works even if I'm sending to my own process, I'm happy. 

Well, as long as the message is not handled synchronously, because (for example) if the receiver is in the same process, it might want the same locks I'm currently holding.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 0 of 13 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 115 at r1](https://reviewable.io/reviews/oscarlab/graphene/2225#-MVuq4MYB8u0GhmXVZWR:-MWABrVg5_HvTEL5c-S5:b-vnjqd2) ([raw file](https://github.com/oscarlab/graphene/blob/cbc27b289fb6b3e42c070a3b3cf6959f32ad8308/LibOS/shim/src/ipc/shim_ipc_helper.c#L115)):*

Sure, that's what I want! If we have a "send IPC message" operation that works even if I'm sending to my own process, I'm happy. 

Well, as long as the message is not handled synchronously, because (for example) if the receiver is in the same process, it might want the same locks I'm currently holding.
</blockquote></details>

I don't have code ready yet, but from what I've planned, it should only take locks internal to IPC subsystem, so it should be good.

mkow (changes requested):
> Reviewed 12 of 13 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWB3Efh9TLA92gNPGNh:-MWB3Efh9TLA92gNPGNi:b-dago94) (no related file):*
In the commit message: `disconnect of the leader` -> `disconnection of the leader`

---
___
*[LibOS/shim/test/ltp/ltp.cfg, line 2040 at r2](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWByMkE5NLmHXfBckGC:-MWByMkE5NLmHXfBckGD:b8azd8e) ([raw file](https://github.com/oscarlab/graphene/blob/e047bfd818f8159bb7780b5cef7db23d709b62bb/LibOS/shim/test/ltp/ltp.cfg#L2040)):*
> ```INI
> 
> # Fails from time to time in CI on an assert in IPC code:
> # assert failed ipc/shim_ipc_ranges.c:706 !qstrempty(&g_process_ipc_info.ns->uri)
> ```

Nice! Any idea where was the bug?
___
*[LibOS/shim/test/regression/double_fork.c, line 7 at r2](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWByqhxESG9LAIYBknR:-MWByqhxESG9LAIYBknS:bfur1f3) ([raw file](https://github.com/oscarlab/graphene/blob/db6cfc274265b034d57d9c3c2c5b1835a3823fde/LibOS/shim/test/regression/double_fork.c#L7)):*
> ```C
> #include <stdint.h>
> #include <stdio.h>
> #include <signal.h>
> ```

please sort
___
*[LibOS/shim/test/regression/double_fork.c, line 94 at r2](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWC-K5S9FX3nRDc1jaF:-MWC-K5S9FX3nRDc1jaG:b-ljkdc1) ([raw file](https://github.com/oscarlab/graphene/blob/db6cfc274265b034d57d9c3c2c5b1835a3823fde/LibOS/shim/test/regression/double_fork.c#L94)):*
> ```C
>     };
>     errno = 0;
>     long ret = syscall(SYS_futex, &last_sig, FUTEX_WAIT, 0, &ts, NULL, 0);
> ```

clever trick

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 12 of 13 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWB3Efh9TLA92gNPGNh:-MWC2_nl0yMmnfkz4OIt:be6mdj4) (no related file):*

In the commit message: `disconnect of the leader` -> `disconnection of the leader`
</blockquote></details>

TODO: at rebase

---
___
*[LibOS/shim/test/ltp/ltp.cfg, line 2040 at r2](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWByMkE5NLmHXfBckGC:-MWC2thxEbCQdGL9R5ZG:bluaooa) ([raw file](https://github.com/oscarlab/graphene/blob/e047bfd818f8159bb7780b5cef7db23d709b62bb/LibOS/shim/test/ltp/ltp.cfg#L2040)):*

Nice! Any idea where was the bug?
</blockquote></details>

In Graphene.
___
*[LibOS/shim/test/regression/double_fork.c, line 7 at r2](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWByqhxESG9LAIYBknR:-MWC3Ia9CcSwvnUSr_PT:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/db6cfc274265b034d57d9c3c2c5b1835a3823fde/LibOS/shim/test/regression/double_fork.c#L7)):*

please sort
</blockquote></details>

Done.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 12 of 13 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/test/ltp/ltp.cfg, line 2040 at r2](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWByMkE5NLmHXfBckGC:-MWC3qhU8CYMOfz7Nthb:b-set3rh) ([raw file](https://github.com/oscarlab/graphene/blob/e047bfd818f8159bb7780b5cef7db23d709b62bb/LibOS/shim/test/ltp/ltp.cfg#L2040)):*

In Graphene.
</blockquote></details>

:)

But for real: no idea, but I highly suspect the discovery code, it was a mess.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

mkow (dismissed):
> Reviewed 1 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

boryspoplawski:
Jenkins, retest Jenkins-SGX-18.04 please (`udp` from LibOS regression failed, triggered already before, not reltated to this PR, but we need to debug it somehow; could not reproduce locally)

dimakuv (changes requested):
> Reviewed 11 of 13 files at r1, 1 of 1 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/include/shim_ipc.h, line 51 at r3](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWU_0Lb0_NUixItnC93:-MWU_0Lc8JujfqF8QR1M:b-6i6wb7) ([raw file](https://github.com/oscarlab/graphene/blob/e047bfd818f8159bb7780b5cef7db23d709b62bb/LibOS/shim/include/shim_ipc.h#L51)):*
> ```objc
>     IPC_MSG_LEASE,
>     IPC_MSG_OFFER,
>     IPC_MSG_RENEW,
> ```

Isn't removing this `RENEW` stuff unrelated to the rest of the PR?

Or in another way: why do you remove `RENEW` but keep `SUBLEASE`? Aren't these related, and if one is removed, the second one should also be removed? (AFAIU, `RENEW` is to inform the leader process that some process wants to keep using some resources, and `SUBLEASE` is to inform the leader process that some process wants to give a subset of its resources to some other process.)

RESOLVING: Found that `RENEW` logic was dead code completely.
___
*[LibOS/shim/include/shim_ipc.h, line 90 at r3](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWUdX9U5WOkEIHIK6To:-MWUdX9U5WOkEIHIK6Tp:b8wtn1q) ([raw file](https://github.com/oscarlab/graphene/blob/e047bfd818f8159bb7780b5cef7db23d709b62bb/LibOS/shim/include/shim_ipc.h#L90)):*
> ```objc
> struct shim_process_ipc_info {
>     IDTYPE vmid;
>     struct shim_lock lock;
> ```

This lock was completely useless? Because all the fields here were created during startup of a Graphene instance?
___
*[LibOS/shim/src/shim_init.c, line 453 at r3](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWU_li9DI-UYd8HGPtW:-MWU_liA7IjBFAHUULf0:b-7d1glg) ([raw file](https://github.com/oscarlab/graphene/blob/bb52ab16855977cf806e9d76f291834677b2d793/LibOS/shim/src/shim_init.c#L453)):*
> ```C
> 
>     /* FIXME: `PAL_CB(parent_process)` was handed over to IPC code above so we should't be using it
>      * here. */
> ```

Why not just move `RUN_INIT(init_ipc_helper)` below this piece of code?
___
*[LibOS/shim/src/ipc/shim_ipc_helper.c, line 121 at r3](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWUd7fU2f05qi6NZIto:-MWUd7fU2f05qi6NZItp:b-bte722) ([raw file](https://github.com/oscarlab/graphene/blob/bb52ab16855977cf806e9d76f291834677b2d793/LibOS/shim/src/ipc/shim_ipc_helper.c#L121)):*
> ```C
>         }
>         assert(!g_process_ipc_info.ns->port);
>         return 0;
> ```

What is the logic here? Why don't we create a `ns->port` here? Is it because this process is the leader process itself, so doesn't need to send messages to itself? And there is some logic like `if (!ns->port) <do stuff locally since I'm the leader>`?

RESOLVING: Found that the rest of the code uses logic like `if (leader == g_process_ipc_info.vmid) ...`, so yeah, we don't need to create `ns->port` in this case.
___
*[LibOS/shim/test/regression/double_fork.c, line 1 at r3](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWUgXQWBZXcU1fWeOMo:-MWUgXQX-_lmjdZmbdZa:b-a5t2ox) ([raw file](https://github.com/oscarlab/graphene/blob/bb52ab16855977cf806e9d76f291834677b2d793/LibOS/shim/test/regression/double_fork.c#L1)):*
> ```C
> #define _GNU_SOURCE
> ```

Can you add a quick top-level comment in this file that: the grandparent (main) process creates a child which creates a grandchild. Then the child terminates, and only then the grandparent kicks the grandchild, and the grandchild sends SIGALRM to the grandparent. So this test checks that there is a working connection between grandparent and grandchild even when the intermediate child process already terminated.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 12 of 13 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[LibOS/shim/include/shim_ipc.h, line 51 at r3](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWU_0Lb0_NUixItnC93:-MWXPV_aEteecw5iAStf:bx6yjiu) ([raw file](https://github.com/oscarlab/graphene/blob/e047bfd818f8159bb7780b5cef7db23d709b62bb/LibOS/shim/include/shim_ipc.h#L51)):*

Isn't removing this `RENEW` stuff unrelated to the rest of the PR?

Or in another way: why do you remove `RENEW` but keep `SUBLEASE`? Aren't these related, and if one is removed, the second one should also be removed? (AFAIU, `RENEW` is to inform the leader process that some process wants to keep using some resources, and `SUBLEASE` is to inform the leader process that some process wants to give a subset of its resources to some other process.)

RESOLVING: Found that `RENEW` logic was dead code completely.
</blockquote></details>

Yeah, I've just removed dead code recursively. More to come!
___
*[LibOS/shim/include/shim_ipc.h, line 90 at r3](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWUdX9U5WOkEIHIK6To:-MWXPdOcE15RoGwVI2Cx:bbx0dsb) ([raw file](https://github.com/oscarlab/graphene/blob/e047bfd818f8159bb7780b5cef7db23d709b62bb/LibOS/shim/include/shim_ipc.h#L90)):*

This lock was completely useless? Because all the fields here were created during startup of a Graphene instance?
</blockquote></details>

I think no, connecting to ns leader was done lazily (+ renewals and finding new leader, but that didn't work). It would be useless now tho (all of self,parent and ns are connected at startup).
___
*[LibOS/shim/src/shim_init.c, line 453 at r3](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWU_li9DI-UYd8HGPtW:-MWXQKVXDse4rxZvgod2:b-8l0qlt) ([raw file](https://github.com/oscarlab/graphene/blob/bb52ab16855977cf806e9d76f291834677b2d793/LibOS/shim/src/shim_init.c#L453)):*

Why not just move `RUN_INIT(init_ipc_helper)` below this piece of code?
</blockquote></details>

Uh, I remember thinking about it and finding some reasons not to, but I've already forgotten...
Let me think about it some more.
___
*[LibOS/shim/test/regression/double_fork.c, line 1 at r3](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWUgXQWBZXcU1fWeOMo:-MWXSJqsCb0Ic-o91D9I:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/bb52ab16855977cf806e9d76f291834677b2d793/LibOS/shim/test/regression/double_fork.c#L1)):*

Can you add a quick top-level comment in this file that: the grandparent (main) process creates a child which creates a grandchild. Then the child terminates, and only then the grandparent kicks the grandchild, and the grandchild sends SIGALRM to the grandparent. So this test checks that there is a working connection between grandparent and grandchild even when the intermediate child process already terminated.
</blockquote></details>

Done.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: 12 of 13 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[LibOS/shim/src/shim_init.c, line 453 at r3](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWU_li9DI-UYd8HGPtW:-MWXUj3W9vbZe50pXHBD:bafj5yj) ([raw file](https://github.com/oscarlab/graphene/blob/bb52ab16855977cf806e9d76f291834677b2d793/LibOS/shim/src/shim_init.c#L453)):*

Uh, I remember thinking about it and finding some reasons not to, but I've already forgotten...
Let me think about it some more.
</blockquote></details>

Ah, I think there is only one reason: here we inform the parent process, that the initialization is done and we (the child process) are spawned and ready to execute (moving `RUN_INIT` below would break this assumption).
This might or might not be what we want, so I've added the comment - I would rather not deal with it in this PR.

boryspoplawski:
Jenkins, retest Jenkins-SGX-18.04 please (again `udp`, see the explanation above)

boryspoplawski:
Managed to reproduce the `udp` issue (took a loooong time): spurious EINTR

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/include/shim_ipc.h, line 90 at r3](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWUdX9U5WOkEIHIK6To:-MWYGB0R7GfzC03Du5Yy:bv56fo0) ([raw file](https://github.com/oscarlab/graphene/blob/e047bfd818f8159bb7780b5cef7db23d709b62bb/LibOS/shim/include/shim_ipc.h#L90)):*

I think no, connecting to ns leader was done lazily (+ renewals and finding new leader, but that didn't work). It would be useless now tho (all of self,parent and ns are connected at startup).
</blockquote></details>

OK, makes sense.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2225#-MWB3Efh9TLA92gNPGNh:-MWYINMY3UcxCoTIHvjX:bb6un1j) (no related file):*

TODO: at rebase
</blockquote></details>

Done

---

mkow (approved):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2225)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #2076, opened 2021-01-12
Redesign manifest


woju:
This is split from a discussion in #2050. We should rework manifest specification, to have it more clear, and possibly take advantage of TOML structures (e.g. list), which were unavailable in previous manifest syntax.

Ideas:
- decide on entrypoints (`[libos.]entrypoint`, `pal.entrypoint/loader.preload`)
- `sgx.trusted_files = []`
- ?

dimakuv:
`sgx.trusted_files` is already a TOML table. You want it to be a TOML array, iiuc. But I actually prefer the current syntax.

In other words, you advocate smth like this:
```
sgx.trusted_files = ["filepath1", "filepath2"]
```

Whereas the current syntax also allows to have keys:
```
sgx.trusted_files.uniquefileid1 = "filepath1"
sgx.trusted_files.uniquefileid2 = "filepath2"

# or the same in another form
[sgx.trusted_files]
uniquefileid1 = "filepath1"
uniquefileid2 = "filepath2"
```

I find the current TOML table syntax quite nice.

mkow:
Some ideas/comments:

* I definitely don't like all those keyed arrays in our current syntax, they can be just plain arrays (current syntax forces users to invent key names for each file). Also, they are verbose without (IMO) any advantages of this verbosity.
* What do you think about using top-level tables? Something like:
  ```Toml
  [pal]
  entrypoint = "asdfg"
  
  [pal.sgx]
  something_sgx_specific = "12312313"
  
  [libos]
  entrypoint = "qwerty"
  ```
  (taking aside the entrypoint discussion, this is just an example to show the idea)
* We should use bools instead of 0/1 (AFAIR this was left temporarily to minimize the breaking changes done at once, but we want to change it).
* Executable envs and args could very nicely fit into TOML syntax, instead of the keyed "arrays" we use right now.

dimakuv:
> What do you think about using top-level tables?

All the above examples are already possible. We're just using a different format of presenting TOML tables (a "chained keys" format).

woju:
What for do you need the unique file id in the manifest?

dimakuv:
I guess I'm just used to it and use them as comments. E.g. I sometimes do
```
sgx.trusted_files.lib_needed_for_python_when_it_fails = "file:/usr/lib/python/pythonapt"
```

Well, yes, this doesn't really make sense :) I'm retracting my comment.

mkow:
[...]

PR #2249, opened 2021-03-17
[Pal/Linux-SGX] Allow specifying whole directories in trusted files

mkow:
## Description of the changes 

Now sgx_sign.py expands trusted files paths recursively if they point to a directories. This allowed to greatly simplify all our examples and made them more portable (they should now work at least on all Ubuntus, without version-specific `if`s).

One downside is that CI will run a bit slower, but the differences were usually negligible (just a few seconds more) except for a few examples with a lot of dependencies (e.g. GCC running a minute longer). The slowdowns are mostly because tomlc99 is not optimized for huge dictionaries and does linear lookup, but this won't be a problem after we migrate to the new, more TOML-ish syntax in the manifests and just have an array of trusted files paths (see #2076).

## How to test this PR? 

CI.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)


boryspoplawski (changes requested):
> Reviewed 70 of 70 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow)

___
*[Documentation/tutorials/pytorch/index.rst, line 255 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW5mUBN316O-JXeqT38:-MW5mUBO1qX0NOmzlQ9n:b-rnegt6) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Documentation/tutorials/pytorch/index.rst#L255)):*
> ```reStructuredText
>    ...
> 
> ``sgx.trusted_files.<name>`` specifies a file or a directory that will be
> ```

The semantics of "verified and trusted directory" are not clear, I would add something about how it works (i.e. that it recursively expands the fs tree and adds all files).
___
*[Examples/apache/httpd.manifest.template, line 75 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW5msb37mhUavKMPHCT:-MW5msb37mhUavKMPHCU:b-frpb8w) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/apache/httpd.manifest.template#L75)):*
> ```
> # Other libraries required by Apache
> sgx.trusted_files.arch_libdir = "file:$(ARCH_LIBDIR)/"
> sgx.trusted_files.usr_arch_libdir = "file:/usr$(ARCH_LIBDIR)/"
> ```

Doesn't it lack a `/` before `$(ARCH_LIBDIR)`?
Btw, old version didn't seem to need this entry?
This applies to all manifests in this PR.
___
*[Examples/python-scipy-insecure/python.manifest.template, line 90 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6qPOVA_fne80pgqMj:-MW6qPOVA_fne80pgqMk:bwhauqh) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/python-scipy-insecure/python.manifest.template#L90)):*
> ```
> sgx.trusted_files.runtime = "file:$(GRAPHENEDIR)/Runtime/"
> sgx.trusted_files.arch_libdir = "file:$(ARCH_LIBDIR)/"
> sgx.trusted_files.usr_libdir = "file:/usr/lib/"
> ```

Is this needed?
___
*[Examples/python-simple/python.manifest.template, line 91 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6r8PYBwyrz7hsgEh6:-MW6r8PYBwyrz7hsgEh7:b7xcuzj) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/python-simple/python.manifest.template#L91)):*
> ```
> sgx.trusted_files.python_dir = "file:$(PYTHONHOME)/"
> sgx.trusted_files.dist = "file:$(PYTHONDISTHOME)/"
> sgx.trusted_files.scripts = "file:scripts"
> ```

NP: please add `/` at the end (for visual indication that this is a directory).
___
*[Examples/r/R.manifest.template, line 85 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6tus7ARL14H-384TA:-MW6tus7ARL14H-384TB:bxtub52) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/r/R.manifest.template#L85)):*
> ```
> sgx.trusted_files.runtime = "file:$(GRAPHENEDIR)/Runtime/"
> sgx.trusted_files.arch_libdir = "file:$(ARCH_LIBDIR)/"
> sgx.trusted_files.usr_libdir = "file:/usr/lib/"
> ```

Is it needed?
___
*[Examples/r/R.manifest.template, line 89 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6tsGX4_C_6wOTdTkA:-MW6tsGX4_C_6wOTdTkB:b-wv15p6) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/r/R.manifest.template#L89)):*
> ```
> # Other libraries as dependencies of R
> sgx.trusted_files.stats_lib = "file:$(R_HOME)/"
> sgx.trusted_files.scripts = "file:scripts"
> ```

ditto (ending /)
___
*[Examples/ra-tls-mbedtls/Makefile, line 111 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6vAeR0t9JqZEDkPoW:-MW6vAeS4hbZXvxhLw2m:b-67noi6) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-mbedtls/Makefile#L111)):*
> ```Makefile
> 
> .SECONDARY: .mbedtls_libs_copied
> .mbedtls_libs_copied: .mbedtls_built
> ```

You usually put such dot files in `libs` (target directory), I guess you can then just remove the whole directory without breaking build system.
___
*[Examples/ra-tls-secret-prov/Makefile, line 17 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6y4cv5Y329VmZw6MU:-MW6y4cv5Y329VmZw6MV:b-olm0rd) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-secret-prov/Makefile#L17)):*
> ```Makefile
> GRAPHENEKEY ?= $(GRAPHENEDIR)/Pal/src/host/Linux-SGX/signer/enclave-key.pem
> 
> ARCH_LIBDIR ?= /lib/x86_64-linux-gnu
> ```

Why not include `Scripts/Makefile.configs` like other makefiles do in these `Examples`?
___
*[Examples/ra-tls-secret-prov/Makefile, line 76 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6wsUg4EisYYu8csgY:-MW6wsUg4EisYYu8csgZ:b-ox5ilu) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-secret-prov/Makefile#L76)):*
> ```Makefile
> 	$(GRAPHENEDIR)/Scripts/download --output $@ $(foreach mirror,$(MBEDCRYPTO_URI),--url $(mirror)/$(MBEDCRYPTO_SRC)) --sha256 $(MBEDCRYPTO_CHECKSUM)
> 
> .SECONDARY: .mbedtls_downloaded
> ```

ditto (dotfiles)
___
*[Examples/ra-tls-secret-prov/Makefile, line 103 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6xGWE90_gbO-cpTpL:-MW6xGWF5KpDXe6M6yWN:b-7av04x) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-secret-prov/Makefile#L103)):*
> ```Makefile
> 
> secret_prov_server_epid: src/secret_prov_server.c .mbedtls_libs_copied .epid_libs_copied
> 	LD_LIBRARY_PATH=./libs/ $(CC) $< $(CFLAGS) $(LFLAGS) -lsecret_prov_verify_epid -pthread -o $@
> ```

Why do we need these `LD_LIBRARY_PATH`s ?
___
*[Examples/tensorflow-lite/Makefile, line 56 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6yhgC-MKTSn5NCCHN:-MW6yhgD0w9iHu3ZtxEe:b2ni78k) ([raw file](https://github.com/oscarlab/graphene/blob/1102fee8f5f5ea85316b1d20a736500b0f1bc04d/Examples/tensorflow-lite/Makefile#L56)):*
> ```Makefile
> label_image.manifest: libtensorflow_framework.so label_image inception_v3.tflite labels.txt image.bmp
> label_image.manifest: $(GRAPHENEDIR)/Runtime/libgcc_s.so.1 $(GRAPHENEDIR)/Runtime/libstdc++.so.6
> label_image.manifest: label_image.manifest.template
> ```

lol
___
*[LibOS/shim/test/regression/openmp.manifest.template, line 7 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6zAyW2jnmsADxqtND:-MW6zAyW2jnmsADxqtNE:b-nu8ztu) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/LibOS/shim/test/regression/openmp.manifest.template#L7)):*
> ```
> # Graphene optionally provides patched OpenMP runtime library that runs faster inside SGX enclaves
> # (execute `make -C LibOS gcc` to generate it). This library will replace the native one because
> # Graphene's Runtime path has priority in LD_LIBRARY_PATH.
> ```

Hmm, isn't it always replaced now (if it's built)? I think previously you could have built but commented the line enabling it, and it would use the host version (since opening the Graphene one would faile due to it not being trusted).
___
*[python/graphenelibos/sgx_sign.py, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW7-3JAEqgxCsoEsUPH:-MW7-3JAEqgxCsoEsUPI:b-p07jnt) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/python/graphenelibos/sgx_sign.py#L165)):*
> ```Python
> def get_hash(filename):
>     # TODO: Remove `str()` after dropping support for Ubuntu 16.04 and its ancient Python.
>     with open(str(filename), 'rb') as file:
> ```

Why do we need this `str` in any of python versions? It should work without it...
___
*[python/graphenelibos/sgx_sign.py, line 194 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW70jrGBelwqE1B01MC:-MW70jrGBelwqE1B01MD:bbifi0z) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/python/graphenelibos/sgx_sign.py#L194)):*
> ```Python
>         if path.is_dir():
>             for sub_path in list_dir(path):
>                 sub_key = path_to_key(key + str(sub_path))
> ```

Why not just `path_to_key(str(sub_path))`?
___
*[python/graphenelibos/sgx_sign.py, line 750 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW71ZAW2fV22PTR5Coq:-MW71ZAW2fV22PTR5Cor:b-llqbmr) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/python/graphenelibos/sgx_sign.py#L750)):*
> ```Python
>     # Get trusted hashes and measurements
> 
>     # use `list()` to ensure non-laziness
> ```

Please expand the explanation. You got me scratching my head until I realized `manifest_sgx` is actually contained in `manifest`

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: 60 of 70 files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Documentation/tutorials/pytorch/index.rst, line 255 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW5mUBN316O-JXeqT38:-MW79nIU-Hqc2v24Csst:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Documentation/tutorials/pytorch/index.rst#L255)):*

The semantics of "verified and trusted directory" are not clear, I would add something about how it works (i.e. that it recursively expands the fs tree and adds all files).
</blockquote></details>

Done.
___
*[Examples/apache/httpd.manifest.template, line 75 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW5msb37mhUavKMPHCT:-MW79yoZ4OIaDe__6Yls:ba0v5pa) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/apache/httpd.manifest.template#L75)):*

Doesn't it lack a `/` before `$(ARCH_LIBDIR)`?
Btw, old version didn't seem to need this entry?
This applies to all manifests in this PR.
</blockquote></details>

`$(ARCH_LIBDIR)` already contains the slash, I didn't want to emit a doubled one. It's quite hacky, we probably should have a separate variable for this, but it's out of scope for this PR.

ad. adding this dir here: I don't remember the exact reason, but it was either them not working on 20.04, or `get_deps.sh` emitting some entries from it.
___
*[Examples/python-scipy-insecure/python.manifest.template, line 90 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6qPOVA_fne80pgqMj:-MW7AZWH8lDpMMK6Q0eO:byeahk0) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/python-scipy-insecure/python.manifest.template#L90)):*

Is this needed?
</blockquote></details>

Yep, but only on 16.04 (libblas is located right there on 16.04).
___
*[Examples/python-simple/python.manifest.template, line 91 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6r8PYBwyrz7hsgEh6:-MW7AiP49exY0rB6GNuS:b-o419a3) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/python-simple/python.manifest.template#L91)):*

NP: please add `/` at the end (for visual indication that this is a directory).
</blockquote></details>

Right, fixed.
___
*[Examples/r/R.manifest.template, line 85 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6tus7ARL14H-384TA:-MW7Az1k2saIibsS_EDJ:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/r/R.manifest.template#L85)):*

Is it needed?
</blockquote></details>

ditto
___
*[Examples/r/R.manifest.template, line 89 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6tsGX4_C_6wOTdTkA:-MW7B-fk9KWo2VOlz0WZ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/r/R.manifest.template#L89)):*

ditto (ending /)
</blockquote></details>

Done.
___
*[Examples/ra-tls-mbedtls/Makefile, line 111 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6vAeR0t9JqZEDkPoW:-MW7B6QZB2H79ULUueEF:b-fznogt) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-mbedtls/Makefile#L111)):*

You usually put such dot files in `libs` (target directory), I guess you can then just remove the whole directory without breaking build system.
</blockquote></details>

Good idea, done.
___
*[Examples/ra-tls-secret-prov/Makefile, line 17 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6y4cv5Y329VmZw6MU:-MW7FkApEQEP0Wo2pG5E:b-jk4c6p) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-secret-prov/Makefile#L17)):*

Why not include `Scripts/Makefile.configs` like other makefiles do in these `Examples`?
</blockquote></details>

Ah, that's why it was missing...
___
*[Examples/ra-tls-secret-prov/Makefile, line 76 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6wsUg4EisYYu8csgY:-MW7CwVE2aIQR2VfP1vX:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-secret-prov/Makefile#L76)):*

ditto (dotfiles)
</blockquote></details>

Done.
___
*[Examples/ra-tls-secret-prov/Makefile, line 103 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6xGWE90_gbO-cpTpL:-MW7Cy4d6un2oX3ydUNj:b-223dbo) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-secret-prov/Makefile#L103)):*

Why do we need these `LD_LIBRARY_PATH`s ?
</blockquote></details>

For some reason the compiler uses these .so files for linking, I'm not sure why exactly.
___
*[LibOS/shim/test/regression/openmp.manifest.template, line 7 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6zAyW2jnmsADxqtND:-MW7DJu127Ctg0dAC_IA:b751jj5) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/LibOS/shim/test/regression/openmp.manifest.template#L7)):*

Hmm, isn't it always replaced now (if it's built)? I think previously you could have built but commented the line enabling it, and it would use the host version (since opening the Graphene one would faile due to it not being trusted).
</blockquote></details>

It is, but I guess this is a feature, not a bug? There's no reason to use the system OpenMP with SGX.
___
*[python/graphenelibos/sgx_sign.py, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW7-3JAEqgxCsoEsUPH:-MW7DXSY1awqflPGOxGS:b-qvrm6n) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/python/graphenelibos/sgx_sign.py#L165)):*

Why do we need this `str` in any of python versions? It should work without it...
</blockquote></details>

Not on 16.04 :) `open()` didn't support Pathlib back then.
___
*[python/graphenelibos/sgx_sign.py, line 194 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW70jrGBelwqE1B01MC:-MW7Dky57bkNhlLw5qY6:b-gl96pz) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/python/graphenelibos/sgx_sign.py#L194)):*

Why not just `path_to_key(str(sub_path))`?
</blockquote></details>

I think there were some collisions between these auto-generated preload items and Runtime/ listing, and this hacky way resolves it.
This code will go to trash pretty soon, we will be dropping the keys altogether for just an array.
___
*[python/graphenelibos/sgx_sign.py, line 750 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW71ZAW2fV22PTR5Coq:-MW7Ff4I2F45ozq8NFRc:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/python/graphenelibos/sgx_sign.py#L750)):*

Please expand the explanation. You got me scratching my head until I realized `manifest_sgx` is actually contained in `manifest`
</blockquote></details>

Done.

boryspoplawski (changes requested):
> Reviewed 11 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Examples/apache/httpd.manifest.template, line 75 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW5msb37mhUavKMPHCT:-MW80fhGERhoY3SZ2iDL:b6x28uf) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/apache/httpd.manifest.template#L75)):*
> I didn't want to emit a doubled one.

Why not? Imo it's better to have two than none. Also it would look nicer, now it's weird.
> not working on 20.04

They are just symlinks there, so not this.
> or get_deps.sh emitting some entries from it.

This example did not use `get_deps.sh`, as far as I can see.

Ok, it might be needed on ubuntu 16 and 18.
___
*[Examples/python-scipy-insecure/python.manifest.template, line 90 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6qPOVA_fne80pgqMj:-MW82Hdy88BraBMH8sJr:bb52uds) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/python-scipy-insecure/python.manifest.template#L90)):*

Yep, but only on 16.04 (libblas is located right there on 16.04).
</blockquote></details>

Then please add a comment, so we can remove it once we drop support for ubuntu16.
___
*[Examples/r/R.manifest.template, line 85 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6tus7ARL14H-384TA:-MW82WjG711QiWO4FFhs:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/r/R.manifest.template#L85)):*

ditto
</blockquote></details>

ditto
___
*[Examples/ra-tls-secret-prov/Makefile, line 103 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6xGWE90_gbO-cpTpL:-MW83K8T7NMIysHRJNOC:b-7h9ngb) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-secret-prov/Makefile#L103)):*

For some reason the compiler uses these .so files for linking, I'm not sure why exactly.
</blockquote></details>

what.
___
*[Examples/ra-tls-secret-prov/Makefile, line 30 at r2](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW831ex16B046wesrpn:-MW831ey3LJXxX9-EaEo:b33eow2) ([raw file](https://github.com/oscarlab/graphene/blob/ea8180007a71bb165d2136b66f02d4430003ef5c/Examples/ra-tls-secret-prov/Makefile#L30)):*
> ```Makefile
> endif
> 
> include ../../Scripts/Makefile.configs
> ```

I would say: "please move to the top of the file", but I think I know the reason for it being here...
___
*[LibOS/shim/test/regression/openmp.manifest.template, line 7 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6zAyW2jnmsADxqtND:-MW83iYDD6HzuBYmZvbM:bwevvub) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/LibOS/shim/test/regression/openmp.manifest.template#L7)):*

It is, but I guess this is a feature, not a bug? There's no reason to use the system OpenMP with SGX.
</blockquote></details>

I guess.
___
*[python/graphenelibos/sgx_sign.py, line 194 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW70jrGBelwqE1B01MC:-MW83yCNEJUeacIqfBCl:bq0ykje) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/python/graphenelibos/sgx_sign.py#L194)):*

I think there were some collisions between these auto-generated preload items and Runtime/ listing, and this hacky way resolves it.
This code will go to trash pretty soon, we will be dropping the keys altogether for just an array.
</blockquote></details>

How there could be a collision? If there are two `sub_path`s with the same value (even with different `key`), then all values here (`uri`, `sub_key`) will be the same, so it's just 2x the same entry, which should be fine, wouldn't it?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: 47 of 71 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[Examples/ra-tls-secret-prov/Makefile, line 103 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6xGWE90_gbO-cpTpL:-MW9YvVJCpEE86xz6S4w:b-109cfm) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-secret-prov/Makefile#L103)):*

what.
</blockquote></details>

Do you want me to debug this? We can also summon @dimakuv to answer this, he knows what's happening here :)
___
*[Examples/ra-tls-secret-prov/Makefile, line 30 at r2](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW831ex16B046wesrpn:-MW9YqYu-RW6Mr9AQzCI:b-teuhl5) ([raw file](https://github.com/oscarlab/graphene/blob/ea8180007a71bb165d2136b66f02d4430003ef5c/Examples/ra-tls-secret-prov/Makefile#L30)):*

I would say: "please move to the top of the file", but I think I know the reason for it being here...
</blockquote></details>

Yeah :)
___
*[python/graphenelibos/sgx_sign.py, line 194 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW70jrGBelwqE1B01MC:-MW9Z7jr-ZDBZikznlQu:b-tt5oo2) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/python/graphenelibos/sgx_sign.py#L194)):*

How there could be a collision? If there are two `sub_path`s with the same value (even with different `key`), then all values here (`uri`, `sub_key`) will be the same, so it's just 2x the same entry, which should be fine, wouldn't it?
</blockquote></details>

No, you can't have a duplicated key in TOML, the parser errors out (and this happens if you remove this `key + `).

Hmm, actually, after writing this, I realized that I written this before we had a `dict` and a proper TOML parser here, and we were just appending entries manually... I'll remove it and we'll see if it passes all the tests.

pwmarcz (changes requested):
> Reviewed 2 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: 47 of 71 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW9ZOb09kdXtvLbMOsb:-MW9ZOb09kdXtvLbMOsc:b7mwgft) (no related file):*
I tried to build busybox and got:
```
~/graphene/Examples/busybox$ make SGX=1
cp src/busybox busybox
sed -e 's|$(GRAPHENEDIR)|'"../.."'|g' \
	-e 's|$(GRAPHENE_LOG_LEVEL)|'"error"'|g' \
	-e 's|$(ARCH_LIBDIR)|'"/lib/x86_64-linux-gnu"'|g' \
	busybox.manifest.template > busybox.manifest
ln -s ../../Runtime/pal_loader pal_loader
make: *** No rule to make target 'busybox.sig', needed by 'all'.  Stop.
make: *** Waiting for unfinished jobs....
```

(It does actually build `busybox.sig`, and the second run of `make SGX=1` works. Looks like there is a dependency missing.)

Doesn't look like a problem with your code, but I wonder why CI works.

---
___
*[python/graphenelibos/sgx_sign.py, line 176 at r2](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW9_Vay7iSL_e4JAGEF:-MW9_Vaz77ysyXrLHpQD:bfmfucu) ([raw file](https://github.com/oscarlab/graphene/blob/ea8180007a71bb165d2136b66f02d4430003ef5c/python/graphenelibos/sgx_sign.py#L176)):*
> ```Python
> 
> 
> def list_dir(path):
> ```

* `list_dir` sounds like it lists a single directory, not all subdirectories. Maybe `walk_dir`?

* This will skip paths without permissions, is that OK?

* It will also skip directories that are symlinks, is that OK? (i.e. we will not include `foo/bar.txt` in manifest if `foo/` is a symlink)

* What do you think about sorting the list, to make it deterministic? Right now it's in the order in which the filesystem gives you results.  I can't think of a specific use case right now but probably diffing the manifests might be surprising (for example, if you're comparing across machines).

boryspoplawski (changes requested):
> Reviewed 24 of 24 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[Examples/ra-tls-secret-prov/Makefile, line 103 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6xGWE90_gbO-cpTpL:-MW9m1KO4dTjVktjpTf7:b-n2dt66) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-secret-prov/Makefile#L103)):*

Do you want me to debug this? We can also summon @dimakuv to answer this, he knows what's happening here :)
</blockquote></details>

Yeah.
Btw, did old version use them as well? If yes how?
___
*[python/graphenelibos/sgx_sign.py, line 176 at r2](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW9_Vay7iSL_e4JAGEF:-MW9mLdT698QnK1siVWk:b-260q25) ([raw file](https://github.com/oscarlab/graphene/blob/ea8180007a71bb165d2136b66f02d4430003ef5c/python/graphenelibos/sgx_sign.py#L176)):*

* `list_dir` sounds like it lists a single directory, not all subdirectories. Maybe `walk_dir`?

* This will skip paths without permissions, is that OK?

* It will also skip directories that are symlinks, is that OK? (i.e. we will not include `foo/bar.txt` in manifest if `foo/` is a symlink)

* What do you think about sorting the list, to make it deterministic? Right now it's in the order in which the filesystem gives you results.  I can't think of a specific use case right now but probably diffing the manifests might be surprising (for example, if you're comparing across machines).
</blockquote></details>

- +1
- I think it's ok
- Hmm, this might be an issue.
- +0.5

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: 70 of 71 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW9ZOb09kdXtvLbMOsb:-MWAb22I9h4CMNeJTBss:ba7uozl) (no related file):*

busybox is not tested in CI :)
Yeah, it lacks an intermediate step or `&:` support, but it's not in scope of this PR, I suppose
</blockquote></details>

I'll take a note and fix it separately, as it looks unrelated.

---
___
*[Examples/ra-tls-secret-prov/Makefile, line 103 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6xGWE90_gbO-cpTpL:-MWAdL2T0rEqiar2cIPJ:bgkd7g8) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-secret-prov/Makefile#L103)):*

Yeah.
Btw, did old version use them as well? If yes how?
</blockquote></details>

Without setting `LD_LIBRARY_PATH` this you get:
```
/usr/bin/ld: warning: libsgx_util.so, needed by ./libs//libsecret_prov_verify_epid.so, not found (try using -rpath or -rpath-link)
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `hexdump_mem_to_buffer'
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `ias_cleanup'
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `ias_init'
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `verify_quote'
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `parse_hex'
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `verify_ias_report_extract_quote'
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `ias_verify_quote_raw'
collect2: error: ld returned 1 exit status
```
Seems that if the files were in the main directory it was able to find them. But I'm not really sure what's happening here...
___
*[python/graphenelibos/sgx_sign.py, line 176 at r2](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW9_Vay7iSL_e4JAGEF:-MWAbehG3CEaiq7Q60PF:b77bn6c) ([raw file](https://github.com/oscarlab/graphene/blob/ea8180007a71bb165d2136b66f02d4430003ef5c/python/graphenelibos/sgx_sign.py#L176)):*

- +1
- I think it's ok
- Hmm, this might be an issue.
- +0.5
</blockquote></details>

* good idea, done
* yup, I think we have to do this, because a common use case is mounting things from the host /usr/lib, where you may not have permissions to read everything
* I'm not sure we want to support this, doing it correctly and with reasonable semantics seems hard.
* That's actually _not_ this place which makes the resulting manifests indeterministic, but Python's TOML library itself when doing `.dumps()`. It's a WONTFIX from their side, as it makes no difference in parsed TOML. But it's a problem if we wanted to have reproducible builds.

pwmarcz (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: 70 of 71 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[python/graphenelibos/sgx_sign.py, line 176 at r2](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW9_Vay7iSL_e4JAGEF:-MWAfUlO1MV_Wi0xOdE1:b-3tmh8z) ([raw file](https://github.com/oscarlab/graphene/blob/ea8180007a71bb165d2136b66f02d4430003ef5c/python/graphenelibos/sgx_sign.py#L176)):*
Python dicts preserve order since Python 3.6, are you sure their WONTFIX is not outdated?

* I see an issue from 2013, which is pre-Python 3.4 (https://github.com/uiri/toml/pull/2)
* A quick look at [the source](https://github.com/uiri/toml/blob/master/toml/encoder.py) doesn't show anything weird, just dicts
* A simple experiment works:
```
>>> toml.dumps({'a': 1, 'b': 2})
'a = 1\nb = 2\n'
>>> toml.dumps({'b': 2, 'a': 1})
'b = 2\na = 1\n'
```

(I didn't look deep so I might still be wrong, though)

I don't think it's very important but I would still sort the results: either now, or after this becomes an array instead of dict.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: 70 of 71 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[python/graphenelibos/sgx_sign.py, line 176 at r2](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW9_Vay7iSL_e4JAGEF:-MWAn4QU5cZTM5VJ724U:b-aiuj83) ([raw file](https://github.com/oscarlab/graphene/blob/ea8180007a71bb165d2136b66f02d4430003ef5c/python/graphenelibos/sgx_sign.py#L176)):*
> Python dicts preserve order since Python 3.6, are you sure their WONTFIX is not outdated?

Oh... That explains why the initial version of the TOML-in-sgx_sign PR was failing only on 16.04 pipelines :)

If it's actually this way, then sorting seems like a good idea.

mkow:
Jenkins, retest Jenkins-Debug please (`apps.LTP.kill11` timed out, known issue)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Examples/ra-tls-secret-prov/Makefile, line 103 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6xGWE90_gbO-cpTpL:-MWBqQN3C8NeZXOLckSV:bswinzb) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-secret-prov/Makefile#L103)):*

Without setting `LD_LIBRARY_PATH` this you get:
```
/usr/bin/ld: warning: libsgx_util.so, needed by ./libs//libsecret_prov_verify_epid.so, not found (try using -rpath or -rpath-link)
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `hexdump_mem_to_buffer'
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `ias_cleanup'
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `ias_init'
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `verify_quote'
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `parse_hex'
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `verify_ias_report_extract_quote'
/usr/bin/ld: ./libs//libsecret_prov_verify_epid.so: undefined reference to `ias_verify_quote_raw'
collect2: error: ld returned 1 exit status
```
Seems that if the files were in the main directory it was able to find them. But I'm not really sure what's happening here...
</blockquote></details>

Linker is not able to find the shared libs, to validate symbols. There are two acceptable solutions:
1)  Change `rpath` in `LFLAGS` to `-Wl,-rpath,./libs/` and remove **all** `LD_LIBRARY_PATHS` in this file.
2) Change `rpath` in `LFLAGS` to `-Wl,-rpath-link,./libs/` and remove `LD_LIBRARY_PATHS` from compilation steps.
In normal makefile I would insist on 2) but these are tests and we do not export/ship/share created files, so 1) is fine as well.
___
*[python/graphenelibos/sgx_sign.py, line 180 at r4](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWBrV7t66ZFcNIX3NVp:-MWBrV7t66ZFcNIX3NVq:b60y0i9) ([raw file](https://github.com/oscarlab/graphene/blob/61c8b620d24554d31cb9fabb2c5273453e3d7a3b/python/graphenelibos/sgx_sign.py#L180)):*
> ```Python
>     for sub_path in path.rglob('*'):
>         if sub_path.is_file():
>             res.append(sub_path)
> ```

This loop screams to be onelined...

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: 69 of 71 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[Examples/ra-tls-secret-prov/Makefile, line 103 at r1](https://reviewable.io/reviews/oscarlab/graphene/2249#-MW6xGWE90_gbO-cpTpL:-MWC5aIY7uZHq7g23aQu:b4werxv) ([raw file](https://github.com/oscarlab/graphene/blob/f51b23be8656e73bfa0de7eef05b9d97dfa146fe/Examples/ra-tls-secret-prov/Makefile#L103)):*

Linker is not able to find the shared libs, to validate symbols. There are two acceptable solutions:
1)  Change `rpath` in `LFLAGS` to `-Wl,-rpath,./libs/` and remove **all** `LD_LIBRARY_PATHS` in this file.
2) Change `rpath` in `LFLAGS` to `-Wl,-rpath-link,./libs/` and remove `LD_LIBRARY_PATHS` from compilation steps.
In normal makefile I would insist on 2) but these are tests and we do not export/ship/share created files, so 1) is fine as well.
</blockquote></details>

Aah, I get it now. I did 1. for simplicity and commented that this isn't for production.
___
*[python/graphenelibos/sgx_sign.py, line 180 at r4](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWBrV7t66ZFcNIX3NVp:-MWC87hp5RWitiISG7AR:b2nnez8) ([raw file](https://github.com/oscarlab/graphene/blob/61c8b620d24554d31cb9fabb2c5273453e3d7a3b/python/graphenelibos/sgx_sign.py#L180)):*

This loop screams to be onelined...
</blockquote></details>

You mean into list comprehension? I tried and it doesn't look more readable than the current version.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[python/graphenelibos/sgx_sign.py, line 180 at r4](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWBrV7t66ZFcNIX3NVp:-MWCGRmz63u1OsMzapSa:b-qalik7) ([raw file](https://github.com/oscarlab/graphene/blob/61c8b620d24554d31cb9fabb2c5273453e3d7a3b/python/graphenelibos/sgx_sign.py#L180)):*

You mean into list comprehension? I tried and it doesn't look more readable than the current version.
</blockquote></details>

`return sorted(filter(Path.is_file, path.rglob('*')))` ?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: 68 of 71 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[python/graphenelibos/sgx_sign.py, line 180 at r4](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWBrV7t66ZFcNIX3NVp:-MWCQ91F6SEBI0C4Ykd-:beofm2i) ([raw file](https://github.com/oscarlab/graphene/blob/61c8b620d24554d31cb9fabb2c5273453e3d7a3b/python/graphenelibos/sgx_sign.py#L180)):*

`return sorted(filter(Path.is_file, path.rglob('*')))` ?
</blockquote></details>

Ok, I didn't think about passing a method to `filter()`.

mkow:
Jenkins, retest Jenkins please (`apps.LTP.kill11` failed, known issue)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Examples/ra-tls-mbedtls/Makefile, line 98 at r6](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWC_l-o5iNd-m6wEHgh:-MWC_l-p8p4MLYxsu6UH:bddg2l4) ([raw file](https://github.com/oscarlab/graphene/blob/6f2bf3d8f4dc6624e8d16bdac889c943de47fa8d/Examples/ra-tls-mbedtls/Makefile#L98)):*
> ```Makefile
> CFLAGS += -I./mbedtls/install/include -I./mbedtls/crypto/include
> # These flags are *NOT* for production, setting rpath is here only to simplify the test setup.
> LFLAGS += -Wl,-rpath,./libs/ -L./libs/ -ldl -lmbedcrypto -lmbedtls -lmbedx509
> ```

`rpath-link` since we do not need `rpath` now.
___
*[Examples/ra-tls-secret-prov/Makefile, line 101 at r6](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWCa-YaCTV1f9_dIWYd:-MWCa-YaCTV1f9_dIWYe:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/6f2bf3d8f4dc6624e8d16bdac889c943de47fa8d/Examples/ra-tls-secret-prov/Makefile#L101)):*
> ```Makefile
> CFLAGS += -Wall -std=c11 -I$(GRAPHENEDIR)/Pal/src/host/Linux-SGX/tools/ra-tls
> # These flags are *NOT* for production, setting rpath is here only to simplify the test setup.
> LFLAGS += -Wl,-rpath,./libs/ -L./libs/ -lmbedcrypto -lmbedtls -lmbedx509
> ```

ditto

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: 69 of 71 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Examples/ra-tls-mbedtls/Makefile, line 98 at r6](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWC_l-o5iNd-m6wEHgh:-MWCavuH5Vl1KZEJ5oaW:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6f2bf3d8f4dc6624e8d16bdac889c943de47fa8d/Examples/ra-tls-mbedtls/Makefile#L98)):*

`rpath-link` since we do not need `rpath` now.
</blockquote></details>

Done.
___
*[Examples/ra-tls-secret-prov/Makefile, line 101 at r6](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWCa-YaCTV1f9_dIWYd:-MWCaw_63qcOE_hI2wBI:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6f2bf3d8f4dc6624e8d16bdac889c943de47fa8d/Examples/ra-tls-secret-prov/Makefile#L101)):*

ditto
</blockquote></details>

Done.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (changes requested):
> Reviewed 41 of 70 files at r1, 4 of 11 files at r2, 23 of 24 files at r3, 1 of 3 files at r6, 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: all files reviewed, 15 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYTdXVEEm-P3aj4HJT:-MWYTdXWAggaEN8Rcx4X:b-qhr8v5) (no related file):*
Not really related to this PR, but I found that https://github.com/oscarlab/graphene/blob/master/Examples/README.md talks about `Manifest(s)`. Can we fix it to talk about just one central manifest (we don't have several manifests for an example anymore)?

---
___
*[Documentation/tutorials/pytorch/index.rst, line 258 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYJN52EHhLU286tYf1:-MWYJN52EHhLU286tYf2:b-2d1v0c) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Documentation/tutorials/pytorch/index.rst#L258)):*
> ```reStructuredText
> verified and trusted by the SGX enclave (in the latter case it's recursively
> traversed and all files are added as trusted). Note that the key string
> ``<name>`` may be an arbitrary legal string (but without ``-`` and other special
> ```

This "without -" is false for some time (after introduction of TOML) but unrelated. Not blocking.
___
*[Examples/apache/Makefile, line 77 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYJnuyD7SrnV2REQ-h:-MWYJnuyD7SrnV2REQ-i:b-xka5vt) ([raw file](https://github.com/oscarlab/graphene/blob/1102fee8f5f5ea85316b1d20a736500b0f1bc04d/Examples/apache/Makefile#L77)):*
> ```Makefile
> # binaries are already listed in the manifest template, so we can skip them
> # from the ldd results
> GLIBC_DEPS = linux-vdso /lib64/ld-linux-x86-64 libc libm librt libdl libpthread
> ```

This is great! This was a really obscure thing in our Makefiles.
___
*[Examples/bash/manifest.template, line 59 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYK4ak0bdjgr4cYKDb:-MWYK4ak0bdjgr4cYKDc:b-uk213c) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/bash/manifest.template#L59)):*
> ```
> sgx.trusted_files.runtime = "file:$(GRAPHENEDIR)/Runtime/"
> sgx.trusted_files.arch_libdir = "file:$(ARCH_LIBDIR)/"
> sgx.trusted_files.usr_arch_libdir = "file:/usr/$(ARCH_LIBDIR)/"
> ```

This `/usr/ARCH_LIBDIR` doesn't seem to be needed. Did you add it just for uniformity?

UPDATE: Ok, I see that you added it everywhere, just for uniformity. Unblocking.
___
*[Examples/busybox/busybox.manifest.template, line 82 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYQtfe5c53IjyHYbzo:-MWYQtfe5c53IjyHYbzp:b7k1pho) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/busybox/busybox.manifest.template#L82)):*
> ```
> # specified library, measures its hash and outputs the hash in auto-generated
> # entry 'sgx.trusted_checksum.xxx' in auto-generated 'busybox.manifest.sgx'
> # file. Note that this happens on the developer's machine.
> ```

Please add somewhere here that we can also specify directories (and that's what we use in this example).
___
*[Examples/memcached/memcached.manifest.template, line 133 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYQoWF6BwzIdOxkDFr:-MWYQoWF6BwzIdOxkDFs:b7k1pho) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/memcached/memcached.manifest.template#L133)):*
> ```
> # specified file, measures its hash, and outputs the hash in auto-generated
> # entry 'sgx.trusted_checksum.xxx' in auto-generated memcached.manifest.sgx.
> # Note that this happens on the developer machine or a build server.
> ```

Please add somewhere here that we can also specify directories (and that's what we use in this example).
___
*[Examples/python-scipy-insecure/python.manifest.template, line 119 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYTNNQ-6b5MTGhAZUs:-MWYTNNQ-6b5MTGhAZUt:b-s65wdn) ([raw file](https://github.com/oscarlab/graphene/blob/1102fee8f5f5ea85316b1d20a736500b0f1bc04d/Examples/python-scipy-insecure/python.manifest.template#L119)):*
> ```
> sgx.allowed_files.tmp = "file:/tmp"
> sgx.allowed_files.etc = "file:/etc"
> sgx.allowed_files.pyhome = "file:$(PYTHONHOME)"
> ```

The README for this example talks about "insecure allowed files", but this is not true anymore -- you use trusted files now. Please fix the README.
___
*[Examples/python-simple/python.manifest.template, line 86 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYMm_11HcaznzkPLjB:-MWYMm_11HcaznzkPLjC:b-ap7cx9) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/python-simple/python.manifest.template#L86)):*
> ```
> sgx.trusted_files.python = "file:$(PYTHONEXEC)"
> 
> sgx.trusted_files.runtime = "file:$(GRAPHENEDIR)/Runtime/"
> ```

There is now not much difference between `python-simple` and `python-scipy-insecure`. So both can be merged into one `python` example. But this can be done in a future PR.
___
*[Examples/ra-tls-mbedtls/Makefile, line 224 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYOJJBAyZelCPdhuJv:-MWYOJJBAyZelCPdhuJw:bm9gv3l) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/ra-tls-mbedtls/Makefile#L224)):*
> ```Makefile
> 	SGX=1 ./pal_loader server epid dummy-option >/dev/null & SERVER_ID=$$!; \
> 	sleep 30; \
> 	export LD_LIBRARY_PATH=$(LD_LIBRARY_PATH):./libs; \
> ```

You need to update the README file as well, adding `LD_LIBRARY_PATH=...` there.
___
*[Examples/ra-tls-secret-prov/Makefile, line 233 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYP9VO4jCeMUGpTttq:-MWYP9VO4jCeMUGpTttr:bjaaoev) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/ra-tls-secret-prov/Makefile#L233)):*
> ```Makefile
> .PHONY: check_epid
> check_epid: app epid files/input.txt
> 	LD_LIBRARY_PATH=./libs/ ./secret_prov_server_epid >/dev/null & SERVER_ID=$$!; \
> ```

Why didn't you use `export LD_LIBRARY_PATH` as in the above example?
___
*[Examples/ra-tls-secret-prov/Makefile, line 233 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYQNSF98Tzhu3-P4pY:-MWYQNSF98Tzhu3-P4pZ:b-hxh7w3) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/ra-tls-secret-prov/Makefile#L233)):*
> ```Makefile
> .PHONY: check_epid
> check_epid: app epid files/input.txt
> 	LD_LIBRARY_PATH=./libs/ ./secret_prov_server_epid >/dev/null & SERVER_ID=$$!; \
> ```

You need to update the README file with `LD_LIBRARY_PATH`
___
*[Examples/ra-tls-secret-prov/Makefile, line 247 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYQ-BjBkITRgff834g:-MWYQ-BjBkITRgff834h:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/ra-tls-secret-prov/Makefile#L247)):*
> ```Makefile
> .PHONY: check_dcap
> check_dcap: app dcap files/input.txt
> 	LD_LIBRARY_PATH=./libs/ ./secret_prov_server_dcap >/dev/null & SERVER_ID=$$!; \
> ```

ditto
___
*[Examples/redis/redis-server.manifest.template, line 105 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYQfNQCRUIX5ZQvTzz:-MWYQfNQCRUIX5ZQvU--:b7k1pho) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/redis/redis-server.manifest.template#L105)):*
> ```
> # specified file, measures its hash, and outputs the hash in auto-generated
> # entry 'sgx.trusted_checksum.xxx' in auto-generated redis-server.manifest.sgx.
> # Note that this happens on the developer machine or a build server.
> ```

Please add somewhere here that we can also specify directories (and that's what we use in this example).
___
*[Examples/tensorflow-lite/Makefile, line 17 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYRIwz1Fk5rcdBOcuq:-MWYRIwz1Fk5rcdBOcur:b-f0cz0g) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/tensorflow-lite/Makefile#L17)):*
> ```Makefile
> else
> GRAPHENE_LOG_LEVEL = error
> endif
> ```

So there was a bug before? Hm, how could this work... Probably nobody tested this example in a while
___
*[tests/benchmarks/redis-server.manifest.template, line 36 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYRtDt2G_L1qLIotly:-MWYRtDt2G_L1qLIotlz:bq97k8e) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/tests/benchmarks/redis-server.manifest.template#L36)):*
> ```
> sgx.trusted_files.libgpgerror = file:@ARCH_LIBDIR@/libgpg-error.so.0
> sgx.trusted_files.liblz4 = file:/usr/@ARCH_LIBDIR@/liblz4.so.1
> sgx.trusted_files.libjemaloc = file:/usr/@ARCH_LIBDIR@/libjemalloc.so.1
> ```

Why not add `ARCH_LIBDIR` as a trusted directory?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: 67 of 77 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYTdXVEEm-P3aj4HJT:-MWYugU30fXUCuZNbtUc:b-896fix) (no related file):*

Not really related to this PR, but I found that https://github.com/oscarlab/graphene/blob/master/Examples/README.md talks about `Manifest(s)`. Can we fix it to talk about just one central manifest (we don't have several manifests for an example anymore)?
</blockquote></details>

Done.

---
___
*[Documentation/tutorials/pytorch/index.rst, line 258 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYJN52EHhLU286tYf1:-MWYvBO22jh4Em5Bhom5:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Documentation/tutorials/pytorch/index.rst#L258)):*

This "without -" is false for some time (after introduction of TOML) but unrelated. Not blocking.
</blockquote></details>

Done.
___
*[Examples/apache/Makefile, line 77 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYJnuyD7SrnV2REQ-h:-MWYvGuf0xeYI_Qh9zv1:b32vhw) ([raw file](https://github.com/oscarlab/graphene/blob/1102fee8f5f5ea85316b1d20a736500b0f1bc04d/Examples/apache/Makefile#L77)):*

This is great! This was a really obscure thing in our Makefiles.
</blockquote></details>

:)
___
*[Examples/busybox/busybox.manifest.template, line 82 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYQtfe5c53IjyHYbzo:-MWYv_sSEAckMlkmvlMh:b-1c1t4) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/busybox/busybox.manifest.template#L82)):*

Please add somewhere here that we can also specify directories (and that's what we use in this example).
</blockquote></details>

I'd skip this, as the current description is correct. I plan to get rid of this comment duplication hell in our examples and have only one "richly commented" (and linked to in the readmes as the go-to place for newcomers).
___
*[Examples/memcached/memcached.manifest.template, line 133 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYQoWF6BwzIdOxkDFr:-MWYvw_rBzZX5ZEkoyjs:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/memcached/memcached.manifest.template#L133)):*

Please add somewhere here that we can also specify directories (and that's what we use in this example).
</blockquote></details>

ditto
___
*[Examples/python-simple/python.manifest.template, line 86 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYMm_11HcaznzkPLjB:-MWYwZBg8rNWrCAzhUZW:bgulf79) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/python-simple/python.manifest.template#L86)):*

There is now not much difference between `python-simple` and `python-scipy-insecure`. So both can be merged into one `python` example. But this can be done in a future PR.
</blockquote></details>

I added this to my todolist.
___
*[Examples/ra-tls-mbedtls/Makefile, line 224 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYOJJBAyZelCPdhuJv:-MWYxksHCahi1k1-MJkm:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/ra-tls-mbedtls/Makefile#L224)):*

You need to update the README file as well, adding `LD_LIBRARY_PATH=...` there.
</blockquote></details>

Done.
___
*[Examples/ra-tls-secret-prov/Makefile, line 233 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYP9VO4jCeMUGpTttq:-MWYxpSKB0LjThHIJJcD:b-ojjx46) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/ra-tls-secret-prov/Makefile#L233)):*

Why didn't you use `export LD_LIBRARY_PATH` as in the above example?
</blockquote></details>

Because here it's just 2 places, so we can have it nicely constrained (i.e. other commands won't get this directory).
___
*[Examples/ra-tls-secret-prov/Makefile, line 233 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYQNSF98Tzhu3-P4pY:-MWYy6MNAOAn-v_occk_:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/ra-tls-secret-prov/Makefile#L233)):*

You need to update the README file with `LD_LIBRARY_PATH`
</blockquote></details>

Done.
___
*[Examples/ra-tls-secret-prov/Makefile, line 247 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYQ-BjBkITRgff834g:-MWYy7NR28U5Npd6JPlJ:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/ra-tls-secret-prov/Makefile#L247)):*

ditto
</blockquote></details>

ditto
___
*[Examples/redis/redis-server.manifest.template, line 105 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYQfNQCRUIX5ZQvTzz:-MWYyDU43ZgLt1wfJbwl:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/redis/redis-server.manifest.template#L105)):*

Please add somewhere here that we can also specify directories (and that's what we use in this example).
</blockquote></details>

ditto
___
*[Examples/tensorflow-lite/Makefile, line 17 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYRIwz1Fk5rcdBOcuq:-MWYyKw47NLZcVrEHEoc:b-x0r7w0) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/Examples/tensorflow-lite/Makefile#L17)):*

So there was a bug before? Hm, how could this work... Probably nobody tested this example in a while
</blockquote></details>

Yep :)
___
*[tests/benchmarks/redis-server.manifest.template, line 36 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYRtDt2G_L1qLIotly:-MWYyR2LDcwf9SEnVlDa:b-41f9yy) ([raw file](https://github.com/oscarlab/graphene/blob/1fed91910b3c1483770742fd6e0a8f4f94cb7f5b/tests/benchmarks/redis-server.manifest.template#L36)):*

Why not add `ARCH_LIBDIR` as a trusted directory?
</blockquote></details>

Oh, I missed this. Fixed.
___
*[Examples/python-scipy-insecure/python.manifest.template, line 119 at r7](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWYTNNQ-6b5MTGhAZUs:-MWYwS9WBRxWOBPbxrb8:b47hegd) ([raw file](https://github.com/oscarlab/graphene/blob/1102fee8f5f5ea85316b1d20a736500b0f1bc04d/Examples/python-scipy-insecure/python.manifest.template#L119)):*

The README for this example talks about "insecure allowed files", but this is not true anymore -- you use trusted files now. Please fix the README.
</blockquote></details>

Done. Also renamed the example.

boryspoplawski (dismissed):
> Reviewed 10 of 10 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: all files reviewed, 10 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (changes requested):
> Reviewed 10 of 10 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWZ2Om3BAwg8Oci8Cpj:-MWZ2Om4Bc1E7owocyvn:bfbl6bl) (no related file):*
> python-scipy-insecure

There are a bunch of places that use this old name, see https://github.com/oscarlab/graphene/search?q=python-scipy-insecure. Please fix them to use the new name.

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: 77 of 81 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2249#-MWZ2Om3BAwg8Oci8Cpj:-MWZ3F90-U92d8N76eiQ:bql6urd) (no related file):*

> python-scipy-insecure

There are a bunch of places that use this old name, see https://github.com/oscarlab/graphene/search?q=python-scipy-insecure. Please fix them to use the new name.
</blockquote></details>

Should be fixed now.

---

boryspoplawski (dismissed):
> Reviewed 4 of 4 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

dimakuv (approved):
> Reviewed 4 of 4 files at r9, 1 of 1 files at r10.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2249)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2269, opened 2021-03-24
Add musl support

### LINKED ISSUE #49: [LibOS] Add support for Musl libc
Currently Graphene supports only Glibc as the standard C library. Add support for Musl libc.

This will be beneficial for IoT/embedded users.

mkow:
## Description of the changes 

Disclaimer: musl is not binary-compatible with most programs, only the simplest can use musl binaries without recompilation. Others have to be compiled against musl headers (see [musl documentation](https://musl.libc.org/doc/1.1.24/manual.html)).

Resolves https://github.com/gramineproject/gramine/issues/49.

## How to test this PR? 

See the testing discussion on Reviewable.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/graphene/2269)

### DIFF
--- a/Documentation/building.rst
+++ b/Documentation/building.rst
@@ -14,7 +14,7 @@ Graphene consists of several components:
 - The Library OS itself (a shared library named ``libsysdb.so``, called the
   "shim" in our source code)
 - The Platform Adaptation Layer, or PAL (a shared library named ``libpal.so``)
-- A patched GNU C or musl library (a set of shared libraries ``libc.so``,
+- A patched glibc or musl library (a set of shared libraries ``libc.so``,
   ``libpthread.so``, ``libm.so``, etc.)
 
 The build of Graphene implies building at least the first two components. The
--- a/Documentation/building.rst
+++ b/Documentation/building.rst
@@ -14,17 +14,17 @@ Graphene consists of several components:
 - The Library OS itself (a shared library named ``libsysdb.so``, called the
   "shim" in our source code)
 - The Platform Adaptation Layer, or PAL (a shared library named ``libpal.so``)
-- A patched GNU C Library (a set of shared libraries ``libc.so``,
+- A patched GNU C or musl library (a set of shared libraries ``libc.so``,
   ``libpthread.so``, ``libm.so``, etc.)
 
 The build of Graphene implies building at least the first two components. The
 build of the patched C library is optional but highly recommended for
-performance reasons. The patched C library is built by default.
+performance reasons. Patched glibc is built by default.
 
 Graphene currently only works on the x86_64 architecture. Graphene is currently
 tested on Ubuntu 16.04 and 18.04 (both server and desktop version), along with
-Linux kernel versions 3.x/4.x/5.x. We recommend building and installing Graphene
-on the same host platform. If you find problems with Graphene on other Linux
+Linux kernel versions 4.x/5.x. We recommend building and installing Graphene on
+the same host platform. If you find problems with Graphene on other Linux
 distributions, please contact us with a |~| detailed `bug report
 <https://github.com/oscarlab/graphene/issues/new>`__.
 
@@ -198,6 +198,13 @@ Running :command:`make SGX=1 sgx-tokens` in the test or regression directory
 will automatically generate the required manifest signatures (``.sig`` files)
 and EINITTOKENs (``.token`` files).
 
+Building with musl instead of glibc
+-----------------------------------
+
+To build with a patched musl, add ``LIBC=MUSL`` to the ``make`` invocation.
+You'll need to run :command:`make clean` in ``Runtime/`` directory if you
+already have Graphene built with glibc.
+
 Additional build options
 ------------------------
 
@@ -207,8 +214,8 @@ Additional build options
 
 - To build with ``-Werror``, run :command:`make WERROR=1`.
 
-- To specify custom mirrors for downloading the Glibc source, use :command:`make
-  GLIBC_MIRRORS=...`.
+- To specify custom mirrors for downloading the glibc source, use :command:`make
+  GLIBC_MIRRORS=...` (and MUSL_MIRRORS for musl).
 
 - Each part of Graphene can be built separately in the subdirectories. For
   example, to build only the Pal component, use :command:`make -c Pal`.
--- a/Examples/apache/httpd.manifest.template
+++ b/Examples/apache/httpd.manifest.template
@@ -20,7 +20,7 @@ sys.enable_sigterm_injection = 1
 # Mounted FSes. The following "chroot" FSes mount a part of the host FS into the
 # guest. Other parts of the host FS will not be available in the guest.
 
-# Default glibc files, mounted from the Runtime directory in GRAPHENEDIR.
+# Default libc files, mounted from the Runtime directory in GRAPHENEDIR.
 fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:$(GRAPHENEDIR)/Runtime"
--- a/Examples/bash/manifest.template
+++ b/Examples/bash/manifest.template
@@ -17,7 +17,7 @@ loader.env.PATH = "$(EXECDIR)"
 # Mounted FSes. The following "chroot" FSes mount a part of the host FS into the
 # guest. Other parts of the host FS will not be available in the guest.
 
-# Default glibc files, mounted from the Runtime directory in GRAPHENEDIR.
+# Default libc files, mounted from the Runtime directory in GRAPHENEDIR.
 fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:$(GRAPHENEDIR)/Runtime"
--- a/Examples/busybox/busybox.manifest.template
+++ b/Examples/busybox/busybox.manifest.template
@@ -31,7 +31,7 @@ loader.log_level = "$(GRAPHENE_LOG_LEVEL)"
 # fs.mount.xxx.path, not fs.mount.xxx.uri).
 #
 # In case of Busybox:
-# - /lib is searched for Glibc libraries (ld, libc, libresolv) provided by
+# - /lib is searched for libc libraries (ld, libc, libresolv) provided by
 #   graphene,
 # - $(ARCH_LIBDIR) is searched for libcrypt, NSS libraries.
 loader.env.LD_LIBRARY_PATH = "/lib:$(ARCH_LIBDIR)"
@@ -51,7 +51,7 @@ loader.env.HOSTNAME = "test"
 # - Names of mount entries (lib1, lib2) are irrelevant but must be unique.
 # - In-Graphene visible path names may be arbitrary.
 
-# Mount host-OS directory containing Graphene glibc/runtime libraries
+# Mount host-OS directory containing Graphene libc/runtime libraries
 # (specified as 'uri', GRAPHENEDIR macro is expanded in the Makefile as part
 # of the build process) into in-Graphene visible directory /lib (specified
 # as 'path').
--- a/Examples/capnproto/addressbook.manifest.template
+++ b/Examples/capnproto/addressbook.manifest.template
@@ -17,7 +17,7 @@ loader.insecure__use_cmdline_argv = 1
 # Specify paths to search for libraries (usual LD_LIBRARY_PATH syntax)
 loader.env.LD_LIBRARY_PATH = "/lib:$(ARCH_LIBDIR):/usr/$(ARCH_LIBDIR)"
 
-# Mount host-OS directory to Graphene glibc/runtime libraries
+# Mount host-OS directory to Graphene libc/runtime libraries
 fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:$(GRAPHENEDIR)/Runtime"
--- a/Examples/lighttpd/lighttpd.manifest.template
+++ b/Examples/lighttpd/lighttpd.manifest.template
@@ -26,7 +26,7 @@ sys.enable_sigterm_injection = 1
 # Mounted FSes. The following "chroot" FSes mount a part of the host FS into the
 # guest. Other parts of the host FS will not be available in the guest.
 
-# Default glibc files, mounted from the Runtime directory in GRAPHENEDIR.
+# Default libc files, mounted from the Runtime directory in GRAPHENEDIR.
 fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:$(GRAPHENEDIR)/Runtime"
--- a/Examples/memcached/memcached.manifest.template
+++ b/Examples/memcached/memcached.manifest.template
@@ -44,7 +44,7 @@ loader.insecure__use_cmdline_argv = 1
 # paths must be taken from fs.mount.xxx.path, not fs.mount.xxx.uri).
 #
 # In case of Memcached:
-# - /lib is searched for Glibc libraries (ld, libc, libpthread)
+# - /lib is searched for libc libraries (ld, libc, libpthread)
 # - $(ARCH_LIBDIR) is searched for Name Service Switch (NSS) libraries
 # - /usr/$(ARCH_LIBDIR) is searched for libevent (the only explicit
 #                             Memcached dependency)
@@ -67,7 +67,7 @@ sys.enable_sigterm_injection = 1
 # - In-Graphene visible path names may be arbitrary but we reuse host-OS URIs
 #   for simplicity (except for the first "lib" case).
 
-# Mount host-OS directory to Graphene glibc/runtime libraries (in "uri") into
+# Mount host-OS directory to Graphene libc/runtime libraries (in "uri") into
 # in-Graphene visible directory /lib (in "path"). Note that GRAPHENEDIR macro
 # is expanded to relative path to Graphene repository in the Makefile as part
 # of the build process.
--- a/Examples/nginx/nginx.manifest.template
+++ b/Examples/nginx/nginx.manifest.template
@@ -22,7 +22,7 @@ sys.enable_sigterm_injection = 1
 # Mounted FSes. The following "chroot" FSes mount a part of the host FS into the
 # guest. Other parts of the host FS will not be available in the guest.
 
-# Default glibc files, mounted from the Runtime directory in GRAPHENEDIR.
+# Default libc files, mounted from the Runtime directory in GRAPHENEDIR.
 fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:$(GRAPHENEDIR)/Runtime"
--- a/Examples/openvino/object_detection_sample_ssd.manifest.template
+++ b/Examples/openvino/object_detection_sample_ssd.manifest.template
@@ -19,7 +19,7 @@ loader.env.LD_LIBRARY_PATH = "/lib:$(OPENVINO_DIR)/inference-engine/temp/tbb/lib
 # Mounted FSes. The following "chroot" FSes mount a part of the host FS into the
 # guest. Other parts of the host FS will not be available in the guest.
 
-# Default glibc files, mounted from the Runtime directory in GRAPHENE_DIR.
+# Default libc files, mounted from the Runtime directory in GRAPHENE_DIR.
 fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:$(GRAPHENE_DIR)/Runtime"
--- a/Examples/python-scipy/python.manifest.template
+++ b/Examples/python-scipy/python.manifest.template
@@ -23,7 +23,7 @@ loader.env.HOME = "/home/user"
 # Mounted FSes. The following "chroot" FSes mount a part of the host FS into the
 # guest. Other parts of the host FS will not be available in the guest.
 
-# Default glibc files, mounted from the Runtime directory in GRAPHENEDIR.
+# Default libc files, mounted from the Runtime directory in GRAPHENEDIR.
 fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:$(GRAPHENEDIR)/Runtime"
--- a/Examples/python-simple/python.manifest.template
+++ b/Examples/python-simple/python.manifest.template
@@ -27,7 +27,7 @@ sys.enable_sigterm_injection = 1
 # Mounted FSes. The following "chroot" FSes mount a part of the host FS into the
 # guest. Other parts of the host FS will not be available in the guest.
 
-# Default glibc files, mounted from the Runtime directory in GRAPHENEDIR.
+# Default libc files, mounted from the Runtime directory in GRAPHENEDIR.
 fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:$(GRAPHENEDIR)/Runtime"
--- a/Examples/pytorch/pytorch.manifest.template
+++ b/Examples/pytorch/pytorch.manifest.template
@@ -19,7 +19,7 @@ loader.insecure__use_host_env = 1
 # Overwrite some environment variables
 loader.env.LD_LIBRARY_PATH = "/lib:/usr/lib:$(ARCH_LIBDIR):/usr/$(ARCH_LIBDIR)"
 
-# Default glibc files, mounted from the Runtime directory in GRAPHENEDIR
+# Default libc files, mounted from the Runtime directory in GRAPHENEDIR
 fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:$(GRAPHENEDIR)/Runtime/"
--- a/Examples/r/R.manifest.template
+++ b/Examples/r/R.manifest.template
@@ -28,7 +28,7 @@ loader.env.R_HOME = "$(R_HOME)"
 # Mounted FSes. The following "chroot" FSes mount a part of the host FS into the guest. Other parts
 # of the host FS will not be available in the guest.
 
-# Default glibc files, mounted from the Runtime directory in GRAPHENEDIR.
+# Default libc files, mounted from the Runtime directory in GRAPHENEDIR.
 fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:$(GRAPHENEDIR)/Runtime"
--- a/Examples/redis/redis-server.manifest.template
+++ b/Examples/redis/redis-server.manifest.template
@@ -29,7 +29,7 @@ loader.insecure__use_cmdline_argv = 1
 # paths must be taken from fs.mount.xxx.path, not fs.mount.xxx.uri).
 #
 # In case of Redis:
-# - /lib is searched for Glibc libraries (ld, libc, libpthread)
+# - /lib is searched for libc libraries (ld, libc, libpthread)
 # - $(ARCH_LIBDIR) is searched for Name Service Switch (NSS) libraries
 loader.env.LD_LIBRARY_PATH = "/lib:$(ARCH_LIBDIR):/usr/$(ARCH_LIBDIR)"
 
@@ -47,7 +47,7 @@ sys.enable_sigterm_injection = 1
 # - In-Graphene visible path names may be arbitrary but we reuse host-OS URIs
 #   for simplicity (except for the first 'lib' case).
 
-# Mount host-OS directory to Graphene glibc/runtime libraries (in 'uri') into
+# Mount host-OS directory to Graphene libc/runtime libraries (in 'uri') into
 # in-Graphene visible directory /lib (in 'path'). Note that GRAPHENEDIR macro
 # is expanded to relative path to Graphene repository in the Makefile as part
 # of the build process.
--- a/LibOS/.gitignore
+++ b/LibOS/.gitignore
@@ -3,6 +3,10 @@
 /glibc-build/
 /build.log
 
+/musl-*.*.*.tar.gz
+/musl-*.*.*/
+/musl-build/
+
 /gcc-*.tar.gz
 /gcc-*.*.*/
 /gcc-build/
--- a/LibOS/Makefile
+++ b/LibOS/Makefile
@@ -2,212 +2,37 @@ include ../Scripts/Makefile.configs
 include ../Scripts/Makefile.rules
 
 RUNTIME_DIR = $(CURDIR)/../Runtime
+SHIM_DIR = shim
 
 define LN_SF_TO_RUNTIME_DIR_template =
 $(RUNTIME_DIR)/$(notdir $(1)): $(1)
 	$$(call cmd,ln_sfr)
 endef
 
-GNU_MIRRORS ?= https://ftp.gnu.org/gnu/ \
-               https://mirrors.kernel.org/gnu/ \
-               https://mirrors.ocf.berkeley.edu/gnu/
-
-# ------------------------------------------------------------------------------------------------
-# Glibc build: patch libc.so and other C standard libraries to replace raw SYSCALL instructions
-# with function calls into Graphene (plus some smaller patches for Graphene). The resulting
-# libraries are symlinked under $RUNTIME_DIR.
-#
-# Override CC, CXX, AS in case we're compiling the rest of the project with clang.
-# ------------------------------------------------------------------------------------------------
-GLIBC_VERSION ?= 2.31
-GLIBC_SRC = glibc-$(GLIBC_VERSION)
-GLIBC_HASH = $(firstword $(shell grep $(GLIBC_SRC).tar.gz glibc-checksums))
-SHIM_DIR = shim
-BUILD_DIR = glibc-build
-GLIBC_LIBS = \
-	csu/crt1.o \
-	csu/crti.o \
-	csu/crtn.o \
-	dlfcn/libdl.so.2 \
-	libc.so \
-	libc.so.6 \
-	login/libutil.so.1 \
-	math/libm.so.6 \
-	nptl/libpthread.so.0 \
-	nptl_db/libthread_db.so.1 \
-	resolv/libnss_dns.so.2 \
-	resolv/libresolv.so.2 \
-	rt/librt.so.1
-ifeq ($(ARCH),x86_64)
-GLIBC_LIBS += elf/ld-linux-x86-64.so.2
+ifeq ($(LIBC),)
+	# default
+	LIBC = GLIBC
 endif
 
-GLIBC_TARGET = $(addprefix $(BUILD_DIR)/, $(GLIBC_LIBS))
-GLIBC_RUNTIME = $(addprefix $(RUNTIME_DIR)/, $(notdir $(GLIBC_TARGET)))
+# These are set in libc-specific makefiles included below.
+LIBC_TARGETS =
+LIBC_LINK_TARGETS =
+CLEAN_FILES =
+DISTCLEAN_FILES =
 
-GLIBC_CFLAGS = -O2 -Wp,-U_FORTIFY_SOURCE -Wno-unused-value
-ifeq ($(DEBUG),1)
-	GLIBC_CFLAGS += -g
+ifeq ($(LIBC),GLIBC)
+	include glibc.mk
+else ifeq ($(LIBC),MUSL)
+	include musl.mk
+else
+	$(error Error: LIBC must be either GLIBC or MUSL)
 endif
-export GLIBC_CFLAGS
 
-# For LibOS entry API
-GLIBC_CPPFLAGS = -I$(abspath $(SHIM_DIR))/include -I$(abspath $(SHIM_DIR))/include/arch/$(ARCH)
-export GLIBC_CPPFLAGS
-GLIBC_DEPS = shim/include/arch/$(ARCH)/shim_entry_api.h
+$(foreach lib,$(LIBC_LINK_TARGETS),$(eval $(call LN_SF_TO_RUNTIME_DIR_template,$(lib))))
 
 .PHONY: all
-all: $(GLIBC_TARGET) $(GLIBC_RUNTIME)
+all: $(LIBC_TARGETS)
 	$(MAKE) -C $(SHIM_DIR) all
-
-ifeq ($(findstring x86_64,$(SYS))$(findstring linux,$(SYS)),x86_64linux)
-
-.SECONDARY: $(BUILD_DIR)/Build.success
-
-$(BUILD_DIR)/Build.success: $(BUILD_DIR)/Makefile
-	@echo "Building glibc, may take a while to finish. Warning messages may show up. If this process terminates with failures, see \"$(BUILD_DIR)/build.log\" for more information."
-	(CC=gcc CXX=g++ AS=gcc $(MAKE) -C $(BUILD_DIR) 2>&1 > build.log) && touch $@
-
-$(GLIBC_TARGET): $(BUILD_DIR)/Build.success
-
-$(BUILD_DIR)/Makefile: $(GLIBC_SRC)/.configured
-	mkdir -p $(BUILD_DIR)
-	(cd $(BUILD_DIR) || exit 1; \
-	CC=gcc CXX=g++ AS=gcc \
-	CFLAGS=$$GLIBC_CFLAGS CPPFLAGS=$$GLIBC_CPPFLAGS \
-	../$(GLIBC_SRC)/configure --prefix=$(RUNTIME_DIR) \
-		--with-tls \
-		--without-selinux \
-		--disable-test \
-		--disable-nscd \
-	)
-
-$(foreach lib,$(GLIBC_TARGET),$(eval $(call LN_SF_TO_RUNTIME_DIR_template,$(lib))))
-
-GLIBC_PATCHES = \
-	glibc-patches/$(GLIBC_SRC).patch
-
-GLIBC_PATCHES_2.27 = \
-	glibc-patches/hp-timing-2.27.patch
-
-GLIBC_PATCHES_2.31 = \
-	glibc-patches/hp-timing-2.31.patch
-
-GLIBC_PATCHES += $(GLIBC_PATCHES_$(GLIBC_VERSION))
-
-.SECONDARY: $(GLIBC_SRC)/.configured
-$(GLIBC_SRC)/.configured: $(GLIBC_PATCHES) $(GLIBC_SRC).tar.gz $(GLIBC_DEPS)
-	$(RM) -r $(GLIBC_SRC)
-	tar -mxzf $(GLIBC_SRC).tar.gz
-	cd $(GLIBC_SRC) && \
-	for p in $(GLIBC_PATCHES); do \
-		echo applying $$p; \
-		patch -p1 -l < ../$$p || exit 255; \
-	done
-	touch $@
-
-$(GLIBC_SRC).tar.gz:
-	../Scripts/download --output $@ --sha256 $(GLIBC_HASH) $(foreach mirror,$(GNU_MIRRORS),--url $(mirror)glibc/$(GLIBC_SRC).tar.gz)
-
-$(GLIBC_SRC)/elf/Versions: $(GLIBC_SRC)/.configured
-
-$(GLIBC_SRC)/nptl/Versions: $(GLIBC_SRC)/.configured
-
-$(GLIBC_SRC)/dlfcn/Versions: $(GLIBC_SRC)/.configured
-
-# ------------------------------------------------------------------------------------------------
-# GCC build: patch libgomp.so.1 (OpenMP runtime library) to replace raw SYSCALL instruction with
-# function call into Graphene. GCC is not built by default with Graphene; use `make -C LibOS gcc`
-# to build it. The resulting libgomp.so.1 is symlinked under $RUNTIME_DIR. This patched version
-# makes sense only on x86_64 platforms. NOTE: We'd prefer to build libgomp.so.1 alone but it is
-# impossible (the only way to build it is as part of the complete GCC build).
-#
-# We explicitly unset CC, CXX, AS environment variables for the case we're compiling the rest of the
-# project with clang. This is because in GCC build, "defining certain environment variables such as
-# CC can interfere with the functioning of make" (https://gcc.gnu.org/install/build.html). Indeed,
-# defining CC=gcc or CC=clang leads to errors during GCC build.
-# ------------------------------------------------------------------------------------------------
-GCC_VERSION ?= 10.2.0
-GCC_SRC = gcc-$(GCC_VERSION)
-GCC_HASH = 27e879dccc639cd7b0cc08ed575c1669492579529b53c9ff27b0b96265fa867d
-GCC_BUILD_DIR = gcc-build
-GCC_LIBS = x86_64-pc-linux-gnu/libgomp/.libs/libgomp.so.1
-GCC_TARGET = $(addprefix $(GCC_BUILD_DIR)/, $(GCC_LIBS))
-GCC_RUNTIME = $(addprefix $(RUNTIME_DIR)/, $(notdir $(GCC_TARGET)))
-
-.SECONDARY: $(GCC_BUILD_DIR)/Build.success
-
-$(GCC_BUILD_DIR)/Build.success: $(GCC_BUILD_DIR)/Makefile
-	@echo "Building gcc, may take a while to finish. Warning messages may show up. If this process terminates with failures, see \"$(GCC_BUILD_DIR)/gcc-build.log\" for more information."
-	(unset CC CXX AS; $(MAKE) -C $(GCC_BUILD_DIR) 2>&1 > gcc-build.log) && touch $@
-
-$(GCC_TARGET): $(GCC_BUILD_DIR)/Build.success
-
-$(GCC_BUILD_DIR)/Makefile: $(GCC_SRC)/.configured
-	mkdir -p $(GCC_BUILD_DIR)
-	(cd $(GCC_BUILD_DIR) || exit 1; \
-	unset CC CXX AS; \
-	../$(GCC_SRC)/configure --prefix=$(RUNTIME_DIR) \
-		--enable-languages=c \
-		--disable-multilib \
-	)
-
-$(foreach lib,$(GCC_TARGET),$(eval $(call LN_SF_TO_RUNTIME_DIR_template,$(lib))))
-
-GCC_PATCHES = \
-	gcc-patches/libgomp-replace-futex-instruction.patch
-
-# For LibOS entry API (file symlinked in patch)
-GCC_DEPS = shim/include/arch/$(ARCH)/shim_entry_api.h
-
-.SECONDARY: $(GCC_SRC)/.configured
-$(GCC_SRC)/.configured: $(GCC_PATCHES) $(GCC_SRC).tar.gz $(GCC_DEPS)
-	$(RM) -r $(GCC_SRC)
-	tar -mxzf $(GCC_SRC).tar.gz
-	cd $(GCC_SRC) && \
-	for p in $(GCC_PATCHES); do \
-		echo applying $$p; \
-		patch -p1 -l < ../$$p || exit 255; \
-	done
-	cd $(GCC_SRC) && ./contrib/download_prerequisites
-	touch $@
-
-$(GCC_SRC).tar.gz:
-	../Scripts/download --output $@ --sha256 $(GCC_HASH) $(foreach mirror,$(GNU_MIRRORS),--url $(mirror)gcc/$(GCC_SRC)/$(GCC_SRC).tar.gz)
-
-.PHONY: gcc
-gcc: $(GCC_TARGET) $(GCC_RUNTIME)
-
-# ------------------------------------------------------------------------------------------------
-# Common targets
-# ------------------------------------------------------------------------------------------------
-.PHONY: clean_
-clean_:
-	$(RM) -r $(BUILD_DIR) $(GLIBC_SRC) $(GCC_BUILD_DIR) $(GCC_SRC) build.log gcc-build.log
-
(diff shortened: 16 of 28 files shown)


mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: 0 of 27 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MW_Src3EThqsJR-Qtku:-MW_Src48rGVWWyplHce:bs042jq) (no related file):*
There's an issue with testing this PR - how to exactly test it in CI?

Problems:
- 4 LibOS regression tests (`TC_01_Bootstrap::test_110_basic_bootstrapping_cpp`, `TC_02_OpenMP::test_000_simple_for_loop`, `TC_30_Syscall::test_080_sched`, `TC_40_FileSystem::test_050_sysfs`) are failing due to missing functionalities in musl.
- Our examples would need recompilation against musl to work.
- We'd need to somehow swap libraries in Runtime/ if we wanted to test both.

Any ideas? From my side I'd be fine with just one example or the LibOS regression working in CI.

CC: @boryspoplawski @dimakuv @woju

---

woju (commented):
> Reviewed 19 of 27 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: 19 of 27 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MW_Src3EThqsJR-Qtku:-MWa8aWQCxZPgju6w18x:b7wnrk5) (no related file):*

There's an issue with testing this PR - how to exactly test it in CI?

Problems:
- 4 LibOS regression tests (`TC_01_Bootstrap::test_110_basic_bootstrapping_cpp`, `TC_02_OpenMP::test_000_simple_for_loop`, `TC_30_Syscall::test_080_sched`, `TC_40_FileSystem::test_050_sysfs`) are failing due to missing functionalities in musl.
- Our examples would need recompilation against musl to work.
- We'd need to somehow swap libraries in Runtime/ if we wanted to test both.

Any ideas? From my side I'd be fine with just one example or the LibOS regression working in CI.

CC: @boryspoplawski @dimakuv @woju
</blockquote></details>

Ad missing features in musl, you can skip tests if the binaries are not built. I think this is no big deal.
If the Examples and Runtime need recompilation, I think it should be done in a separate pipeline.

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: 19 of 27 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MW_Src3EThqsJR-Qtku:-MWaJ6jC7FrbtE3J1in0:b-y5hhll) (no related file):*
> Ad missing features in musl, you can skip tests if the binaries are not built. I think this is no big deal.

Yeah, but how exactly? Just looking at the env variable? This would assume that the same variable will be passed both to the build and the tests invocation.

>  If the Examples and Runtime need recompilation, I think it should be done in a separate pipeline.

Right now this would involve building the whole Graphene again... But with musl it's quite fast, so I guess we can do this. OTOH, this will result in at least 2 new pipelines (SGX and non-SGX), which gets pretty hard to manage without a real matrix support in CI.

---

yamahata (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: 19 of 27 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[LibOS/musl-patches/musl.patch, line 97 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcJZWm-ZmMH7fl2WKn:-MWcJZWnEMKOYBowH3d_:b-r4gzmq) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl-patches/musl.patch#L97)):*
> ```Diff
>  		install_new_tls();
>  	orig_tail = tail;
> +
> ```

nitpick: useless hunk.
please drop this hunk.

boryspoplawski (changes requested):
> Reviewed 27 of 27 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv, @mkow, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MW_Src3EThqsJR-Qtku:-MWcCn_YCw-PTHk-5Ynf:bu30mfw) (no related file):*

> Ad missing features in musl, you can skip tests if the binaries are not built. I think this is no big deal.

Yeah, but how exactly? Just looking at the env variable? This would assume that the same variable will be passed both to the build and the tests invocation.

>  If the Examples and Runtime need recompilation, I think it should be done in a separate pipeline.

Right now this would involve building the whole Graphene again... But with musl it's quite fast, so I guess we can do this. OTOH, this will result in at least 2 new pipelines (SGX and non-SGX), which gets pretty hard to manage without a real matrix support in CI.
</blockquote></details>

+1 to running it in a separate pipeline (+ it solves all mentioned issues)

---
___
*[LibOS/glibc.mk, line 124 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcMEFM3Huw41enRgD3:-MWcMEFM3Huw41enRgD4:bj65rq8) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/glibc.mk#L124)):*
> ```Makefile
> $(GCC_BUILD_DIR)/Build.success: $(GCC_BUILD_DIR)/Makefile
> 	@echo "Building gcc, may take a while to finish. Warning messages may show up. If this process terminates with failures, see \"$(GCC_BUILD_DIR)/gcc-build.log\" for more information."
> 	(unset CC CXX AS; $(MAKE) -C $(GCC_BUILD_DIR) 2>&1 > gcc-build.log) && touch $@
> ```

Please split into two lines just line in glibc
___
*[LibOS/Makefile, line 24 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcFW9n6iy4c9w4nlmA:-MWcFW9n6iy4c9w4nlmB:b-psg70p) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/Makefile#L24)):*
> ```Makefile
> 
> ifeq ($(LIBC),GLIBC)
> 	include glibc.mk
> ```

Why not invoke a sub-makefile instead of including?
I don't like including makefiles (with some exceptions, most notably config options) and imo it would be cleaner.
___
*[LibOS/musl.mk, line 17 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcMSiN1LCK5DCmJGxD:-MWcMSiN1LCK5DCmJGxE:b-3h9fj9) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl.mk#L17)):*
> ```Makefile
> CLEAN_FILES +=
> MUSL_TARGETS = \
> 	$(MUSL_BUILD_DIR)/lib/crt1.o \
> ```

Please remove the prefix from here and run `addprefix` after defining this list
___
*[LibOS/musl.mk, line 32 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcMvmu9_XrrMjdBR43:-MWcMvmu9_XrrMjdBR44:bgv9y1n) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl.mk#L32)):*
> ```Makefile
> 	$(MUSL_BUILD_DIR)/lib/Scrt1.o
> LIBC_LINK_TARGETS = \
> 	$(MUSL_TARGETS) \
> ```

Why are all the files above needed as link targets? Dont we need only `libc.so`?
___
*[LibOS/musl-patches/musl.patch, line 97 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcJZWm-ZmMH7fl2WKn:-MWcOQby2RkzJj6f8_T2:b32v57) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl-patches/musl.patch#L97)):*

nitpick: useless hunk.
please drop this hunk.
</blockquote></details>

+1
___
*[LibOS/shim/test/regression/stat_invalid_args.c, line 10 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcPXR7AbyCE3jdr05o:-MWcPXR8Cc5dot2xb90X:b-d4k5qm) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/shim/test/regression/stat_invalid_args.c#L10)):*
> ```C
> #include <unistd.h>
> 
> static long inline_stat(const char* filename, struct stat* statbuf) {
> ```

Any explanation for this change? Maybe add a comment?

dimakuv (changes requested):
> Reviewed 27 of 27 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @mkow, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MW_Src3EThqsJR-Qtku:-MWcVbqm9a8lRXv4qKwC:b-jkywc4) (no related file):*

+1 to running it in a separate pipeline (+ it solves all mentioned issues)
</blockquote></details>

+1 to running in a separate pipeline and using the env var for distinction.

What is the problem with these 4 LibOS tests? Can we easily fix them to support both Glibc and musl?

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcVlQnAst0MfYOpFpf:-MWcVlQnAst0MfYOpFpg:b-8x9vci) (no related file):*
What are the libraries musl generates? I'm in particular interested in `libpthread.so` and `libm.so` (since we use them frequently in our LibOS tests) -- is their functionality embedded in a single `libc.so`? How does gcc then understand that it doesn't need to link against an explicit `libpthread.so`?

---
___
*[Documentation/building.rst, line 17 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcPg8X2pAEy3W1gvc1:-MWcPg8X2pAEy3W1gvc2:b-3d32rd) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/Documentation/building.rst#L17)):*
> ```reStructuredText
>   "shim" in our source code)
> - The Platform Adaptation Layer, or PAL (a shared library named ``libpal.so``)
> - A patched GNU C or musl library (a set of shared libraries ``libc.so``,
> ```

Maybe we should replace `GNU C` with `glibc`? I don't think anyone calls glibc that.
___
*[Documentation/building.rst, line 205 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcQ6_xCiiQ_NF5JBu1:-MWcQ6_y9hAhfgE4hPeZ:bol4vqw) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/Documentation/building.rst#L205)):*
> ```reStructuredText
> 
> To build with a patched musl, add ``LIBC=MUSL`` to the ``make`` invocation.
> You'll need to run :command:`make clean` in ``Runtime/`` directory if you
> ```

Is this (`cd Runtime && make clean`) really enough?
___
*[LibOS/glibc.mk, line 101 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcSpTqBA_EaY19bCYK:-MWcSpTqBA_EaY19bCYL:bosh0yz) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/glibc.mk#L101)):*
> ```Makefile
> 
> # ------------------------------------------------------------------------------------------------
> # GCC build: patch libgomp.so.1 (OpenMP runtime library) to replace raw SYSCALL instruction with
> ```

Shouldn't we split GCC build from Glibc build? Why did you put them together in this file?
___
*[LibOS/musl.mk, line 22 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcUqku5cgJfYIIFk1k:-MWcUqku5cgJfYIIFk1l:b-425g6l) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl.mk#L22)):*
> ```Makefile
> 	$(MUSL_BUILD_DIR)/lib/libcrypt.a \
> 	$(MUSL_BUILD_DIR)/lib/libc.so \
> 	$(MUSL_BUILD_DIR)/lib/libdl.a \
> ```

How do we get shared libs like `libdl.so`? Are they just symlinks to `libc.so` generated together with `libc.so` target? Or they are not needed at all (and that's why you required `sgx.trusted_files = <directory>` as a prerequisite PR)?
___
*[LibOS/musl.mk, line 71 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcUkz1-zr-MBMgMPf7:-MWcUkz2-eh44K3zPIFT:b-j7hagl) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl.mk#L71)):*
> ```Makefile
> 		--enable-optimize \
> 		$(MUSL_CONFIGURE_FLAGS) \
> 		--disable-static \
> ```

How does this work? You disable static libs, but you still specify e.g. `libm.a` as a target.
___
*[LibOS/musl.mk, line 89 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcVMaMEyw9loiTNpyw:-MWcVMaMEyw9loiTNpyx:bh9oxg3) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl.mk#L89)):*
> ```Makefile
> 
> $(MUSL_BUILD_DIR)/libresolv.so.2: $(MUSL_BUILD_DIR)/lib/libc.so
> 	ln -sfr $< $@
> ```

Why do we need these symlinks Why don't we have the same for e.g. `libpthread.so.0`?
___
*[LibOS/musl-patches/musl.patch, line 88 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcRTiq1CXR5hpoczCA:-MWcRTiq1CXR5hpoczCB:b-r319du) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl-patches/musl.patch#L88)):*
> ```Diff
>  static void reloc_all(struct dso *p)
>  {
> +	shim_register_library(p->name, (unsigned long)p->base);
> ```

Why do you add this in `reloc_all()`? A strange place to inform Graphene of an added library (this function seems to do relocations).
___
*[LibOS/musl-patches/musl.patch, line 104 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcRvqt76vcMb_XZAbK:-MWcRvqt76vcMb_XZAbL:byavnze) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl-patches/musl.patch#L104)):*
> ```Diff
> similarity index 77%
> rename from src/process/x86_64/vfork.s
> rename to src/process/x86_64/vfork.S
> ```

What's the deal with renames (from `.s` to `.S`)?
___
*[LibOS/shim/test/regression/stat_invalid_args.c, line 10 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcPXR7AbyCE3jdr05o:-MWcQpw34TkkCXCCXTcQ:b-h3rsf6) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/shim/test/regression/stat_invalid_args.c#L10)):*

Any explanation for this change? Maybe add a comment?
</blockquote></details>

Looks like Musl doesn't have a wrapper for `stat/lstat` system calls? A comment would help here.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @mkow, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MW_Src3EThqsJR-Qtku:-MWdHmKrA206B-L8zzeB:b-7pui0k) (no related file):*

+1 to running in a separate pipeline and using the env var for distinction.

What is the problem with these 4 LibOS tests? Can we easily fix them to support both Glibc and musl?
</blockquote></details>

Environment variable is unwieldy, it would be more complicated to run the tests without the Jenkins...

Would this work:
```
if not pathlib.Path('executable').exists():
    self.skipTest('executable not built')
```
?

---

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 16 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @mkow, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MW_Src3EThqsJR-Qtku:-MWdW5fyEqo1PJm0P8QC:b-qraetv) (no related file):*

Environment variable is unwieldy, it would be more complicated to run the tests without the Jenkins...

Would this work:
```
if not pathlib.Path('executable').exists():
    self.skipTest('executable not built')
```
?


</blockquote></details>

Why not just use decorator in python, just like we do for SGX?

---

boryspoplawski:
Jenkins, retest Jenkins-Debug please (`kill11` timedout, known issue)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: 22 of 28 files reviewed, 16 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MW_Src3EThqsJR-Qtku:-MWzDmCd2UJDnOZBlhCa:b6bdg0m) (no related file):*
> `if not pathlib.Path('executable').exists():`

But which executable? Both musl and glibc generate a set of libraries into Runtime/ and that's all.

> What is the problem with these 4 LibOS tests? Can we easily fix them to support both Glibc and musl?

I don't think they are easy to fix, they don't work because of missing features in musl (at least that's my interpretatio of the errors). I added reasons to the skip messages in `test_libos.py`.

> +1 to running it in a separate pipeline (+ it solves all mentioned issues)

This isn't "a pipeline", but rather 4 of them, because Jenkins can't into matrix builds ({SGX, non-SGX} x {16.04, 18.04}). Even if we choose only one Ubuntu, then it's still 2 additional pipelines which I don't like.
Maybe adding it to existing pipelines could do the work, same as we do with glibc patches for other versions (modulo the fact that we only build them), i.e. build with musl, run LibOS regression, clean Runtime/ and then proceed normally. What do you think?

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcVlQnAst0MfYOpFpf:-MWzX6-c9VCQJfz0LEC0:bicwyeh) (no related file):*
musl puts all the code into libc.so, but for compatibility with standard compilation cmdlines (`-lm`, etc) it generates these .a files. I linked these files into Runtime/ to mirror what we do with glibc's .o files, but that probably doesn't make sense. I just dropped all of them, seems everything still works...

> How does gcc then understand that it doesn't need to link against an explicit libpthread.so?

Good question, I need to debug why the current solution works.
Interesting logs: (with log_level set to "all")
```
~/g/L/s/t/regression> ./pal_loader ./multi_pthread 2>&1 | grep 'open\|stat'
[P4066699:T1:multi_pthread] process 4066699 exited with status 0

~/g/L/s/t/regression> strace ./multi_pthread 2>&1 | grep 'open\|stat'
openat(AT_FDCWD, "/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=56278, ...}) = 0
openat(AT_FDCWD, "/lib/x86_64-linux-gnu/libpthread.so.0", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0755, st_size=157224, ...}) = 0
openat(AT_FDCWD, "/lib/x86_64-linux-gnu/libc.so.6", O_RDONLY|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0755, st_size=2029224, ...}) = 0
fstat(1, {st_mode=S_IFIFO|0600, st_size=0, ...}) = 0

```

---
___
*[Documentation/building.rst, line 17 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcPg8X2pAEy3W1gvc1:-MX3MUDN6jB8da79J3l7:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/Documentation/building.rst#L17)):*

Maybe we should replace `GNU C` with `glibc`? I don't think anyone calls glibc that.
</blockquote></details>

Done.
___
*[Documentation/building.rst, line 205 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcQ6_xCiiQ_NF5JBu1:-MX-PBi14rPdAPwXQLfV:b-yza6zk) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/Documentation/building.rst#L205)):*

Is this (`cd Runtime && make clean`) really enough?
</blockquote></details>

Yup, at least it worked in my tests. You also need to regenerate .manifest.sgx files, but that's expected after rebuilding Graphene.
___
*[LibOS/glibc.mk, line 101 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcSpTqBA_EaY19bCYK:-MX-_wDB4ZvXmlx-3vwr:b-b1byyo) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/glibc.mk#L101)):*

Shouldn't we split GCC build from Glibc build? Why did you put them together in this file?
</blockquote></details>

Just because it was like this previously, plus I think OpenMP requires glibc? So, it probably wouldn't make sense to build it without glibc.
Current version makes `gcc` target available only in glibc mode, which I think is what we want?
___
*[LibOS/glibc.mk, line 124 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcMEFM3Huw41enRgD3:-MX-rVbc39yqrBzTIh1U:bdexn53) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/glibc.mk#L124)):*

Please split into two lines just line in glibc
</blockquote></details>

This _is_ glibc and it was already like this. But I can fix it while we're at it.
___
*[LibOS/Makefile, line 24 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcFW9n6iy4c9w4nlmA:-MX1sawtE7YzyIssGvM3:b-6dn6it) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/Makefile#L24)):*

Why not invoke a sub-makefile instead of including?
I don't like including makefiles (with some exceptions, most notably config options) and imo it would be cleaner.
</blockquote></details>

I don't think it's cleaner, why would it be? Problems with sub-makes:
a) They spam with "Entering directory [...]"/"Leaving directory [...]" in the logs.
b) Spawning another instance of Make sounds like an overkill.
c) We'd need to proxy the targets we use to the submake.
___
*[LibOS/musl.mk, line 17 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcMSiN1LCK5DCmJGxD:-MX-X2uUCVA91BewjQcT:boi791l) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl.mk#L17)):*

Please remove the prefix from here and run `addprefix` after defining this list
</blockquote></details>

I thought about this, but the logic gets more complicated then (especially with the ugly and unreadable Make syntax). I find the plain, obvious list more compelling.
Anyways, the list got shrunk to one item now, so I guess we don't need any loops.
___
*[LibOS/musl.mk, line 22 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcUqku5cgJfYIIFk1k:-MX-XrcA7zJhda5BGc9V:b8ndwer) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl.mk#L22)):*

How do we get shared libs like `libdl.so`? Are they just symlinks to `libc.so` generated together with `libc.so` target? Or they are not needed at all (and that's why you required `sgx.trusted_files = <directory>` as a prerequisite PR)?
</blockquote></details>

Let's keep this discussion at the other thread (the one not pinned to any file).
___
*[LibOS/musl.mk, line 32 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcMvmu9_XrrMjdBR43:-MX-Y85j5EdXY5d2jmft:b-w8px8k) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl.mk#L32)):*

Why are all the files above needed as link targets? Dont we need only `libc.so`?
</blockquote></details>

See the discussion with Dima. I removed them now and we'll see if it still works.
___
*[LibOS/musl.mk, line 71 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcUkz1-zr-MBMgMPf7:-MX-YTWjEwcKcEd28wyv:bnjp71e) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl.mk#L71)):*

How does this work? You disable static libs, but you still specify e.g. `libm.a` as a target.
</blockquote></details>

See the main discussion. It's just a fake lib to keep `-lm` work with musl.
___
*[LibOS/musl.mk, line 89 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcVMaMEyw9loiTNpyw:-MX-YgzL0XPVCu5lBTEh:bdl10d2) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl.mk#L89)):*

Why do we need these symlinks Why don't we have the same for e.g. `libpthread.so.0`?
</blockquote></details>

ditto, let's keep this issue in that other thread.
___
*[LibOS/musl-patches/musl.patch, line 88 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcRTiq1CXR5hpoczCA:-MX-ZJpl6Gv0HhZ6kPKw:b-k4o3ni) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl-patches/musl.patch#L88)):*

Why do you add this in `reloc_all()`? A strange place to inform Graphene of an added library (this function seems to do relocations).
</blockquote></details>

Because this is something which is called exactly once per each loaded library. Any ideas for a better place to do this?
___
*[LibOS/musl-patches/musl.patch, line 97 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcJZWm-ZmMH7fl2WKn:-MX1sgqL5eiHDZOn298H:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl-patches/musl.patch#L97)):*

+1
</blockquote></details>

Fixed.
___
*[LibOS/musl-patches/musl.patch, line 104 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcRvqt76vcMb_XZAbK:-MX-ZT4cE-5LMz9-RVQD:bq3vcl0) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl-patches/musl.patch#L104)):*

What's the deal with renames (from `.s` to `.S`)?
</blockquote></details>

Good question! It turns out that `.s` files are normal assembly sources, but `.S` files are "assembly source, but first run a C preprocessor on them" (and our headers are .h, so we need the preprocessing phase).
___
*[LibOS/shim/test/regression/stat_invalid_args.c, line 10 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcPXR7AbyCE3jdr05o:-MX-_FIDBSO7LsEqTLgb:b-2dysbc) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/shim/test/regression/stat_invalid_args.c#L10)):*

Looks like Musl doesn't have a wrapper for `stat/lstat` system calls? A comment would help here.
</blockquote></details>

This test passes invalid pointers to C functions, it was completely wrong in the first place and worked only by an accident on glibc, and on musl it just crashes.
In general I think raw syscalls should be the default state in such tests (which target specific syscalls), it's using libc wrappers which requires explanation/justification.

woju (commented):
> Reviewed 1 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: 23 of 28 files reviewed, 16 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MW_Src3EThqsJR-Qtku:-MX4vcejE8dC1t9JcrTn:b-85gvkl) (no related file):*

> `if not pathlib.Path('executable').exists():`

But which executable? Both musl and glibc generate a set of libraries into Runtime/ and that's all.

> What is the problem with these 4 LibOS tests? Can we easily fix them to support both Glibc and musl?

I don't think they are easy to fix, they don't work because of missing features in musl (at least that's my interpretatio of the errors). I added reasons to the skip messages in `test_libos.py`.

> +1 to running it in a separate pipeline (+ it solves all mentioned issues)

This isn't "a pipeline", but rather 4 of them, because Jenkins can't into matrix builds ({SGX, non-SGX} x {16.04, 18.04}). Even if we choose only one Ubuntu, then it's still 2 additional pipelines which I don't like.
Maybe adding it to existing pipelines could do the work, same as we do with glibc patches for other versions (modulo the fact that we only build them), i.e. build with musl, run LibOS regression, clean Runtime/ and then proceed normally. What do you think?
</blockquote></details>

The executable which is compiled as part of tests (`bootstrap_cpp` and others). If it failed to compile, the test should be skipped.

This `USES_MUSL` thing is wrong, because it ties testsuite to our poor buildsystem.

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: 23 of 28 files reviewed, 16 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MW_Src3EThqsJR-Qtku:-MX50kmh-8JikLunWMIg:b-kqux9u) (no related file):*

The executable which is compiled as part of tests (`bootstrap_cpp` and others). If it failed to compile, the test should be skipped.

This `USES_MUSL` thing is wrong, because it ties testsuite to our poor buildsystem.
</blockquote></details>

But we aren't compiling anything against musl headers here. The current version of this PR compiles them normally (using system headers) and supplies musl only in runtime. So, this tests binary compatibility (which may be good or bad, open to discussion).

---

dimakuv (changes requested):
> Reviewed 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MW_Src3EThqsJR-Qtku:-MX5v_0u3L4M29hk9fFh:bdtq6wq) (no related file):*

But we aren't compiling anything against musl headers here. The current version of this PR compiles them normally (using system headers) and supplies musl only in runtime. So, this tests binary compatibility (which may be good or bad, open to discussion).
</blockquote></details>

The logic of "build application against system-default Glibc, and then run it with our-provided Musl" sounds wrong. Maybe we should just have an Alpine-based Jenkins pipeline (a single one for now)?

Well, maybe we should also check what other cross-libc projects are doing in their testing.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcVlQnAst0MfYOpFpf:-MX5und91wG9XRQ3JzL7:briixew) (no related file):*
> it generates these .a files

Why would we care about static libs (`.a`)? My question was purely about shared libs (`.so`).

> Interesting logs: (with log_level set to "all")

The logs are sad... So our applications actually don't use Musl libs at all currently, they just skip over `graphene/Runtime/` and check the rest of system paths, and find Glibc.

On the other hand, you said that 4 LibOS tests fail, so clearly at least some of them, sometimes, pick up Musl's `libc.so`. So how does all that work?

---
___
*[LibOS/glibc.mk, line 23 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MX5wGjV8PI34NKvuxNb:-MX5wGjV8PI34NKvuxNc:b-ny823d) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/glibc.mk#L23)):*
> ```Makefile
> 	csu/crt1.o \
> 	csu/crti.o \
> 	csu/crtn.o \
> ```

These are not needed completely? Well, it makes sense. I wonder why we had them in the first place.
___
*[LibOS/glibc.mk, line 101 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcSpTqBA_EaY19bCYK:-MX5wS2p2DFgiVriJlJ9:b-qezkjt) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/glibc.mk#L101)):*

Just because it was like this previously, plus I think OpenMP requires glibc? So, it probably wouldn't make sense to build it without glibc.
Current version makes `gcc` target available only in glibc mode, which I think is what we want?
</blockquote></details>

Ok, fair enough.
___
*[LibOS/musl-patches/musl.patch, line 88 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcRTiq1CXR5hpoczCA:-MX5wtkz9r5gqeCt8OBv:bvkri52) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl-patches/musl.patch#L88)):*

Because this is something which is called exactly once per each loaded library. Any ideas for a better place to do this?
</blockquote></details>

OK. No better ideas.
___
*[LibOS/musl-patches/musl.patch, line 104 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcRvqt76vcMb_XZAbK:-MX5x5v7Ch1rWqiCvTrE:bbhhdvl) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl-patches/musl.patch#L104)):*

Good question! It turns out that `.s` files are normal assembly sources, but `.S` files are "assembly source, but first run a C preprocessor on them" (and our headers are .h, so we need the preprocessing phase).
</blockquote></details>

OK. That's a fun side-effect.
___
*[LibOS/shim/test/regression/stat_invalid_args.c, line 10 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcPXR7AbyCE3jdr05o:-MX5xIzz8L2J4Eb9BKW3:b-3q8mkl) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/shim/test/regression/stat_invalid_args.c#L10)):*

This test passes invalid pointers to C functions, it was completely wrong in the first place and worked only by an accident on glibc, and on musl it just crashes.
In general I think raw syscalls should be the default state in such tests (which target specific syscalls), it's using libc wrappers which requires explanation/justification.
</blockquote></details>

OK, good point. My bad, I wrote this test. Indeed, for such tests we must always use raw syscalls.

Still, a comment on this would be helpful (`We can't use libc wrappers for stat/lstat because we test the raw interfaces of these syscalls.`).
___
*[LibOS/shim/test/regression/test_libos.py, line 106 at r2](https://reviewable.io/reviews/oscarlab/graphene/2269#-MX5zKQV350e8glPlEEU:-MX5zKQV350e8glPlEEV:b-r473eh) ([raw file](https://github.com/oscarlab/graphene/blob/9a35c8f503750e4f3b6acfc2771609d30a01d88c/LibOS/shim/test/regression/test_libos.py#L106)):*
> ```Python
>         self.assertIn('argv[0] = bootstrap_pie', stdout)
> 
>     @unittest.skipIf(USES_MUSL, 'musl doesn\'t contain libstdc++.so')
> ```

I guess this is that issue that Musl is bundled with `libc++.so` instead of `libstdc++.so`...
___
*[LibOS/shim/test/regression/test_libos.py, line 487 at r2](https://reviewable.io/reviews/oscarlab/graphene/2269#-MX5yS_mEKdkwLs9qkxt:-MX5yS_n281RfiWXyfex:beugt43) ([raw file](https://github.com/oscarlab/graphene/blob/9a35c8f503750e4f3b6acfc2771609d30a01d88c/LibOS/shim/test/regression/test_libos.py#L487)):*
> ```Python
>         self.assertIn('eventfd_using_fork completed successfully', stdout)
> 
>     @unittest.skipIf(USES_MUSL, 'musl doesn\'t implement scheduling APIs')
> ```

That's surprisingly true: https://git.musl-libc.org/cgit/musl/tree/src/sched/sched_setscheduler.c?id=43b640c2ff30bef9c943a8222008f4bd6f6cbaf6.
___
*[LibOS/shim/test/regression/test_libos.py, line 621 at r2](https://reviewable.io/reviews/oscarlab/graphene/2269#-MX5yXxM8g3i7kY1gHtW:-MX5yXxM8g3i7kY1gHtX:be9kkz6) ([raw file](https://github.com/oscarlab/graphene/blob/9a35c8f503750e4f3b6acfc2771609d30a01d88c/LibOS/shim/test/regression/test_libos.py#L621)):*
> ```Python
> 
>     @unittest.skipIf(USES_MUSL, 'musl is missing gnu_dev_minor and gnu_dev_major symbols if the '
>                                 'binaries weren\'t built against musl headers')
> ```

Yeah, `major` and `minor` are defined as aliases of `gnu_dev_major` and `gnu_dev_minor` functions in Glibc, but as simple macros in Musl: http://git.musl-libc.org/cgit/musl/tree/include/sys/sysmacros.h?id=dbbb3734d8c0176feabd6c46e2e85bbc3b8a60af.

I think building apps against Glibc headers and then using Musl libs at runtime is doomed to fail in such mysterious ways. One suggestion above is to use Alpine docker containers for such tests, another suggestion is to `-I` Musl headers when building these LibOS tests (and disable system headers)?

boryspoplawski (changes requested):
> Reviewed 5 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcVlQnAst0MfYOpFpf:-MX6f1ezBCTxq7Csen8R:b7oe4r9) (no related file):*
> The logs are sad... So our applications actually don't use Musl libs at all currently, they just skip over graphene/Runtime/ and check the rest of system paths, and find Glibc.

Because they are compiled against glibc. Why would they use something else?
Remember that musl libc is one monolitic library? It so happens that it's also the dynamic loader (the very same binary). Dynamic loader is opened+loaded by kernel (here graphene) so you won't find it in the logs. There is no need to load it second time.
The question is why Graphene picks musl over glibc? Then answer is simple: we do not provide glibc inside graphene fs. Now the last question: but the path to loader is hardcoded inside elf, so if we do not provide it (it might be different for glibc vs musl) how does it all work? The answer is scary, but you should know: https://github.com/oscarlab/graphene/blob/master/LibOS/shim/src/shim_rtld.c#L725 (my eyes are bleeding).



> How does gcc then understand that it doesn't need to link against an explicit libpthread.so?
Good question, I need to debug why the current solution works.

As far as I understand it does link against `libpthread.so` (after all it's a glibc build), but the symbols are resolved lazily, so since musl-libc provides all of them, there is never a need to load `libpthread.so`.

---
___
*[LibOS/glibc.mk, line 23 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MX5wGjV8PI34NKvuxNb:-MX6nneO52y8mkjJM2g8:b4lpqsk) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/glibc.mk#L23)):*

These are not needed completely? Well, it makes sense. I wonder why we had them in the first place.
</blockquote></details>

They would be useful if you wanted to build your application against Graphene libc. Which is rarely the case, I guess.
___
*[LibOS/glibc.mk, line 124 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcMEFM3Huw41enRgD3:-MX6nyr-2VRUjbSvG9Qq:bfxt19w) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/glibc.mk#L124)):*

This _is_ glibc and it was already like this. But I can fix it while we're at it.
</blockquote></details>

This is GCC not glibc.
Forgot to fix?
___
*[LibOS/Makefile, line 24 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcFW9n6iy4c9w4nlmA:-MX6oF455HfxsHvhH7VG:b-r7ndo9) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/Makefile#L24)):*
>  why would it be? 

Because make is declarative, so you just include some targets from another file and run them in **this** file. You cannot read this make, without checking also the included onces, whereas using submake you don't have to peek into them. Something like descriptive function name: you don't have to read it, you can just leave it for later to validate that it really does what it says and just assume it for now.

> a) They spam with "Entering directory [...]"/"Leaving directory [...]" in the logs.

Only if you opt to. All make commands are printed, not only entering directory. You can silence all of them using `@` before command or `-s` parameter. We do it sometimes, we can do it for submakes as well.

> b) Spawning another instance of Make sounds like an overkill.

Lol, why? It's not an new enclave heavy process, just a plain linux process. Building libc is rather spacial case enough to allow for another process for it...

> c) We'd need to proxy the targets we use to the submake.

You can do that with oneliner, but yeah we would. And I think that would be cleaner than getting targets form another makefile.
___
*[LibOS/musl.mk, line 17 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcMSiN1LCK5DCmJGxD:-MX6rI1qCn6W93hch9bI:b-q221wc) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/musl.mk#L17)):*

I thought about this, but the logic gets more complicated then (especially with the ugly and unreadable Make syntax). I find the plain, obvious list more compelling.
Anyways, the list got shrunk to one item now, so I guess we don't need any loops.
</blockquote></details>

Ok, with one item it looks better this way.

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 7 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MX7vMEFAdheAe36QlzO:-MX7vMEG78uOFaki0Qaa:b-psr3la) (no related file):*
Can musl-gcc wrapper be used to build test programs?
https://www.musl-libc.org/how.html

---

yamahata (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 7 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MX7vMEFAdheAe36QlzO:-MX7yudSApBrfgaGjj5c:b-pllwf6) (no related file):*

Can musl-gcc wrapper be used to build test programs?
https://www.musl-libc.org/how.html
</blockquote></details>

ubuntu has packages, musl, musl-dev, musl-tools. They can be installed and musl-gcc is ready to use.

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 7 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcVlQnAst0MfYOpFpf:-MXBXbH7ESAj5IxNQMrt:b-fz4zjg) (no related file):*
> Because they are compiled against glibc. Why would they use something else?

But Musl is an (almost) drop-in replacement for glibc. Both glibc and musl have the same name of the (main file) libc: `libc.so`. So even if the app was compiled against Glibc, if there is a Musl code inside `/lib/libc.so`, then the app will use it.

> It so happens that it's also the dynamic loader (the very same binary). Dynamic loader is opened+loaded by kernel (here graphene) so you won't find it in the logs.

Thanks, I completely forgot about it! This makes sense.

> Then answer is simple: we do not provide glibc inside graphene fs.

I guess by "inside graphene fs" you mean "under `graphene/Runtime/`". This makes sense, yes, that's what I meant in my first paragraph here.

> The answer is scary...

Why are these default libs scary? (For context, I'll copy-paste the current code: `const char* default_paths[] = {"/lib", "/lib64", NULL};`)

> As far as I understand it does link against libpthread.so (after all it's a glibc build), but the symbols are resolved lazily, so since musl-libc provides all of them, there is never a need to load libpthread.so.

Thanks, makes sense.

---
___
*[LibOS/shim/test/regression/test_libos.py, line 621 at r2](https://reviewable.io/reviews/oscarlab/graphene/2269#-MX5yXxM8g3i7kY1gHtW:-MXBXTzc2GIjb540HpX3:bco0z2o) ([raw file](https://github.com/oscarlab/graphene/blob/9a35c8f503750e4f3b6acfc2771609d30a01d88c/LibOS/shim/test/regression/test_libos.py#L621)):*
> ubuntu has packages, musl, musl-dev, musl-tools. They can be installed and musl-gcc is ready to use.

Here is yet another suggestion from Isaku.

boryspoplawski (commented):
> Reviewed 1 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 7 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcVlQnAst0MfYOpFpf:-MXDHJM43wfsqumo4L3t:bo3ljzd) (no related file):*
> But Musl is an (almost) drop-in replacement for glibc. Both glibc and musl have the same name of the (main file) libc: libc.so. So even if the app was compiled against Glibc, if there is a Musl code inside /lib/libc.so, then the app will use it.

Afaik they have different paths to the ld.so (dynamic loader) and it's the ld.so that resolves other libraries (including libc.so). Generally speaking if you compile app against glibc and do not probide glibc's ld.so it wont work. In Graphene the situation is different (a bit) - we provide musl ld.so at the same path as glibc's ld.so would lay.
It's not that simple, musl is an almost drop-in replacement for glibc at code level (compile time) not in runtime.
> Why are these default libs scary?

There is no such thing as default loader or loader path. The full path is hardcoded in each elf.
If I understood the code correctly, it strips the path to just a file name and then searched all `LD_LIBRARY_PATH` paths (or defaults it it's not provided) for this file name. This is not how it should work.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MX7vMEFAdheAe36QlzO:-MXDJJR75SDGeMWRtnvX:b-mkyhue) (no related file):*

ubuntu has packages, musl, musl-dev, musl-tools. They can be installed and musl-gcc is ready to use.
</blockquote></details>

I think we could generate `musl-gcc` when building musl, it's just a set of scripts for gcc (e.g. linker script).

---

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 7 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @yamahata)

___
*[LibOS/Makefile, line 24 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcFW9n6iy4c9w4nlmA:-MXDPR1E7eVQEthl95Bq:bal7yfg) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/Makefile#L24)):*

>  why would it be? 

Because make is declarative, so you just include some targets from another file and run them in **this** file. You cannot read this make, without checking also the included onces, whereas using submake you don't have to peek into them. Something like descriptive function name: you don't have to read it, you can just leave it for later to validate that it really does what it says and just assume it for now.

> a) They spam with "Entering directory [...]"/"Leaving directory [...]" in the logs.

Only if you opt to. All make commands are printed, not only entering directory. You can silence all of them using `@` before command or `-s` parameter. We do it sometimes, we can do it for submakes as well.

> b) Spawning another instance of Make sounds like an overkill.

Lol, why? It's not an new enclave heavy process, just a plain linux process. Building libc is rather spacial case enough to allow for another process for it...

> c) We'd need to proxy the targets we use to the submake.

You can do that with oneliner, but yeah we would. And I think that would be cleaner than getting targets form another makefile.
</blockquote></details>

I agree that a separate Makefile would be cleaner. It's much harder to answer questions like "what does this target do" if you cannot just look into one file.

Also, it seems to me it's only an artifact of history that these projects (patched glibc and gcc) reside in LibOS directory instead of their own, especially since they actually keep their own subdirectories for source, patches and build. Maybe they should have their own directories? Something like:

```
LibOS/
  Makefile
  shim/
  glibc/
    Makefile
    glibc-2.27.patch
    glibc-2.31.patch
    src/
    build/
  gcc/
    Makefile
    ...
  musl/
    Makefile
    ...
```

This way it's also easier to perform some operations on just the glibc part: I know that I can run `make glibc` to build it, but how do I clean it? `make glibc-clean`? And so on.

(Not very related, but also: it would be very nice to have something like `make clean KEEP_GLIBC=1`, rebuilding glibc is the longest part of the build).

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 7 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow, @pwmarcz, and @yamahata)

___
*[LibOS/Makefile, line 24 at r1](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcFW9n6iy4c9w4nlmA:-MXDWDHjAbuppqJJ1hoY:b-qusepa) ([raw file](https://github.com/oscarlab/graphene/blob/e95903b7ffcf46bf81571de817315d6c27f07161/LibOS/Makefile#L24)):*
+1 to all of the above
> (Not very related, but also: it would be very nice to have something like make clean KEEP_GLIBC=1, rebuilding glibc is the longest part of the build).

Yes, I've been complaining about it a long time, but got used to rebuilding every time by now :)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2269)* status: all files reviewed, 7 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow, @pwmarcz, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2269#-MWcVlQnAst0MfYOpFpf:-MY42w_FCfSO01OqKsaM:bmo1rot) (no related file):*

> But Musl is an (almost) drop-in replacement for glibc. Both glibc and musl have the same name of the (main file) libc: libc.so. So even if the app was compiled against Glibc, if there is a Musl code inside /lib/libc.so, then the app will use it.

Afaik they have different paths to the ld.so (dynamic loader) and it's the ld.so that resolves other libraries (including libc.so). Generally speaking if you compile app against glibc and do not probide glibc's ld.so it wont work. In Graphene the situation is different (a bit) - we provide musl ld.so at the same path as glibc's ld.so would lay.
It's not that simple, musl is an almost drop-in replacement for glibc at code level (compile time) not in runtime.
> Why are these default libs scary?

There is no such thing as default loader or loader path. The full path is hardcoded in each elf.
If I understood the code correctly, it strips the path to just a file name and then searched all `LD_LIBRARY_PATH` paths (or defaults it it's not provided) for this file name. This is not how it should work.
</blockquote></details>

Thanks for explanations! Now I understand.

---

mkow:
This PR will be re-done in the new repo, for now we're tracking the status in https://github.com/gramineproject/gramine/issues/49.

PR #2288, opened 2021-03-31
Manifest preprocessor

woju:
## Description of the changes 

This is a tool for preprocessing manifest templates using jinja. The templates were previously munged by `sed`. The problem is, `sed` is not very powerful and the config is distributed in many different makefiles, so changing something for all templates requires manual editing of them all.

One notable thing was achieved by having a single `{{ libos }}` substitution which supplies the path to `libsysdb.so`. The substitution does right thing whether Graphene is installed or not. This will later come handy when we'll switch CI to run the tests (incl. `Examples`) using installed graphene (not from repo), which in turn is a prerequisite for a greater effort of having graphene packaged eventually.

RFC: there are two common substitutions which I have left untouched:
- `{{ arch_libdir }}` is problematic, because it is (will be) straightforward in meson, but without installing we'd have to resort to autodetection; I'm inclined to leave this alone for now and fix it at a later stage (when moving to installed graphene only)
- `{{ graphene_log_level }}` is not something standard enough for me, though I'd be open to suggestions that we should "standardise" it as a convention. Maybe we could even think about some general macro like `{{ preamble }}` which would expand to two entry points and loglevel, because this is very common. In any case I think such conventions are out of scope of this PR. 

Also as part of this PR some problematic Python-related variables were corrected like `PYTHONHOME` and `PYTHONPATH` which affect python execution, but they were abused for unrelated things.

There are some currently unused features of the manifest environment like `ldd`, which is a leftover from when the tool was originally written, which is before `sgx.trusted_paths` could glob the whole subtree.

## How to test this PR? 

All `Examples/` should work as they were.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)


dimakuv (changes requested):
> Reviewed 56 of 56 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCAuzgA_tutxBdxv9I:-MXCAuzh3PnGRpK7Dtu9:b-3iuki0) (no related file):*
> Maybe we could even think about some general macro like {{ preamble }} which would expand to two entry points and loglevel, because this is very common. 

I dislike the idea of a `{{ preamble }}`. This will be some "Graphene magic", unlike the current straight-forward sed-like replacement. Since manifest files will be written by users of different levels of proficiency, we will doom ourselves to GitHub issues complaining about errors due to misusing this magic keyword.

> {{ graphene_log_level }} is not something standard enough for me, though I'd be open to suggestions that we should "standardise" it as a convention

Same for this. I would prefer to have straight-forward substitutions.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCCCScCKFrKQXz5Kam:-MXCCCScCKFrKQXz5Kan:b-1h4trw) (no related file):*
Currently we have two predefined keywords: `libos` and `get_runtimedir`. Users may wonder where these magic values come from (the other substitutions come from Makefiles, so the others are obvious). I suggest to give them more descriptive names like `predefined_libos` and `predefined_runtimedir`. Or maybe the `autodetected_` prefix? Or something similar...

Same for `python` keywords.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCDb9v37kaXjKqy24D:-MXCDb9wBJgaPcWUgUtq:bf4zptc) (no related file):*
You need to add some explanations that the manifest template can use Jinja-style substitutions here: https://graphene.readthedocs.io/en/latest/manifest-syntax.html#manifest-syntax. Also list all predefined keywords (`libos`, `get_runtimedir`, etc).

---
___
*[Examples/apache/httpd.manifest.template, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXC4VPK1cZZploYD6p1:-MXC4VPK1cZZploYD6p2:b-9ld0zy) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/Examples/apache/httpd.manifest.template#L26)):*
> ```
> fs.mount.lib.type = "chroot"
> fs.mount.lib.path = "/lib"
> fs.mount.lib.uri = "file:{{ get_runtimedir() }}"
> ```

Can we have it not as a function, but a string? Or what was the rationale to make it a function?

I am afraid that users will make a mistake of `{{get_runtimedir}}` and will be confused by the returned address of the function (or whatever Python/Jinja returns in this case).
___
*[Examples/python-scipy/Makefile, line 13 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCCqQ91MDMlq39cQMZ:-MXCCqQ91MDMlq39cQM_:b-qqiayb) ([raw file](https://github.com/oscarlab/graphene/blob/46da82fbd3abfcbf2a9edcb554773aabb1b7e51f/Examples/python-scipy/Makefile#L13)):*
> ```Makefile
> 
> # Python constants are declared in Makefile.python
> include ../../Scripts/Makefile.python
> ```

Thanks for removing this!
___
*[python/graphenelibos/manifest.py, line 51 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCDO2b9anFAWEJwSv2:-MXCDO2c6n6XmCn33ZoA:b-fi711l) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/python/graphenelibos/manifest.py#L51)):*
> ```Python
>     offsets = Empty() # object() is not sufficient, because it lacks __dict__
> 
> def ldd(*args):
> ```

Can we add a TODO that this is deprecated and should be removed?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: 55 of 65 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCAuzgA_tutxBdxv9I:-MXCKpziAwPpmtxlOlQW:b-h4byme) (no related file):*

> Maybe we could even think about some general macro like {{ preamble }} which would expand to two entry points and loglevel, because this is very common. 

I dislike the idea of a `{{ preamble }}`. This will be some "Graphene magic", unlike the current straight-forward sed-like replacement. Since manifest files will be written by users of different levels of proficiency, we will doom ourselves to GitHub issues complaining about errors due to misusing this magic keyword.

> {{ graphene_log_level }} is not something standard enough for me, though I'd be open to suggestions that we should "standardise" it as a convention

Same for this. I would prefer to have straight-forward substitutions.
</blockquote></details>

I understand, epecially the part about magic and the ticket load.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCCCScCKFrKQXz5Kam:-MXCL2tH7T--RvVp4Qnb:bf73tcm) (no related file):*

Currently we have two predefined keywords: `libos` and `get_runtimedir`. Users may wonder where these magic values come from (the other substitutions come from Makefiles, so the others are obvious). I suggest to give them more descriptive names like `predefined_libos` and `predefined_runtimedir`. Or maybe the `autodetected_` prefix? Or something similar...

Same for `python` keywords.
</blockquote></details>

Would it be OK if those were namespaced: `{{ graphene.libos }}`, `{{ graphene.get_runtimedir() }}` and `{{ python.stdlib }}`?

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCDb9v37kaXjKqy24D:-MXCNIJAAuNFfk_gHRJU:bk8weoj) (no related file):*

You need to add some explanations that the manifest template can use Jinja-style substitutions here: https://graphene.readthedocs.io/en/latest/manifest-syntax.html#manifest-syntax. Also list all predefined keywords (`libos`, `get_runtimedir`, etc).
</blockquote></details>

Will do. And a manpage.

---
___
*[Examples/apache/httpd.manifest.template, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXC4VPK1cZZploYD6p1:-MXCLRI3EIhD7wGzGKvb:bzgh86v) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/Examples/apache/httpd.manifest.template#L26)):*

Can we have it not as a function, but a string? Or what was the rationale to make it a function?

I am afraid that users will make a mistake of `{{get_runtimedir}}` and will be confused by the returned address of the function (or whatever Python/Jinja returns in this case).
</blockquote></details>

The rationale is that this is dependent on the chosen libc, so this leaves an option for `{{ get_runtimedir('musl') }}` and also maybe switching versions: `{{ get_runtimedir('glibc-12.345') }}`. Optimally we should be able to install multiple patched libcs together, and it should be enclave's decision which one to use (for example by `fs.mount`).

If you'd like, I could invent a magic internal object that would do OK whether it got `()` or not, like this:

```jinja
{{ runtimedir }}
{{ runtimedir() }}
{{ runtimedir('musl') }}
```
___
*[Examples/python-scipy/Makefile, line 13 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCCqQ91MDMlq39cQMZ:-MXCMgsA2Ovu0nNWZftN:b-5yn8vs) ([raw file](https://github.com/oscarlab/graphene/blob/46da82fbd3abfcbf2a9edcb554773aabb1b7e51f/Examples/python-scipy/Makefile#L13)):*

Thanks for removing this!
</blockquote></details>

You're welcome. Anything else you'd like me to optimise out?
___
*[python/graphenelibos/manifest.py, line 51 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCDO2b9anFAWEJwSv2:-MXCMquI0UB57Lebuhtg:b8izchb) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/python/graphenelibos/manifest.py#L51)):*

Can we add a TODO that this is deprecated and should be removed?
</blockquote></details>

Is it really? I think this should be provided as an alternative for globbing the `/lib` directory.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: 52 of 69 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCDb9v37kaXjKqy24D:-MXCgMhACqCwTKYOJxtu:b-ei0gx7) (no related file):*

Will do. And a manpage.
</blockquote></details>

I wrote the manpage and added a reference in `manifest-syntax.rst`.

---

mkow (changes requested):
> Reviewed 51 of 56 files at r1, 18 of 18 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCAuzgA_tutxBdxv9I:-MXIhv5z-P57YlUQbl5f:b7hjiu5) (no related file):*

I understand, epecially the part about magic and the ticket load.
</blockquote></details>

Also -1 from me to the `{{ preamble }}` idea.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCCCScCKFrKQXz5Kam:-MXIi70t45S80NItX4Df:b-f33dgq) (no related file):*

Would it be OK if those were namespaced: `{{ graphene.libos }}`, `{{ graphene.get_runtimedir() }}` and `{{ python.stdlib }}`?
</blockquote></details>

I like the namespace idea.

---
___
*[Documentation/manpages/graphene-manifest.rst, line 24 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKBtoRB04rFK8uyhl6:-MXKBtoRB04rFK8uyhl7:b-12bg4k) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Documentation/manpages/graphene-manifest.rst#L24)):*
> ```
> available in template.
> ```

Not "the template"? (the one you invoked the command on)
___
*[Documentation/manpages/graphene-manifest.rst, line 26 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKCDRdE_qBJS_6kh_U:-MXKCDReAAu_ncTqysYX:bs84sgy) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Documentation/manpages/graphene-manifest.rst#L26)):*
> ```
> available in template
> ```

I'd rather use plural here ("templates")
___
*[Documentation/manpages/graphene-manifest.rst, line 91 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKDHsB4gR301TzvkmN:-MXKDHsB4gR301TzvkmO:b34qrxi) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Documentation/manpages/graphene-manifest.rst#L91)):*
> ```reStructuredText
> 
>    loader.preload = "file:{{ libos }}"
>    libos.entrypoint ="file:{{ entrypoint }}"
> ```

missing space after `=`
___
*[Documentation/manpages/graphene-manifest.rst, line 109 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKENKD-FcmbmXkdkZ1:-MXKENKE0DwJd6uv7jHU:bfepbqs) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Documentation/manpages/graphene-manifest.rst#L109)):*
> ```reStructuredText
>    %.manifest: manifest.template
>       graphene-manifest \
>          -Dentrypoint=$(ENTRYPOINT) \
> ```

I guess this is more like a pseudocode, right? I mean, you're not using the wildcard from the target pattern matching, which normally should somehow influence `$(ENTRYPOINT)`.
___
*[Examples/apache/Makefile, line 86 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXMoq8i1hLgppUCMH3l:-MXMoq8i1hLgppUCMH3m:b5ayryi) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Examples/apache/Makefile#L86)):*
> ```Makefile
> 		-Dinstall_dir=$(INSTALL_DIR) \
> 		-Dinstall_dir_abspath=$(abspath $(INSTALL_DIR)) \
> 		-Darch_libdir=$(ARCH_LIBDIR) \
> ```

Don't we need doublequotes here? (same for the whole PR)
___
*[Examples/blender/blender.manifest.template, line 14 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXMptuC0Dl6d-xjc3WB:-MXMptuC0Dl6d-xjc3WC:b-6fac61) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Examples/blender/blender.manifest.template#L14)):*
> ```
> 
> 
> libos.entrypoint = "file:{{ blender_dir}}/blender"
> ```

missing space
___
*[Examples/lighttpd/lighttpd.manifest.template, line 21 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKF-y-Eb0kWcR_v-nF:-MXKF-y-Eb0kWcR_v-nG:b12xnvh) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Examples/lighttpd/lighttpd.manifest.template#L21)):*
> ```
> install_dir
> ```

Now this name became a bit ambiguous: It's not clear that this is lighttpd install path, not Graphene's install path. Please grep the whole PR for similar occurrences.
___
*[Examples/lighttpd/Makefile, line 60 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKFLOg5U-LUkKxHDn3:-MXKFLOh5nixfBhtLUf2:b2jqjtr) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Examples/lighttpd/Makefile#L60)):*
> ```Makefile
> 		-Dinstall_dir=$(INSTALL_DIR) \
> 		-Dinstall_dir_abspath=$(abspath $(INSTALL_DIR)) \
> 	    $< >$@
> ```

you missed this place when converting alignment to indentation (spaces -> tabs)
___
*[python/graphenelibos/manifest.py, line 51 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCDO2b9anFAWEJwSv2:-MXN-LO28vR6b1eL3AfR:b-8jqbte) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/python/graphenelibos/manifest.py#L51)):*

Is it really? I think this should be provided as an alternative for globbing the `/lib` directory.
</blockquote></details>

I'm not sure, there's not really too much uses for this selective adding. Currently the only issue is I think the loading time which grows, because we use keys instead of a TOML array (and this will be fixed soon). Otherwise `ldd` is error-prone (easy to miss dynamic deps) and quite hacky.

Actually, currently this function is unused, right?
___
*[python/graphenelibos/manifest.py, line 36 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXN03AQ5goPIO2FvYmb:-MXN03AQ5goPIO2FvYmc:bn1r5iu) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/python/graphenelibos/manifest.py#L36)):*
> ```Python
>     '''
>     # Be careful: We have to skip vdso, which doesn't have a corresponding file on the disk (we
>     # assume that such files have paths starting with '/', seems ldd aways prints absolute paths).
> ```

`aways` -> `always`
___
*[python/graphenelibos/manifest.py, line 107 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXN1yTZ99iaqP9ZqlIy:-MXN1yTZ99iaqP9ZqlIz:b-gz1afb) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/python/graphenelibos/manifest.py#L107)):*
> ```Python
> def render(template, variables=None):
>     '''Render template, given as string. Optional variables may be given as mapping.'''
>     return _env.from_string(template).render(**(variables or {})) + '\n'
> ```

can we add this newline only when it's actually missing?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: 20 of 69 files reviewed, 13 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCCCScCKFrKQXz5Kam:-MXajjY23BO51Hizt8hn:bjb2l26) (no related file):*

I like the namespace idea.
</blockquote></details>

Done.

To avoid a related confusion, I added `undefined=StrictUndefined` to `jinja2.Environment`, so that any unknown substitution also throws an error (previously it was an error to do operations on an unknown subsitution, but just `{{ unknown }}` was empty string).

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCAuzgA_tutxBdxv9I:-MXaqleNBdTiBqb8Jjb7:b-80tlqk) (no related file):*

Also -1 from me to the `{{ preamble }}` idea.
</blockquote></details>

Sure.

---
___
*[Documentation/manpages/graphene-manifest.rst, line 24 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKBtoRB04rFK8uyhl6:-MXaTAwe09nQDwz2b27b:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Documentation/manpages/graphene-manifest.rst#L24)):*

> ```
> available in template.
> ```

Not "the template"? (the one you invoked the command on)
</blockquote></details>

Done.
___
*[Documentation/manpages/graphene-manifest.rst, line 26 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKCDRdE_qBJS_6kh_U:-MXaTAeH55_8jpoR6mvI:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Documentation/manpages/graphene-manifest.rst#L26)):*

> ```
> available in template
> ```

I'd rather use plural here ("templates")
</blockquote></details>

Done.
___
*[Documentation/manpages/graphene-manifest.rst, line 91 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKDHsB4gR301TzvkmN:-MXaT6mX4S5WHCb3tn0z:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Documentation/manpages/graphene-manifest.rst#L91)):*

missing space after `=`
</blockquote></details>

Done.
___
*[Documentation/manpages/graphene-manifest.rst, line 109 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKENKD-FcmbmXkdkZ1:-MXaTTtG3ffAs0Y36MXd:bch52eq) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Documentation/manpages/graphene-manifest.rst#L109)):*

I guess this is more like a pseudocode, right? I mean, you're not using the wildcard from the target pattern matching, which normally should somehow influence `$(ENTRYPOINT)`.
</blockquote></details>

Yes, it's a pseudocode, but it's more complicated. In a simple example like here, sure it could be inferred from pattern. In practice:

https://github.com/oscarlab/graphene/blob/d1de02f0b1fb393a9be35f4c3640ea1dcf81116a/Scripts/Makefile.rules#L160
https://github.com/oscarlab/graphene/blob/d1de02f0b1fb393a9be35f4c3640ea1dcf81116a/LibOS/shim/test/regression/Makefile#L114

So I don't know. Would you care to suggest a solution?
___
*[Examples/apache/Makefile, line 86 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXMoq8i1hLgppUCMH3l:-MXaUvwV42fBfgyQnPvB:b-9lodnz) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Examples/apache/Makefile#L86)):*

Don't we need doublequotes here? (same for the whole PR)
</blockquote></details>

Yes and no. You're right that if the variable contains some specific characters then this breaks. If you add quotes (single or double), then the set of characters that could break the script changes, but I believe in no case such set is empty. This is a general problem with make+shell. Fortunately, neither `GRAPHENE_LOG_LEVEL` nor `*_{DIR,PATH}`, `ENTRYPOINT` (which is also a path) etc. variables contain whitespace or quotes, so we should be safe.
___
*[Examples/blender/blender.manifest.template, line 14 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXMptuC0Dl6d-xjc3WB:-MXa_NeoDEq4gdyGDa3-:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Examples/blender/blender.manifest.template#L14)):*

missing space
</blockquote></details>

Done.
___
*[Examples/lighttpd/lighttpd.manifest.template, line 21 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKF-y-Eb0kWcR_v-nF:-MXa_Sjz0jC7V4xZ8b7S:bxr8do2) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Examples/lighttpd/lighttpd.manifest.template#L21)):*

> ```
> install_dir
> ```

Now this name became a bit ambiguous: It's not clear that this is lighttpd install path, not Graphene's install path. Please grep the whole PR for similar occurrences.
</blockquote></details>

This is substituted from `-D` in make, so it's explicit, and after the namespace change it even improved. In any case, this is not a regression in the PR.
___
*[Examples/lighttpd/Makefile, line 60 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKFLOg5U-LUkKxHDn3:-MXagZaG3FAVbeStiiDC:bvtqogv) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Examples/lighttpd/Makefile#L60)):*

you missed this place when converting alignment to indentation (spaces -> tabs)
</blockquote></details>

Uh, yes, and in other place too. I retabbed all the Makefiles in Examples.
___
*[python/graphenelibos/manifest.py, line 51 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCDO2b9anFAWEJwSv2:-MXagxM11BLvHKVoCc3M:bil8utz) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/python/graphenelibos/manifest.py#L51)):*

I'm not sure, there's not really too much uses for this selective adding. Currently the only issue is I think the loading time which grows, because we use keys instead of a TOML array (and this will be fixed soon). Otherwise `ldd` is error-prone (easy to miss dynamic deps) and quite hacky.

Actually, currently this function is unused, right?
</blockquote></details>

Yes, it is unused. But I wanted to include it anyway, since replicating this feature outside this script would be less convenient (it would probably involve the shell script that I rewrote here plus some unholy `-D` substitution with space-separated value, which would then be `.split()` in the template -- doable but harder).
___
*[python/graphenelibos/manifest.py, line 36 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXN03AQ5goPIO2FvYmb:-MXahEzs7jYIWDftBHcL:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/python/graphenelibos/manifest.py#L36)):*

`aways` -> `always`
</blockquote></details>

Done.
___
*[python/graphenelibos/manifest.py, line 107 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXN1yTZ99iaqP9ZqlIy:-MXahLEBA5WaECf0A6Ae:b-zi3ao6) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/python/graphenelibos/manifest.py#L107)):*

can we add this newline only when it's actually missing?
</blockquote></details>

I fixed this using `keep_trailing_newline=` argument to `Environment`. This way it will be present if it is present in the file.

mkow (changes requested):
> Reviewed 54 of 54 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCCCScCKFrKQXz5Kam:-MXbYwglAmMw-wrP4GGO:b-7hz0bu) (no related file):*

Done.

To avoid a related confusion, I added `undefined=StrictUndefined` to `jinja2.Environment`, so that any unknown substitution also throws an error (previously it was an error to do operations on an unknown subsitution, but just `{{ unknown }}` was empty string).
</blockquote></details>

Good, this sounds like something which we should enable by default.

---
___
*[Documentation/manpages/graphene-manifest.rst, line 109 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKENKD-FcmbmXkdkZ1:-MXbZN9j5tig6nNyRszK:b-qbf4ls) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Documentation/manpages/graphene-manifest.rst#L109)):*

Yes, it's a pseudocode, but it's more complicated. In a simple example like here, sure it could be inferred from pattern. In practice:

https://github.com/oscarlab/graphene/blob/d1de02f0b1fb393a9be35f4c3640ea1dcf81116a/Scripts/Makefile.rules#L160
https://github.com/oscarlab/graphene/blob/d1de02f0b1fb393a9be35f4c3640ea1dcf81116a/LibOS/shim/test/regression/Makefile#L114

So I don't know. Would you care to suggest a solution?
</blockquote></details>

Nah, I think I'm fine with the current over-simplified version. Just wanted to ensure that it was intended to be like this.
___
*[Examples/apache/Makefile, line 86 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXMoq8i1hLgppUCMH3l:-MXbZbQsEreq_PEaxP8j:b-ynjuh1) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Examples/apache/Makefile#L86)):*

Yes and no. You're right that if the variable contains some specific characters then this breaks. If you add quotes (single or double), then the set of characters that could break the script changes, but I believe in no case such set is empty. This is a general problem with make+shell. Fortunately, neither `GRAPHENE_LOG_LEVEL` nor `*_{DIR,PATH}`, `ENTRYPOINT` (which is also a path) etc. variables contain whitespace or quotes, so we should be safe.
</blockquote></details>

Ok. But I'd like this to get all places like this be fixed when migrating to Meson ;) (not necessarily by you, just in general)
___
*[Examples/apache/Makefile, line 122 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXbZtWl8BeHRg6PKty8:-MXbZtWl8BeHRg6PKty9:bl79pda) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/apache/Makefile#L122)):*
> ```Makefile
> 		-e "s|^#Include conf/extra/httpd-ssl.conf|Include conf/extra/httpd-ssl-graphene.conf|g" \
> 		-e "s|#EnableMMAP off|EnableMMAP off|g" \
> 		-e "s|#EnableSendfile on|EnableSendfile on|g" \
> ```

Here the spaces were on purpose (for alignment with `-e`), please undo.
___
*[Examples/apache/Makefile, line 143 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXb_-pZDMqDBzXhprzN:-MXb_-pZDMqDBzXhprzO:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/apache/Makefile#L143)):*
> ```Makefile
> 		-e "s|^<VirtualHost _default_:443>|<VirtualHost 127.0.0.1:8443>|g" \
> 		-e "s|^ServerName www.example.com:443|ServerName www.example.com:8443|g" \
> 		-e "s|^SSLSessionCache|#SSLSessionCache|g" \
> ```

ditto
___
*[Examples/blender/Makefile, line 90 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXb_7S5CnANaZjpDm1j:-MXb_7S5CnANaZjpDm1k:b-ixofv4) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/blender/Makefile#L90)):*
> ```Makefile
> clean:
> 	$(RM) -r $(RUN_DIR) $(DATA_DIR)/images blender.manifest blender.manifest.sgx blender.sig \
> 		blender.token
> ```

ditto (this was an alignment for readability)
___
*[Examples/capnproto/Makefile, line 54 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXb_JLr3kv18G5D3de_:-MXb_JLr3kv18G5D3dea:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/capnproto/Makefile#L54)):*
> ```Makefile
> 	cd $(SRCDIR) && capnp compile -oc++ addressbook.capnp
> 	cd $(SRCDIR) && c++ -std=c++14 -Wall addressbook.c++ addressbook.capnp.c++ \
> 		`pkg-config --cflags --libs capnp` -o addressbook
> ```

ditto
___
*[Examples/lighttpd/Makefile, line 60 at r2](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXKFLOg5U-LUkKxHDn3:-MXb_Ycx2dTwShulUxQ9:b-k5dwwe) ([raw file](https://github.com/oscarlab/graphene/blob/aaf9df2d9c7bbccc4805139018cd26b9193e5acf/Examples/lighttpd/Makefile#L60)):*

Uh, yes, and in other place too. I retabbed all the Makefiles in Examples.
</blockquote></details>

A few too many, I pointed them out :)
___
*[Examples/lighttpd/Makefile, line 123 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXb_eji2J0hjVnEMsC9:-MXb_eji2J0hjVnEMsCA:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/lighttpd/Makefile#L123)):*
> ```Makefile
> clean:
> 	$(RM) *.manifest *.manifest.sgx *.token *.sig pal_loader OUTPUT result-* $(CONF_FILES) \
> 		pal_loader
> ```

ditto
___
*[Examples/ra-tls-mbedtls/Makefile, line 240 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXba3og910UH7FqsMyf:-MXba3og910UH7FqsMyg:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/ra-tls-mbedtls/Makefile#L240)):*
> ```Makefile
> clean:
> 	$(RM) -r *.token *.sig *.manifest.sgx *.manifest pal_loader server client *.so *.so.* OUTPUT \
> 		libs/
> ```

ditto
___
*[Examples/ra-tls-secret-prov/Makefile, line 250 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXba7aJ0g124fsDneAD:-MXba7aJ0g124fsDneAE:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/ra-tls-secret-prov/Makefile#L250)):*
> ```Makefile
> 	$(RM) -r *.token *.sig *.manifest.sgx *.manifest *.so *.so.* OUTPUT pal_loader pf_crypt \
> 		secret_prov_server_epid secret_prov_server_dcap \
> 		secret_prov_client secret_prov_min_client secret_prov_pf_client libs/
> ```

ditto
___
*[python/graphenelibos/manifest.py, line 51 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCDO2b9anFAWEJwSv2:-MXbakHT-kyKLF0YzYF7:b-grtzrt) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/python/graphenelibos/manifest.py#L51)):*

Yes, it is unused. But I wanted to include it anyway, since replicating this feature outside this script would be less convenient (it would probably involve the shell script that I rewrote here plus some unholy `-D` substitution with space-separated value, which would then be `.split()` in the template -- doable but harder).
</blockquote></details>

But I don't expect anyone using this feature after we got the directory support in trusted files. Anyways, we can leave it for some time if you want, we'll see if it's used.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

___
*[Examples/apache/Makefile, line 122 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXbZtWl8BeHRg6PKty8:-MXdRiR-8z4FAHEEKhdB:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/apache/Makefile#L122)):*

Here the spaces were on purpose (for alignment with `-e`), please undo.
</blockquote></details>

Done.
___
*[Examples/apache/Makefile, line 143 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXb_-pZDMqDBzXhprzN:-MXdRjjdEudvfgmNLhpe:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/apache/Makefile#L143)):*

ditto
</blockquote></details>

Done.
___
*[Examples/blender/Makefile, line 90 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXb_7S5CnANaZjpDm1j:-MXdRPcxCL3a6tMdFnNw:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/blender/Makefile#L90)):*

ditto (this was an alignment for readability)
</blockquote></details>

Done.
___
*[Examples/capnproto/Makefile, line 54 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXb_JLr3kv18G5D3de_:-MXdRsD34wKeS2LN0tVa:b-98sbbb) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/capnproto/Makefile#L54)):*

ditto
</blockquote></details>

There is no good way to do it here. This version is only slightly better.
___
*[Examples/lighttpd/Makefile, line 123 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXb_eji2J0hjVnEMsC9:-MXdSQe28pO5nsBPV0T1:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/lighttpd/Makefile#L123)):*

ditto
</blockquote></details>

Done.
___
*[Examples/ra-tls-mbedtls/Makefile, line 240 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXba3og910UH7FqsMyf:-MXdSZQu5MS-IeQNbgk3:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/ra-tls-mbedtls/Makefile#L240)):*

ditto
</blockquote></details>

Done.
___
*[Examples/ra-tls-secret-prov/Makefile, line 250 at r3](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXba7aJ0g124fsDneAD:-MXdSi399A4WwbROcDBx:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7ac69df771444d0c96276e9e349cfc8d4d2d0e91/Examples/ra-tls-secret-prov/Makefile#L250)):*

ditto
</blockquote></details>

Done.

mkow (dismissed):
> Reviewed 6 of 6 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

dimakuv (changes requested):
> Reviewed 6 of 18 files at r2, 48 of 54 files at r3, 6 of 6 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: all files reviewed, 4 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCCCScCKFrKQXz5Kam:-MY4MsJI8qazaEjcMDaB:bn8qrc4) (no related file):*

Good, this sounds like something which we should enable by default.
</blockquote></details>

Namespaces look good, but now we have this sed-like replacement: `graphene_log_level`. I understand that this is not a predefined keyword (it is specified in the manifest files), but the name is confusing for newbies: they will see `graphene_log_level` and `graphene.libos` and will wonder why there's a difference (underscore vs dot).

To avoid confusion of newbies, I suggest to simply rename `graphene_log_level` to `log_level`.

---
___
*[Examples/apache/httpd.manifest.template, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXC4VPK1cZZploYD6p1:-MY4KRaI9ZlTeG-033Qs:b5hvp1t) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/Examples/apache/httpd.manifest.template#L26)):*

The rationale is that this is dependent on the chosen libc, so this leaves an option for `{{ get_runtimedir('musl') }}` and also maybe switching versions: `{{ get_runtimedir('glibc-12.345') }}`. Optimally we should be able to install multiple patched libcs together, and it should be enclave's decision which one to use (for example by `fs.mount`).

If you'd like, I could invent a magic internal object that would do OK whether it got `()` or not, like this:

```jinja
{{ runtimedir }}
{{ runtimedir() }}
{{ runtimedir('musl') }}
```
</blockquote></details>

It would be great to have such a magic internal object (with the name `runtimedir` which will alias to `get_runtimedir()`, just like you suggested).
___
*[Examples/python-scipy/Makefile, line 13 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCCqQ91MDMlq39cQMZ:-MY4LFZe2l0dTxoQ4PLE:byqe7mm) ([raw file](https://github.com/oscarlab/graphene/blob/46da82fbd3abfcbf2a9edcb554773aabb1b7e51f/Examples/python-scipy/Makefile#L13)):*

You're welcome. Anything else you'd like me to optimise out?
</blockquote></details>

No, I can't think of anything else.
___
*[Examples/python-scipy/python.manifest.template, line 18 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MY4LWChEmEdamP8m184:-MY4LWChEmEdamP8m185:br9yd3m) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/Examples/python-scipy/python.manifest.template#L18)):*
> ```
> # Environment variables for Python
> loader.env.LD_LIBRARY_PATH = "{{ python_stdlib }}/lib:/lib:{{ arch_libdir }}:/usr/lib:/usr/{{ arch_libdir }}"
> loader.env.HOME = "/home/user"
> ```

Actually, what is this? I guess it is simply unused by Python? Can we remove it?
___
*[Examples/python-simple/python.manifest.template, line 22 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MY4LbDmEMY0GSl8-93H:-MY4LbDmEMY0GSl8-93I:br9yd3m) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/Examples/python-simple/python.manifest.template#L22)):*
> ```
> #loader.env.PYTHONHOME = "$(PYTHONHOME)"
> #loader.env.PYTHONPATH = "$(PYTHONHOME):$(PYTHONHOME)/plat-$(ARCH_LONG):$(PYTHONDISTHOME):$(PYTHONHOME)/lib-dynload"
> loader.env.HOME = "/home/user"
> ```

Actually, what is this? I guess it is simply unused by Python? Can we remove it?

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: 16 of 70 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXCCCScCKFrKQXz5Kam:-MY5G8MeEuqrziQ25mB1:b-896fix) (no related file):*

Namespaces look good, but now we have this sed-like replacement: `graphene_log_level`. I understand that this is not a predefined keyword (it is specified in the manifest files), but the name is confusing for newbies: they will see `graphene_log_level` and `graphene.libos` and will wonder why there's a difference (underscore vs dot).

To avoid confusion of newbies, I suggest to simply rename `graphene_log_level` to `log_level`.
</blockquote></details>

Done.

---
___
*[Examples/apache/httpd.manifest.template, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MXC4VPK1cZZploYD6p1:-MY5Fg1s4nASTlxbgmHY:b-r5y06a) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/Examples/apache/httpd.manifest.template#L26)):*

It would be great to have such a magic internal object (with the name `runtimedir` which will alias to `get_runtimedir()`, just like you suggested).
</blockquote></details>

I think either 'get_runtimedir()` or `runtimedir` as above. Done.
___
*[Examples/python-scipy/python.manifest.template, line 18 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MY4LWChEmEdamP8m184:-MY53L4SDeAYIc9ONIpU:b-ygeclz) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/Examples/python-scipy/python.manifest.template#L18)):*

Actually, what is this? I guess it is simply unused by Python? Can we remove it?
</blockquote></details>

I don't know. 
___
*[Examples/python-simple/python.manifest.template, line 22 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MY4LbDmEMY0GSl8-93H:-MY5G2v09To3hougj1Nd:bikumrk) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/Examples/python-simple/python.manifest.template#L22)):*

Actually, what is this? I guess it is simply unused by Python? Can we remove it?
</blockquote></details>

Let's try and see what happens.

woju (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: 16 of 70 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

___
*[Examples/python-simple/python.manifest.template, line 22 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MY4LbDmEMY0GSl8-93H:-MY6H_rl6UCBqidW6ovX:b-2yefke) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/Examples/python-simple/python.manifest.template#L22)):*

Let's try and see what happens.
</blockquote></details>

Looks like nothing happened, tests passed. (Also fixed sibling python-scipy).

dimakuv (approved):
> Reviewed 54 of 54 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[Examples/python-simple/python.manifest.template, line 22 at r1](https://reviewable.io/reviews/oscarlab/graphene/2288#-MY4LbDmEMY0GSl8-93H:-MY6D5DR04ckk3bUa-M9:bgrgpc7) ([raw file](https://github.com/oscarlab/graphene/blob/a59bda8d698f95df5d4a1c1e5cc2a664e5f43209/Examples/python-simple/python.manifest.template#L22)):*

Let's try and see what happens.
</blockquote></details>

Everything works, so apparently it is not needed.

woju:
Jenkins, test Jenkins-SGX-18.04 please

mkow (approved):
> Reviewed 1 of 54 files at r6, 57 of 57 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

mkow:
Jenkins, retest Jenkins-SGX please (`test_pal.TC_23_SendHandle.test_000_send_handle` failed, known issue fixed by #2309)

dimakuv (approved):
> Reviewed 57 of 57 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2288)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2304, opened 2021-04-12
[Pal] Flush logging buffers after newline

pwmarcz:
This makes sure the PAL and untrusted SGX PAL logging subsystems always flush the buffer after encountering a newline, same as LibOS.

This minimizes the chance of log lines from different sources interleaving, and makes sure all (full) log lines are flushed before process exit.

Should fix test failures like this one: https://leeroy.cs.unc.edu/job/graphene-sgx-18.04/5067/

```
Send Handle OK
debug: Receive Handle OK
sock_getopt (fd = 7, sockopt addr = 0x7ffe1115e6a0) is not implemented and always returns 0
```

The above output was produced by `pal_printf("Receive Handle OK")` (from inner PAL) and `urts_log_debug("sock_getopt ...")` (from outer PAL).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2304)

### DIFF
--- a/Pal/src/host/Linux-SGX/sgx_log.c
+++ b/Pal/src/host/Linux-SGX/sgx_log.c
@@ -55,7 +55,7 @@ static int output_char(void* f, int ch, void* buf_) {
     struct log_buf* buf = (struct log_buf*)buf_;
 
     buf->buf[buf->end++] = ch;
-    if (buf->end == LOG_BUF_SIZE) {
+    if (ch == '\n' || buf->end == LOG_BUF_SIZE) {
         ret = write_all(buf->fd, buf->buf, buf->end);
         buf->end = 0;
     }
--- a/Pal/src/host/Linux-SGX/sgx_log.c
+++ b/Pal/src/host/Linux-SGX/sgx_log.c
@@ -68,7 +68,6 @@ static void print_to_fd(int fd, const char* fmt, va_list ap) {
     buf.fd = fd;
     buf.end = 0;
     vfprintfmt(output_char, NULL, &buf, fmt, ap);
-    write_all(fd, buf.buf, buf.end);
     // No error handling, as `_urts_log` doesn't return errors anyways.
 }
 
--- a/Pal/src/printf.c
+++ b/Pal/src/printf.c
@@ -33,7 +33,7 @@ static int fputch(void* f, int ch, void* buf_) {
     struct printbuf* buf = buf_;
 
     buf->buf[buf->idx++] = ch;
-    if (buf->idx == PRINTBUF_SIZE - 1) {
+    if (ch == '\n' || buf->idx == PRINTBUF_SIZE - 1) {
         _DkPrintConsole(buf->buf, buf->idx);
         buf->idx = 0;
     }


dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2304)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required) (waiting on @mkow)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2304)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2304)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2304#-MY4gcsv63cmn2n_AHrl:-MY4gcsv63cmn2n_AHrm:bk4ta43) (no related file):*
We've already spotted this issue before, @mkow  was about to fix it (I think). Afaik flushing on newlines does not solve it, the only ways I can think of are:
1) introduce global/per thread buffer (just like in LibOS)
2) use snprintf to append log_level prefixes
3) fix the faulty test (Pal/regression)

---
___
*[Pal/src/host/Linux-SGX/sgx_log.c, line 71 at r2](https://reviewable.io/reviews/oscarlab/graphene/2304#-MY4gKTm-uS7DK-C9mCc:-MY4gKTo9PQcwfzHkAnl:b-9jy2ze) ([raw file](https://github.com/oscarlab/graphene/blob/189f99f62bc6cbd5d368b1ef26760f71ffe4e967/Pal/src/host/Linux-SGX/sgx_log.c#L71)):*
> ```C
>     buf.end = 0;
>     vfprintfmt(output_char, NULL, &buf, fmt, ap);
>     write_all(fd, buf.buf, buf.end);
> ```

I don't think this is correct - if you write `log_error("ABC")` it will never be printed and just disappear.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2304)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2304#-MY4gcsv63cmn2n_AHrl:-MY4nTAF87PezOfzo-05:b-a1mby5) (no related file):*

We've already spotted this issue before, @mkow  was about to fix it (I think). Afaik flushing on newlines does not solve it, the only ways I can think of are:
1) introduce global/per thread buffer (just like in LibOS)
2) use snprintf to append log_level prefixes
3) fix the faulty test (Pal/regression)
</blockquote></details>

Good point, this doesn't help when there are multiple threads. The status quo is worse, though, as it doesn't guarantee unbroken messages even for a single thread. So how about merging this before @mkow fixes it?

(Using snprintf sounds like the simplest solution, I don't think there is a need for global/thread-local buffer? Except LibOS syscall logging which still depends on it.)

---
___
*[Pal/src/host/Linux-SGX/sgx_log.c, line 71 at r2](https://reviewable.io/reviews/oscarlab/graphene/2304#-MY4gKTm-uS7DK-C9mCc:-MY4miyYDFXty4TABU65:b-yavaa6) ([raw file](https://github.com/oscarlab/graphene/blob/189f99f62bc6cbd5d368b1ef26760f71ffe4e967/Pal/src/host/Linux-SGX/sgx_log.c#L71)):*

I don't think this is correct - if you write `log_error("ABC")` it will never be printed and just disappear.
</blockquote></details>

Yes, the assumption is that people will do `log_error("...\n")` (or issue multiple calls, with newline in the last one). That's how it works in the other two logging subsystems. 

I agree it's not very good, but the current code, with a hacky buffer flush in just 1 of 3 of these systems, is worse, I think. It causes logging to work in a surprising way (`urts_log_error("ABC")` is flushed immediately), and only in the outer PAL. It also goes against the rest of that code - we use a global buffer but work around it.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2304)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2304#-MY4gcsv63cmn2n_AHrl:-MY4rHcm7OkToh6lxLE1:b8u7gju) (no related file):*
> The status quo is worse, though, as it doesn't guarantee unbroken messages even for a single thread.

How can a single thread break it?

There's also another way: move all logging into Pal and have a special `DkSomething(log_level, fmt, ...)` function for logging (so that we have one logging buffer per thread, managed by Pal)

---
___
*[Pal/src/host/Linux-SGX/sgx_log.c, line 71 at r2](https://reviewable.io/reviews/oscarlab/graphene/2304#-MY4gKTm-uS7DK-C9mCc:-MY4s3ir6UYNzBndFstT:b-ietefk) ([raw file](https://github.com/oscarlab/graphene/blob/189f99f62bc6cbd5d368b1ef26760f71ffe4e967/Pal/src/host/Linux-SGX/sgx_log.c#L71)):*

Yes, the assumption is that people will do `log_error("...\n")` (or issue multiple calls, with newline in the last one). That's how it works in the other two logging subsystems. 

I agree it's not very good, but the current code, with a hacky buffer flush in just 1 of 3 of these systems, is worse, I think. It causes logging to work in a surprising way (`urts_log_error("ABC")` is flushed immediately), and only in the outer PAL. It also goes against the rest of that code - we use a global buffer but work around it.
</blockquote></details>

No, now if you do:
```
log_error("ABC");
log_error("123\n");
```
only `123\n` will be printed, `ABC` will just silently disappear (even on a single thread).
`struct log_buf buf` is local to this function, so you lose everything that's not flushed.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2304)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2304#-MY4gcsv63cmn2n_AHrl:-MY4tA2ZAMD3smxfssYn:b-3kobng) (no related file):*
>How can a single thread break it?

As mentioned in the PR, if you use two of these logging systems (for example inner PAL and outer PAL), you can get `debug:` prefix from system A, then message from system B, then the rest of message from system A.

Anyway, yes, this fix is broken. I'll close the PR.

---
___
*[Pal/src/host/Linux-SGX/sgx_log.c, line 71 at r2](https://reviewable.io/reviews/oscarlab/graphene/2304#-MY4gKTm-uS7DK-C9mCc:-MY4sqda4wW7gfwfL2Uy:b-6lhvna) ([raw file](https://github.com/oscarlab/graphene/blob/189f99f62bc6cbd5d368b1ef26760f71ffe4e967/Pal/src/host/Linux-SGX/sgx_log.c#L71)):*

No, now if you do:
```
log_error("ABC");
log_error("123\n");
```
only `123\n` will be printed, `ABC` will just silently disappear (even on a single thread).
`struct log_buf buf` is local to this function, so you lose everything that's not flushed.
</blockquote></details>

Ouch, right... I thought this was a global buffer. So that's right, this fix doesn't make sense. Closing.

Issue #2321, opened 2021-04-19
RFC: Dentry cleanup


pwmarcz:
Master issue: #1803

I started looking at the filesystem and want to do the following cleanup. It is intended to be a clear improvement, but a much smaller effort than a complete rewrite of all filesystems.

This issue is me asking for a **sanity check**: is there anything that sounds like a bad idea, or will turn out to be impossible on closer look?

## Summary

- Get rid of the hashes from the API, at least for now. This is an (only partially implemented) optimization that can be added later.

- Introduce proper locking discipline for the data stored in dentry: `dcache_lock` for modifying the actual tree structure, `dent->lock` for operations on a file. Keep it simple; we might think about optimizations later, but for now let's make sure this is correct.

- Trim down the list of flags: `DENTRY_ISLINK`, `DENTRY_ISDIRECTORY`, and `DENTRY_MOUNTPOINT` are reduntant with the already-present dentry fields (`type` and `mounted`).

- Do *not* split off an `inode` structure yet; this might be useful, but is too big a change yet. I am aware this will complicate implementation of several features (hardlinks, renaming, unlinking an open file).

- Refactor the "lookup" API, and change it so that it distinguishes between "file does not exist but can be created" (return 0 and a negative dentry) and "file does not exist and cannot be created" (return `-ENOENT`).

- Make sure the reference counting is correct, and we at least delete unused *non-positive* dentries. Later we might want to also delete positive dentries, but the tmpfs filesystem stores data in them, so they need to persist for now.

- Keep the current mount semantics for now (creating synthetic directories whenever necessary), but make sure the mount points actually appear; currently `readdir("/")` doesn't show `/lib`, `/proc`, etc.

  Later, we might want to get rid of the synthetic directories, and instead make the root-level filesystem a tmpfs. This will bring the semantics closer to normal Linux.

- Fix lookup flags: there is `LOOKUP_FOLLOW` (disabled most of the time), but actually it means following symlinks along the path to target, whereas Linux `O_NOFOLLOW` and `lstat` are only concerned with whether the target file (at the end of the path) is a symlink or not

## API draft

```c
/*
 * Directory entries. This object currently serves two purposes:
 *
 * 1. Information about file names, to help with path lookups (dentry in Linux).
 * 2. Information about files themselves (inode in Linux).
 *
 * This has some disadvantages (e.g. hardlinks, or renaming, is harder to implement), so later we
 * might want to create a separate `shim_inode` struct.
 */
struct shim_dentry {
[...]

PR #2324, opened 2021-04-21
[LibOS] Simplify low-level dentry functions

pwmarcz:
Related to issue #2321 (dentry cleanup)

See the commit message for full list.

This takes care of some low hanging fruit in the dentry cache system - the things that, hopefully, do not cascade to a bigger change.

Next steps are going to be a bit bigger:
- Rewrite `mount_fs`. The current mount semantics are pretty weird: `DENTRY_ANCESTOR` synthetic entries, confusion between parent and child filesystem dentry, deleting old dentries under a mountpoint, mountpoints are invisible in listings (`readdir("/")` doesn't contain `"proc"` even though listing `/proc` works), etc.
  - Fun fact: the regression tests contain a `dev` binary which somehow gets mounted under `/dev`, alongside (???) the dev filesystem
- Rewrite `path_lookupat` / `open_namei` functions, as outlined in #2321 (saner semantics around files that do not exist, and following links); unfortunately this means rewriting all of the call sites, so I want to do it in a separate PR
- Get serious about reference counting and debug whatever is happening there; probably I'll write some nice function to dump the full dentry tree

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -93,7 +93,6 @@ struct shim_fs_ops {
 #define DENTRY_NEGATIVE    0x0002 /* recently deleted or inaccessible */
 #define DENTRY_RECENTLY    0x0004 /* recently used */
 #define DENTRY_PERSIST     0x0008 /* added as a persistent dentry */
-#define DENTRY_HASHED      0x0010 /* added in the dcache */
 #define DENTRY_MOUNTPOINT  0x0040 /* this dentry is a mount point */
 #define DENTRY_ISLINK      0x0080 /* this dentry is a link */
 #define DENTRY_ISDIRECTORY 0x0100 /* this dentry is a directory */
@@ -218,7 +217,7 @@ struct shim_mount {
     LIST_TYPE(shim_mount) list;
 };
 
-extern struct shim_dentry* dentry_root;
+extern struct shim_dentry* g_dentry_root;
 
 #define LOOKUP_FOLLOW    001
 #define LOOKUP_DIRECTORY 002
@@ -296,24 +295,24 @@ int walk_mounts(int (*walk)(struct shim_mount* mount, void* arg), void* arg);
 /* functions for dcache supports */
 int init_dcache(void);
 
-extern struct shim_lock dcache_lock;
+extern struct shim_lock g_dcache_lock;
 
 /* Checks permission (specified by mask) of a dentry. If force is not set, permission is considered
  * granted on invalid dentries.
- * Assumes that caller has acquired dcache_lock. */
+ * Assumes that caller has acquired g_dcache_lock. */
 int __permission(struct shim_dentry* dent, mode_t mask);
 
 /* This function looks up a single dentry based on its parent dentry pointer and the name. `namelen`
  * is the length of char* name. The dentry is returned in pointer *new.
  *
- * The caller should hold the dcache_lock.
+ * The caller should hold the g_dcache_lock.
  */
 int lookup_dentry(struct shim_dentry* parent, const char* name, int namelen,
                   struct shim_dentry** new, struct shim_mount* fs);
 
 /* Looks up path under start dentry. Saves in dent.
  *
- * Assumes dcache_lock is held; main difference from path_lookupat is that dcache_lock is not
+ * Assumes g_dcache_lock is held; main difference from path_lookupat is that g_dcache_lock is not
  * released on return.
  *
  * The refcount is raised by one on the returned dentry.
@@ -330,7 +329,7 @@ int __path_lookupat(struct shim_dentry* start, const char* path, int flags,
                     struct shim_dentry** dent, int link_depth, struct shim_mount* fs,
                     bool make_ancestor);
 
-/* Just wraps __path_lookupat, but also acquires and releases the dcache_lock. */
+/* Just wraps __path_lookupat, but also acquires and releases the g_dcache_lock. */
 int path_lookupat(struct shim_dentry* start, const char* name, int flags, struct shim_dentry** dent,
                   struct shim_mount* fs);
 
@@ -386,8 +385,34 @@ int list_directory_handle(struct shim_dentry* dent, struct shim_handle* hdl);
 void get_dentry(struct shim_dentry* dent);
 /* Decrement the reference count on dent */
 void put_dentry(struct shim_dentry* dent);
-/* Decrement the reference count by one and delete `dent` if this is the last reference. */
-void put_dentry_maybe_delete(struct shim_dentry* dent);
+
+/*!
+ * \brief Garbage-collect a dentry, if possible
+ *
+ * \param dentry the dentry (has to have a parent)
+ *
+ * This function checks if a dentry is unused, and deletes it if that's true. The caller must hold
+ * `g_dcache_lock`.
+ *
+ * A dentry is unused if it has no external references and does not correspond to a real file
+ * (i.e. is invalid or negative). Such dentries can remain after failed lookups or file deletion.
+ *
+ * The function should be called when processing a list of children, after you're done with a given
+ * dentry. It guarantees that the amortized cost of processing such dentries is constant, i.e. they
+ * will be only encountered once.
+ *
+ * \code
+ * struct shim_dentry* child;
+ * struct shim_dentry* tmp;
+ *
+ * LISTP_FOR_EACH_ENTRY_SAFE(child, tmp, &dent->children, siblings) {
+ *     // do something with `child`, increase ref count if used
+ *     ...
+ *     dentry_gc(child);
+ * }
+ * \endcode
+ */
+void dentry_gc(struct shim_dentry* dent);
 
 /* Size of the path constructed by dentry_get_path(), including null terminator. */
 size_t dentry_get_path_size(struct shim_dentry* dent);
@@ -408,31 +433,51 @@ static inline const char* dentry_get_name(struct shim_dentry* dent) {
     return qstrgetstr(&dent->name);
 }
 
-/* Allocate and initialize a new dentry for path name, under parent.  Return the dentry.
+/*!
+ * \brief Allocate and initialize a new dentry
  *
- * `mount` is the mountpoint the dentry is under; this is typically the parent->fs, but is passed
- * explicitly for initializing the dentry of a mountpoint.
+ * \param parent the parent node, or NULL if this is supposed to be the dentry root
+ * \param fs the filesystem the dentry is under, or NULL
+ * \param name name of the new dentry
+ * \param name_len length of the name
  *
- * If hashptr is passed (as an optimization), this is a hash of the name.
+ * \returns the new dentry, or NULL in case of allocation failure
  *
- * If parent is non-null, the ref count is 2; else it is 1.
+ * The caller should hold `g_dcache_lock`.
  *
- * This function also sets up both a name and a relative path
+ * The function will initialize the following fields: `fs` (if provided), `name`, `rel_path`, and
+ * parent/children links.
+ *
+ * The reference count of the returned dentry will be 2 if `parent` was provided, 1 otherwise.
+ *
+ * The `fs` parameter should typically be `parent->fs`, but is passed explicitly to support
+ * initializing the root dentry of a newly mounted filesystem. If `fs` is NULL, the resulting
+ * dentry's filesystem will be left as NULL.
+ *
+ * TODO: This function sets `rel_path` of a newly created dentry to:
+ * - `parent->rel_path + "/" + name` if parent exists and has a relative path,
+ * - `name` otherwise.
+ * This is usually right, but is wrong in the case of mounting of a new filesystem, in which case
+ * the `rel_path` has to be manually reset to empty. This should be fixed together with the mount
+ * semantics.
  */
-struct shim_dentry* get_new_dentry(struct shim_mount* mount, struct shim_dentry* parent,
-                                   const char* name, int namelen, HASHTYPE* hashptr);
+struct shim_dentry* get_new_dentry(struct shim_mount* fs, struct shim_dentry* parent,
+                                   const char* name, size_t name_len);
 
-/* This function searches for name/namelen (as the relative path).
+/*!
+ * \brief Search for a child of a dentry with a given name
  *
- * If requested, the expected hash of the dentry is returned in hashptr, primarily so that the
- * hashing can be reused to add the dentry later.
+ * \param parent the dentry to search under
+ * \param name name of searched dentry
+ * \param name_len length of the name
  *
- * The reference count on the found dentry is incremented by one.
+ * \returns the new dentry, or NULL if not found
  *
- * Used only by shim_namei.c
+ * The caller should hold `g_dcache_lock`.
+ *
+ * If found, the reference count on the returned dentry is incremented.
  */
-struct shim_dentry* __lookup_dcache(struct shim_dentry* start, const char* name, int namelen,
-                                    HASHTYPE* hashptr);
+struct shim_dentry* lookup_dcache(struct shim_dentry* parent, const char* name, size_t name_len);
 
 /* This function recursively deletes and frees all dentries under root
  *
--- a/LibOS/shim/include/shim_types.h
+++ b/LibOS/shim/include/shim_types.h
@@ -365,7 +365,6 @@ struct shim_str {
 /* Use qstr for names. This has fixed size string + string object
  * if len > SHIM_QSTR_SIZE then use overflow string */
 struct shim_qstr {
-    HASHTYPE hash;
     size_t len;
     char name[QSTR_SIZE];
     struct shim_str* oflow;
--- a/LibOS/shim/include/shim_utils.h
+++ b/LibOS/shim/include/shim_utils.h
@@ -123,7 +123,6 @@ static inline int qstrempty(const struct shim_qstr* qstr) {
 
 static inline void qstrcopy(struct shim_qstr* to, const struct shim_qstr* from) {
     qstrsetstr(to, qstrgetstr(from), from->len);
-    to->hash = from->hash;
 }
 
 static inline int qstrcmpstr(const struct shim_qstr* qstr, const char* str, size_t len) {
--- a/LibOS/shim/src/bookkeep/shim_handle.c
+++ b/LibOS/shim/src/bookkeep/shim_handle.c
@@ -465,7 +465,7 @@ void put_handle(struct shim_handle* hdl) {
         }
 
         if (hdl->dentry)
-            put_dentry_maybe_delete(hdl->dentry);
+            put_dentry(hdl->dentry);
 
         if (hdl->fs)
             put_mount(hdl->fs);
--- a/LibOS/shim/src/fs/shim_dcache.c
+++ b/LibOS/shim/src/fs/shim_dcache.c
@@ -26,15 +26,11 @@ static struct shim_lock dcache_mgr_lock;
 #define OBJ_TYPE struct shim_dentry
 #include "memmgr.h"
 
-struct shim_lock dcache_lock;
+struct shim_lock g_dcache_lock;
 
 static MEM_MGR dentry_mgr = NULL;
 
-struct shim_dentry* dentry_root = NULL;
-
-static inline HASHTYPE hash_dentry(struct shim_dentry* start, const char* path, int len) {
-    return rehash_path(start ? start->rel_path.hash : 0, path, len);
-}
+struct shim_dentry* g_dentry_root = NULL;
 
 static struct shim_dentry* alloc_dentry(void) {
     struct shim_dentry* dent =
@@ -59,30 +55,30 @@ static struct shim_dentry* alloc_dentry(void) {
 }
 
 int init_dcache(void) {
-    if (!create_lock(&dcache_mgr_lock) || !create_lock(&dcache_lock)) {
+    if (!create_lock(&dcache_mgr_lock) || !create_lock(&g_dcache_lock)) {
         return -ENOMEM;
     }
 
     dentry_mgr = create_mem_mgr(init_align_up(DCACHE_MGR_ALLOC));
 
-    dentry_root = alloc_dentry();
-    if (!dentry_root) {
+    g_dentry_root = alloc_dentry();
+    if (!g_dentry_root) {
         return -ENOMEM;
     }
 
     /* The root is special; we assume it won't change or be freed, so we artificially increase its
      * refcount by 1. */
-    get_dentry(dentry_root);
+    get_dentry(g_dentry_root);
 
     /* Initialize the root to a valid state, as a low-level lookup
      *  will fail. */
-    dentry_root->state |= DENTRY_VALID;
+    g_dentry_root->state |= DENTRY_VALID;
 
     /* The root should be a directory too*/
-    dentry_root->state |= DENTRY_ISDIRECTORY;
+    g_dentry_root->state |= DENTRY_ISDIRECTORY;
 
-    qstrsetstr(&dentry_root->name, "", 0);
-    qstrsetstr(&dentry_root->rel_path, "", 0);
+    qstrsetstr(&g_dentry_root->name, "", 0);
+    qstrsetstr(&g_dentry_root->rel_path, "", 0);
 
     return 0;
 }
@@ -90,9 +86,9 @@ int init_dcache(void) {
 /* Increment the reference count for a dentry */
 void get_dentry(struct shim_dentry* dent) {
 #ifdef DEBUG_REF
-    int count = REF_INC(dent->ref_count);
+    int64_t count = REF_INC(dent->ref_count);
 
-    log_debug("get dentry %p(%s/%s) (ref_count = %d)\n", dent,
+    log_debug("get dentry %p(%s/%s) (ref_count = %lld)\n", dent,
               dent->fs ? qstrgetstr(&dent->fs->path) : "", qstrgetstr(&dent->rel_path), count);
 #else
     REF_INC(dent->ref_count);
@@ -127,189 +123,93 @@ static void free_dentry(struct shim_dentry* dent) {
     free_mem_obj_to_mgr(dentry_mgr, dent);
 }
 
-/*
- * Decrement the reference count on dent.
- *
- * If we notice that dentry is negative and there are only 2 references left (ours and 1 for
- * parent's children list) we try to free this dentry - try, because somebody could get another
- * reference asynchronously.
- */
-void put_dentry_maybe_delete(struct shim_dentry* dent) {
-    long count = REF_GET(dent->ref_count);
-    assert(count >= 0);
-
-    /* First try without the lock. This check is racy - in some cases dentry might be left alive,
-     * but this is a opportunistic free anyway. */
-    if (count == 2) {
-        lock(&dcache_lock);
-
-        count = REF_GET(dent->ref_count);
-        /* If the dentry does not exist on fs anymore, let's delete it. */
-        if (count == 2 && dent->state & DENTRY_NEGATIVE && dent->parent) {
-            LISTP_DEL_INIT(dent, &dent->parent->children, siblings);
-            put_dentry(dent);
-        }
-
-        unlock(&dcache_lock);
-    }
-
-    /* This might free `dent`. */
-    put_dentry(dent);
-}
-
 void put_dentry(struct shim_dentry* dent) {
-    int count = REF_DEC(dent->ref_count);
+    int64_t count = REF_DEC(dent->ref_count);
+#ifdef DEBUG_REF
+    log_debug("put dentry %p(%s/%s) (ref_count = %lld)\n", dent,
+              dent->fs ? qstrgetstr(&dent->fs->path) : "", qstrgetstr(&dent->rel_path), count);
+#endif
     assert(count >= 0);
+
     if (count == 0) {
         assert(LIST_EMPTY(dent, siblings));
+        assert(LISTP_EMPTY(&dent->children));
         free_dentry(dent);
     }
 }
 
-/* Allocate and initialize a new dentry for path name, under
- * parent.  Return the dentry.
- *
- * mount is the mountpoint the dentry is under; this is typically
- * the parent->fs, but is passed explicitly for initializing
- * the dentry of a mountpoint.
- *
- * If hashptr is passed (as an optimization), this is a hash
- * of the name.
- *
- * If parent is non-null, the ref count is 2; else it is 1.
- *
- * This function also sets up both a name and a relative path
- */
-struct shim_dentry* get_new_dentry(struct shim_mount* mount, struct shim_dentry* parent,
-                                   const char* name, int namelen, HASHTYPE* hashptr) {
-    assert(locked(&dcache_lock));
+void dentry_gc(struct shim_dentry* dent) {
+    assert(locked(&g_dcache_lock));
+    assert(dent->parent);
+
+    if (REF_GET(dent->ref_count) != 1)
+        return;
+
+    if ((dent->state & DENTRY_VALID) && !(dent->state & DENTRY_NEGATIVE))
+        return;
+
+    LISTP_DEL_INIT(dent, &dent->parent->children, siblings);
+    dent->parent->nchildren--;
+    /* This should delete `dent` */
+    put_dentry(dent);
+}
+
+struct shim_dentry* get_new_dentry(struct shim_mount* fs, struct shim_dentry* parent,
+                                   const char* name, size_t name_len) {
+    assert(locked(&g_dcache_lock));
 
     struct shim_dentry* dent = alloc_dentry();
-    HASHTYPE hash;
 
     if (!dent)
         return NULL;
 
-    if (hashptr) {
-#ifdef DEBUG
-        // For debug builds, assert that the hash passed in is correct.
-        assert(*hashptr == hash_dentry(parent, name, namelen));
-#endif
-        hash = *hashptr;
-    } else {
-        hash = hash_dentry(parent, name, namelen);
+    if (fs) {
+        get_mount(fs);
+        dent->fs = fs;
     }
 
-    qstrsetstr(&dent->name, name, namelen);
-    dent->rel_path.hash = hash;
-    /* DEP 6/16/17: Not sure this flag is strictly necessary.
-     * But keeping it for now.
-     */
-    dent->state |= DENTRY_HASHED;
-
-    if (mount) {
-        get_mount(mount);
-        dent->fs = mount;
-    }
+    qstrsetstr(&dent->name, name, name_len);
 
     if (parent) {
-        // Increment both dentries' ref counts once they are linked
         get_dentry(parent);
+        dent->parent = parent;
+
         get_dentry(dent);
         LISTP_ADD_TAIL(dent, &parent->children, siblings);
-        dent->parent = parent;
         parent->nchildren++;
 
         if (!qstrempty(&parent->rel_path)) {
             const char* strs[] = {qstrgetstr(&parent->rel_path), "/", name};
-            size_t lens[]      = {parent->rel_path.len, 1, namelen};
+            size_t lens[]      = {parent->rel_path.len, 1, name_len};
             assert(lens[0] + lens[1] + lens[2] < STR_SIZE);
             qstrsetstrs(&dent->rel_path, 3, strs, lens);
         } else {
-            qstrsetstr(&dent->rel_path, name, namelen);
+            qstrsetstr(&dent->rel_path, name, name_len);
         }
     } else {
-        qstrsetstr(&dent->rel_path, name, namelen);
+        qstrsetstr(&dent->rel_path, name, name_len);
     }
 
     return dent;
 }
 
-/* This function searches for name/namelen (as the relative path) under
- * the parent directory (start).
- *
- * If requested, the expected hash of the dentry is returned in hashptr,
- * primarily so that the hashing can be reused to add the dentry later.
- *
- * The reference count on the found dentry is incremented by one.
- *
- * Used only by shim_namei.c
- */
-struct shim_dentry* __lookup_dcache(struct shim_dentry* start, const char* name, int namelen,
-                                    HASHTYPE* hashptr) {
-    assert(locked(&dcache_lock));
+struct shim_dentry* lookup_dcache(struct shim_dentry* parent, const char* name, size_t name_len) {
+    assert(locked(&g_dcache_lock));
 
-    /* In this implementation, we just look at the children
-     * under the parent and see if there are matches.  It so,
-     * return it; if not, don't.
-     *
-     * To minimize disruption (and possibly for future optimization)
-     * we are keeping hashes, so let's start with that for a marginally
-     * faster comparison
-     */
-    HASHTYPE hash = hash_dentry(start, name, namelen);
-    struct shim_dentry *dent, *found = NULL;
+    assert(parent);
+    assert(name_len > 0);
 
-    /* If start is NULL, there will be no hit in the cache.
-     * This mainly happens when boostrapping; in general, we assume the
-     * caller will use the current root or cwd.
-     */
-    if (!start)
-        return NULL;
-
-    /* If we are looking up an empty string, return start */
-    if (namelen == 0) {
-        get_dentry(start);
-        found = start;
-        goto out;
+    struct shim_dentry* tmp;
+    struct shim_dentry* dent;
+    LISTP_FOR_EACH_ENTRY_SAFE(dent, tmp, &parent->children, siblings) {
+        if (qstrcmpstr(&dent->name, name, name_len) == 0) {
+            get_dentry(dent);
+            return dent;
+        }
+        dentry_gc(dent);
     }
 
-    LISTP_FOR_EACH_ENTRY(dent, &start->children, siblings) {
-        /* DEP 6/20/XX: The old code skipped mountpoints; I don't see any good
-         * reason for mount point lookup to fail, at least in this code.
-         * Keeping a note just in case.  That is why you always leave a note.
-         */
-        // if (dent->state & DENTRY_MOUNTPOINT)
-        //     continue;
-
-        // Check for memory corruption
-        assert((dent->state & DENTRY_INVALID_FLAGS) == 0);
-
-        /* Compare the hash first */
-        if (dent->rel_path.hash != hash)
-            continue;
-
-        /* I think comparing the relative path is adequate; with a global
-         * hash table, a full path comparison may be needed, but I think
-         * we can assume a parent has children with unique names */
-        const char* filename = get_file_name(name, namelen);
-        const char* dname    = dentry_get_name(dent);
-        int dname_len        = strlen(dname);
-        int fname_len        = name + namelen - filename;
-        if (dname_len != fname_len || memcmp(dname, filename, fname_len))
-            continue;
-
-        /* If we get this far, we have a match */
-        get_dentry(dent);
-        found = dent;
-        break;
-    }
-
-out:
-    if (hashptr)
-        *hashptr = hash;
-
-    return found;
+    return NULL;
 }
 
 /* This function recursively removes children and drops the reference count
@@ -322,7 +222,7 @@ out:
  * structure on the heap to track progress.
  */
 int __del_dentry_tree(struct shim_dentry* root) {
-    assert(locked(&dcache_lock));
+    assert(locked(&g_dcache_lock));
 
     struct shim_dentry *cursor, *n;
 
@@ -334,9 +234,6 @@ int __del_dentry_tree(struct shim_dentry* root) {
         LISTP_DEL_INIT(cursor, &root->children, siblings);
         cursor->parent = NULL;
         root->nchildren--;
-        // Clear the hashed flag, in case there is any vestigial code based
-        //  on this state machine (where hased == valid).
-        cursor->state &= ~DENTRY_HASHED;
         put_dentry(cursor);
     }
 
@@ -445,7 +342,7 @@ BEGIN_RS_FUNC(dentry) {
 #if DEBUG_RESUME == 1
     char buffer[dentry_get_path_size(dent)];
 #endif
-    DEBUG_RS("hash=%08lx,path=%s,fs=%s", dent->rel_path.hash, dentry_get_path(dent, buffer),
+    DEBUG_RS("path=%s,fs=%s", dentry_get_path(dent, buffer),
              dent->fs ? qstrgetstr(&dent->fs->path) : NULL);
 }
 END_RS_FUNC(dentry)
--- a/LibOS/shim/src/fs/shim_fs.c
+++ b/LibOS/shim/src/fs/shim_fs.c
@@ -405,7 +405,7 @@ int search_builtin_fs(const char* type, struct shim_mount** fs) {
 }
 
 static int __mount_fs(struct shim_mount* mount, struct shim_dentry* dent) {
-    assert(locked(&dcache_lock));
+    assert(locked(&g_dcache_lock));
 
     int ret = 0;
 
@@ -418,7 +418,7 @@ static int __mount_fs(struct shim_mount* mount, struct shim_dentry* dent) {
 
     if (!mount_root) {
         /* mount_root->state |= DENTRY_VALID; */
-        mount_root = get_new_dentry(mount, NULL, "", 0, NULL);
+        mount_root = get_new_dentry(mount, /*fs=*/NULL, /*name=*/"", /*name_len=*/0);
         assert(mount->d_ops && mount->d_ops->lookup);
         ret = mount->d_ops->lookup(mount_root);
         if (ret < 0) {
@@ -526,7 +526,7 @@ int mount_fs(const char* type, const char* uri, const char* mount_point, struct 
     find_last_component(mount_point, &last, &last_len);
 
     bool need_parent_put = false;
-    lock(&dcache_lock);
+    lock(&g_dcache_lock);
 
     if (!parent) {
         // See if we are not at the root mount
@@ -536,7 +536,7 @@ int mount_fs(const char* type, const char* uri, const char* mount_point, struct 
             char* parent_path = __alloca(parent_len + 1);
             memcpy(parent_path, mount_point, parent_len);
             parent_path[parent_len] = 0;
-            if ((ret = __path_lookupat(dentry_root, parent_path, 0, &parent, 0, dentry_root->fs,
+            if ((ret = __path_lookupat(g_dentry_root, parent_path, 0, &parent, 0, g_dentry_root->fs,
                                        make_ancestor)) < 0) {
                 log_error("Path lookup failed %d\n", ret);
                 goto out_with_unlock;
@@ -575,16 +575,16 @@ int mount_fs(const char* type, const char* uri, const char* mount_point, struct 
     struct shim_dentry* dent2;
     /* Special case the root */
     if (last_len == 0)
-        dent = dentry_root;
+        dent = g_dentry_root;
     else {
-        dent = __lookup_dcache(parent, last, last_len, NULL);
+        dent = lookup_dcache(parent, last, last_len);
 
         if (!dent) {
-            dent = get_new_dentry(mount, parent, last, last_len, NULL);
+            dent = get_new_dentry(mount, parent, last, last_len);
         }
     }
 
-    if (dent != dentry_root && dent->state & DENTRY_VALID) {
+    if (dent != g_dentry_root && dent->state & DENTRY_VALID) {
         log_error("Mount %s already exists, verify that there are no duplicate mounts in manifest\n"
                   "(note that /proc and /dev are automatically mounted in Graphene).\n",
                   mount_point);
@@ -595,11 +595,10 @@ int mount_fs(const char* type, const char* uri, const char* mount_point, struct 
     // We need to fix up the relative path to this mount, but only for
     // directories.
     qstrsetstr(&dent->rel_path, "", 0);
-    mount->path.hash = dent->rel_path.hash;
 
     /*Now go ahead and do a lookup so the dentry is valid */
-    if ((ret = __path_lookupat(dentry_root, mount_point, 0, &dent2, 0, parent ? parent->fs : mount,
-                               make_ancestor)) < 0)
+    if ((ret = __path_lookupat(g_dentry_root, mount_point, 0, &dent2, 0,
+                               parent ? parent->fs : mount, make_ancestor)) < 0)
         goto out_with_unlock;
 
     assert(dent == dent2);
@@ -626,7 +625,7 @@ int mount_fs(const char* type, const char* uri, const char* mount_point, struct 
     }
 
 out_with_unlock:
-    unlock(&dcache_lock);
+    unlock(&g_dcache_lock);
     if (need_parent_put) {
         put_dentry(parent);
     }
--- a/LibOS/shim/src/fs/shim_namei.c
+++ b/LibOS/shim/src/fs/shim_namei.c
@@ -49,9 +49,9 @@ static inline int __lookup_flags(int flags) {
  *
  * Returns 0 on success, negative on failure.
  */
-/* Assume caller has acquired dcache_lock */
+/* Assume caller has acquired g_dcache_lock */
 int __permission(struct shim_dentry* dent, mode_t mask) {
-    assert(locked(&dcache_lock));
+    assert(locked(&g_dcache_lock));
 
     mode_t mode = 0;
 
@@ -109,10 +109,6 @@ int __permission(struct shim_dentry* dent, mode_t mask) {
  * The dentry is returned in pointer *new.  The refcount of new is incremented
  * by one.
  *
- * Parent can be null when mounting the root file system.  In this case, the
- * function creates a new, negative dentry, which will then be initialized by
- * the mount code and made non-negative.
- *
  * The fs argument specifies the file system type to use on a miss; typically,
  * this will be parent->fs.
  *
@@ -125,18 +121,19 @@ int __permission(struct shim_dentry* dent, mode_t mask) {
  * Returns zero if the file is found; -ENOENT if it doesn't exist,
  * possibly other errors.
  *
- * The caller should hold the dcache_lock.
+ * The caller should hold the g_dcache_lock.
  */
 int lookup_dentry(struct shim_dentry* parent, const char* name, int namelen,
                   struct shim_dentry** new, struct shim_mount* fs) {
-    assert(locked(&dcache_lock));
+    assert(locked(&g_dcache_lock));
+    assert(parent);
 
     struct shim_dentry* dent = NULL;
     int do_fs_lookup = 0;
     int err = 0;
 
     /* Look up the name in the dcache first, one atom at a time. */
-    dent = __lookup_dcache(parent, name, namelen, NULL);
+    dent = lookup_dcache(parent, name, namelen);
 
     if (!dent) {
         if (parent) {
@@ -149,7 +146,7 @@ int lookup_dentry(struct shim_dentry* parent, const char* name, int namelen,
             }
         }
 
-        dent = get_new_dentry(fs, parent, name, namelen, NULL);
+        dent = get_new_dentry(fs, parent, name, namelen);
         if (!dent) {
             err = -ENOMEM;
             goto out;
@@ -215,8 +212,8 @@ out:
  * are cases where start may be absent (e.g., in bootstrapping).  Fs can be
  * null only if the current thread is defined.
  *
- * Assumes dcache_lock is held; main difference from path_lookupat is that
- * dcache_lock is acquired/released.
+ * Assumes g_dcache_lock is held; main difference from path_lookupat is that
+ * g_dcache_lock is acquired/released.
  *
  * We assume the caller has incremented the reference count on start and its
  * associated file system mount by one
@@ -242,7 +239,7 @@ out:
 int __path_lookupat(struct shim_dentry* start, const char* path, int flags,
                     struct shim_dentry** dent, int link_depth, struct shim_mount* fs,
                     bool make_ancestor) {
-    assert(locked(&dcache_lock));
+    assert(locked(&g_dcache_lock));
     // Basic idea: recursively iterate over path, peeling off one atom at a
     // time.
     /* Chia-Che 12/5/2014:
@@ -278,7 +275,7 @@ int __path_lookupat(struct shim_dentry* start, const char* path, int flags,
             /* Start at the global root if we have no fs and no start dentry.
              * This should only happen as part of initialization.
              */
-            start = dentry_root;
+            start = g_dentry_root;
         }
         get_dentry(start);
         unlock(&g_process.fs_lock);
@@ -465,14 +462,14 @@ out:
     return err;
 }
 
-/* Just wraps __path_lookupat, but also acquires and releases the dcache_lock.
+/* Just wraps __path_lookupat, but also acquires and releases the g_dcache_lock.
  */
 int path_lookupat(struct shim_dentry* start, const char* name, int flags, struct shim_dentry** dent,
                   struct shim_mount* fs) {
     int ret = 0;
-    lock(&dcache_lock);
+    lock(&g_dcache_lock);
     ret = __path_lookupat(start, name, flags, dent, 0, fs, 0);
-    unlock(&dcache_lock);
+    unlock(&g_dcache_lock);
     return ret;
 }
 
@@ -502,7 +499,7 @@ int open_namei(struct shim_handle* hdl, struct shim_dentry* start, const char* p
         return -ENOENT;
     }
 
-    lock(&dcache_lock);
+    lock(&g_dcache_lock);
 
     // lookup the path from start, passing flags
     err = __path_lookupat(start, path, lookup_flags, &mydent, 0, NULL, 0);
@@ -586,7 +583,7 @@ out:
         put_dentry(mydent);
     }
 
-    unlock(&dcache_lock);
+    unlock(&g_dcache_lock);
 
     return err;
 }
@@ -707,14 +704,14 @@ static inline void set_dirent_type(mode_t* type, int d_type) {
 int list_directory_dentry(struct shim_dentry* dent) {
     int ret = 0;
     struct shim_mount* fs = dent->fs;
-    lock(&dcache_lock);
+    lock(&g_dcache_lock);
 
     /* DEP 8/4/17: Another process could list this directory
      * while we are waiting on the dcache lock.  This is ok,
      * no need to blow an assert.
      */
     if (dent->state & DENTRY_LISTED) {
-        unlock(&dcache_lock);
+        unlock(&g_dcache_lock);
         return 0;
     }
 
@@ -723,7 +720,7 @@ int list_directory_dentry(struct shim_dentry* dent) {
     // expect to learn is beyond me, but be careful with blowing assert
     // and tell the program something to keep it moving.
     if (dent->state & DENTRY_NEGATIVE) {
-        unlock(&dcache_lock);
+        unlock(&g_dcache_lock);
         return 0;
     }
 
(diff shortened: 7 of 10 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoHUzRBFUHXfXESVE4:-MYoHUzRBFUHXfXESVE5:bmca08g) (no related file):*
> Fun fact: the regression tests contain a dev binary which somehow gets mounted under /dev, alongside (???) the dev filesystem

Well, that's a hack/feature that is allowed by our current FS mount point semantics. It is legitimate to mount not only directories, but also separate files. It is also legit to mount a subdirectory of another mounter directory (its ancestor). Finally, it is legit to mount subdirectories even under pseudo-FSes like proc and dev. But this abusing of `/dev` can be removed if it is too hacky.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoIKl14TFfFEJLkXFU:-MYoIKl14TFfFEJLkXFV:b-exbyah) (no related file):*
Can you also rename `dcache_lock` -> `g_dcache_lock` and `dentry_root` -> `g_dentry_root` (to make them visibly global).

---
___
*[LibOS/shim/include/shim_fs.h, line 416 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoOtOQ84ky1lVApoMn:-MYoOtORE-I354MiP-ZE:balm5e3) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L416)):*
> ```objc
>  * `dcache_lock`.
>  *
>  * `parent` is the parent node, or NULL if this is supposed to be a root dentry.
> ```

By the way, why can't we enforce `parent != NULL` and pass `dentry_root` in root-dentry cases? 
___
*[LibOS/shim/include/shim_fs.h, line 418 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoIdyg13Xrl9r2Qwbv:-MYoIdyg13Xrl9r2Qwbw:bhteqf7) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L418)):*
> ```objc
>  * `parent` is the parent node, or NULL if this is supposed to be a root dentry.
>  *
>  * `fs` is the mountpoint the dentry is under, or NULL. This should typically be `parent->fs`, but
> ```

Could you please add a comment what happens if `fs == NULL`? Is the mountpoint taken from `parent->fs`?
___
*[LibOS/shim/include/shim_fs.h, line 419 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoJB6ZCXpj1sStKy3M:-MYoJB6_CP31fzgkJUTq:bwbghlq) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L419)):*
> ```objc
>  *
>  * `fs` is the mountpoint the dentry is under, or NULL. This should typically be `parent->fs`, but
>  * is passed explicitly for initializing the dentry of a mountpoint.
> ```

I'm confused. Do you mean:
- "`fs` is used to initialize to mountpoint field of the dentry" or
- "the dentry field of `fs` is initialized as a side-effect"
?
___
*[LibOS/shim/include/shim_fs.h, line 422 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoJ8ND63U2Owd1vVmT:-MYoJ8ND63U2Owd1vVmU:bk82yyk) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L422)):*
> ```objc
>  *
>  * Returns the new dentry, or NULL in case of allocation failure. The function will initialize
>  * `fs` (if provided), `name`, `rel_path`, and parent/children links. The reference count of the
> ```

What is the `rel_path`?
___
*[LibOS/shim/include/shim_fs.h, line 423 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoJb0l87dUQAgYM2wj:-MYoJb0l87dUQAgYM2wk:b-rvz4rc) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L423)):*
> ```objc
>  * Returns the new dentry, or NULL in case of allocation failure. The function will initialize
>  * `fs` (if provided), `name`, `rel_path`, and parent/children links. The reference count of the
>  * returned dentry will be incremented.
> ```

The old comment mentions 2 or 1, depending on whether there is a parent or not. This looks like an important detail to mention.
___
*[LibOS/shim/include/shim_fs.h, line 436 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoKvxc-nJ5iRM-0Uhp:-MYoKvxc-nJ5iRM-0Uhq:b-kaz55r) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L436)):*
> ```objc
> 
> /*
>  * Search for a child of a dentry with a given name. The caller should hold `dcache_lock`.
> ```

Accidentally copied a wrong comment here? You didn't change this function (and it recursively deletes the whole dentry tree).
___
*[LibOS/shim/src/fs/shim_dcache.c, line 142 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoNVBt2YPxB2EaQOZm:-MYoNVBt2YPxB2EaQOZn:b-7n3nbu) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L142)):*
> ```C
> }
> 
> static void maybe_delete_dentry(struct shim_dentry* dent) {
> ```

The name is misleading... Maybe `unlink_nonexisting_dentry_from_parent()`?

Well, my proposed name is also long, but at least very specific.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 164 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoN3ecBF9guPabY6GI:-MYoN3ed3qkdMwsd-Qne:bywld3l) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L164)):*
> ```C
>     if (count == 2) {
>         /* If the ref count is exactly 2, we are holding the last reference to dentry (except for parent
>          * or filesystem, in case of root). Try to delete it. Make sure we check the conditions while
> ```

`except for parent or filesystem, in case of root` -- I don't get this sentence. Looks like you're trying to say that we can have count = 2 in two cases:
- Last reference to non-root dentry: legitimate case, can delete and free
- Two independent references to root dentry: don't do anything, just put one reference

If I understood you correctly, please re-write the comment to be more precise about these two cases.
___
*[LibOS/shim/src/sys/shim_socket.c, line 428 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoKfTmEslQbKenTThZ:-MYoKfTnArDvcEwvyuVK:b-esd87v) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/sys/shim_socket.c#L428)):*
> ```C
> 
> static int hash_to_hex_string(struct shim_dentry* dent, char* buf, size_t size) {
>     char buffer[dentry_get_path_size(dent)];
> ```

I would suggest to put an assert on the maximum path size here. We're using the LibOS stack, it could overflow if the path is very long... I believe there is already a macro constant like this (`PATH_SIZE`?).

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: 4 of 10 files reviewed, 11 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoHUzRBFUHXfXESVE4:-MYoPvAl-dTJVKvskUEw:b2ci49h) (no related file):*

> Fun fact: the regression tests contain a dev binary which somehow gets mounted under /dev, alongside (???) the dev filesystem

Well, that's a hack/feature that is allowed by our current FS mount point semantics. It is legitimate to mount not only directories, but also separate files. It is also legit to mount a subdirectory of another mounter directory (its ancestor). Finally, it is legit to mount subdirectories even under pseudo-FSes like proc and dev. But this abusing of `/dev` can be removed if it is too hacky.
</blockquote></details>

Okay, so the `/dev` directory shadows the `dev` executable, but somehow we load the `dev` executable earlier, while it's still accessible...

I ran into it while trying to refactor the mount system, because in other cases it complains about duplicate mounts, so I assumed mounting under an existing file/directory should be disallowed. But apparently it's not always disallowed.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoIKl14TFfFEJLkXFU:-MYojC32-CSzzP_sJlmD:b-9h485d) (no related file):*

Can you also rename `dcache_lock` -> `g_dcache_lock` and `dentry_root` -> `g_dentry_root` (to make them visibly global).
</blockquote></details>

Done. But I don't want this PR to get too big, so let's not bundle more unrelated changes here. I'd rather follow up with next PRs quickly, this is already a part of a much bigger branch.

---
___
*[LibOS/shim/include/shim_fs.h, line 416 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoOtOQ84ky1lVApoMn:-MYoT6ElEsO48jcX6c2n:b4pinco) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L416)):*

By the way, why can't we enforce `parent != NULL` and pass `dentry_root` in root-dentry cases? 
</blockquote></details>

I'm not convinced that is better... If the root dentry has itself as parent, you trade all `if(dent->parent)` checks for `if (dent->parent != dent)`, but you still need to make the checks in all cases that use the parent (detaching entry from parent, computing the path, etc.)

(Also, if we're creating the root dentry, it doesn't exist yet, so wouldn't we need to bootstrap it with NULL anyway, and only then add the cyclic reference?)
___
*[LibOS/shim/include/shim_fs.h, line 418 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoIdyg13Xrl9r2Qwbv:-MYoUryf6_Nb_NGgmxyH:bpogyqp) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L418)):*

Could you please add a comment what happens if `fs == NULL`? Is the mountpoint taken from `parent->fs`?
</blockquote></details>

It defaults to null, I think taking it from`parent` would be too clever (too many ways of achieving the same thing). I added a comment.

Maybe I'll be able get rid of that `fs` parameter when rewriting mounts. Or split that off to another mount-making function.
___
*[LibOS/shim/include/shim_fs.h, line 419 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoJB6ZCXpj1sStKy3M:-MYoVs-VBzPtD8BzBFdR:b-i1f1oc) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L419)):*

I'm confused. Do you mean:
- "`fs` is used to initialize to mountpoint field of the dentry" or
- "the dentry field of `fs` is initialized as a side-effect"
?
</blockquote></details>

I reworded it to "to support initializing the root dentry of a newly mounted filesystem." I hope that's more clear.
___
*[LibOS/shim/include/shim_fs.h, line 422 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoJ8ND63U2Owd1vVmT:-MYoVy1Y8SSkyj7I7-PW:b-anddbf) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L422)):*

What is the `rel_path`?
</blockquote></details>

`rel_path` is path relative to dentry's mountpoint, as documented in `struct shim_dentry`. It's also a horrible mess - the mount code does this:

```c
    // We need to fix up the relative path to this mount, but only for
    // directories.
    qstrsetstr(&dent->rel_path, "", 0);
```

Because I do not want to modify the mount code yet, I'll document what happens here, and leave the behavior of `get_new_dentry` as-is.
___
*[LibOS/shim/include/shim_fs.h, line 423 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoJb0l87dUQAgYM2wj:-MYo_2ew2VBSmckwS2vb:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L423)):*

The old comment mentions 2 or 1, depending on whether there is a parent or not. This looks like an important detail to mention.
</blockquote></details>

Done.
___
*[LibOS/shim/include/shim_fs.h, line 436 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoKvxc-nJ5iRM-0Uhp:-MYo_bLD1emN-_drcVDC:blqu2zo) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L436)):*

Accidentally copied a wrong comment here? You didn't change this function (and it recursively deletes the whole dentry tree).
</blockquote></details>

Yeah, I don't want to touch that yet. Fixed.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 142 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoNVBt2YPxB2EaQOZm:-MYo_qr5EP_PWel1XN0V:b7brstq) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L142)):*

The name is misleading... Maybe `unlink_nonexisting_dentry_from_parent()`?

Well, my proposed name is also long, but at least very specific.
</blockquote></details>

I don't want to say "unlink", because it suggests unlinking an actual file from directory.

I changed it to `detach_dentry_if_possible` and added a better explanation of what happens.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 164 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoN3ecBF9guPabY6GI:-MYod4GW3ShFcB0Zc6xj:b-6jvb1r) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L164)):*

`except for parent or filesystem, in case of root` -- I don't get this sentence. Looks like you're trying to say that we can have count = 2 in two cases:
- Last reference to non-root dentry: legitimate case, can delete and free
- Two independent references to root dentry: don't do anything, just put one reference

If I understood you correctly, please re-write the comment to be more precise about these two cases.
</blockquote></details>

Is the explanation right now OK? I'm not talking about the dentry root, but about whether the dentry has a parent (which is what this function should care about).

I don't want to enumerate these 2 cases here, because (at least right now) there is also the `__del_dentry_tree` function that detaches parent manually. There might be other such cases in the future, and as long as the reference counts are maintained correctly, `put_dentry` should do the right thing.
___
*[LibOS/shim/src/sys/shim_socket.c, line 428 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoKfTmEslQbKenTThZ:-MYof0XiDp2Xty4h6jUJ:bblj79u) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/sys/shim_socket.c#L428)):*

I would suggest to put an assert on the maximum path size here. We're using the LibOS stack, it could overflow if the path is very long... I believe there is already a macro constant like this (`PATH_SIZE`?).
</blockquote></details>

No, apparently we don't enforce any limit on path length, only the `STR_MAX` limit on _relative_ path. Maybe we should have one (and probably we should be aware of `PATH_MAX` when handling system calls...)

So stack allocation is not okay, and an assert is not correct either, until we have such a limit in place. I changed this to malloc and added a comment.

yamahata (dismissed):
> Reviewed 4 of 9 files at r1, 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 12 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 187 at r2](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoxrEyAy7qm10fGoRK:-MYoxrEyAy7qm10fGoRL:buz6mo6) ([raw file](https://github.com/oscarlab/graphene/blob/67c6867ee4927df737292bb35abd46ba5e82e0fe/LibOS/shim/src/fs/shim_dcache.c#L187)):*
> ```C
>          * holding `g_dcache_lock`, as someone might have acquired a reference in meantime.
>          */
>         if (locked(&g_dcache_lock)) {
> ```

This doesn't seem right. What if other thread is holding this lock?
If we know this thread has this lock, it should be passed as argument.

anyway this PR is preparation for future PR, the original code is a sort of broken, whole rewrite is planned, I make this comment non-blocking.

dimakuv (changes requested):
> Reviewed 5 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoHUzRBFUHXfXESVE4:-MYoxicr10znluBInzHT:bzhr6c3) (no related file):*

Okay, so the `/dev` directory shadows the `dev` executable, but somehow we load the `dev` executable earlier, while it's still accessible...

I ran into it while trying to refactor the mount system, because in other cases it complains about duplicate mounts, so I assumed mounting under an existing file/directory should be disallowed. But apparently it's not always disallowed.
</blockquote></details>

Oops. I'm sorry, I misunderstood you. I thought you meant `/dev/kmsg` device that we mount in our LibOS regression tests.

Now I understand that you meant the `dev` executable (one of our tests). Because `dev` is located in the current dir, it automatically goes under root dir and becomes `/dev`. Yes, this is surprising and is a fun fact. I don't know how/why this happens.

---
___
*[LibOS/shim/include/shim_fs.h, line 416 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoOtOQ84ky1lVApoMn:-MYoy9RG5VDecAAR0n2S:bmctigb) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L416)):*

I'm not convinced that is better... If the root dentry has itself as parent, you trade all `if(dent->parent)` checks for `if (dent->parent != dent)`, but you still need to make the checks in all cases that use the parent (detaching entry from parent, computing the path, etc.)

(Also, if we're creating the root dentry, it doesn't exist yet, so wouldn't we need to bootstrap it with NULL anyway, and only then add the cyclic reference?)
</blockquote></details>

OK, true.
___
*[LibOS/shim/include/shim_fs.h, line 419 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoJB6ZCXpj1sStKy3M:-MYoyIWI-h1MSHY7pksn:b-j2lh2c) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L419)):*

I reworded it to "to support initializing the root dentry of a newly mounted filesystem." I hope that's more clear.
</blockquote></details>

OK, thanks. These semantics are just insane...
___
*[LibOS/shim/include/shim_fs.h, line 422 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoJ8ND63U2Owd1vVmT:-MYoyOQe2pSeKQx7QsIC:btm9clh) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L422)):*

`rel_path` is path relative to dentry's mountpoint, as documented in `struct shim_dentry`. It's also a horrible mess - the mount code does this:

```c
    // We need to fix up the relative path to this mount, but only for
    // directories.
    qstrsetstr(&dent->rel_path, "", 0);
```

Because I do not want to modify the mount code yet, I'll document what happens here, and leave the behavior of `get_new_dentry` as-is.
</blockquote></details>

OK. I wrote this comment because `rel_path` is not in the argument list of this function, so I thought you made a typo. Now I understand that you meant initializing `rel_path` of the dentry.
___
*[LibOS/shim/include/shim_fs.h, line 425 at r2](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoyh8T63EPK-LQ9kfV:-MYoyh8T63EPK-LQ9kfW:b8tg65m) ([raw file](https://github.com/oscarlab/graphene/blob/67c6867ee4927df737292bb35abd46ba5e82e0fe/LibOS/shim/include/shim_fs.h#L425)):*
> ```objc
>  * `fs` (if provided), `name`, `rel_path`, and parent/children links.
>  *
>  * The reference count of the returned dentry will 2 if `parent` was provided, 1 otherwise.
> ```

`will be`
___
*[LibOS/shim/src/fs/shim_dcache.c, line 164 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoN3ecBF9guPabY6GI:-MYoz7-b4KdNZOjPfCzG:bxa31k) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L164)):*

Is the explanation right now OK? I'm not talking about the dentry root, but about whether the dentry has a parent (which is what this function should care about).

I don't want to enumerate these 2 cases here, because (at least right now) there is also the `__del_dentry_tree` function that detaches parent manually. There might be other such cases in the future, and as long as the reference counts are maintained correctly, `put_dentry` should do the right thing.
</blockquote></details>

OK, thanks, the comments are very good now.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: 9 of 10 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoHUzRBFUHXfXESVE4:-MYpEJcY8yDntsANP3Pt:b46bk3b) (no related file):*

Oops. I'm sorry, I misunderstood you. I thought you meant `/dev/kmsg` device that we mount in our LibOS regression tests.

Now I understand that you meant the `dev` executable (one of our tests). Because `dev` is located in the current dir, it automatically goes under root dir and becomes `/dev`. Yes, this is surprising and is a fun fact. I don't know how/why this happens.
</blockquote></details>

I'll look into it when rewriting mounts, and probably rename the example for sanity.

And at some point I would like the default directory (where the executable is mounted, etc.) to not be `/`, but something like `/home/user`.

---
___
*[LibOS/shim/include/shim_fs.h, line 419 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoJB6ZCXpj1sStKy3M:-MYpCkIr9f9zeFffmyXr:b-l4xeyu) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L419)):*

OK, thanks. These semantics are just insane...
</blockquote></details>

It's going to get better :)
___
*[LibOS/shim/include/shim_fs.h, line 425 at r2](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoyh8T63EPK-LQ9kfV:-MYpEynD67eca7SWwe4L:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/67c6867ee4927df737292bb35abd46ba5e82e0fe/LibOS/shim/include/shim_fs.h#L425)):*

`will be`
</blockquote></details>

Fixed.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 187 at r2](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoxrEyAy7qm10fGoRK:-MYpCw9PDef4KopycA_y:btev8yx) ([raw file](https://github.com/oscarlab/graphene/blob/67c6867ee4927df737292bb35abd46ba5e82e0fe/LibOS/shim/src/fs/shim_dcache.c#L187)):*

This doesn't seem right. What if other thread is holding this lock?
If we know this thread has this lock, it should be passed as argument.

anyway this PR is preparation for future PR, the original code is a sort of broken, whole rewrite is planned, I make this comment non-blocking.
</blockquote></details>

`locked()` checks if our thread is holding the lock (`get_cur_tid() == lock->owner`), it's my understanding that it is accurate (otherwise the asserts that use it are not accurate either, and could sometimes pass when they shouldn't).

The point is that I'd like `put_dentry` to be usable in all circumstances, and be guaranteed to detach the dentry sooner or later. If this check is no good, there could be `put_dentry_with_lock()` and `put_dentry_without_lock()` and you're supposed to call the right one, but it sounds pretty inconvenient.

Or maybe there is some better way of garbage-collecting the dentries...

boryspoplawski (changes requested):
> Reviewed 4 of 9 files at r1, 5 of 6 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @pwmarcz, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoHUzRBFUHXfXESVE4:-MYpGhLABWXsN7kQpgiW:brevglv) (no related file):*

I'll look into it when rewriting mounts, and probably rename the example for sanity.

And at some point I would like the default directory (where the executable is mounted, etc.) to not be `/`, but something like `/home/user`.
</blockquote></details>

Imo there should be no default main-exec mounting. If you want to run something mount it explicitly. We should also get rid of this mounting host `.` as `/` completely.

---
___
*[LibOS/shim/include/shim_fs.h, line 389 at r2](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYpEubj7lLp8-lh2YJ4:-MYpEubk2knOScHHhR15:bllow90) ([raw file](https://github.com/oscarlab/graphene/blob/67c6867ee4927df737292bb35abd46ba5e82e0fe/LibOS/shim/include/shim_fs.h#L389)):*
> ```objc
>  * Decrement the reference count on `dent`, and delete it from the tree if possible (i.e. there are
>  * no references to it except parent, and it does not represent a real file (it's either invalid or
>  * negative).
> ```

Is Graphene really ready for this without other reworks?
E.g. what happens if you mount host file `a` on `/a` and then open+close it (so that `put_dentry` triggers on last reference)?
___
*[LibOS/shim/include/shim_fs.h, line 412 at r2](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYpFWH13CqBhYwCUNxR:-MYpFWH13CqBhYwCUNxS:b-u42v0a) ([raw file](https://github.com/oscarlab/graphene/blob/67c6867ee4927df737292bb35abd46ba5e82e0fe/LibOS/shim/include/shim_fs.h#L412)):*
> ```objc
> }
> 
> /*
> ```

Could you rewrite this to use sphinx syntax?
Actually ditto for all comments you've rewritten in `.h`.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 187 at r2](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoxrEyAy7qm10fGoRK:-MYpJTO51pmten0aZNFI:b-5yrbdt) ([raw file](https://github.com/oscarlab/graphene/blob/67c6867ee4927df737292bb35abd46ba5e82e0fe/LibOS/shim/src/fs/shim_dcache.c#L187)):*
Hard veto from my side.

`locked()` is only meant to be used inside `assert` in debug builds and nowhere else (e.g. spinlocks define such function only on debug builds).
> otherwise the asserts that use it are not accurate either, and could sometimes pass when they shouldn't

Yes, they sometimes can pass when they shouldn't. These asserts have some purpose tho:
1) there are no false-negatives, so if some trigger it means we have a bug (and given enough runs they should),
2) they document which locks a function needs.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 126 at r3](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYpIMRJ2qMUXoUa-gtK:-MYpIMRJ2qMUXoUa-gtL:b-t7jsjc) ([raw file](https://github.com/oscarlab/graphene/blob/0d8335b7ef5cf644ea6991b8ded3c7d2c2f58747/LibOS/shim/src/fs/shim_dcache.c#L126)):*
> ```C
> }
> 
> static int __put_dentry(struct shim_dentry* dent) {
> ```

Names starting with double underscore are reserved in C.
___
*[LibOS/shim/src/sys/shim_socket.c, line 427 at r3](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYpMDi9-okt1FRBaxcb:-MYpMDiA1jGcfQOj4swt:b-4ttp46) ([raw file](https://github.com/oscarlab/graphene/blob/0d8335b7ef5cf644ea6991b8ded3c7d2c2f58747/LibOS/shim/src/sys/shim_socket.c#L427)):*
> ```C
> }
> 
> static int hash_to_hex_string(struct shim_dentry* dent, char* buf, size_t size) {
> ```

Now this function name is wrong, isn't it?

yamahata (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 187 at r2](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoxrEyAy7qm10fGoRK:-MYpPYmo7tU-cJss_t9E:b-ntcb5t) ([raw file](https://github.com/oscarlab/graphene/blob/67c6867ee4927df737292bb35abd46ba5e82e0fe/LibOS/shim/src/fs/shim_dcache.c#L187)):*

Hard veto from my side.

`locked()` is only meant to be used inside `assert` in debug builds and nowhere else (e.g. spinlocks define such function only on debug builds).
> otherwise the asserts that use it are not accurate either, and could sometimes pass when they shouldn't

Yes, they sometimes can pass when they shouldn't. These asserts have some purpose tho:
1) there are no false-negatives, so if some trigger it means we have a bug (and given enough runs they should),
2) they document which locks a function needs.
</blockquote></details>

Ok, then it makes sense as long as lock owner is checked.

mkow (changes requested):
> Reviewed 9 of 9 files at r1, 1 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoHUzRBFUHXfXESVE4:-MYoiXeA8egr-kohkBeV:b-62wnk5) (no related file):*
> Well, that's a hack/feature that is allowed by our current FS mount point semantics.

I'd say it's a feature, I think on Linux you can also mount single files.

> But this abusing of /dev can be removed if it is too hacky.

In this case it's not abusing it, it's just a bug that this is possible.

Anyways, I see that you just misunderstood Paweł :)

> Imo there should be no default main-exec mounting. If you want to run something mount it explicitly. We should also get rid of this mounting host `.` as `/` completely.

+1, although the main-exec is not "mounted", it's just the current directory which is (I think).

---
___
*[LibOS/shim/src/fs/shim_dcache.c, line 127 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYqbKZP1jbQ5k4fjy7p:-MYqbKZP1jbQ5k4fjy7q:b-ib6du7) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L127)):*
> ```C
> 
> static int __put_dentry(struct shim_dentry* dent) {
>     int count = REF_DEC(dent->ref_count);
> ```

Huh, I just noticed that our ref counters are `int64_t`, but `REF_DEC` trims the counts to `int`. @pwmarcz Could you fix this in another PR? Looks simple.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 164 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYqby4HDR9iP3fx7_J8:-MYqby4HDR9iP3fx7_J9:b-n66qs2) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L164)):*
> ```C
>     if (count == 2) {
>         /* If the ref count is exactly 2, we are holding the last reference to dentry (except for parent
>          * or filesystem, in case of root). Try to delete it. Make sure we check the conditions while
> ```

Please rewrap
___
*[LibOS/shim/src/fs/shim_dcache.c, line 187 at r2](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoxrEyAy7qm10fGoRK:-MYqcVNOCBwVj3a2pOQb:b-rmlfvk) ([raw file](https://github.com/oscarlab/graphene/blob/67c6867ee4927df737292bb35abd46ba5e82e0fe/LibOS/shim/src/fs/shim_dcache.c#L187)):*

Ok, then it makes sense as long as lock owner is checked.
</blockquote></details>

I agree with Borys, `locked()` is racy and usable only in asserts.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 127 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYqbKZP1jbQ5k4fjy7p:-MYqx3jqAXulsUZ_m3UP:b49phva) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L127)):*

Huh, I just noticed that our ref counters are `int64_t`, but `REF_DEC` trims the counts to `int`. @pwmarcz Could you fix this in another PR? Looks simple.
</blockquote></details>

I've already done that in IPC PR (well, probably should have split that...)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 127 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYqbKZP1jbQ5k4fjy7p:-MYr3lht05tSVAbldvYD:b-lkgykz) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L127)):*

I've already done that in IPC PR (well, probably should have split that...)
</blockquote></details>

Great! But I'll leave this blocking, as rebasing may introduce (probably silent?) bugs here. Not sure if GCC would warn about such an assignment.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 127 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYqbKZP1jbQ5k4fjy7p:-MYr4rIAAkrgjMWmwYXE:bt8p1ky) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L127)):*

Great! But I'll leave this blocking, as rebasing may introduce (probably silent?) bugs here. Not sure if GCC would warn about such an assignment.
</blockquote></details>

Ah, no, I've just change the function signature/return value, but I've not touched every call site (which means probably every `put_something`).

dimakuv (approved):
> Reviewed 1 of 6 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 7 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: 6 of 10 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, @mkow, and @yamahata)

___
*[LibOS/shim/include/shim_fs.h, line 389 at r2](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYpEubj7lLp8-lh2YJ4:-MYtgjRY3fbeL9sPJrY8:b7k0d90) ([raw file](https://github.com/oscarlab/graphene/blob/67c6867ee4927df737292bb35abd46ba5e82e0fe/LibOS/shim/include/shim_fs.h#L389)):*

Is Graphene really ready for this without other reworks?
E.g. what happens if you mount host file `a` on `/a` and then open+close it (so that `put_dentry` triggers on last reference)?
</blockquote></details>

We are only deleting invalid and negative dentries, so in your example, nothing will happen. "Real" dentries are managed by their filesystem, so we would have to ask filesystem to release the data first. This is only a solution for "ephemeral" dentries such as result of a failed lookup.

 (I refactored this, but the point still stands).
___
*[LibOS/shim/include/shim_fs.h, line 412 at r2](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYpFWH13CqBhYwCUNxR:-MYthB_94-KyU1_HP4YM:b-o1qu32) ([raw file](https://github.com/oscarlab/graphene/blob/67c6867ee4927df737292bb35abd46ba5e82e0fe/LibOS/shim/include/shim_fs.h#L412)):*

Could you rewrite this to use sphinx syntax?
Actually ditto for all comments you've rewritten in `.h`.
</blockquote></details>

Sure, done. I'll keep that in mind when rewriting more functions.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 127 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYqbKZP1jbQ5k4fjy7p:-MYthFGJB9d7nyo8cObg:bi2jhpk) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L127)):*

Ah, no, I've just change the function signature/return value, but I've not touched every call site (which means probably every `put_something`).
</blockquote></details>

I changed `get_dentry` and `put_dentry` to use `int64_t`.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 164 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYqby4HDR9iP3fx7_J8:-MYtRN749sV8wmS7pL4R:bic92yp) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L164)):*

Please rewrap
</blockquote></details>

I rewrote this code, but what is wrong here? This comment is wrapped to 100 columns.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 187 at r2](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoxrEyAy7qm10fGoRK:-MYthZtC0lDIg4hi1QTV:bm7fo7b) ([raw file](https://github.com/oscarlab/graphene/blob/67c6867ee4927df737292bb35abd46ba5e82e0fe/LibOS/shim/src/fs/shim_dcache.c#L187)):*

I agree with Borys, `locked()` is racy and usable only in asserts.
</blockquote></details>

OK, got it. I am reverting this.

I don't want to make the user of `put_dentry` think whether they're holding a lock or not, though. A solution with two `put_dentry` functions (or a parametrized one) seems pretty inconvenient to me. So maybe it's not a good idea for reference counting to be too clever.

I think that a better approach (at least for now) is to delete ephemeral dentries during traversal. It doesn't guarantee we will drop them immediately, but performing many failed lookups will not slow down future operations. And it's a more internal mechanism now - the rest of filesystem code doesn't need to do anything.

(This will probably change later, if we decide that a linear directory lookup is too slow, or that we want to keep some negative dentries around, or purge non-negative dentries).
___
*[LibOS/shim/src/fs/shim_dcache.c, line 126 at r3](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYpIMRJ2qMUXoUa-gtK:-MYtPkqCD-GCl9Gh_FSI:b-fz45lv) ([raw file](https://github.com/oscarlab/graphene/blob/0d8335b7ef5cf644ea6991b8ded3c7d2c2f58747/LibOS/shim/src/fs/shim_dcache.c#L126)):*

Names starting with double underscore are reserved in C.
</blockquote></details>

This code is now gone, but this issue will come up often as Graphene makes heavy use of `__name` for internal/helper functions - any ideas for an alternative convention?
___
*[LibOS/shim/src/sys/shim_socket.c, line 427 at r3](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYpMDi9-okt1FRBaxcb:-MYtkJX20cKblwXffzDv:b-bvindf) ([raw file](https://github.com/oscarlab/graphene/blob/0d8335b7ef5cf644ea6991b8ded3c7d2c2f58747/LibOS/shim/src/sys/shim_socket.c#L427)):*

Now this function name is wrong, isn't it?
</blockquote></details>

Renamed to `hash_dentry_path`.

mkow (changes requested):
> Reviewed 1 of 6 files at r2, 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @pwmarcz)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 164 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYqby4HDR9iP3fx7_J8:-MYttpIT265xxqNYY4Mt:bj0h6ot) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L164)):*

I rewrote this code, but what is wrong here? This comment is wrapped to 100 columns.
</blockquote></details>

Hmm, I don't think it was. I have Reviewable set to show 100 chars per line and these two got wrapped.

These lines in r1 are:
```C
        /* If the ref count is exactly 2, we are holding the last reference to dentry (except for parent
         * or filesystem, in case of root). Try to delete it. Make sure we check the conditions while
```

The first one has 104 characters and the second 101. Maybe your editor has the count incorrectly set? Or maybe I was looking at a different revision than you?
___
*[LibOS/shim/src/fs/shim_dcache.c, line 126 at r3](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYpIMRJ2qMUXoUa-gtK:-MYtspMD0wp6xA6Yqb2s:bqulh24) ([raw file](https://github.com/oscarlab/graphene/blob/0d8335b7ef5cf644ea6991b8ded3c7d2c2f58747/LibOS/shim/src/fs/shim_dcache.c#L126)):*

This code is now gone, but this issue will come up often as Graphene makes heavy use of `__name` for internal/helper functions - any ideas for an alternative convention?
</blockquote></details>

Maybe a single underscore? It's not reserved if the second char is lowercase.
___
*[LibOS/shim/src/sys/shim_socket.c, line 436 at r4](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYtuvyV0DUkO0sEEy41:-MYtuvyV0DUkO0sEEy42:b-rcd3t4) ([raw file](https://github.com/oscarlab/graphene/blob/6bff9e6921538f2dfe060a6cff778e773bcfa624/LibOS/shim/src/sys/shim_socket.c#L436)):*
> ```C
> 
>     /* TODO: this could be a stack allocation, but LibOS full paths do not respect any limit (only
>      * mount-relative paths do). */
> ```

Is this really something we want to change? I'm fine with internal support for unlimited paths and only filtering them where original Linux interface does.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: 9 of 10 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 164 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYqby4HDR9iP3fx7_J8:-MYuWPHN2kwhyVKGs_SB:bkp9c0h) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L164)):*

Hmm, I don't think it was. I have Reviewable set to show 100 chars per line and these two got wrapped.

These lines in r1 are:
```C
        /* If the ref count is exactly 2, we are holding the last reference to dentry (except for parent
         * or filesystem, in case of root). Try to delete it. Make sure we check the conditions while
```

The first one has 104 characters and the second 101. Maybe your editor has the count incorrectly set? Or maybe I was looking at a different revision than you?
</blockquote></details>

I see, we're talking about different revisions. Reviewable showed me your comment over the code for r2 (which is properly wrapped), but there is a small "r1" in the corner. Next time something like that happens, I'll know where to look :)

(I did post r2 before your comment, though, judging by the timestamps.)
___
*[LibOS/shim/src/fs/shim_dcache.c, line 126 at r3](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYpIMRJ2qMUXoUa-gtK:-MYuVzCkEtudD7mPeds7:b-k93gqi) ([raw file](https://github.com/oscarlab/graphene/blob/0d8335b7ef5cf644ea6991b8ded3c7d2c2f58747/LibOS/shim/src/fs/shim_dcache.c#L126)):*

Maybe a single underscore? It's not reserved if the second char is lowercase.
</blockquote></details>

Oh, that's good to know.
___
*[LibOS/shim/src/sys/shim_socket.c, line 436 at r4](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYtuvyV0DUkO0sEEy41:-MYuY7P48PUoK8mWEoyP:b-rajwva) ([raw file](https://github.com/oscarlab/graphene/blob/6bff9e6921538f2dfe060a6cff778e773bcfa624/LibOS/shim/src/sys/shim_socket.c#L436)):*

Is this really something we want to change? I'm fine with internal support for unlimited paths and only filtering them where original Linux interface does.
</blockquote></details>

I don't know. There are a few places where we allocate a buffer on stack, but they can be fixed to also use malloc (or even trim, in case of debugging output). @boryspoplawski @dimakuv What do you think?

And by the way, the same question applies to mount-relative path. Currently it has a limit of `STR_SIZE = 4096`. If it's better to construct it on-demand than store in the dentries, should it still have a limit?

(And for now, I'll remove this TODO. If this is not an obvious thing, then there's no point keeping it in a comment.)

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 164 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYqby4HDR9iP3fx7_J8:-MYuawwA3oURra3NFNG7:ba80wni) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/src/fs/shim_dcache.c#L164)):*

I see, we're talking about different revisions. Reviewable showed me your comment over the code for r2 (which is properly wrapped), but there is a small "r1" in the corner. Next time something like that happens, I'll know where to look :)

(I did post r2 before your comment, though, judging by the timestamps.)
</blockquote></details>

Ah, seems we had a race condition, I didn't see that you pushed an update while I was reviewing :)
___
*[LibOS/shim/src/sys/shim_socket.c, line 436 at r4](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYtuvyV0DUkO0sEEy41:-MYubA6FEVbSPR831YVo:b-w1najs) ([raw file](https://github.com/oscarlab/graphene/blob/6bff9e6921538f2dfe060a6cff778e773bcfa624/LibOS/shim/src/sys/shim_socket.c#L436)):*

I don't know. There are a few places where we allocate a buffer on stack, but they can be fixed to also use malloc (or even trim, in case of debugging output). @boryspoplawski @dimakuv What do you think?

And by the way, the same question applies to mount-relative path. Currently it has a limit of `STR_SIZE = 4096`. If it's better to construct it on-demand than store in the dentries, should it still have a limit?

(And for now, I'll remove this TODO. If this is not an obvious thing, then there's no point keeping it in a comment.)
</blockquote></details>

I'd say we should keep such things on heap. 4KB is quite a lot, and it's easy to spin out of control and get stack overflow, especially in non-main threads, which has smaller and non-growable stacks.

boryspoplawski (changes requested):
> Reviewed 3 of 4 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_fs.h, line 423 at r1](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoJb0l87dUQAgYM2wj:-MYulVp-2h0AWFHKtSdp:boui2t) ([raw file](https://github.com/oscarlab/graphene/blob/feae2d7794a9c75ea5e6814d5745416fec78756e/LibOS/shim/include/shim_fs.h#L423)):*

Done.
</blockquote></details>

TBH that doesn't sound important, every reference is counted, so it's obvious that if `parent` got one, then the ref count was increased as well. The only thing that may be worth mentioning (but might be obvious too, idk) is that we pass the ownership of the returned pointer to the caller (i.e. it's a new pointer which caller will be now owner of).
(not blocking - the current version has all information needed as well)
___
*[LibOS/shim/include/shim_fs.h, line 438 at r5](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYukMsj-dTYrLGK80D4:-MYukMsj-dTYrLGK80D5:b-5u2qw9) ([raw file](https://github.com/oscarlab/graphene/blob/0e7ad02c6ed35f1e3cd36b56a49a979dc415f92d/LibOS/shim/include/shim_fs.h#L438)):*
> ```objc
> /*!
>  * \brief Allocate and initialize a new dentry for path name, under parent. The caller should hold
>  * `g_dcache_lock`.
> ```

I think info about locks fits better in general description, not `brief`.
___
*[LibOS/shim/include/shim_fs.h, line 447 at r5](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYuksSz6cB0YDF91B9V:-MYuksT-7yp38GVbMrCN:b-13z9md) ([raw file](https://github.com/oscarlab/graphene/blob/0e7ad02c6ed35f1e3cd36b56a49a979dc415f92d/LibOS/shim/include/shim_fs.h#L447)):*
> ```objc
>  * \returns the new dentry, or NULL in case of allocation failure
>  *
>  * The function will initialize `fs` (if provided), `name`, `rel_path`, and parent/children links.
> ```

This is confusing (at least to me): you mean `fs` (as in pointer passed to this function) or `dent->fs`? (I guess the latter)
___
*[LibOS/shim/src/fs/shim_dcache.c, line 126 at r3](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYpIMRJ2qMUXoUa-gtK:-MYur5xpAEHNahr02Zyr:bvg46fs) ([raw file](https://github.com/oscarlab/graphene/blob/0d8335b7ef5cf644ea6991b8ded3c7d2c2f58747/LibOS/shim/src/fs/shim_dcache.c#L126)):*

Oh, that's good to know.
</blockquote></details>

Yeah, in the new code we tend to use one underscore.
I personally add a leading underscore only if we have 2 functions, doing basically the same thing and one of them assumes that a lock is taken (with `_`) and the other that it's not taken (the one without `_`).
___
*[LibOS/shim/src/fs/shim_dcache.c, line 175 at r5](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYuusYjDCvbwMDZZdHX:-MYuusYjDCvbwMDZZdHY:b-y4j50z) ([raw file](https://github.com/oscarlab/graphene/blob/0e7ad02c6ed35f1e3cd36b56a49a979dc415f92d/LibOS/shim/src/fs/shim_dcache.c#L175)):*
> ```C
>     if (parent) {
>         /* Increment both dentries' ref counts, because they will be linked (through `dent->parent`
>          * and `parent->children`) */
> ```

NP: if you would move each `get_dentry` after corresponding copy, then this comment would be obvious and unnecessary.
___
*[LibOS/shim/src/sys/shim_socket.c, line 436 at r4](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYtuvyV0DUkO0sEEy41:-MYuiPPl19OdUrHx4SYw:b-56gvkq) ([raw file](https://github.com/oscarlab/graphene/blob/6bff9e6921538f2dfe060a6cff778e773bcfa624/LibOS/shim/src/sys/shim_socket.c#L436)):*

I'd say we should keep such things on heap. 4KB is quite a lot, and it's easy to spin out of control and get stack overflow, especially in non-main threads, which has smaller and non-growable stacks.
</blockquote></details>

Even if we had any limits, allocating 4kb on stack is not a good thing to do (as @mkow mentioned above). Also in most cases these paths are small (or at least well below 4k), so no point in storing such huge buffers.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/include/shim_fs.h, line 438 at r5](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYukMsj-dTYrLGK80D4:-MYx_lNxBf4wCoqOBO7w:b-58vy0b) ([raw file](https://github.com/oscarlab/graphene/blob/0e7ad02c6ed35f1e3cd36b56a49a979dc415f92d/LibOS/shim/include/shim_fs.h#L438)):*

I think info about locks fits better in general description, not `brief`.
</blockquote></details>

OK, moved to description.
___
*[LibOS/shim/include/shim_fs.h, line 447 at r5](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYuksSz6cB0YDF91B9V:-MYx_yLp6Wd1n4Sg5aII:b-dif9sb) ([raw file](https://github.com/oscarlab/graphene/blob/0e7ad02c6ed35f1e3cd36b56a49a979dc415f92d/LibOS/shim/include/shim_fs.h#L447)):*

This is confusing (at least to me): you mean `fs` (as in pointer passed to this function) or `dent->fs`? (I guess the latter)
</blockquote></details>

Reworded to make it clear it's about fields.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 175 at r5](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYuusYjDCvbwMDZZdHX:-MYxaV2H1daeNMtl1kHe:bk95eja) ([raw file](https://github.com/oscarlab/graphene/blob/0e7ad02c6ed35f1e3cd36b56a49a979dc415f92d/LibOS/shim/src/fs/shim_dcache.c#L175)):*

NP: if you would move each `get_dentry` after corresponding copy, then this comment would be obvious and unnecessary.
</blockquote></details>

Right, I moved the `get_dentry`s around so that it's clear.

dimakuv (approved):
> Reviewed 1 of 4 files at r4, 1 of 1 files at r5, 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/include/shim_fs.h, line 415 at r6](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYyPxMf3PXu_ZJ_r1Nx:-MYyPxMf3PXu_ZJ_r1Ny:b-flpoi3) ([raw file](https://github.com/oscarlab/graphene/blob/2c875af06ffe36f8ca8a98b032cb1334257f5797/LibOS/shim/include/shim_fs.h#L415)):*
> ```objc
>  * \endcode
>  */
> void dentry_gc(struct shim_dentry* dent);
> ```

This is cool! Nice idea.
___
*[LibOS/shim/src/sys/shim_socket.c, line 436 at r4](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYtuvyV0DUkO0sEEy41:-MYyPn5C1ncP__d3KhVY:b-5hegyc) ([raw file](https://github.com/oscarlab/graphene/blob/6bff9e6921538f2dfe060a6cff778e773bcfa624/LibOS/shim/src/sys/shim_socket.c#L436)):*

Even if we had any limits, allocating 4kb on stack is not a good thing to do (as @mkow mentioned above). Also in most cases these paths are small (or at least well below 4k), so no point in storing such huge buffers.
</blockquote></details>

I'm absolutely fine with mallocs everywhere and having unlimited paths.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYzUz0J-PWfJ52I4bkw:-MYzUz0K92t4pbUFwMaI:byvlmnj) (no related file):*
By the way, I'm experimenting with visualising the dentry tree. I'll submit some code with the next PR. Here is an example state after running `./pal_loader busybox ls /lib`:

```
[P64909:T1:busybox] [chroot V..S 10] * /
[P64909:T1:busybox] [  proc V..S  1] *   proc/
[P64909:T1:busybox] [   dev V..S  3] *   dev/
[P64909:T1:busybox] [chroot V.IS  3] *     tty
[P64909:T1:busybox] [   sys V..S  1] *   sys/
[P64909:T1:busybox] [chroot VLIS 26] *   lib/
[P64909:T1:busybox] [chroot V.IS  1] *     x86_64-linux-gnu/
[P64909:T1:busybox] [chroot V.I.  3]       ld-linux-x86-64.so.2
[P64909:T1:busybox] [chroot V.I.  2]       libm.so.6
[P64909:T1:busybox] [chroot V.I.  2]       libresolv.so.2
[P64909:T1:busybox] [chroot V.I.  2]       libc.so.6
[P64909:T1:busybox] [chroot V.I.  1]       pal-Linux-SGX
[P64909:T1:busybox] [chroot V.I.  1]       libnss_dns.so.2
[P64909:T1:busybox] [chroot V.I.  1]       Makefile
[P64909:T1:busybox] [chroot V.I.  1]       meson.build
[P64909:T1:busybox] [chroot V.I.  1]       libthread_db.so.1
[P64909:T1:busybox] [chroot V.I.  1]       libdl.so.2
[P64909:T1:busybox] [chroot V.I.  1]       libutil.so.1
[P64909:T1:busybox] [chroot V.I.  1]       pal-Linux
[P64909:T1:busybox] [chroot V.I.  1]       libpal-Linux.so
[P64909:T1:busybox] [chroot V.I.  1]       librt.so.1
[P64909:T1:busybox] [chroot V.I.  1]       libpal-Linux-SGX.so
[P64909:T1:busybox] [chroot V.I.  1]       crtn.o
[P64909:T1:busybox] [chroot V.I.  1]       crt1.o
[P64909:T1:busybox] [chroot V.I.  1]       libsysdb.a
[P64909:T1:busybox] [chroot V.I.  1]       libsysdb.so
[P64909:T1:busybox] [chroot V.I.  1]       libc.so
[P64909:T1:busybox] [chroot V.I.  1]       crti.o
[P64909:T1:busybox] [chroot V.I.  1]       libpthread.so.0
[P64909:T1:busybox] [chroot V...  1]       .gitignore
[P64909:T1:busybox] [chroot V.I.  1]       pal_loader
[P64909:T1:busybox] [chroot V.IS  2] *   etc/
[P64909:T1:busybox] [chroot V...  1]  -    ld.so.preload
[P64909:T1:busybox] [chroot V...  2]     busybox
```

(`*` is mountpoint, `-` is negative, remaining flags are `V`alid, `L`isted, `I`no updated, `S`ynthetic)

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYzUz0J-PWfJ52I4bkw:-MZ-ScfBAo5tRPbkMHp_:b-ed0vma) (no related file):*

By the way, I'm experimenting with visualising the dentry tree. I'll submit some code with the next PR. Here is an example state after running `./pal_loader busybox ls /lib`:

```
[P64909:T1:busybox] [chroot V..S 10] * /
[P64909:T1:busybox] [  proc V..S  1] *   proc/
[P64909:T1:busybox] [   dev V..S  3] *   dev/
[P64909:T1:busybox] [chroot V.IS  3] *     tty
[P64909:T1:busybox] [   sys V..S  1] *   sys/
[P64909:T1:busybox] [chroot VLIS 26] *   lib/
[P64909:T1:busybox] [chroot V.IS  1] *     x86_64-linux-gnu/
[P64909:T1:busybox] [chroot V.I.  3]       ld-linux-x86-64.so.2
[P64909:T1:busybox] [chroot V.I.  2]       libm.so.6
[P64909:T1:busybox] [chroot V.I.  2]       libresolv.so.2
[P64909:T1:busybox] [chroot V.I.  2]       libc.so.6
[P64909:T1:busybox] [chroot V.I.  1]       pal-Linux-SGX
[P64909:T1:busybox] [chroot V.I.  1]       libnss_dns.so.2
[P64909:T1:busybox] [chroot V.I.  1]       Makefile
[P64909:T1:busybox] [chroot V.I.  1]       meson.build
[P64909:T1:busybox] [chroot V.I.  1]       libthread_db.so.1
[P64909:T1:busybox] [chroot V.I.  1]       libdl.so.2
[P64909:T1:busybox] [chroot V.I.  1]       libutil.so.1
[P64909:T1:busybox] [chroot V.I.  1]       pal-Linux
[P64909:T1:busybox] [chroot V.I.  1]       libpal-Linux.so
[P64909:T1:busybox] [chroot V.I.  1]       librt.so.1
[P64909:T1:busybox] [chroot V.I.  1]       libpal-Linux-SGX.so
[P64909:T1:busybox] [chroot V.I.  1]       crtn.o
[P64909:T1:busybox] [chroot V.I.  1]       crt1.o
[P64909:T1:busybox] [chroot V.I.  1]       libsysdb.a
[P64909:T1:busybox] [chroot V.I.  1]       libsysdb.so
[P64909:T1:busybox] [chroot V.I.  1]       libc.so
[P64909:T1:busybox] [chroot V.I.  1]       crti.o
[P64909:T1:busybox] [chroot V.I.  1]       libpthread.so.0
[P64909:T1:busybox] [chroot V...  1]       .gitignore
[P64909:T1:busybox] [chroot V.I.  1]       pal_loader
[P64909:T1:busybox] [chroot V.IS  2] *   etc/
[P64909:T1:busybox] [chroot V...  1]  -    ld.so.preload
[P64909:T1:busybox] [chroot V...  2]     busybox
```

(`*` is mountpoint, `-` is negative, remaining flags are `V`alid, `L`isted, `I`no updated, `S`ynthetic)
</blockquote></details>

Wow, that looks awesome! (and ofc very useful!)

---

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> Reviewed 1 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: :shipit: complete! all files reviewed, all discussions resolved

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: :shipit: complete! all files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoHUzRBFUHXfXESVE4:-MZYaER-9f0UXR5aRvkl:bktk0ai) (no related file):*

> Well, that's a hack/feature that is allowed by our current FS mount point semantics.

I'd say it's a feature, I think on Linux you can also mount single files.

> But this abusing of /dev can be removed if it is too hacky.

In this case it's not abusing it, it's just a bug that this is possible.

Anyways, I see that you just misunderstood Paweł :)

> Imo there should be no default main-exec mounting. If you want to run something mount it explicitly. We should also get rid of this mounting host `.` as `/` completely.

+1, although the main-exec is not "mounted", it's just the current directory which is (I think).
</blockquote></details>

Several branches later, I finally hit that again... It looks like this worked because a dentry for executable was created sort of "on the side", not in the actual dentry tree.

The gory details: https://github.com/oscarlab/graphene/pull/2342/commits/441ef9cfd8e881ad507d608031b2aa851db40788

(part of upcoming PR #2342)

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2324)* status: :shipit: complete! all files reviewed, all discussions resolved

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2324#-MYoHUzRBFUHXfXESVE4:-MZZzyXz9VffmZXadHW_:bg2mzue) (no related file):*

Several branches later, I finally hit that again... It looks like this worked because a dentry for executable was created sort of "on the side", not in the actual dentry tree.

The gory details: https://github.com/oscarlab/graphene/pull/2342/commits/441ef9cfd8e881ad507d608031b2aa851db40788

(part of upcoming PR #2342)
</blockquote></details>

yay, finally fixed! this mess was really annoying :)

---

PR #2333, opened 2021-04-26
[LibOS] Rewrite path lookup functions

pwmarcz:
This is the next step in FS rewrite, after #2324 (the first commit should disappear after merging #2324).

See also the issue for dentry cleanup (#2321).

Initially, I wanted to reverse the "follow symlinks" semantics (`LOOKUP_FOLLOW` -> `LOOKUP_NO_FOLLOW`). I decided against that after seeing Linux and BSD have a flag like that, but found that FreeBSD has a convention to *also* use `NOFOLLOW` (defined as zero) to make the code clearer.

Apart from that (and general cleanup), the most important change is `LOOKUP_CREATE` and not returning `-ENOENT` when the file can be created.

Next step: rewrite `mount_fs` and make mount semantics more sane (as mentioned in #2324).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -102,7 +102,7 @@ struct shim_fs_ops {
 //#define DENTRY_UNREACHABLE  0x0800  /* permission checked to be unreachable */
 #define DENTRY_LISTED      0x1000 /* children in directory listed */
 #define DENTRY_INO_UPDATED 0x2000 /* ino updated */
-#define DENTRY_ANCESTOR    0x4000 /* Auto-generated dentry to connect a mount point in the        \
+#define DENTRY_SYNTHETIC   0x4000 /* Auto-generated dentry to connect a mount point in the        \
                                    * manifest to the root, when one or more intermediate          \
                                    * directories do not exist on the underlying FS. The semantics \
                                    * of subsequent changes to such directories (or attempts to    \
@@ -217,12 +217,10 @@ struct shim_mount {
     LIST_TYPE(shim_mount) list;
 };
 
+/* TODO: This actually does not get migrated after a fork. We migrate `g_process.root`, which is
+ * enough for Graphene to function, but leaves `g_dentry_root` in child process pointing to an empty
+ * directory. */
 extern struct shim_dentry* g_dentry_root;
-
-#define LOOKUP_FOLLOW    001
-#define LOOKUP_DIRECTORY 002
-#define LOOKUP_CONTINUE  004  // No longer needed
-#define LOOKUP_PARENT    010  // Not sure we need this
 
 #define F_OK 0
 // XXX: Duplicate definition; should probably weed out includes of host system
@@ -242,28 +240,6 @@ extern struct shim_dentry* g_dentry_root;
 #define ACC_MODE(x)                                        \
     ((((x) == O_RDONLY || (x) == O_RDWR) ? MAY_READ : 0) | \
      (((x) == O_WRONLY || (x) == O_RDWR) ? MAY_WRITE : 0))
-
-#define LOOKUP_OPEN   0100  // Appears to be ignored
-#define LOOKUP_CREATE 0200
-#define LOOKUP_ACCESS 0400  // Appears to be ignored
-#define LOOKUP_SYNC   (LOOKUP_OPEN | LOOKUP_CREATE | LOOKUP_ACCESS)
-
-enum lookup_type {
-    LAST_NORM,
-    LAST_ROOT,
-    LAST_DOT,
-    LAST_DOTDOT,
-    LAST_BIND
-};
-
-struct lookup {
-    struct shim_dentry* dentry;
-    struct shim_mount* mount;
-    const char* last;
-    int depth;
-    int flags;
-    enum lookup_type last_type;
-};
 
 /* initialization for fs and mounts */
 int init_fs(void);
@@ -297,41 +273,99 @@ int init_dcache(void);
 
 extern struct shim_lock g_dcache_lock;
 
-/* Checks permission (specified by mask) of a dentry. If force is not set, permission is considered
- * granted on invalid dentries.
- * Assumes that caller has acquired g_dcache_lock. */
-int __permission(struct shim_dentry* dent, mode_t mask);
-
-/* This function looks up a single dentry based on its parent dentry pointer and the name. `namelen`
- * is the length of char* name. The dentry is returned in pointer *new.
+/*!
+ * \brief Dump dentry cache
  *
- * The caller should hold the g_dcache_lock.
+ * \param dent the starting dentry, or NULL (will default to dentry root)
+ *
+ * Dumps the dentry cache using `log_always`, starting from the provided dentry. Intended for
+ * debugging the filesystem - just add it manually to the code.
  */
-int lookup_dentry(struct shim_dentry* parent, const char* name, int namelen,
-                  struct shim_dentry** new, struct shim_mount* fs);
+void dump_dcache(struct shim_dentry* dent);
 
-/* Looks up path under start dentry. Saves in dent.
+/*!
+ * \brief Check file permissions, similar to Unix access
  *
- * Assumes g_dcache_lock is held; main difference from path_lookupat is that g_dcache_lock is not
- * released on return.
+ * \param dentry the dentry to check
+ * \param mask mask, same as for Unix access
  *
- * The refcount is raised by one on the returned dentry.
+ * Checks permissions for a dentry. Because Graphene currently has no notion of users, this will
+ * always use the "user" part of file mode.
  *
- * The make_ancestor flag creates pseudo-dentries for any parent paths that are not in cache and do
- * not exist on the underlying file system. This is intended for use only in setting up the
- * file system view specified in the manifest.
+ * The caller should hold `g_dcache_lock`.
  *
- * If the file isnt' found, returns -ENOENT.
- *
- * If the LOOKUP_DIRECTORY flag is set, and the found file isn't a directory, returns -ENOTDIR.
+ * `dentry` should be a valid dentry, but can be negative (in which case the function will return
+ * -ENOENT).
  */
-int __path_lookupat(struct shim_dentry* start, const char* path, int flags,
-                    struct shim_dentry** dent, int link_depth, struct shim_mount* fs,
-                    bool make_ancestor);
+int check_permissions(struct shim_dentry* dent, mode_t mask);
 
-/* Just wraps __path_lookupat, but also acquires and releases the g_dcache_lock. */
-int path_lookupat(struct shim_dentry* start, const char* name, int flags, struct shim_dentry** dent,
-                  struct shim_mount* fs);
+/*
+ * Flags for `path_lookupat`.
+ *
+ * Note that, opposite to user-level O_NOFOLLOW, we define LOOKUP_FOLLOW as a positive flag, and add
+ * LOOKUP_NO_FOLLOW as a pseudo-flag for readability.
+ *
+ * This is modeled after Linux and BSD codebases, which define a positive FOLLOW flag, and a
+ * negative pseudo-flag was introduced by FreeBSD.
+ */
+#define LOOKUP_NO_FOLLOW       0
+#define LOOKUP_FOLLOW          0x1
+#define LOOKUP_CREATE          0x2
+#define LOOKUP_DIRECTORY       0x4
+#define LOOKUP_MAKE_SYNTHETIC  0x8
+
+/* Maximum number of nested symlinks that `path_lookupat` and related functions will follow */
+#define MAX_LINK_DEPTH 8
+
+/*!
+ * \brief Look up a path, retrieving a dentry
+ *
+ * \param start the start dentry for relative paths, or NULL (in which case it will default to
+ * process' cwd)
+ * \param path the path to look up
+ * \param flags lookup flags (see description below)
+ * \param[out] found pointer to retrieved dentry
+ *
+ * The caller should hold `g_dcache_lock`.
+ *
+ * On success, returns 0, and puts the retrieved dentry in `*found`. The reference count of the
+ * dentry will be increased by one.
+ *
+ * The retrieved dentry is always valid, and can only be negative if LOOKUP_CREATE is set.
+ *
+ * On failure, returns a negative error code, and sets `*found` to NULL.
+ *
+ * Supports the following flags:
+ *
+ * - LOOKUP_FOLLOW: if `path` refers to a symbolic link, follow it (the default is to return the
+ *   dentry to the link). Note that symbolic links for intermediate path segments are always
+ *   followed.
+ *
+ * - LOOKUP_NO_FOLLOW: this is a pseudo-flag defined as 0. You can use it to indicate to the reader
+ *   that symbolic links are intentionally not being followed.
+ *
+ * - LOOKUP_CREATE: if the file under `path` does not exist, but can be created (i.e. the parent
+ *   directory exists), the function will succeed and a negative dentry will be put in `*found`. If
+ *   the parent directory also does not exist, the function will still fail with -ENOENT.
+ *
+ * - LOOKUP_DIRECTORY: expect the file under `path` to be a directory, and fail with -ENOTDIR
+ *   otherwise
+ *
+ * - LOOKUP_MAKE_SYNTHETIC: create pseudo-files (DENTRY_SYNTHETIC) for any components on the path
+ *   that do not exist. This is intended for use when creating mountpoints specified in manifest.
+ *
+ * Note that a path with trailing slash is always treated as a directory, and LOOKUP_FOLLOW /
+ * LOOKUP_CREATE do not apply.
+ *
+ * TODO: This function doesn't check any permissions. It should return -EACCES on inaccessible
+ * directories.
+ */
+int _path_lookupat(struct shim_dentry* start, const char* path, int flags,
+                   struct shim_dentry** found);
+
+/* Same as path_lookupat, but also acquires and releases  `g_dcache_lock`. */
+int path_lookupat(struct shim_dentry* start, const char* path, int flags,
+                  struct shim_dentry** found);
 
 /*
  * This function returns a dentry (in *dir) from a handle corresponding to dirfd.
@@ -343,24 +377,53 @@ int path_lookupat(struct shim_dentry* start, const char* name, int flags, struct
  */
 int get_dirfd_dentry(int dirfd, struct shim_dentry** dir);
 
-/* Open path with given flags, in mode, similar to Unix open.
+/*
+ * \brief Open a file under a given path, similar to Unix open
  *
- * The start dentry specifies where to begin the search.
- * `hdl` is an optional argument; if passed in, it is initialized to refer to the opened path.
+ * \param[out] hdl handle to associate with dentry, can be NULL
+ * \param start the start dentry for relative paths, or NULL (in which case it will default to
+ * process' cwd)
+ * \param path the path to look up
+ * \param flags Unix open flags (see below)
+ * \param mode Unix file mode, used when creating a new file/directory
+ * \param[out] found pointer to retrieved dentry, can be NULL
  *
- * The result is stored in `dent`.
+ * If `hdl` is provided, on success it will be associated with the dentry. Otherwise, the file will
+ * just be retrieved or created.
+ *
+ * If `found` is provided, on success it will be set to the file's dentry (and its reference count
+ * will be increased), and on failure it will be set to NULL.
+ *
+ * Similar to Unix open, `flags` must include one of O_RDONLY, O_WRONLY or O_RDWR. In addition,
+ * the following flags are supported by this function:
+ * - O_CREAT: create a new file if one does not exist
+ * - O_EXCL: fail if the file already exists
+ * - O_DIRECTORY: expect/create a directory instead of regular file
+ * - O_NOFOLLOW: don't follow symbolic links when resolving a path
+ * - O_TRUNC: truncate the file after opening
+ *
+ * The flags (including any not listed above), as well as file mode, are passed to the underlying
+ * filesystem.
+ *
+ * TODO: This function checks permissions of the opened file (if it exists) and parent directory (if
+ * the file is being created), but not permissions for the whole path. That check probably should be
+ * added to `path_lookupat`.
  */
-
 int open_namei(struct shim_handle* hdl, struct shim_dentry* start, const char* path, int flags,
-               int mode, struct shim_dentry** dent);
+               int mode, struct shim_dentry** found);
 
-/* This function calls the low-level file system to do the work of opening file indicated by dent,
- * and initializing it in hdl. Flags are standard open flags.
+/*
+ * \brief Open an already retrieved dentry, and associate a handle with it
  *
- * If O_TRUNC is specified, this function is responsible for calling the underlying truncate
- * function.
+ * \param[out] hdl handle to associate with dentry
+ * \param dent the dentry to open
+ * \param flags Unix open flags
+ *
+ * The dentry has to already correspond to a file (i.e. has to be valid and non-negative).
+ *
+ * The `flags` parameter will be passed to the underlying filesystem's `open` function. If O_TRUNC
+ * flag is specified, the filesystem's `truncate` function will also be called.
  */
-
 int dentry_open(struct shim_handle* hdl, struct shim_dentry* dent, int flags);
 
 /* This function enumerates a directory and caches the results in the dentry.
@@ -471,7 +534,7 @@ struct shim_dentry* get_new_dentry(struct shim_mount* fs, struct shim_dentry* pa
  * \param name name of searched dentry
  * \param name_len length of the name
  *
- * \returns the new dentry, or NULL if not found
+ * \returns the dentry, or NULL if not found
  *
  * The caller should hold `g_dcache_lock`.
  *
--- a/LibOS/shim/src/bookkeep/shim_handle.c
+++ b/LibOS/shim/src/bookkeep/shim_handle.c
@@ -38,7 +38,7 @@ static int init_tty_handle(struct shim_handle* hdl, bool write) {
     hdl->flags = write ? (O_WRONLY | O_APPEND) : O_RDONLY;
 
     struct shim_dentry* dent = NULL;
-    if ((ret = path_lookupat(NULL, "/dev/tty", LOOKUP_OPEN, &dent, NULL)) < 0)
+    if ((ret = path_lookupat(/*start=*/NULL, "/dev/tty", LOOKUP_FOLLOW, &dent)) < 0)
         return ret;
 
     ret = dent->fs->d_ops->open(hdl, dent, hdl->flags);
@@ -82,7 +82,7 @@ static inline int init_exec_handle(void) {
         while (*p == '/') {
             p++;
         }
-        path_lookupat(fs->root, p, 0, &exec->dentry, fs);
+        path_lookupat(fs->root, p, LOOKUP_FOLLOW, &exec->dentry);
         set_handle_fs(exec, fs);
         if (exec->dentry)
             dentry_get_path_into_qstr(exec->dentry, &exec->path);
--- a/LibOS/shim/src/bookkeep/shim_process.c
+++ b/LibOS/shim/src/bookkeep/shim_process.c
@@ -46,7 +46,7 @@ int init_process(int argc, const char** argv) {
     g_process.umask = 0;
 
     struct shim_dentry* dent = NULL;
-    int ret = path_lookupat(NULL, "/", 0, &dent, NULL);
+    int ret = path_lookupat(/*start=*/NULL, "/", LOOKUP_FOLLOW | LOOKUP_DIRECTORY, &dent);
     if (ret < 0) {
         log_error("Could not set up dentry for \"/\", something is seriously broken.\n");
         return ret;
--- a/LibOS/shim/src/fs/proc/ipc-thread.c
+++ b/LibOS/shim/src/fs/proc/ipc-thread.c
@@ -98,12 +98,6 @@ do_ipc:
     if (dentptr) {
         /* XXX: Not sure how to handle this case yet */
         __abort();
-        ret = path_lookupat(NULL, (char*)ipc_data, 0, &dent, NULL);
-        if (ret < 0)
-            goto out;
-
-        get_dentry(dent);
-        *dentptr = dent;
     }
 
 out:
--- a/LibOS/shim/src/fs/proc/thread.c
+++ b/LibOS/shim/src/fs/proc/thread.c
@@ -110,7 +110,7 @@ static int find_thread_link(const char* name, struct shim_qstr* link,
     if (nextnext) {
         struct shim_dentry* next_dent = NULL;
 
-        ret = path_lookupat(dent, nextnext, 0, &next_dent, dent->fs);
+        ret = path_lookupat(dent, nextnext, LOOKUP_FOLLOW, &next_dent);
         if (ret < 0)
             goto out;
 
@@ -344,7 +344,7 @@ static int find_thread_each_fd(const char* name, struct shim_qstr* link,
     if (rest) {
         struct shim_dentry* next_dent = NULL;
 
-        ret = path_lookupat(dent, rest, 0, &next_dent, dent->fs);
+        ret = path_lookupat(dent, rest, LOOKUP_NO_FOLLOW, &next_dent);
         if (ret < 0)
             goto out;
 
--- a/LibOS/shim/src/fs/shim_dcache.c
+++ b/LibOS/shim/src/fs/shim_dcache.c
@@ -250,6 +250,56 @@ bool dentry_is_ancestor(struct shim_dentry* anc, struct shim_dentry* dent) {
     return false;
 }
 
+static int dump_dentry_write_all(const char* str, size_t size, void* arg) {
+    __UNUSED(arg);
+    log_always("%.*s\n", (int)size, str);
+    return 0;
+}
+
+#define DUMP_FLAG(flag, s, empty) buf_puts(&buf, (dent->state & (flag)) ? (s) : (empty))
+
+static void dump_dentry(struct shim_dentry* dent, unsigned int level) {
+    assert(locked(&g_dcache_lock));
+
+    struct print_buf buf = INIT_PRINT_BUF(dump_dentry_write_all);
+
+    buf_printf(&buf, "[%6.6s ", dent->fs ? dent->fs->type : "");
+
+    DUMP_FLAG(DENTRY_VALID, "V", ".");
+    DUMP_FLAG(DENTRY_LISTED, "L", ".");
+    DUMP_FLAG(DENTRY_INO_UPDATED, "I", ".");
+    DUMP_FLAG(DENTRY_SYNTHETIC, "S", ".");
+    buf_printf(&buf, "%3d] ", (int)REF_GET(dent->ref_count));
+
+    DUMP_FLAG(DENTRY_MOUNTPOINT, "*", " ");
+    DUMP_FLAG(DENTRY_NEGATIVE, "-", " ");
+
+    for (unsigned int i = 0; i < level; i++)
+        buf_puts(&buf, "  ");
+
+    buf_puts(&buf, qstrgetstr(&dent->name));
+    DUMP_FLAG(DENTRY_ISDIRECTORY, "/", "");
+    DUMP_FLAG(DENTRY_ISLINK, " -> ", "");
+    buf_flush(&buf);
+
+    struct shim_dentry* child;
+    LISTP_FOR_EACH_ENTRY(child, &dent->children, siblings) {
+        dump_dentry(child, level + 1);
+    }
+}
+
+#undef DUMP_FLAG
+
+void dump_dcache(struct shim_dentry* dent) {
+    lock(&g_dcache_lock);
+
+    if (!dent)
+        dent = g_dentry_root;
+
+    dump_dentry(dent, 0);
+    unlock(&g_dcache_lock);
+}
+
 BEGIN_CP_FUNC(dentry) {
     __UNUSED(size);
     assert(size == sizeof(struct shim_dentry));
--- a/LibOS/shim/src/fs/shim_fs.c
+++ b/LibOS/shim/src/fs/shim_fs.c
@@ -363,7 +363,7 @@ int init_mount(void) {
 
     if (fs_start_dir) {
         struct shim_dentry* dent = NULL;
-        ret = path_lookupat(NULL, fs_start_dir, 0, &dent, NULL);
+        ret = path_lookupat(/*start=*/NULL, fs_start_dir, LOOKUP_FOLLOW | LOOKUP_DIRECTORY, &dent);
         free(fs_start_dir);
         if (ret < 0) {
             log_error("Invalid 'fs.start_dir' in manifest.\n");
@@ -456,12 +456,12 @@ static int __mount_fs(struct shim_mount* mount, struct shim_dentry* dent) {
     do {
         struct shim_dentry* parent = dent->parent;
 
-        if (dent->state & DENTRY_ANCESTOR) {
+        if (dent->state & DENTRY_SYNTHETIC) {
             put_dentry(dent);
             break;
         }
 
-        dent->state |= DENTRY_ANCESTOR;
+        dent->state |= DENTRY_SYNTHETIC;
         if (parent)
             get_dentry(parent);
         put_dentry(dent);
@@ -501,7 +501,7 @@ out:
 /* Parent is optional, but helpful.
  * dentp (optional) memoizes the dentry of the newly-mounted FS, on success.
  *
- * The make_ancestor flag creates pseudo-dentries for any missing paths (passed to __path_lookupat).
+ * The make_ancestor flag creates pseudo-dentries for any missing paths (passed to _path_lookupat).
  * This is only intended for use to connect mounts specified in the manifest when an intervening
  * path is missing.
  */
@@ -509,6 +509,10 @@ int mount_fs(const char* type, const char* uri, const char* mount_point, struct 
              struct shim_dentry** dentp, bool make_ancestor) {
     int ret = 0;
     struct shim_fs* fs = find_fs(type);
+
+    int lookup_flags = LOOKUP_NO_FOLLOW;
+    if (make_ancestor)
+        lookup_flags |= LOOKUP_MAKE_SYNTHETIC;
 
     if (!fs || !fs->fs_ops || !fs->fs_ops->mount) {
         ret = -ENODEV;
@@ -536,8 +540,7 @@ int mount_fs(const char* type, const char* uri, const char* mount_point, struct 
             char* parent_path = __alloca(parent_len + 1);
             memcpy(parent_path, mount_point, parent_len);
             parent_path[parent_len] = 0;
-            if ((ret = __path_lookupat(g_dentry_root, parent_path, 0, &parent, 0, g_dentry_root->fs,
-                                       make_ancestor)) < 0) {
+            if ((ret = _path_lookupat(g_dentry_root, parent_path, lookup_flags, &parent)) < 0) {
                 log_error("Path lookup failed %d\n", ret);
                 goto out_with_unlock;
             }
@@ -597,8 +600,7 @@ int mount_fs(const char* type, const char* uri, const char* mount_point, struct 
     qstrsetstr(&dent->rel_path, "", 0);
 
     /*Now go ahead and do a lookup so the dentry is valid */
-    if ((ret = __path_lookupat(g_dentry_root, mount_point, 0, &dent2, 0,
-                               parent ? parent->fs : mount, make_ancestor)) < 0)
+    if ((ret = _path_lookupat(g_dentry_root, mount_point, lookup_flags, &dent2)) < 0)
         goto out_with_unlock;
 
     assert(dent == dent2);
--- a/LibOS/shim/src/fs/shim_namei.c
+++ b/LibOS/shim/src/fs/shim_namei.c
@@ -26,53 +26,27 @@ static inline const char* eat_slashes(const char* string) {
     return string;
 }
 
-static inline int __lookup_flags(int flags) {
-    int retval = LOOKUP_FOLLOW;
-
-    if (flags & O_NOFOLLOW)
-        retval &= ~LOOKUP_FOLLOW;
-
-    if ((flags & (O_CREAT | O_EXCL)) == (O_CREAT | O_EXCL))
-        retval &= ~LOOKUP_FOLLOW;
-
-    if (flags & O_DIRECTORY)
-        retval |= LOOKUP_DIRECTORY;
-
-    return retval;
-}
-
-/* check permission (specified by mask) of a dentry. If force is not set,
- * permission is considered granted on invalid dentries
- *
- * mask is the same as mode in the manual for access(2): F_OK, R_OK, W_OK,
- * X_OK
- *
- * Returns 0 on success, negative on failure.
- */
-/* Assume caller has acquired g_dcache_lock */
-int __permission(struct shim_dentry* dent, mode_t mask) {
+int check_permissions(struct shim_dentry* dent, mode_t mask) {
     assert(locked(&g_dcache_lock));
-
-    mode_t mode = 0;
-
-    /* Pseudo dentries don't really have permssions.  I wonder if
-     * we could tighten up the range of allowed calls.
-     */
-    if (dent->state & DENTRY_ANCESTOR)
-        return 0;
+    assert(dent->state & DENTRY_VALID);
 
     if (dent->state & DENTRY_NEGATIVE)
         return -ENOENT;
 
-    /* At this point, we can just return zero if we are only
-     * checking F_OK (the dentry isn't negative). */
+    /* If we only check if the file exists, at this point we know that */
     if (mask == F_OK)
         return 0;
 
-    /* A dentry may not have the mode stored.  The original code
-     * used both NO_MODE and !DENTRY_VALID; let's try to consolidate to
-     * just NO_MODE.
+    /*
+     * Synthetic directories don't really have permissions.
+     *
+     * TODO: current mount operation marks the mount root as synthetic. Once that's fixed, we'll be
+     * able to disallow W_OK here.
      */
+    if (dent->state & DENTRY_SYNTHETIC)
+        return 0;
+
+    /* A dentry may not have the mode stored yet. Query the underlying filesystem. */
     if (dent->mode == NO_MODE) {
         assert(dent->fs);
         if (!dent->fs->d_ops || !dent->fs->d_ops->mode) {
@@ -81,6 +55,7 @@ int __permission(struct shim_dentry* dent, mode_t mask) {
         }
 
         /* Fall back to the low-level file system */
+        mode_t mode = 0;
         int err = dent->fs->d_ops->mode(dent, &mode);
 
         /*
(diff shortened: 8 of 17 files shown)


pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: 0 of 17 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 261 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZDWNKH7MPCVry6rT_-:-MZDWNKH7MPCVry6rT_0:b34u3t5) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_dcache.c#L261)):*
> ```C
> #define DUMP_FLAG(flag, s, empty) buf_puts(&buf, (dent->state & flag) ? s : empty)
> 
> static void dump_dentry(struct shim_dentry* dent, unsigned int level) {
> ```

For now, I'm manually adding `dump_dentry(NULL);` in `libos_clean_and_exit()`. Any idea how best I can enable it without having to make local modifications to the code? I would stick it at `log_trace()`, but it's probably not useful if you're not working on the filesystem specifically.

dimakuv (changes requested):
> Reviewed 17 of 17 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: all files reviewed, 25 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHd2KcCckxkdQH3Uao:-MZHd2Kd8zhmedQuuweJ:bu9se5) (no related file):*
> Instead, the code now treats such cases as a success and create a

Typo in commit message: `create` -> `creates`.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHdTJI9xzzUoP1RJyL:-MZHdTJI9xzzUoP1RJyM:b-zxr5h) (no related file):*
>   This also adds a LOOKUP_NO_FOLLOW value to make the the call sites

One more typo in commit message: `the the` -> `the`

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZILG_UCwtx7w7zUnHQ:-MZILG_UCwtx7w7zUnHR:b-51d0kj) (no related file):*
Awesome PR!

---
___
*[LibOS/shim/include/shim_fs.h, line 220 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHe29OAvhy_xrZQmNK:-MZHe29OAvhy_xrZQmNL:b-ebc9gp) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L220)):*
> ```objc
> };
> 
> /* TODO: This actually does not get migrated after a fork, we just copy `g_process.root` */
> ```

Is `g_process.root` equal to `g_dentry_root`? In this case, you're right, we should simply use the migration mechanism instead of copying.
___
*[LibOS/shim/include/shim_fs.h, line 225 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHeDatAVVeQIBzxqKD:-MZHeDatAVVeQIBzxqKE:b4ejiz2) ([raw file](https://github.com/oscarlab/graphene/blob/658de0e002ac782e2effc9334db359a4e09c879a/LibOS/shim/include/shim_fs.h#L225)):*
> ```objc
> #define LOOKUP_DIRECTORY 002
> #define LOOKUP_CONTINUE  004  // No longer needed
> #define LOOKUP_PARENT    010  // Not sure we need this
> ```

I just love these old comments! :)
___
*[LibOS/shim/include/shim_fs.h, line 308 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHeuL1CSwAmrsiSopp:-MZHeuL24Uk3lW4BYQ2S:b-tg93bt) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L308)):*
> ```objc
>  * -ENOENT).
>  */
> int check_permissions(struct shim_dentry* dent, mode_t mask);
> ```

Shouldn't you start this function name with `_` because it assumes a lock?
___
*[LibOS/shim/include/shim_fs.h, line 316 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHemLW6grBwucNuZgm:-MZHemLW6grBwucNuZgn:b97mf7b) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L316)):*
> ```objc
>  * LOOKUP_NO_FOLLOW as a pseudo-flag for readability.
>  *
>  * This is modeled after Linux and and BSD codebases, which define a positive FOLLOW flag, and a
> ```

typo: `and and` -> `and`
___
*[LibOS/shim/include/shim_fs.h, line 337 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHg-AiD2COycKnqUd6:-MZHg-AiD2COycKnqUd7:b9k1syy) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L337)):*
> ```objc
>  *
>  * On success, returns 0, and puts the found dentry in `*new`. The reference count of the dentry
>  * will be increased by one.
> ```

`increased by one` == `set to one`? The wording is confusing -- we just created the dentry and we set the refcount, there is nothing to increase.
___
*[LibOS/shim/include/shim_fs.h, line 340 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHgHs02-HBFtSp2DOz:-MZHgHs02-HBFtSp2DP-:b4lmu7h) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L340)):*
> ```objc
>  *
>  * On failure, returns a negative error code, and sets `*new` to NULL. (Note that in case of
>  * LOOKUP_CREATE, not all file-not-found errors are considered a failure).
> ```

What does this mean? Could you give an example? By `file-not-found errors`, do you mean `-ENOENT` or do you mean `file isn't there but can be created`?
___
*[LibOS/shim/include/shim_fs.h, line 385 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHhh5b5orpdzyO6T-8:-MZHhh5c4-aSgZkUlINZ:bk732z1) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L385)):*
> ```objc
>  * \param[out] hdl handle to initialize, can be NULL
>  * \param start the start dentry for relative paths, or NULL (in which case it will default to
>  * process' cwd)
> ```

You missed `\param path`
___
*[LibOS/shim/include/shim_fs.h, line 388 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHiDyf4nk1nWUwEHEm:-MZHhjd_0JGWHyM8kWSc:bw1g971) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L388)):*
> ```objc
>  * \param flags Unix open flags (see below)
>  * \param mode Unix file mode, used when creating a new file/directory
>  * \param[out] new pointer to retrieved dentry, can be NULL
> ```

Why `new`? Should be `dent`
___
*[LibOS/shim/include/shim_fs.h, line 396 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHiNVfEfAeX7CAc29D:-MZHiNVfEfAeX7CAc29E:bw1g971) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L396)):*
> ```objc
>  * will be just retrieved or created.
>  *
>  * If `new` is provided, on success it will be set to the files's dentry (and its reference count
> ```

Why `new`? Should be `dent`
___
*[LibOS/shim/include/shim_fs.h, line 396 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHiOMmEAvZWVJSYO4o:-MZHiOMmEAvZWVJSYO4p:bgknppj) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L396)):*
> ```objc
>  * will be just retrieved or created.
>  *
>  * If `new` is provided, on success it will be set to the files's dentry (and its reference count
> ```

Typo: `files's dentry` -> `file's dentry`
___
*[LibOS/shim/include/shim_fs.h, line 407 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHicylDakcKWq6iQQf:-MZHicylDakcKWq6iQQg:b-a3ene) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L407)):*
> ```objc
>  * - O_TRUNC: truncate the file after opening
>  *
>  * The flags (including any not listed above), as well as file mode, are passed to the underlying
> ```

Why "including any not listed above"? Is this how Linux does it? Shouldn't we check and restrict the possible flags, just like we do in other places (e.g. `mmap` and `futex`)?
___
*[LibOS/shim/include/shim_fs.h, line 414 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHjF5E7RXk5okTiD_Z:-MZHjF5E7RXk5okTiD__:b-lgzwyf) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L414)):*
> ```objc
> 
> /*
>  * \brief Open an already retrieved dentry
> ```

`...and associate hdl with it`
___
*[LibOS/shim/include/shim_fs.h, line 416 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHjKH_73SXG2AWECAb:-MZHjKH_73SXG2AWECAc:b-3wh0of) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L416)):*
> ```objc
>  * \brief Open an already retrieved dentry
>  *
>  * \param[out] hdl handle to initialize
> ```

Maybe not `initialize` but `associate with the opened dent`?
___
*[LibOS/shim/src/fs/shim_dcache.c, line 261 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZDWNKH7MPCVry6rT_-:-MZHvu533cm_bog2sC-Q:b-ml3kzt) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_dcache.c#L261)):*

For now, I'm manually adding `dump_dentry(NULL);` in `libos_clean_and_exit()`. Any idea how best I can enable it without having to make local modifications to the code? I would stick it at `log_trace()`, but it's probably not useful if you're not working on the filesystem specifically.
</blockquote></details>

Maybe just add this line that you manually add, and comment it out with a `NOTE: ...` comment?

I don't think we want to introduce any runtime flags for this. This is too FS-debugging-specific. To me it's fine to manually uncomment the line you'll add and do the debugging.
___
*[LibOS/shim/src/fs/shim_namei.c, line 44 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHxrS22icbZoswLo0g:-MZHxrS22icbZoswLo0h:bup5x90) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L44)):*
> ```C
>      *
>      * TODO: current mount operation marks the mount root as synthetic. Once that's fixed, we'll be
>      * able to disallow writes here.
> ```

I don't understand this comment. What writes?
___
*[LibOS/shim/src/fs/shim_namei.c, line 66 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHyGsP97TBS6M7lsQ6:-MZHyGsP97TBS6M7lsQ7:b49zxad) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L66)):*
> ```C
>          * I'm ok with a file system that doesn't care setting the
>          * permission to all.
>          */
> ```

What is this comment? Can we just remove it? It seems to be completely out of sync with the code.
___
*[LibOS/shim/src/fs/shim_namei.c, line 150 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHzPfZ0yQLYGoqvfJF:-MZHzPfZ0yQLYGoqvfJG:b-yml5q1) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L150)):*
> ```C
> err:
>     if (dent)
>         get_dentry(dent);
> ```

Why are we increasing the refcount on the error path? Even if we want to keep the dentry, wasn't its refcount already incremented in one of the above functions (`lookup()`, `lookup_dcache()`)?
___
*[LibOS/shim/src/fs/shim_namei.c, line 182 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIER2ACo_zSsHGbVga:-MZIER2B1tvbjvR_l0PZ:bm4es3q) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L182)):*
> ```C
> 
> 
> #define MAX_LINK_DEPTH 8
> ```

Looks like you forgot to undef it. But also, maybe we should move this macro to a header file? Looks like a rather important constant.
___
*[LibOS/shim/src/fs/shim_namei.c, line 186 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIEbx524XtyJ6X6ilH:-MZIEbx524XtyJ6X6ilI:be2zr7d) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L186)):*
> ```C
> /*
>  * This implementation is mostly iterative, but uses recursion to follow symlinks (which is why the
>  * link depth is capped at 8).
> ```

`8` -> `MAX_LINK_DEPTH`
___
*[LibOS/shim/src/fs/shim_namei.c, line 352 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIJ7-F9Ig-gschBGNl:-MZIJ7-G-_9Oi-K4hiFN:b9kxhen) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L352)):*
> ```C
> 
>     if ((flags & (O_CREAT | O_EXCL)) == (O_CREAT | O_EXCL))
>         retval &= ~LOOKUP_FOLLOW;
> ```

Could you add a comment why this subtle semantics ("do not follow symlinks if want to open exclusive")?
___
*[LibOS/shim/src/fs/shim_namei.c, line 510 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIKxSm7NvhWIrJpcQg:-MZIKxSn-4AL5JUWjCQ3:b9xesa7) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L510)):*
> ```C
>         /* Check permissions, but only if we didn't create the file: create/O_CREAT have
>          * idiosyncratic semantics about opening a newly-created, read-only file for writing, but
>          * only the first time. */
> ```

Could you explain this? I don't really follow this. If there is special treatment of `O_CREAT` during checking permissions, then where is it reflected in our code?
___
*[LibOS/shim/src/sys/shim_stat.c, line 171 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHxUba5UtYwu5NsPdf:-MZHxUbbAw684XA9yzxv:bdt3ys7) ([raw file](https://github.com/oscarlab/graphene/blob/658de0e002ac782e2effc9334db359a4e09c879a/LibOS/shim/src/sys/shim_stat.c#L171)):*
> ```C
>     struct shim_dentry* dent = NULL;
> 
>     if ((ret = path_lookupat(NULL, path, LOOKUP_ACCESS | LOOKUP_FOLLOW, &dent, NULL)) < 0)
> ```

This was a bug (`LOOKUP_FOLLOW`)?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: 13 of 17 files reviewed, 25 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHd2KcCckxkdQH3Uao:-MZIhjcy9LER4MgxNH8K:bt2vr1b) (no related file):*

> Instead, the code now treats such cases as a success and create a

Typo in commit message: `create` -> `creates`.
</blockquote></details>

Fixed.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHdTJI9xzzUoP1RJyL:-MZIhmDJ9FiZ5_IPdfbc:bt2vr1b) (no related file):*

>   This also adds a LOOKUP_NO_FOLLOW value to make the the call sites

One more typo in commit message: `the the` -> `the`
</blockquote></details>

Fixed.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZILG_UCwtx7w7zUnHQ:-MZIhnF_-LsbjoG0o5bL:b-ucfknh) (no related file):*

Awesome PR!
</blockquote></details>

Thanks!

---
___
*[LibOS/shim/include/shim_fs.h, line 220 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHe29OAvhy_xrZQmNK:-MZISaT1Enbo5eJQ8ks1:bczms3k) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L220)):*

Is `g_process.root` equal to `g_dentry_root`? In this case, you're right, we should simply use the migration mechanism instead of copying.
</blockquote></details>

I meant we migrate `g_process.root`, but leave `g_dentry_root` alone and it points to an empty directory. I reworded the comment.

Additional context:

- After initialization, we use `g_process.root` pretty much everywhere, which is why this went undetected
- The two are not necessarily the same, because we support chroot (although I'm not sure how good that support is).
___
*[LibOS/shim/include/shim_fs.h, line 308 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHeuL1CSwAmrsiSopp:-MZITjquDW6QIUGeVuP5:b-1a3p4x) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L308)):*

Shouldn't you start this function name with `_` because it assumes a lock?
</blockquote></details>

Do we have this convention? I thought `_func` is a version of `func` that assumes a lock, but if there's only one of them, there's no need to mark them this way.

And from the looks of it (there are other functions such as `lookup_dcache`, `get_new_dentry` etc. that also assume the lock) I think it's not worth it to enforce a naming convention for all functions requiring the lock. (Also, while it's not checked statically, in DEBUG mode mistakes are easy to notice at runtime because of the asserts).
___
*[LibOS/shim/include/shim_fs.h, line 316 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHemLW6grBwucNuZgm:-MZIV8of5KWzqBPiq3eU:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L316)):*

typo: `and and` -> `and`
</blockquote></details>

Fixed.
___
*[LibOS/shim/include/shim_fs.h, line 337 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHg-AiD2COycKnqUd6:-MZIVA5R7q6SYgxqe4Rd:b-e5yvah) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L337)):*

`increased by one` == `set to one`? The wording is confusing -- we just created the dentry and we set the refcount, there is nothing to increase.
</blockquote></details>

This function does not necessarily create a new dentry. And even if it does, it adds it to the cache first, so there is one extra reference above the ones in cache.

I guess `new` is a pretty bad name. Changed to`found`.
___
*[LibOS/shim/include/shim_fs.h, line 340 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHgHs02-HBFtSp2DOz:-MZIXbmrBbJ3ZKUXhroU:b-i8344n) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L340)):*

What does this mean? Could you give an example? By `file-not-found errors`, do you mean `-ENOENT` or do you mean `file isn't there but can be created`?
</blockquote></details>

Yeah, it's not clear here. I meant that there are circumstances when the file does not exist but this function succeeds (if `LOOKUP_CREATE` is set).

I reworded the description: the description of `LOOKUP_CREATE` elaborates on this case, and here I just say that it is the only case in which we return a negative dentry. Is that better?
___
*[LibOS/shim/include/shim_fs.h, line 385 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHhh5b5orpdzyO6T-8:-MZIYq5k03pxXwDtLoYA:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L385)):*

You missed `\param path`
</blockquote></details>

Fixed.
___
*[LibOS/shim/include/shim_fs.h, line 388 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHiDyf4nk1nWUwEHEm:-MZIYqo92E8bAidXuGcd:b-csuwfw) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L388)):*

Why `new`? Should be `dent`
</blockquote></details>

I renamed it to `found`, for consistency with `path_lookupat`.
___
*[LibOS/shim/include/shim_fs.h, line 396 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHiNVfEfAeX7CAc29D:-MZIZ5x74KOtxIa5G69m:b-csuwfw) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L396)):*

Why `new`? Should be `dent`
</blockquote></details>

I renamed it to `found`, for consistency with `path_lookupat`.
___
*[LibOS/shim/include/shim_fs.h, line 396 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHiOMmEAvZWVJSYO4o:-MZIhvZmA_7dPpxnPq27:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L396)):*

Typo: `files's dentry` -> `file's dentry`
</blockquote></details>

Fixed.
___
*[LibOS/shim/include/shim_fs.h, line 407 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHicylDakcKWq6iQQf:-MZIZXqe2KPfNo0YDt5o:b-u4cukv) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L407)):*

Why "including any not listed above"? Is this how Linux does it? Shouldn't we check and restrict the possible flags, just like we do in other places (e.g. `mmap` and `futex`)?
</blockquote></details>

Is this the right place, though? This function is a proxy, and needs to handle only some of the flags. It's not the syscall entrypoint such as `shim_do_openat` - maybe we should restrict the flags there?

The analogous function in Linux (`lookup_open`) assumes that the flags are already validated, and passes them to underlying filesystem.
___
*[LibOS/shim/include/shim_fs.h, line 414 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHjF5E7RXk5okTiD_Z:-MZIhzoo7yFy_y-DQD7O:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L414)):*

`...and associate hdl with it`
</blockquote></details>

Done.
___
*[LibOS/shim/include/shim_fs.h, line 416 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHjKH_73SXG2AWECAb:-MZI_Wkk-MH4WBxiBwX0:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L416)):*

Maybe not `initialize` but `associate with the opened dent`?
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 261 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZDWNKH7MPCVry6rT_-:-MZIa1XM3rTRHUnaVEHC:bxpzz4m) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_dcache.c#L261)):*

Maybe just add this line that you manually add, and comment it out with a `NOTE: ...` comment?

I don't think we want to introduce any runtime flags for this. This is too FS-debugging-specific. To me it's fine to manually uncomment the line you'll add and do the debugging.
</blockquote></details>

I'd rather not add a commented-out line to `libos_clean_and_exit` (it's not so easy to discover, and I might need to print this from inside other places too), but I added a comment explaining that this function is to be added manually to code for debugging.
___
*[LibOS/shim/src/fs/shim_namei.c, line 44 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHxrS22icbZoswLo0g:-MZIag_xE1Gj6K1LdD1S:biepicl) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L44)):*

I don't understand this comment. What writes?
</blockquote></details>

Changed to `disallow W_OK here.`
___
*[LibOS/shim/src/fs/shim_namei.c, line 66 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHyGsP97TBS6M7lsQ6:-MZIb-BC8WUnh66lrP7U:b7trxlq) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L66)):*

What is this comment? Can we just remove it? It seems to be completely out of sync with the code.
</blockquote></details>

I understand it as a suggestion to remove `d_ops->mode()`, and instead initialize the mode together with the rest of dentry (in `d_ops->lookup()`). A filesystem that doesn't disallow anything could set the mode to something like 0777, and this entire branch could be removed.

I think it's a good idea, and the same might be done with `d_ops->stat()`. That way, the initial `lookup()` fills all the necessary information about a file. But I don't want to do it until I understand this part of the filesystem better.
___
*[LibOS/shim/src/fs/shim_namei.c, line 150 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHzPfZ0yQLYGoqvfJF:-MZId4d15mBcggGBl4yD:b-3pgct8) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L150)):*

Why are we increasing the refcount on the error path? Even if we want to keep the dentry, wasn't its refcount already incremented in one of the above functions (`lookup()`, `lookup_dcache()`)?
</blockquote></details>

Oops, this should be `put_dentry`. Fixed.
___
*[LibOS/shim/src/fs/shim_namei.c, line 182 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIER2ACo_zSsHGbVga:-MZIdfI--hChoZpXolqK:b-w2xbxf) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L182)):*

Looks like you forgot to undef it. But also, maybe we should move this macro to a header file? Looks like a rather important constant.
</blockquote></details>

I moved this constant to `shim_fs.h`.
___
*[LibOS/shim/src/fs/shim_namei.c, line 186 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIEbx524XtyJ6X6ilH:-MZIdnrL51kaKZdeBliP:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L186)):*

`8` -> `MAX_LINK_DEPTH`
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/shim_namei.c, line 352 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIJ7-F9Ig-gschBGNl:-MZIeRJR6wechZgxsI6o:bhmztwu) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L352)):*

Could you add a comment why this subtle semantics ("do not follow symlinks if want to open exclusive")?
</blockquote></details>

Okay, I added a quote from the man page.

And I just noticed that we should fail with when opening a link (with `-ELOOP`, apparently). I added that and tested it by trying to open `/proc/self/cwd` with `O_NOFOLLOW`.
___
*[LibOS/shim/src/fs/shim_namei.c, line 510 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIKxSm7NvhWIrJpcQg:-MZIe_Bd9YRvLJDQuOaN:b6blnfe) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L510)):*
We are at the place where it's reflected in our code: this comment explains why the check is done here, and not outside of the if.

Here is the use case described in the comment, as replicated in Python:
```py
# creating a new file
>>> os.open('xyz', os.O_CREAT | os.O_WRONLY, stat.S_IRUSR)
3
# opening it again
>>> os.open('xyz', os.O_CREAT | os.O_WRONLY, stat.S_IRUSR)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
PermissionError: [Errno 13] Permission denied: 'xyz'
```
___
*[LibOS/shim/src/sys/shim_fs.c, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIi5JU8w-V48oYc7Jh:-MZIi5JU8w-V48oYc7Ji:b-pccvj0) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/sys/shim_fs.c#L1)):*
> ```C
> ```

It's annoying to have both `sys/shim_fs.h` and `fs/shim_fs.h`, especially since they are different layers of the same thing. What do you think about renaming this file (later) to `sys/shim_files.h`?
___
*[LibOS/shim/src/sys/shim_stat.c, line 171 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHxUba5UtYwu5NsPdf:-MZIgiR492ssohT_1xme:bb8h8ko) ([raw file](https://github.com/oscarlab/graphene/blob/658de0e002ac782e2effc9334db359a4e09c879a/LibOS/shim/src/sys/shim_stat.c#L171)):*

This was a bug (`LOOKUP_FOLLOW`)?
</blockquote></details>

The man page for `statfs` doesn't say anything about symlinks, so I assumed it does not follow them.

But I just double-checked (in Linux sources, and experimentally) and I was wrong, it does follow symlinks. Fixed.

dimakuv (changes requested):
> Reviewed 3 of 4 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHd2KcCckxkdQH3Uao:-MZJCsiP97bljMzh0Am4:bc25yqy) (no related file):*

Fixed.
</blockquote></details>

I don't see any changes in the commit messages?

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHdTJI9xzzUoP1RJyL:-MZJD-Oy8krAk6ZPELgQ:b-7ropln) (no related file):*

Fixed.
</blockquote></details>

ditto

---
___
*[LibOS/shim/include/shim_fs.h, line 220 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHe29OAvhy_xrZQmNK:-MZJD6yq6FhNGf96XxdD:b1sdurx) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L220)):*

I meant we migrate `g_process.root`, but leave `g_dentry_root` alone and it points to an empty directory. I reworded the comment.

Additional context:

- After initialization, we use `g_process.root` pretty much everywhere, which is why this went undetected
- The two are not necessarily the same, because we support chroot (although I'm not sure how good that support is).
</blockquote></details>

OK, thanks, now the comment is good and I understand the issue.
___
*[LibOS/shim/include/shim_fs.h, line 308 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHeuL1CSwAmrsiSopp:-MZJDLeA70TrwJFfyn0T:b-1xfk04) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L308)):*

Do we have this convention? I thought `_func` is a version of `func` that assumes a lock, but if there's only one of them, there's no need to mark them this way.

And from the looks of it (there are other functions such as `lookup_dcache`, `get_new_dentry` etc. that also assume the lock) I think it's not worth it to enforce a naming convention for all functions requiring the lock. (Also, while it's not checked statically, in DEBUG mode mistakes are easy to notice at runtime because of the asserts).
</blockquote></details>

OK, agreed. I had a feeling that "global function + assumes lock" means "start with underscore". But apparently not the case.
___
*[LibOS/shim/include/shim_fs.h, line 337 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHg-AiD2COycKnqUd6:-MZJDh7ODeCox9-jkCEO:b48iuai) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L337)):*

This function does not necessarily create a new dentry. And even if it does, it adds it to the cache first, so there is one extra reference above the ones in cache.

I guess `new` is a pretty bad name. Changed to`found`.
</blockquote></details>

OK. Thanks, this is better.
___
*[LibOS/shim/include/shim_fs.h, line 340 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHgHs02-HBFtSp2DOz:-MZJDtHg4amJc6VL2DO_:b1lnnh2) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L340)):*

Yeah, it's not clear here. I meant that there are circumstances when the file does not exist but this function succeeds (if `LOOKUP_CREATE` is set).

I reworded the description: the description of `LOOKUP_CREATE` elaborates on this case, and here I just say that it is the only case in which we return a negative dentry. Is that better?
</blockquote></details>

OK, perfect.
___
*[LibOS/shim/include/shim_fs.h, line 407 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHicylDakcKWq6iQQf:-MZJE8UJ--E126V4eOte:b-a96i19) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/include/shim_fs.h#L407)):*

Is this the right place, though? This function is a proxy, and needs to handle only some of the flags. It's not the syscall entrypoint such as `shim_do_openat` - maybe we should restrict the flags there?

The analogous function in Linux (`lookup_open`) assumes that the flags are already validated, and passes them to underlying filesystem.
</blockquote></details>

OK, fair enough. I guess not for this PR, so not blocking.
___
*[LibOS/shim/src/fs/shim_namei.c, line 66 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHyGsP97TBS6M7lsQ6:-MZJEagaB3e_WdL1TobJ:b7aovdi) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L66)):*

I understand it as a suggestion to remove `d_ops->mode()`, and instead initialize the mode together with the rest of dentry (in `d_ops->lookup()`). A filesystem that doesn't disallow anything could set the mode to something like 0777, and this entire branch could be removed.

I think it's a good idea, and the same might be done with `d_ops->stat()`. That way, the initial `lookup()` fills all the necessary information about a file. But I don't want to do it until I understand this part of the filesystem better.
</blockquote></details>

OK, let's keep it.
___
*[LibOS/shim/src/fs/shim_namei.c, line 510 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIKxSm7NvhWIrJpcQg:-MZJFPnWCloLg3PlNBVE:b-n54xjx) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L510)):*

We are at the place where it's reflected in our code: this comment explains why the check is done here, and not outside of the if.

Here is the use case described in the comment, as replicated in Python:
```py
# creating a new file
>>> os.open('xyz', os.O_CREAT | os.O_WRONLY, stat.S_IRUSR)
3
# opening it again
>>> os.open('xyz', os.O_CREAT | os.O_WRONLY, stat.S_IRUSR)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
PermissionError: [Errno 13] Permission denied: 'xyz'
```
</blockquote></details>

My confusion is more about the part of the comment `opening a newly-created, read-only file for writing` -- you don't check permissions at all when creating a file, but in the comment you have a very specific case. So the comment is too restrictive, whereas the code is simply "don't check permissions on newly-created file". So why do you mention this specific example in the comment (or maybe it's the only case?)?
___
*[LibOS/shim/src/sys/shim_fs.c, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIi5JU8w-V48oYc7Jh:-MZJG0df13ZpAx-lq4S1:b-tnejhv) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/sys/shim_fs.c#L1)):*

It's annoying to have both `sys/shim_fs.h` and `fs/shim_fs.h`, especially since they are different layers of the same thing. What do you think about renaming this file (later) to `sys/shim_files.h`?
</blockquote></details>

Which file? This file is `shim_fs.c`. I guess you meant these two:
- `shim/src/fs/shim_fs.c`
- `shim/src/sys/shim_fs.c`

I'm fine with renaming one of them. Your suggestion of `sys/shim_files.c` is fine.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[LibOS/shim/src/sys/shim_fs.c, line 1 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIi5JU8w-V48oYc7Jh:-MZJGnmi6Nc5hwOGIFrr:b-dlwmba) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/sys/shim_fs.c#L1)):*

Which file? This file is `shim_fs.c`. I guess you meant these two:
- `shim/src/fs/shim_fs.c`
- `shim/src/sys/shim_fs.c`

I'm fine with renaming one of them. Your suggestion of `sys/shim_files.c` is fine.
</blockquote></details>

Yes, sorry, I meant `.c` (not `.h`) in all instances.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: 16 of 17 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHd2KcCckxkdQH3Uao:-MZMHOuHDr85iv6X6MN7:b-73ldwg) (no related file):*

I don't see any changes in the commit messages?
</blockquote></details>

Oh, I didn't save the changed message before rebasing, sorry.

Should be good now.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZHdTJI9xzzUoP1RJyL:-MZMQUf33gaY4L56IAE3:b-896fix) (no related file):*

ditto
</blockquote></details>

Done.

---
___
*[LibOS/shim/src/fs/shim_namei.c, line 510 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIKxSm7NvhWIrJpcQg:-MZMHt_A8mBQHnDBWRi0:b-5n860g) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L510)):*

My confusion is more about the part of the comment `opening a newly-created, read-only file for writing` -- you don't check permissions at all when creating a file, but in the comment you have a very specific case. So the comment is too restrictive, whereas the code is simply "don't check permissions on newly-created file". So why do you mention this specific example in the comment (or maybe it's the only case?)?
</blockquote></details>

That's not my comment, but I agree: this case is more general, you can even set `mode` to 0.

I removed the example, and reworded the comment to describe the case directly (the mode can be incompatible with `acc_mode`).

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[LibOS/shim/src/fs/shim_namei.c, line 510 at r1](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZIKxSm7NvhWIrJpcQg:-MZMSl-S-nf4elbL7nu3:blluxvb) ([raw file](https://github.com/oscarlab/graphene/blob/3aaef34c2472f5e8e5aa491ec2283a38841f123c/LibOS/shim/src/fs/shim_namei.c#L510)):*

That's not my comment, but I agree: this case is more general, you can even set `mode` to 0.

I removed the example, and reworded the comment to describe the case directly (the mode can be incompatible with `acc_mode`).
</blockquote></details>

OK. I didn't notice that you copied this comment from the previous code (I didn't dare to look at the old code). Now I understand the context for this :) Thanks for rewording the comment!

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 13 of 17 files at r1, 2 of 4 files at r2, 1 of 1 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: all files reviewed, 5 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_fs.h, line 389 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZp1fRN0i9ZOYmL5-7K:-MZp1fRN0i9ZOYmL5-7L:b-v0eczh) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/include/shim_fs.h#L389)):*
> ```
> but not permissions for the whole path.
> ```

So, the caller is expected to do this? Is it actually implemented in the caller or do we just skip these checks? If the latter, there should be a TODO here, I think.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 259 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZtWqKH92x97sF0VTY6:-MZtWqKI8tDhYf2eNFem:b-5v6ese) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/src/fs/shim_dcache.c#L259)):*
> ```C
> }
> 
> #define DUMP_FLAG(flag, s, empty) buf_puts(&buf, (dent->state & flag) ? s : empty)
> ```

I know this macro is used only in the function below, but please still write it safely :) I.e. parenthesize the parameters.
___
*[LibOS/shim/src/fs/shim_fs.c, line 509 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZtY8jd8Fe2lI__pM3C:-MZtY8jd8Fe2lI__pM3D:bvx0hmp) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/src/fs/shim_fs.c#L509)):*
> ```
> bool make_ancestor
> ```

I think we could remove this argument and always assume it's `true`. But I guess you'll refactor this when fixing mountpoints.
___
*[LibOS/shim/src/fs/shim_namei.c, line 222 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZtkZwV3hZPH99Y4nlL:-MZtkZwV3hZPH99Y4nlM:b-a8hu8t) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/src/fs/shim_namei.c#L222)):*
> ```C
>     assert(start);
>     dent = start;
>     get_dentry(dent);
> ```

Any reason why couldn't we simplify this logic and just `get` the final, to-be-returned dentry right before returning?
___
*[LibOS/shim/src/fs/shim_namei.c, line 368 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZtnJ2UB14O9aQVDOUl:-MZtnJ2UB14O9aQVDOUm:b-guc46w) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/src/fs/shim_namei.c#L368)):*
> ```
> is_opened
> ```

I think grammatically correct form is either `is_open` or `is_opened`.
___
*[LibOS/shim/src/fs/shim_namei.c, line 627 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZtrlyp8O0OzB2AFbHQ:-MZtrlyq-7mzVWIWCnwJ:btl4fb7) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/src/fs/shim_namei.c#L627)):*
> ```C
>         if ((ret = lookup_dentry(dent, d->name, strlen(d->name), &child)) < 0) {
>             /* ENOENT from underlying lookup should be handled as DENTRY_NEGATIVE */
>             assert(ret != ENOENT);
> ```

Not `-ENOENT`?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: 14 of 17 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/include/shim_fs.h, line 389 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZp1fRN0i9ZOYmL5-7K:-MZvP9aaD39QYkbnlYqb:b-rqlc74) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/include/shim_fs.h#L389)):*

> ```
> but not permissions for the whole path.
> ```

So, the caller is expected to do this? Is it actually implemented in the caller or do we just skip these checks? If the latter, there should be a TODO here, I think.
</blockquote></details>

That's right, the check should be here. Or actually, it probably belongs to `path_lookupat`. I added a TODO to both of these functions.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 259 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZtWqKH92x97sF0VTY6:-MZvQcgeBcO7QcF6akD3:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/src/fs/shim_dcache.c#L259)):*

I know this macro is used only in the function below, but please still write it safely :) I.e. parenthesize the parameters.
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/shim_fs.c, line 509 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZtY8jd8Fe2lI__pM3C:-MZvQdZqESKJnL8926W3:b-uyj8q1) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/src/fs/shim_fs.c#L509)):*

> ```
> bool make_ancestor
> ```

I think we could remove this argument and always assume it's `true`. But I guess you'll refactor this when fixing mountpoints.
</blockquote></details>

Yes, I will be rewriting this function completely, when I change how mounts work. I don't want to touch the current implementation too much until then. And I agree, it should always create a synthetic node.
___
*[LibOS/shim/src/fs/shim_namei.c, line 222 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZtkZwV3hZPH99Y4nlL:-MZvTY1_4bBIV01XEPOu:b9vvrv6) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/src/fs/shim_namei.c#L222)):*

Any reason why couldn't we simplify this logic and just `get` the final, to-be-returned dentry right before returning?
</blockquote></details>

That's because this function calls other functions that give it a dentry with increased reference count (`lookup_dentry`, and itself, recursively, when following a symlink). So it needs to adjust some reference counts in the middle anyway.

Also, _not_ keeping a reference for the intermediate dentries is only correct assuming nothing can delete them in the meantime. This assumption is true now (because we keep `g_dcache_lock` for the whole time), but might not be later (if we e.g. release the lock during low-level filesystem lookup), and since keeping the ref counts is not a big deal here, I'd rather make the code not depend on it.

Overall, I'd say it's better to keep the "correct" reference counting discipline, unless a function is really simple (ideally doesn't call anything else).
___
*[LibOS/shim/src/fs/shim_namei.c, line 368 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZtnJ2UB14O9aQVDOUl:-MZvYu0Y-peMFl2ii5QK:b-cgkycx) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/src/fs/shim_namei.c#L368)):*

> ```
> is_opened
> ```

I think grammatically correct form is either `is_open` or `is_opened`.
</blockquote></details>

I thought about it a bit more, and I can just check `hdl->dentry`. The handle really should be unassociated before that (looking at the call sites confirms it, and I added an assert just in case).

(I did not like the `is_open` name because it somehow didn't suggest to me that it was opened in this function. So I started thinking about names like `initialized`. But it's better to avoid this variable completely).
___
*[LibOS/shim/src/fs/shim_namei.c, line 627 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZtrlyp8O0OzB2AFbHQ:-MZv_P-Y8PmQT7IIc1h_:b-sr7xhk) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/src/fs/shim_namei.c#L627)):*

Not `-ENOENT`?
</blockquote></details>

That's right, this assert did nothing. Fixed.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/fs/shim_namei.c, line 222 at r5](https://reviewable.io/reviews/oscarlab/graphene/2333#-MZtkZwV3hZPH99Y4nlL:-MZx0goZ1KeaNu7Jg_gB:b-j40ifj) ([raw file](https://github.com/oscarlab/graphene/blob/a4bf8a49acba539d47ada2ecd583cdf43166925c/LibOS/shim/src/fs/shim_namei.c#L222)):*

That's because this function calls other functions that give it a dentry with increased reference count (`lookup_dentry`, and itself, recursively, when following a symlink). So it needs to adjust some reference counts in the middle anyway.

Also, _not_ keeping a reference for the intermediate dentries is only correct assuming nothing can delete them in the meantime. This assumption is true now (because we keep `g_dcache_lock` for the whole time), but might not be later (if we e.g. release the lock during low-level filesystem lookup), and since keeping the ref counts is not a big deal here, I'd rather make the code not depend on it.

Overall, I'd say it's better to keep the "correct" reference counting discipline, unless a function is really simple (ideally doesn't call anything else).
</blockquote></details>

Ok, makes sense, I was just curious why you wrote it this way :)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv (approved):
> Reviewed 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: :shipit: complete! all files reviewed, all discussions resolved

boryspoplawski (commented):
> Reviewed 8 of 17 files at r1, 2 of 4 files at r2, 3 of 3 files at r6, 4 of 4 files at r7.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: all files reviewed, 12 unresolved discussions (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_-vIrPAtkuJQZzbPSy:-M_-vIrQ7WX1kieKx_WS:bcdvb5p) (no related file):*
Sorry I'm late to the party but I have a couple of comments.

---
___
*[LibOS/shim/include/shim_fs.h, line 383 at r7](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_-wlSX-OsXMCO1U26z:-M_-wlSX-OsXMCO1U27-:b-unojl3) ([raw file](https://github.com/oscarlab/graphene/blob/51dcf9681bbb2e009698b22e47f870cff8b37a21/LibOS/shim/include/shim_fs.h#L383)):*
> ```objc
>  * \brief Open a file under a given path, similar to Unix open
>  *
>  * \param[out] hdl handle to associate with dentry, can be NULL
> ```

I'm not sure if this is really `out` param. It's just modified/passed to another object IIUC, so I would omit it or use `in/out` (but I dont think it fits that case)
___
*[LibOS/shim/include/shim_fs.h, line 418 at r7](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_-yqbD0tle9Sm-h3Z_:-M_-yqbD0tle9Sm-h3Za:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/51dcf9681bbb2e009698b22e47f870cff8b37a21/LibOS/shim/include/shim_fs.h#L418)):*
> ```objc
>  * \brief Open an already retrieved dentry, and associate a handle with it
>  *
>  * \param[out] hdl handle to associate with dentry
> ```

ditto
___
*[LibOS/shim/src/fs/shim_namei.c, line 93 at r7](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_-zsrt06R2vbHBwlLy:-M_-zsrt06R2vbHBwlLz:b-5mei1m) ([raw file](https://github.com/oscarlab/graphene/blob/51dcf9681bbb2e009698b22e47f870cff8b37a21/LibOS/shim/src/fs/shim_namei.c#L93)):*
> ```C
>  * The caller should hold `g_dcache_lock`.
>  *
>  * On success, returns 0 and sets `*new` to the found dentry (whose reference count is increased). A
> ```

`*new` -> `*found`
___
*[LibOS/shim/src/fs/shim_namei.c, line 94 at r7](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_-zxIs-nyykolAxSkr:-M_-zxIs-nyykolAxSks:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/51dcf9681bbb2e009698b22e47f870cff8b37a21/LibOS/shim/src/fs/shim_namei.c#L94)):*
> ```C
>  *
>  * On success, returns 0 and sets `*new` to the found dentry (whose reference count is increased). A
>  * negative filesystem lookup (ENOENT) is also considered a success, and `*new` is set to a negative
> ```

ditto
___
*[LibOS/shim/src/fs/shim_namei.c, line 98 at r7](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_-zyIL9mZVZoy54BLK:-M_-zyIL9mZVZoy54BLL:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/51dcf9681bbb2e009698b22e47f870cff8b37a21/LibOS/shim/src/fs/shim_namei.c#L98)):*
> ```C
>  *
>  * On failure (including lookup failing with any other error than ENOENT) returns the negative error
>  * code, and sets `*new` to NULL.
> ```

ditto
___
*[LibOS/shim/src/fs/shim_namei.c, line 212 at r7](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_0-nqm50sXdpgil2wv:-M_0-nqm50sXdpgil2ww:b-iznrk9) ([raw file](https://github.com/oscarlab/graphene/blob/51dcf9681bbb2e009698b22e47f870cff8b37a21/LibOS/shim/src/fs/shim_namei.c#L212)):*
> ```C
>         /* Relative part with no start dentry provided, use process current working directory */
>         lock(&g_process.fs_lock);
>         start = g_process.cwd;
> ```

This seems incorrect: we do not get this dentry and then unlock `g_process.fs_lock`. What if another thread changes `cwd` and this dentry is destroyed?

Maybe change this to set `dent` directly and just add an `else` clause at the end to handle `start != NULL` case (and just do `dent = start` there)?
___
*[LibOS/shim/src/fs/shim_namei.c, line 228 at r7](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_06gcI5yjg0tjLxAfV:-M_06gcI5yjg0tjLxAfW:b63vseh) ([raw file](https://github.com/oscarlab/graphene/blob/51dcf9681bbb2e009698b22e47f870cff8b37a21/LibOS/shim/src/fs/shim_namei.c#L228)):*
> ```C
>         name++;
> 
>     while (*name != '\0') {
> ```

If `path == "/"` and `!(flags & LOOKUP_DIRECTORY)` this will still succeed and return root dentry.
___
*[LibOS/shim/src/fs/shim_namei.c, line 334 at r7](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_08LJc26nTP3hpa8Gi:-M_08LJd0qVGKqbDJ82t:b-snlvd2) ([raw file](https://github.com/oscarlab/graphene/blob/51dcf9681bbb2e009698b22e47f870cff8b37a21/LibOS/shim/src/fs/shim_namei.c#L334)):*
> ```C
>                    struct shim_dentry** found) {
>     lock(&g_dcache_lock);
>     int ret = do_path_lookupat(start, path, flags, found, /*link_depth=*/0);
> ```

NP: `_path_lookupat`?
___
*[LibOS/shim/src/fs/shim_namei.c, line 467 at r7](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_0Hd59-PisyAFiOwa5:-M_0Hd5A7Zu9Rnvn-Nu7:b-tnrpw2) ([raw file](https://github.com/oscarlab/graphene/blob/51dcf9681bbb2e009698b22e47f870cff8b37a21/LibOS/shim/src/fs/shim_namei.c#L467)):*
> ```C
> 
>     if (dent->state & DENTRY_ISLINK) {
>         /* Can happen if user specified O_NOFOLLOW, or O_TRUNC | O_EXCL. Posix requires us to fail
> ```

I guess we do not support `O_PATH`?
___
*[LibOS/shim/src/fs/shim_namei.c, line 490 at r7](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_0IE8G6GxFlIUyTs2n:-M_0IE8G6GxFlIUyTs2o:b1be03r) ([raw file](https://github.com/oscarlab/graphene/blob/51dcf9681bbb2e009698b22e47f870cff8b37a21/LibOS/shim/src/fs/shim_namei.c#L490)):*
> ```C
>             goto err;
> 
>         /* Create directory or file, depending on O_DIRECTORY. Return -EINVAL if the operation is
> ```

man for `open(2)` states that `When both O_CREAT and O_DIRECTORY are specified in flags and the file specified by pathname does not exist, open() will create a regular file (i.e., O_DIRECTORY is ignored).` Is that not the case on newer kernel? Because if it is, we might need to emulate it...
___
*[LibOS/shim/src/fs/shim_namei.c, line 632 at r7](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_09j6m977CsgUwULgI:-M_09j6n78XgnaWZX9nq:b-4ei442) ([raw file](https://github.com/oscarlab/graphene/blob/51dcf9681bbb2e009698b22e47f870cff8b37a21/LibOS/shim/src/fs/shim_namei.c#L632)):*
> ```C
> 
>             /* Ignore inaccessible files */
>             if (ret == -EACCES)
> ```

How can we have access to only part of directory?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2333)* status: all files reviewed, 12 unresolved discussions (waiting on @boryspoplawski and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2333#-M_-vIrPAtkuJQZzbPSy:-M_1aQsm8rOV2wR0mbPa:b-d17khr) (no related file):*

Sorry I'm late to the party but I have a couple of comments.
</blockquote></details>

Thanks. I created a PR in #2354, and I will respond to your comments there.

---

PR #2336, opened 2021-04-27
[Pal] Rework DkEvent* functions

boryspoplawski:
## Description of the changes 
This additionally fixes a race in PAL SGX that triggered after the rework.

## How to test this PR? 
Usual test suite (one test was rewritten).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)


boryspoplawski:
Jenkins, retest this please (just making sure)

boryspoplawski:
Jenkins, retest this please (just making sure)

dimakuv (changes requested):
> Reviewed 30 of 30 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHGPFI7WAdbadEDJjE:-MZHGPFJ1olCF3FHabME:bnrzwsz) (no related file):*
Awesome PR! Regarding the malicious futex in Linux-SGX PAL: this was known for some time, but surprisingly I didn't find any issue on that in our GitHub. I should always create issues for things like this...

---
___
*[Pal/regression/Event.c, line 78 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHB5Ie92W2ewAtLtPF:-MZHB5Ie92W2ewAtLtPG:b-lsx8kf) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/regression/Event.c#L78)):*
> ```C
> 
>     CHECK(DkSystemTimeQuery(&start));
>     DkSynchronizationObjectWait(event, NO_TIMEOUT);
> ```

? You don't update `ret` here. Also, shouldn't `DkSynchronizationObjectWait()` actually return 0, because the other thread set this event?
___
*[Pal/regression/Event.c, line 79 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHBFfK5E830NG_Ya4b:-MZHBFfK5E830NG_Ya4c:b-a2rkob) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/regression/Event.c#L79)):*
> ```C
>     CHECK(DkSystemTimeQuery(&start));
>     DkSynchronizationObjectWait(event, NO_TIMEOUT);
>     if (ret != -PAL_ERROR_TRYAGAIN) {
> ```

This currently checks the old `ret`.
___
*[Pal/src/host/Linux/db_events.c, line 35 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHC7ApEcFgiGpqeLNO:-MZHC7ApEcFgiGpqeLNP:b-sg3ox6) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux/db_events.c#L35)):*
> ```C
>     __atomic_store_n(&handle->event.signaled, 1, __ATOMIC_RELEASE);
>     int ret = INLINE_SYSCALL(futex, 6, &handle->event.signaled, FUTEX_WAKE,
>                              handle->event.auto_clear ? 1 : INT_MAX, NULL, NULL, 0);
> ```

This semantics is confusing to me. Why do we wake up only one waiter if `auto_clear == 1`, and all waiters otherwise?
___
*[Pal/src/host/Linux-SGX/db_events.c, line 52 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHDxqT48orSVJYoxym:-MZHDxqT48orSVJYoxyn:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux-SGX/db_events.c#L52)):*
> ```C
>         do {
>             ret = ocall_futex(handle->event.signaled_untrusted, FUTEX_WAKE,
>                               handle->event.auto_clear ? 1 : INT_MAX, -1);
> ```

ditto
___
*[Pal/src/host/Linux-SGX/db_events.c, line 66 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHFpIzBvXWfjM30oIM:-MZHFpIzBvXWfjM30oIN:b-5cyp2n) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux-SGX/db_events.c#L66)):*
> ```C
> }
> 
> static int wait_timeout(PAL_HANDLE handle, int64_t timeout_us) {
> ```

I would like some explicit comment that we use `handle->event.signaled` as the trusted information whether the event was raised, and that we use `handle->event.signaled_untrusted` purely as an optimization to sleep on a futex (not to burn cycles in a while loop).

Well, this was verbose and obvious. But I want some quick comment that we closed this security vulnerability.
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 934 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZH763i8gT2pf2nNaet:-MZH763i8gT2pf2nNaeu:b-lf7e8b) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux-SGX/enclave_ocalls.c#L934)):*
> ```C
> 
>     if (op == FUTEX_WAIT) {
>         retval = sgx_ocall(OCALL_FUTEX, ms);
> ```

Thanks for this. But did you encounter a bug, or this is just by manual inspection?
___
*[Pal/src/host/Linux-SGX/pal_host.h, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHFZb_20s7sYM03RAc:-MZHFZbaDqzLiGERPhTM:bmnc3qr) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux-SGX/pal_host.h#L165)):*
> ```objc
>              * still be able to call futex on it. */
>             spinlock_t lock;
>             /* Current number of waiters - used solely as an optimization. `uint32_t` because futex
> ```

By optimization -- you mean to avoid `ocall_futex(FUTEX_WAKE)` during `DkEventSet()`?
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 119 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZH8HM76EqywbGpatP9:-MZH8HM76EqywbGpatPA:bw8cb5x) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux-SGX/sgx_exception.c#L119)):*
> ```C
> }
> 
> int g_in_aex_profiling = 0;
> ```

Isn't it possible that two enclave threads get interrupted and end up in this code? Then one thread will finish and reset `g_in_aex_profiling`, breaking the other thread.

Or is this scenario impossible because we disallow async signals during async handling? If so, please add a comment about this detail. (And if this scenario is possible, then we can simply switch to per-thread var.)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: 27 of 32 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[Pal/regression/Event.c, line 78 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHB5Ie92W2ewAtLtPF:-MZHJn3AE9mz3dEXCy1r:b-kjw1js) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/regression/Event.c#L78)):*

? You don't update `ret` here. Also, shouldn't `DkSynchronizationObjectWait()` actually return 0, because the other thread set this event?
</blockquote></details>

Oops.
Done.
___
*[Pal/regression/Event.c, line 79 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHBFfK5E830NG_Ya4b:-MZHK0Ak6OKpY2eIn44b:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/regression/Event.c#L79)):*

This currently checks the old `ret`.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_events.c, line 35 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHC7ApEcFgiGpqeLNO:-MZHK3fS9UvW3tF--xs7:by6nwfk) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux/db_events.c#L35)):*

This semantics is confusing to me. Why do we wake up only one waiter if `auto_clear == 1`, and all waiters otherwise?
</blockquote></details>

Because with `auto_clear` the waiting thread consumes the event - so there is no need to wake all threads since only one will be able to proceed anyway.
___
*[Pal/src/host/Linux-SGX/db_events.c, line 66 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHFpIzBvXWfjM30oIM:-MZHKSjR8wgqAuXpRj4v:b-yvynms) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux-SGX/db_events.c#L66)):*

I would like some explicit comment that we use `handle->event.signaled` as the trusted information whether the event was raised, and that we use `handle->event.signaled_untrusted` purely as an optimization to sleep on a futex (not to burn cycles in a while loop).

Well, this was verbose and obvious. But I want some quick comment that we closed this security vulnerability.
</blockquote></details>

There is a similar comment in `pal_host.h`. You want me to expand it in some way or add another one here?
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 934 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZH763i8gT2pf2nNaet:-MZHKfhBBTzk4RnQ9VUt:bg9cjym) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux-SGX/enclave_ocalls.c#L934)):*

Thanks for this. But did you encounter a bug, or this is just by manual inspection?
</blockquote></details>

Just manually when checking what exactly `ocall_futex` supports. There was no real bug here, in the sense of correctness.
___
*[Pal/src/host/Linux-SGX/pal_host.h, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHFZb_20s7sYM03RAc:-MZHKzo09jDucCCWyt1d:b-y7q727) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux-SGX/pal_host.h#L165)):*

By optimization -- you mean to avoid `ocall_futex(FUTEX_WAKE)` during `DkEventSet()`?
</blockquote></details>

Yes. Should I add this to the comment?
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 119 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZH8HM76EqywbGpatP9:-MZHL7dKC61LNXb5Np6O:b-p6rtgr) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux-SGX/sgx_exception.c#L119)):*

Isn't it possible that two enclave threads get interrupted and end up in this code? Then one thread will finish and reset `g_in_aex_profiling`, breaking the other thread.

Or is this scenario impossible because we disallow async signals during async handling? If so, please add a comment about this detail. (And if this scenario is possible, then we can simply switch to per-thread var.)
</blockquote></details>

Ops, yes, this is wrong.
Fixed.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux-SGX/db_events.c, line 66 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHFpIzBvXWfjM30oIM:-MZHOsc61M0w40E7u63O:b9jg3c2) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux-SGX/db_events.c#L66)):*

There is a similar comment in `pal_host.h`. You want me to expand it in some way or add another one here?
</blockquote></details>

I would add another one here (a quick one, just saying "Yes, we thought about this problem and solved it").
___
*[Pal/src/host/Linux-SGX/pal_host.h, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHFZb_20s7sYM03RAc:-MZHOYpA1O8w0Pf_ojI0:b-hc3ivy) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux-SGX/pal_host.h#L165)):*

Yes. Should I add this to the comment?
</blockquote></details>

No, that was just for my own understanding. Resolving.

yamahata (dismissed):
> Reviewed 27 of 30 files at r1, 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

boryspoplawski:
Jenkins, retest Jenkins-SGX-18.04 please (`udp` timedout, unfortunately cannot reproduce locally)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: 31 of 32 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @yamahata)

___
*[Pal/src/host/Linux-SGX/db_events.c, line 66 at r1](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZHFpIzBvXWfjM30oIM:-MZHj17Y5Rn2zfmEwI4e:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/02c05702b0b2c8c40c315b47e60606f2303b647e/Pal/src/host/Linux-SGX/db_events.c#L66)):*

I would add another one here (a quick one, just saying "Yes, we thought about this problem and solved it").
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

boryspoplawski:
The failed `udp` failure was most likely due to multiple instanced running at the same time - this test is broken and always binds to a constant udp port.

pwmarcz (dismissed):
> Reviewed 26 of 30 files at r1, 5 of 5 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux-SGX/db_events.c, line 73 at r3](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZI2KV9BSnTaX4i3Kgq:-MZI2KV9BSnTaX4i3Kgr:bnrwswq) ([raw file](https://github.com/oscarlab/graphene/blob/6f8273439116f07865105a80f29deba7b7d649f1/Pal/src/host/Linux-SGX/db_events.c#L73)):*
> ```C
>         bool needs_sleep = false;
>         spinlock_lock(&handle->event.lock);
>         if (handle->event.auto_clear) {
> ```

I found the logic hard to follow because of `needs_sleep`: it's checked in several places, and some of them negate it or have an else-branch, which means a double-negation of `signaled`.

How about this? I tried to write it for my own understanding, and it looks simpler to me (provided I got it right):

```c
spinlock_lock(&handle->event.lock);
if (handle->event.signaled) {
    // Signaled - clear (if auto_clear) and return
    if (handle->event.auto_clear) {
        handle->event.signaled = false;
       __atomic_store_n(handle->event.signaled_untrusted, 0, __ATOMIC_RELEASE);
    }

    if (added_to_count)
        handle->event.waiters_cnt--;

    spinlock_unlock(&handle->event.lock);
    return 0;
}

// Not signaled - sleep and retry the check afterwards
if (!added_to_count) {
    handle->event.waiters_cnt++;
    added_to_count = true;
}
spinlock_unlock(&handle->event.lock);

... FUTEX_WAIT ...
```

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: 31 of 32 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[Pal/src/host/Linux-SGX/db_events.c, line 73 at r3](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZI2KV9BSnTaX4i3Kgq:-MZIt5yR8ps1RCFI5c92:bvff4cj) ([raw file](https://github.com/oscarlab/graphene/blob/6f8273439116f07865105a80f29deba7b7d649f1/Pal/src/host/Linux-SGX/db_events.c#L73)):*

I found the logic hard to follow because of `needs_sleep`: it's checked in several places, and some of them negate it or have an else-branch, which means a double-negation of `signaled`.

How about this? I tried to write it for my own understanding, and it looks simpler to me (provided I got it right):

```c
spinlock_lock(&handle->event.lock);
if (handle->event.signaled) {
    // Signaled - clear (if auto_clear) and return
    if (handle->event.auto_clear) {
        handle->event.signaled = false;
       __atomic_store_n(handle->event.signaled_untrusted, 0, __ATOMIC_RELEASE);
    }

    if (added_to_count)
        handle->event.waiters_cnt--;

    spinlock_unlock(&handle->event.lock);
    return 0;
}

// Not signaled - sleep and retry the check afterwards
if (!added_to_count) {
    handle->event.waiters_cnt++;
    added_to_count = true;
}
spinlock_unlock(&handle->event.lock);

... FUTEX_WAIT ...
```

</blockquote></details>

Yeah, it was a leftover from previous versions which did have locks.
I reworked it and it seems it produced the exact same code as you pasted here :)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: 31 of 32 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[Pal/src/host/Linux-SGX/db_events.c, line 73 at r3](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZI2KV9BSnTaX4i3Kgq:-MZIu0wXEH0hBD4ah3Ql:b9sqsah) ([raw file](https://github.com/oscarlab/graphene/blob/6f8273439116f07865105a80f29deba7b7d649f1/Pal/src/host/Linux-SGX/db_events.c#L73)):*

Yeah, it was a leftover from previous versions which did have locks.
I reworked it and it seems it produced the exact same code as you pasted here :)

</blockquote></details>

* didn't have locks

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 26 of 30 files at r1, 5 of 5 files at r2, 1 of 1 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[Pal/include/lib/api.h, line 371 at r3](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZIiz6KBzpDWYB4ge4O:-MZIiz6L59sUR_0ZhDAZ:bpxy7ri) ([raw file](https://github.com/oscarlab/graphene/blob/6f8273439116f07865105a80f29deba7b7d649f1/Pal/include/lib/api.h#L371)):*
> ```objc
> 
> #define US_IN_S 1000000ul
> #define NS_IN_US 1000ul
> ```

Could you prefix these with "TIME_" or something similar? These are globally visible and the current names aren't obvious to decipher.
___
*[Pal/src/host/Linux/db_events.c, line 65 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK41Li6LVCXMpKD4_2:-MZK41Lj92q6yeOJkmcX:b-ow6s7d) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux/db_events.c#L65)):*
> ```C
>             return 0;
>         }
>         int ret = INLINE_SYSCALL(futex, 6, &handle->event.signaled, FUTEX_WAIT, 0, timeout_ptr,
> ```

This timeout (`timeout_us`) doesn't seem to work properly - you don't decrease it, so you can cross the deadline specified by the caller (even infinitely if you're very unlucky).
___
*[Pal/src/host/Linux-SGX/db_events.c, line 54 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK2JM57TyiIE6fUDdZ:-MZK2JM6-A7OaQk5seue:b-wt4m9o) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/db_events.c#L54)):*
> ```C
>                               handle->event.auto_clear ? 1 : INT_MAX, -1);
>         } while (ret == -EINTR);
>         /* This `FUTEX_WAKE` cannot really fail. */
> ```

But an untrusted host may return failure, what then? This comment suggests that this is impossible.
___
*[Pal/src/host/Linux-SGX/db_events.c, line 90 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK2ikeBUrLpetg7c4N:-MZK2ikeBUrLpetg7c4O:beajyef) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/db_events.c#L90)):*
> ```C
>         spinlock_unlock(&handle->event.lock);
> 
>         int ret = ocall_futex(handle->event.signaled_untrusted, FUTEX_WAIT, 0, timeout_us);
> ```

ditto (no timeout reduction)
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 934 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK4pb-4UIizjG-arKm:-MZK4pb-4UIizjG-arKn:bdrb5qd) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/enclave_ocalls.c#L934)):*
> ```C
> 
>     if (op == FUTEX_WAIT) {
>         retval = sgx_ocall(OCALL_FUTEX, ms);
> ```

Please explain in a comment why exitless doesn't make sense for FUTEX_WAIT.
___
*[Pal/src/host/Linux-SGX/sgx_entry.S, line 75 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK5Y-dDWyl5usLncp6:-MZK5Y-dDWyl5usLncp7:b7d89g6) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/sgx_entry.S#L75)):*
> ```
> 
> #ifdef DEBUG
> 	# Inform that we are in aex profiling code
> ```

`aex` -> `AEX`
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 173 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK5woH1ZCd7S9QDZYm:-MZK5woIA7_4HfOMWvr1:b-r5t2qk) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/sgx_exception.c#L173)):*
> ```C
> 
>     if (interrupted_in_enclave(uc) || interrupted_in_aex_profiling()) {
>         /* signal arrived while in app/LibOS/trusted PAL code, handle signal inside enclave */
> ```

This comment is out of date now. Also, is this change fine w.r.t. security? Untrusted host has full control over `interrupted_in_aex_profiling()` result...
___
*[Pal/src/host/Linux-SGX/tools/common/Makefile, line 26 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK6ggC09v1Nwk17dx6:-MZK6ggC09v1Nwk17dx7:b-nvh0et) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/tools/common/Makefile#L26)):*
> ```Makefile
> CJSON_URI ?= \
> 	https://github.com/DaveGamble/cJSON/archive/$(CJSON_SRC) \
> 	https://packages.grapheneproject.io/distfiles/cJSON-$(CJSON_VERSION).tar.gz
> ```

lol :)

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @mkow)

___
*[Pal/src/host/Linux-SGX/db_events.c, line 54 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK2JM57TyiIE6fUDdZ:-MZM3GBK9jg6er3xeGWZ:befplx5) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/db_events.c#L54)):*

But an untrusted host may return failure, what then? This comment suggests that this is impossible.
</blockquote></details>

+1. I forgot about this. I guess we can simply error out if `ret < 0` with something like `PAL_ERROR_DENIED`.
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 173 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK5woH1ZCd7S9QDZYm:-MZM3pZ13JKw1L8I1Nik:b-uoo5uh) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/sgx_exception.c#L173)):*

This comment is out of date now. Also, is this change fine w.r.t. security? Untrusted host has full control over `interrupted_in_aex_profiling()` result...
</blockquote></details>

Borys, correct me if I'm wrong, but if the untrusted host maliciously modifies untrusted RIP, then we have two cases:

- In reality, signal arrived while in enclave/AEX profiling code, but malicious host changed RIP as if signal arrived in untrusted code (during signal handling) -- then enclave expects second-level EENTER (to handle signal) but gets first-level EENTER (interrupted syscall) and this leads to what Borys observed -- either hangs or weird failure with `RIP = 0x0` at some point. This probably needs more hardening in our sgx-entry asm code (but seems to be out of scope of this PR).

- In reality, signal arrived while in untrusted PAL code (during signal handling), but malicious host changed RIP as if signal arrived while in enclave/AEX profiling code -- then enclave expects first-level EENTER (interrupted syscall) but gets second-level EENTER (to handle signal). This will lead to `return_from_ocall` logic in our asm code (https://github.com/oscarlab/graphene/blob/483a292e676f2a0bbdeeab65fd8c08582c396316/Pal/src/host/Linux-SGX/enclave_entry.S#L665), and this can be considered as incorrect return values of the syscall. In other words, a Iago attack, which must be caught by our Linux-SGX PAL and LibOS layers.

boryspoplawski:
Jenkins, retest Jenkins-Debug-20.04 please (`kill11` failed, known issue)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: 26 of 33 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, @pwmarcz, and @yamahata)

___
*[Pal/include/lib/api.h, line 371 at r3](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZIiz6KBzpDWYB4ge4O:-MZVfI8gD7NLeqJ0sIJO:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6f8273439116f07865105a80f29deba7b7d649f1/Pal/include/lib/api.h#L371)):*

Could you prefix these with "TIME_" or something similar? These are globally visible and the current names aren't obvious to decipher.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux/db_events.c, line 65 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK41Li6LVCXMpKD4_2:-MZVfaSE9YEan-yHZTlB:b-ell9yk) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux/db_events.c#L65)):*

This timeout (`timeout_us`) doesn't seem to work properly - you don't decrease it, so you can cross the deadline specified by the caller (even infinitely if you're very unlucky).
</blockquote></details>

Yes, but the only way to solve it is with querying the time before futex (and either after too or changing it to `FUTEX_WAIT_BITSET`, which has absolute timer). I have that planned for another PR (to support getting the remaining time in the caller of this function), added a TODO for now.
___
*[Pal/src/host/Linux-SGX/db_events.c, line 54 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK2JM57TyiIE6fUDdZ:-MZVhF1FE1vbZJFsgXiI:b-rp16ug) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/db_events.c#L54)):*
> But an untrusted host may return failure, what then? This comment suggests that this is impossible.

Nothing then. Malicious host can always lie about doing or not doing a wake on the futex word, but it's just a DoS, so we don't care.

Expanded the comment.
___
*[Pal/src/host/Linux-SGX/db_events.c, line 90 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK2ikeBUrLpetg7c4N:-MZVi6ps6HB-miV53i65:bnw6i9g) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/db_events.c#L90)):*

ditto (no timeout reduction)
</blockquote></details>

ditto (no interface for it atm)
___
*[Pal/src/host/Linux-SGX/enclave_ocalls.c, line 934 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK4pb-4UIizjG-arKm:-MZViesnAPhoL0slZeH8:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/enclave_ocalls.c#L934)):*

Please explain in a comment why exitless doesn't make sense for FUTEX_WAIT.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_entry.S, line 75 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK5Y-dDWyl5usLncp6:-MZVinnm99Y9aJFvvujl:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/sgx_entry.S#L75)):*

`aex` -> `AEX`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 173 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK5woH1ZCd7S9QDZYm:-MZVj6oXCJKpHwgpiz-j:b-y84746) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/sgx_exception.c#L173)):*
> is this change fine w.r.t. security? Untrusted host has full control over interrupted_in_aex_profiling() result...

This all happens in untrusted part, so it also has complete control over `uc`, or actually everything. Yes it's all ok (or at least not worse than before).

> In reality, signal arrived while in enclave/AEX profiling code, but malicious host changed RIP as if signal arrived in untrusted code (during signal handling) -- then enclave expects second-level EENTER (to handle signal) but gets first-level EENTER (interrupted syscall) and this leads to what Borys observed -- either hangs or weird failure with RIP = 0x0 at some point. This probably needs more hardening in our sgx-entry asm code (but seems to be out of scope of this PR).

I do not understand. Enclave cannot get 1st level EENTER after an AEX. Malicious host cannot spoof nesting level of EENTER. Btw, `sgx_entry.S` is untrusted part, `enclave_entry.S` is trusted. Great naming, I know.
Failures with `RIP = 0` were in untrusted part and due to nesting level confusion in **untrusted** part.

> In reality, signal arrived while in untrusted PAL code (during signal handling), but malicious host changed RIP as if signal arrived while in enclave/AEX profiling code -- then enclave expects first-level EENTER (interrupted syscall) but gets second-level EENTER (to handle signal). This will lead to return_from_ocall logic in our asm code (...), and this can be considered as incorrect return values of the syscall. In other words, a Iago attack, which must be caught by our Linux-SGX PAL and LibOS layers.

We have nested signals disabled (just FYI, it doesnt change anything). Again, malicious host can spoof any value it wants, being host-level signal arrival or ocall return value.

All of this happens in untrusted part, the host can do whatever it wants, it doesn't matter. All `sgx_something` files are outside of the enclave. The only thing we need to worry about is our enclave interface not handling malicious EENTERs properly.

dimakuv (approved):
> Reviewed 1 of 1 files at r5, 7 of 7 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 173 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK5woH1ZCd7S9QDZYm:-MZWRVMTAjYomeRg9ZpP:b-km2md6) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/sgx_exception.c#L173)):*
@boryspoplawski , yes, I agree with all your points. My sloppy wording in the previous comment is at fault :(

> Enclave cannot get 1st level EENTER after an AEX. Malicious host cannot spoof nesting level of EENTER.

Yes, this is correct. The malicious host cannot spoof anything that happens inside the enclave. So if the enclave is entered via EENTER while it was AEX-interrupted, the enclave will go into 2nd level EENTER (increasing the trusted CSSA field correspondingly, which we examine in our enclave-asm code).

What I meant by my `enclave expects second-level EENTER (to handle signal) but gets first-level EENTER (interrupted syscall)` phrase was: the enclave expects *values in GPRs* (arguments in our custom enclave<->untrusted asm interface) that correspond to second-level EENTER but receives values in GPRs that correspond to the first-level EENTER.

My whole point was about "We must make sure that if the values in GPRs are malicious (and tailored to mimic the correct behavior of another legit flow, i.e., interrupt handling instead of returning-from-ocall), does our enclave-asm code still catch these attacks?".

> Btw, sgx_entry.S is untrusted part, enclave_entry.S is trusted. Great naming, I know.

Yeah, we all remember this :) I guess my phrase `our sgx-entry asm code` was confusing. I meant literally "our asm code that is the entry point inside the SGX enclave", which is inside the `enclave_entry.S` file. So yeah, the names are extremely bad, maybe it's time to change them finally so we never get into these misunderstandings again.

> Failures with RIP = 0 were in untrusted part and due to nesting level confusion in untrusted part.

Ok, this is the part that I didn't fully realize. Thanks, this makes it clear to me. I thought that our in-enclave code gets confused because of this "attack" and ultimately segfaults, but it is actually the untrusted part that gets confused. As you mention, we don't care about such cases. That's a relief, this means that at least this "attack" doesn't compromise our enclaves.

> The only thing we need to worry about is our enclave interface not handling malicious EENTERs properly.

Yes, this is what I worried about. I guess my explanations now make it clear that I misunderstood the bug this PR tries to fix (that this bug happens in *untrusted part* whereas the enclave part behaves correctly).

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: all files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 173 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK5woH1ZCd7S9QDZYm:-MZX_SQG18wtxVrYMeIc:bxy3b63) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/sgx_exception.c#L173)):*
> What I meant by my enclave expects second-level EENTER (to handle signal) but gets first-level EENTER (interrupted syscall) phrase was: the enclave expects values in GPRs (arguments in our custom enclave<->untrusted asm interface) that correspond to second-level EENTER but receives values in GPRs that correspond to the first-level EENTER.
> My whole point was about "We must make sure that if the values in GPRs are malicious (and tailored to mimic the correct behavior of another legit flow, i.e., interrupt handling instead of returning-from-ocall), does our enclave-asm code still catch these attacks?".

Ah, now I understand. I think it catches such cases, because the verification of ocall return vs async event happening is made solely based on CSSA. So the only thing the host can do is insert malicious values to usual interfaces (e.g. weird ocall return value), but that should be a very well known attack vector (and hopefully we sanitize args everywhere).

mkow (dismissed):
> Reviewed 7 of 7 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 173 at r5](https://reviewable.io/reviews/oscarlab/graphene/2336#-MZK5woH1ZCd7S9QDZYm:-MZXi66o67dvKn3TS9lQ:bx1kdnv) ([raw file](https://github.com/oscarlab/graphene/blob/7486d180965fd5211376256d7045ea42a4ed76e1/Pal/src/host/Linux-SGX/sgx_exception.c#L173)):*

> What I meant by my enclave expects second-level EENTER (to handle signal) but gets first-level EENTER (interrupted syscall) phrase was: the enclave expects values in GPRs (arguments in our custom enclave<->untrusted asm interface) that correspond to second-level EENTER but receives values in GPRs that correspond to the first-level EENTER.
> My whole point was about "We must make sure that if the values in GPRs are malicious (and tailored to mimic the correct behavior of another legit flow, i.e., interrupt handling instead of returning-from-ocall), does our enclave-asm code still catch these attacks?".

Ah, now I understand. I think it catches such cases, because the verification of ocall return vs async event happening is made solely based on CSSA. So the only thing the host can do is insert malicious values to usual interfaces (e.g. weird ocall return value), but that should be a very well known attack vector (and hopefully we sanitize args everywhere).
</blockquote></details>

Thanks for the explanations, I also had the same confusion as Dmitrii :)

dimakuv:
Jenkins, retest Jenkins-18.04 please (`kill11` LTP test failed, known and unrelated)

boryspoplawski:
Jenkins, retest this please (2/3, just making sure)

boryspoplawski:
Jenkins, retest Jenkins-SGX-18.04 please (SIGPWR, known IPC issue)

boryspoplawski:
Jenkins, retest this please (3/3, just making sure)

boryspoplawski:
Jenkins, retest this please (4/3, just making sure)

pwmarcz (approved):
> Reviewed 1 of 1 files at r5, 7 of 7 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2336)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2355, opened 2021-05-06
[CI] Add AlmaLinux 8 pipeline

woju:
## Description of the changes 

Meson installation paths are different on different flavours of distros, especially Python paths. This pipeline aims to catch at least some of bugs related to differences between distros.


woju:
RHEL-like distros don't have at least shellcheck, so it will require more tweaking. I'm postponing this PR.

PR #2357, opened 2021-05-07
[Pal/Linux-SGX] Do not include `api.h` in stdlib builds of Protected FS

dimakuv:
## Description of the changes 

Some files that implement Protected FS feature are built both in Graphene mode (without stdlib, instead using a subset of stdlib
functions provided in Graphene-internal `api.h`) and in normal stdlib mode. If stdlib headers and the `api.h` header are included together, this may lead to mismatches in function declarations, e.g., in `string.h` functions.

This PR fixes Protected-FS files to always use either `api.h` or stdlib `string.h` header but not two simultaneously.

## How to test this PR? 

All tests (especially `make pf-test`) must pass.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2357)

### DIFF
--- a/Pal/src/host/Linux-SGX/protected-files/protected_files.c
+++ b/Pal/src/host/Linux-SGX/protected-files/protected_files.c
@@ -4,18 +4,18 @@
  * Copyright (C) 2011-2019 Intel Corporation
  */
 
-#ifndef IN_PAL
+#ifdef IN_PAL
+#include "api.h"
+#else
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#define USE_STDLIB
 #endif
 
 #include "protected_files.h"
 #include "protected_files_format.h"
 #include "protected_files_internal.h"
 
-#include "api.h"
 
 /* Function for scrubbing sensitive memory buffers.
  * memset() can be optimized away and memset_s() is not available in PAL.
@@ -79,8 +79,7 @@ static bool ipf_import_metadata_key(pf_context_t* pf, bool restore, pf_key_t* ou
     pf_status_t status;
 
     buf.index = 1;
-    if (!strcpy_static(buf.label, METADATA_KEY_NAME, MAX_LABEL_SIZE))
-        return false;
+    memcpy(buf.label, METADATA_KEY_NAME, sizeof(METADATA_KEY_NAME));
 
     if (!restore) {
         status = g_cb_random((uint8_t*)&buf.nonce, sizeof(buf.nonce));
@@ -89,7 +88,7 @@ static bool ipf_import_metadata_key(pf_context_t* pf, bool restore, pf_key_t* ou
             return false;
         }
     } else {
-        COPY_ARRAY(buf.nonce, pf->file_metadata.plain_part.metadata_key_id);
+        memcpy(buf.nonce, pf->file_metadata.plain_part.metadata_key_id, sizeof(buf.nonce));
     }
 
     // length of output (128 bits)
@@ -103,7 +102,8 @@ static bool ipf_import_metadata_key(pf_context_t* pf, bool restore, pf_key_t* ou
     }
 
     if (!restore) {
-        COPY_ARRAY(pf->file_metadata.plain_part.metadata_key_id, buf.nonce);
+        memcpy(pf->file_metadata.plain_part.metadata_key_id, buf.nonce,
+               sizeof(pf->file_metadata.plain_part.metadata_key_id));
     }
 
     erase_memory(&buf, sizeof(buf));
@@ -188,7 +188,7 @@ static pf_context_t* ipf_open(const char* path, pf_file_mode_t mode, bool create
     // for new file, this value will later be saved in the meta data plain part (init_new_file)
     // for existing file, we will later compare this value with the value from the file
     // (init_existing_file)
-    COPY_ARRAY(pf->user_kdk_key, *kdk_key);
+    memcpy(pf->user_kdk_key, *kdk_key, sizeof(pf->user_kdk_key));
 
     // omeg: we require a canonical full path to file, so no stripping path to filename only
     // omeg: Intel's implementation opens the file, we get the fd and size from the Graphene handler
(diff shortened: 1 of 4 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2357)* status: 0 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2357#-M_4uBxyDCU7UcrKcJlj:-M_4uBxyDCU7UcrKcJlk:bkrhaxw) (no related file):*
@pwmarcz You added `USE_STDLIB` macro to circumvent issues with fortified C functions. Could you check if I correctly removed this macro from here? These were the only places where we needed this macro, so maybe just remove `USE_STDLIB` alltogether?

---
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 83 at r1](https://reviewable.io/reviews/oscarlab/graphene/2357#-M_4twMFAjBKURAZFeIj:-M_4twMFAjBKURAZFeIk:bo8kb3r) ([raw file](https://github.com/oscarlab/graphene/blob/da0bd9585a4209c4cf769fa65bcad93d6cb02c2c/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L83)):*
> ```C
>     buf.index = 1;
>     if (!strcpy_static(buf.label, METADATA_KEY_NAME, MAX_LABEL_SIZE))
>         return false;
> ```

FYI: Checking for the result didn't make sense, we have static buffers so this always succeeds.

In general, we should finally remove this hacky `strcpy_static()` function. And replace with standard function.

pwmarcz (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2357)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2357#-M_4uBxyDCU7UcrKcJlj:-M_5ET2J6rKl3g8yC6C_:bqzlgly) (no related file):*

@pwmarcz You added `USE_STDLIB` macro to circumvent issues with fortified C functions. Could you check if I correctly removed this macro from here? These were the only places where we needed this macro, so maybe just remove `USE_STDLIB` alltogether?
</blockquote></details>

If you don't get linking errors, it should be OK.

I agree we should remove `USE_STDLIB`. I see one remaining place that uses it: `LibOS/shim/test/regression/spinlock.c`. Looks like spinlocks pull in `api.h`. Probably that can (and should) be removed.

---

yamahata (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2357)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv)

mkow (changes requested):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2357)* status: all files reviewed, 2 unresolved discussions (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 92 at r1](https://reviewable.io/reviews/oscarlab/graphene/2357#-M_8feyG3paNlmBb24hm:-M_8feyHBLwV3ymiprks:b-mrwvto) ([raw file](https://github.com/oscarlab/graphene/blob/da0bd9585a4209c4cf769fa65bcad93d6cb02c2c/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L92)):*
> ```C
>         }
>     } else {
>         COPY_ARRAY(buf.nonce, pf->file_metadata.plain_part.metadata_key_id);
> ```

I don't like this change, `COPY_ARRAY` had a bunch of safety checks which `memcpy` lacks.

Overall, I think it's rather string.h which we need to get rid of, not api.h. And yup, "nostd" is not really well supported in the "at least good for low-level stuff" C language.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2357)* status: 2 of 4 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow, @pwmarcz, and @yamahata)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2357#-M_4uBxyDCU7UcrKcJlj:-M_KKCXwBLHGaofw7FRg:b-6smota) (no related file):*

If you don't get linking errors, it should be OK.

I agree we should remove `USE_STDLIB`. I see one remaining place that uses it: `LibOS/shim/test/regression/spinlock.c`. Looks like spinlocks pull in `api.h`. Probably that can (and should) be removed.

</blockquote></details>

OK, that one is not simple to remove. I'll keep `USE_STDLIB` then.

---
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 92 at r1](https://reviewable.io/reviews/oscarlab/graphene/2357#-M_8feyG3paNlmBb24hm:-M_KKa161tD-JYSOlpgZ:b-x6zu7) ([raw file](https://github.com/oscarlab/graphene/blob/da0bd9585a4209c4cf769fa65bcad93d6cb02c2c/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L92)):*
Done. I added similar checks as in `COPY_ARRAY` to the relevant places. I didn't want to copy-paste the `COPY_ARRAY` macro in Protected FS files because it's a lot of code, and I don't want this duplication.

On the other hand, I also don't think that we should keep using the `api.h` utility macros in this file. This file is probably the only one (unique) in Graphene that works both inside Graphene and outside of it (as part of the `pf_crypt` Linux program). So this file shouldn't rely on too many Graphene macros.

> Overall, I think it's rather string.h which we need to get rid of, not api.h.

I don't understand this. This file is used both inside Graphene (and there it has to use `api.h`) and outside of Graphene (and there it has to use "normal" C headers). Why would we use `api.h` in the "normal" outside-of-graphene program?

dimakuv:
Jenkins, retest Jenkins-20.04 please (`abort01` LTP test failed with `Test killed by SIGPWR!`)

PR #2342, opened 2021-04-30
[LibOS] Don't store dentry paths 

pwmarcz:
This is the next step in FS rewrite, after #2333. See also the issue for dentry cleanup (#2321).

This change removes `shim_dentry.rel_path` and `shim_handle.path` fields. Instead, the path is computed whenever necessary. This also means that the length limit for relative path is lifted.

To obtain a path for a dentry, use `dentry_abs_path` or `dentry_rel_path`. The constructed path is allocated on heap and must be
freed afterwards.

Because computing the path can fail when out of memory, the functions for hashing dentries do not compute the path explicitly, but follow the dentry chain to the root.

(There were many places in the code was responsible for maintaining the stored path, or enforcing the length. It looked pretty brittle. So I think this is a positive change).

**Next step**: rewrite `mount_fs` and make mount semantics more sane (as mentioned in #2324). More specifically, I want mounting to not _replace_ a dentry, but attach a new tree under it. This way, both the original mountpoint and the mounted directory can exist in the dentry tree.
* if a dentry is marked as a mountpoint, you're supposed to go to `dentry->mounted->root`,
* `dentry->parent` can be null not only for the global root, but also for the filesystem root, in which case you can go to `dentry->fs->mountpoint`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2342)

PR #2370, opened 2021-05-12
[LibOS] Change mount semantics to not replace mountpoint

pwmarcz:
This is the next branch after #2342 (I'll rebase if necessary).

This changes the mount semantics to be more Unix-like: the mountpoint is not deleted/replaced, but overriden. See the commit message for details.

*Next steps*: I'm going to return to sync engine (#2267). Also, I want to take a closer look at the filesystem interface. It looks like I can remove the `stat` and `mode` operations, making `lookup` already retrieve the right information (and store the information in the dentry structure).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2370)


pwmarcz:
Okay... it's not that simple, because we sometimes override `dentry->fs` to `pipe_builtin_fs` or `socket_builtin_fs`. I'll investigate..

PR #2372, opened 2021-05-12
[Pal/Linux-SGX] Fix refcounting on open/close of Protected Files

### LINKED ISSUE #2360: SGX protected files possible vulnerability
There might be a possible vulnerability in the implementation of the protected files feature. Specifically, we tested the following code that opens two protected files (configured the tmp dir to be protected according to the very intuitive examples you provide)

`int fd1 = open("tmp/sec1.txt", O_CREAT | O_RDWR, 0644);`
`int fd2 = open("tmp/sec2.txt", O_CREAT | O_RDWR, 0644);`

Then we tested what would happen if an attacker changed the return value of open so that the same file descriptor value was used for both files. Interestingly that led to a crash inside the enclave, debugging showed that it was due to accessing invalid memory in the ipf_close function. The specific culprit is the following line:
https://github.com/oscarlab/graphene/blob/1d5dfb4018d865894e0cd959a1c2a91ebfe8749d/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L354

dimakuv:
## Description of the changes 

Previously, `file_open()` and `file_close()` on a Protected File had weird semnatics: if the same PF was opened twice, there was only one PF context created, but there were several underlying host FDs. Even weirder, the PF context tied itself to the host FD (the very first opened one) using the pointer to this FD: `&pal_handle->file.fd`.

This resulted in bugs and cumbersome workarounds: the associated PAL handle could be closed and freed by our common PAL code, and the PF context would contain a dangling pointer. And for `file_attrquery()`, there was a workaround to destroy the PAL context if it seemed to be created only for this attrquery flow (this was based on this pointer-to-pal-handle-fd).

This PR fixes these bugs: the PF context stores the underlying host FD directly, and doesn't open new FDs on subsequent `file_open()`. So, there is always only one host FD for one PF context. As a side effect, the logic of `file_open()` and `file_close()` is refactored.

A new LibOS regression test `protected_file` is added for GDB debugging because the dedicated `fs/` tests are harder to debug.

Fixes #2360.

## How to test this PR? 

A new test is added. Also, `fs/pf-test` should succeed.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/graphene/2372)

### DIFF
--- a/LibOS/shim/test/regression/.gitignore
+++ b/LibOS/shim/test/regression/.gitignore
@@ -1,5 +1,6 @@
 /*.manifest
 /*.xml
+/*.dat
 
 /.cache
 /abort
@@ -76,6 +77,7 @@
 /proc_common
 /proc_cpuinfo
 /proc_path
+/protected_file
 /pselect
 /pthread_set_get_affinity
 /rdtsc
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -177,6 +177,7 @@ clean-tmp:
 		*.sig \
 		*.tmp \
 		*.token \
+		*.dat \
 		*~ \
 		.cache \
 		.pytest_cache \
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -67,6 +67,7 @@ c_executables = \
 	proc_common \
 	proc_cpuinfo \
 	proc_path \
+	protected_file \
 	pselect \
 	pthread_set_get_affinity \
 	readdir \
--- a/LibOS/shim/test/regression/manifest.template
+++ b/LibOS/shim/test/regression/manifest.template
@@ -40,3 +40,7 @@ sgx.allowed_files.testfile = "file:testfile" # for mmap_file test
 sgx.thread_num = 16
 
 sgx.nonpie_binary = 1
+
+# for protected_file test
+sgx.protected_files.pf1 = "file:protected_file_1.dat"
+sgx.protected_files_key = "ffeeddccbbaa99887766554433221100"
--- a/LibOS/shim/test/regression/protected_file.c
+++ b/LibOS/shim/test/regression/protected_file.c
@@ -1,0 +1,84 @@
+/* This test opens the same file twice, reads from one FD, reads from another FD, and closes both
+ * FDs. This test exists mainly to test Protected Files Linux-SGX feature. */
+
+#include <err.h>
+#include <errno.h>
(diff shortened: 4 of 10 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2372)* status: 0 of 10 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[Pal/src/host/Linux-SGX/db_files.c, line 613 at r1](https://reviewable.io/reviews/oscarlab/graphene/2372#-M_W31N5DR7k_OJiawSm:-M_W31N5DR7k_OJiawSn:brb3mi) ([raw file](https://github.com/oscarlab/graphene/blob/7f758af2da2760b157c215aa61de52d4e317455f/Pal/src/host/Linux-SGX/db_files.c#L613)):*
> ```C
>     assert(PF_SUCCESS(pfs));
> 
>     if (fd_from_attrquery == *(int*)pf_handle) { /* this is a PF opened just for us, close it */
> ```

FYI: If you want to understand what was checked here, think of it this way:
- The first time Graphene opened this PF, it put `pf->context.pf_handle = &first_host_fd`
- The second time Graphene opened the same PF, the ``pf->context.pf_handle` is not changed
- So when we arrive here, we basically check "Was it the first time Graphene opened this PF, just to do attrquery?"

This was super-brittle and led to dangling pointers, as found in #2360.

The new code simply opens the PF again and closes it afterwards. This works because PF has a refcount.
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.h, line 244 at r1](https://reviewable.io/reviews/oscarlab/graphene/2372#-M_W3dp5AjR0Bh0gvKO6:-M_W3dp5AjR0Bh0gvKO7:b-hnenq6) ([raw file](https://github.com/oscarlab/graphene/blob/fc2abe0d160890a33d8a83946f2a1ade78dd86bc/Pal/src/host/Linux-SGX/protected-files/protected_files.h#L244)):*
> ```objc
>  * \return PF status
>  */
> pf_status_t pf_set_mode(pf_context_t* pf, pf_file_mode_t mode);
> ```

FYI: I had to add these helpers to prevent the case of "two handles try to open the same PF in writable mode" -- this is prohibited by PFS.

dimakuv:
Jenkins, retest Jenkins-20.04 please (too old, Jenkins doesn't have it in the logs anymore)

dimakuv:
This fix is hacky. We need to refactor Protected Files implementation completely.

PR #2380, opened 2021-05-17
[Pal/Linux-SGX] Use SHA256 to hash session key into SGX report data

dimakuv:
## Description of the changes 

Previously, to verify the remote enclave during SGX local attestation, AES-CMAC was computed over the random enclave ID using the newly established session key (via Diffie-Hellman key exchange). This was overly complicated and incorrect because the session key was re-used. This PR simplifies this logic by performing SHA256 over the session key and sending it in the SGX report's user-defined data field.

## How to test this PR? 

All tests must succeed.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 0 of 8 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-M_u7wP3CCOwIPP24SQZ:bnxfmf1) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*
> ```C
>      */
>     ret = _DkRandomBitsRead(&g_pal_enclave_state.enclave_id,
>                             sizeof(g_pal_enclave_state.enclave_id));
> ```

FYI: `enclave_id` wasn't used for anything other than "CMAC over it" during SGX local attestation (see old `generate_sign_data()`)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 0 of 8 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2380#-Mb1bgZO685B6odtNkl2:-Mb1bgZP4vxE3MsXSjlD:b-hl91ze) (no related file):*
Rebased to the current master.

---

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2380#-Mb1bgZO685B6odtNkl2:-Mb1u3iC9SOfApCTFxsb:b-qbsjll) (no related file):*

Rebased to the current master.
</blockquote></details>

The commit message states that to verify remote enclave, but the flow is similar even to verify local enclaves, right? That is enclaves running on the same system.

---
___
*[Pal/src/host/Linux-SGX/db_process.c, line 76 at r3](https://reviewable.io/reviews/oscarlab/graphene/2380#-Mb1ws3m76QNXYZbuZHp:-Mb1ws3m76QNXYZbuZHq:b-m21ux2) ([raw file](https://github.com/oscarlab/graphene/blob/30a628d415bf584ab8f6585181cb326d3aa1104a/Pal/src/host/Linux-SGX/db_process.c#L76)):*
> ```C
>  */
> 
> static int hash_session_key(const PAL_SESSION_KEY* session_key, sgx_report_data_t* data) {
> ```

Trying to understand how we resolved the session key reuse? In the earlier implementation, we used `AESCMAC(session key, enclave id)` to get the report and now we use `SHA256(session key)`.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-Mb1uQxfE5doOXoIWBEn:bou129y) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*

FYI: `enclave_id` wasn't used for anything other than "CMAC over it" during SGX local attestation (see old `generate_sign_data()`)
</blockquote></details>

Why don't we want to use `enclave_id` to compute the report data? wouldn't unique `enclave_id` add more authenticity?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 0 of 8 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @vijaydhanraj)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2380#-Mb1bgZO685B6odtNkl2:-Mb2aH7X5Cao_kZnDJvX:b-3r2cg7) (no related file):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

The commit message states that to verify remote enclave, but the flow is similar even to verify local enclaves, right? That is enclaves running on the same system.
</blockquote></details>

By "remote" I meant "the other" enclave on the same host. So this whole logic is about **local attestation**. I see how using the word "remote" could confuse people, I will change it to "the other" or something like this during rebase.

---
___
*[Pal/src/host/Linux-SGX/db_process.c, line 76 at r3](https://reviewable.io/reviews/oscarlab/graphene/2380#-Mb1ws3m76QNXYZbuZHp:-Mb2bQerBiyCfVg0LJu9:b5cal1s) ([raw file](https://github.com/oscarlab/graphene/blob/30a628d415bf584ab8f6585181cb326d3aa1104a/Pal/src/host/Linux-SGX/db_process.c#L76)):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

Trying to understand how we resolved the session key reuse? In the earlier implementation, we used `AESCMAC(session key, enclave id)` to get the report and now we use `SHA256(session key)`.
</blockquote></details>

Your understanding of the previous implementation is correct. However, the "session key reuse" is **not** about reusing the same key between different enclave pairs. The "session key reuse" remark is about the fact that using AES-CMAC to sign `enclave id` message with `session key` leads to a reuse of this key **in the same session** (between the same enclaves in the pair).
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-Mb2b3Oy9txuTjapFiKZ:bcxpigc) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

Why don't we want to use `enclave_id` to compute the report data? wouldn't unique `enclave_id` add more authenticity?
</blockquote></details>

There is no need in this additional info. We already generate a unique session key for each enclave pair. The unique session key is generated via Diffie-Hellman, which in turn uses a source of entropy (which is just a random byte sequence from `rdrand` x86 instruction).

So since we already have enough entropy from `rdrand`, we don't need to throw in anything else. The result is already "random enough".
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 875 at r3](https://reviewable.io/reviews/oscarlab/graphene/2380#-Mb2aiZ-5YD3YsVWlXnA:-Mb2aiZ07y4AaIPmvsN3:b-5aqqpa) ([raw file](https://github.com/oscarlab/graphene/blob/30a628d415bf584ab8f6585181cb326d3aa1104a/Pal/src/host/Linux-SGX/enclave_framework.c#L875)):*
> ```C
> }
> 
> int _DkStreamKeyExchange(PAL_HANDLE stream, PAL_SESSION_KEY* key) {
> ```

FYI: @vijaydhanraj This is where we generate a brand new key for each pair of enclaves. We use Diffie-Hellman to create a shared secret and then use a KDF (currently SHA256, but see https://github.com/oscarlab/graphene/pull/2410) to generate the final session key. Note that the generated session key is unique for each pair of enclaves, so there is no reuse of the key between different enclave pairs.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 0 of 8 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/db_process.c, line 76 at r3](https://reviewable.io/reviews/oscarlab/graphene/2380#-Mb1ws3m76QNXYZbuZHp:-Mb2sCXG5LZGmeioMLUl:bez2zfz) ([raw file](https://github.com/oscarlab/graphene/blob/30a628d415bf584ab8f6585181cb326d3aa1104a/Pal/src/host/Linux-SGX/db_process.c#L76)):*

Your understanding of the previous implementation is correct. However, the "session key reuse" is **not** about reusing the same key between different enclave pairs. The "session key reuse" remark is about the fact that using AES-CMAC to sign `enclave id` message with `session key` leads to a reuse of this key **in the same session** (between the same enclaves in the pair).
</blockquote></details>

got it, thanks for the clarification.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-Mb2wxPJ44efySsCUmNc:bsbudrr) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*

There is no need in this additional info. We already generate a unique session key for each enclave pair. The unique session key is generated via Diffie-Hellman, which in turn uses a source of entropy (which is just a random byte sequence from `rdrand` x86 instruction).

So since we already have enough entropy from `rdrand`, we don't need to throw in anything else. The result is already "random enough".
</blockquote></details>

Sorry, I didn't explain myself clearly earlier. Here we are trying to verify the report from`parent` and `child` between themselves and since the session key is already common between the `parent` and `child` pair, felt that adding a unique field like `enclave_id` would make the report verification more authentic.
https://github.com/oscarlab/graphene/pull/2380/files#diff-4f43e3fbcfc0a9a863ccbdaf6c52b210f6394570ef51130322603659bf2b319eR112

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-Mb38i7V3Pv09KA976Fc:b4httix) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*
> would make the report verification more authentic

Sorry, I still don't understand. What do you mean by "more authentic"? Why is the hash of the session key (as part of report verification) is not sufficient?

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-Mb3i9Q6CV9KYjXNtF_-:b8kstop) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*

> would make the report verification more authentic

Sorry, I still don't understand. What do you mean by "more authentic"? Why is the hash of the session key (as part of report verification) is not sufficient?
</blockquote></details>

By authentic I mean that it is indeed from the desired enclave whose identity we are trying to validate.  Maybe the `session key` is enough but my comment stems from the following text in our SDM for `EREPORT` instruction which says that "RCX contains the effective address of a 64-byte `REPORTDATA` structure, which allows the caller of the instruction to *associate data with the enclave* from which the instruction is called". 

I see that `session key` isn't unique to a particular enclave. It is shared between the two parties trying to validate each other. If an adversary were to crack this `session key` there is nothing stopping the local attestation to fail. But if we include `enclave_id`, which is unique to an enclave, it gives us another chance to catch the offender when the `report_data` is examined.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-Mb5aiOi0L2yaWAp06pe:b-1veio8) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*
Hm, I don't think including `enclave_id` helps in any way.

`session key` is indeed unique not to an enclave, but to a **pair of enclaves**. But it is not a problem -- enclave A doesn't need to verify itself (and it never verifies itself), so for enclave A the only other enclave that has the expected `session key` is enclave B. In other words, from the perspective of enclave A, `session key` uniquely identifies enclave B. So we can say that `session key` is indeed unique to a particular enclave (from enclave A's viewpoint).

> it gives us another chance to catch the offender when the report_data is examined.

No, it doesn't. For enclave A, to validate `enclave_id` of enclave B, this `enclave_id` must be first sent to enclave A. The only way to send it is in plaintext. So `enclave_id` is known to everyone, including the adversary. Doesn't help at all.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-Mb85fm_B12BDSJn2Ud8:b-h4a45p) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*
>For enclave A, to validate enclave_id of enclave B, this enclave_id must be first sent to enclave A.

We don't need to send the `enclave_id` right. The child will generate the report with its `enclave_id` and will store this information in the `reportdata`. Now when the child sends the report (even using an unsecured stream), the `reportdata` will be protected by `report's` mac which we verify anyway and also SGX guarantees this (SDM sec 37.16.1).
https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/enclave_framework.c#L186
https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/enclave_framework.c#L197

If we don't use a unique field like `enclave_id`, then we may be susceptible to man-in-the-middle attack.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-MbARSWc-AX9JSFgXKhk:br5iqdt) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*
<details><summary><i>Previously, vijaydhanraj (Vijay Dhanraj) wrote…</i></summary><blockquote>

>For enclave A, to validate enclave_id of enclave B, this enclave_id must be first sent to enclave A.

We don't need to send the `enclave_id` right. The child will generate the report with its `enclave_id` and will store this information in the `reportdata`. Now when the child sends the report (even using an unsecured stream), the `reportdata` will be protected by `report's` mac which we verify anyway and also SGX guarantees this (SDM sec 37.16.1).
https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/enclave_framework.c#L186
https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/enclave_framework.c#L197

If we don't use a unique field like `enclave_id`, then we may be susceptible to man-in-the-middle attack.
</blockquote></details>

Something doesn't add up here. The parent enclave must compare the received child's `reportdata` against something. If the child doesn't send `enclave_id` but derives `reportdata` from `enclave_id`, then the parent enclave sees a random sequence of bytes in `reportdata` and has no way of verifying it. Then what's the point of `reportdata` at all in your scenario?

Of course the `report`'s mac verifies the integrity of `reportdata`, but what is the use of this if `reportdata` itself is just an opaque blob for the parent enclave? The whole point of `reportdata` is to have some meaningful information known to both parties, so that one of the parties can check the other party.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-MbCgs4z4K96HrqWwed5:b-sneg6y) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*
> The parent enclave must compare the received child's reportdata against something.

Current code does it in the following way, right?
1. Child adds its generated signed data (session key + child enclave_id), enclave state (flags, child enclave_id) as part of `reportdata` and generates the `report`.
2. Sends it to the parent over the unsecured stream.
3. Parent gets the report key, computes the MAC using this generated report key, and report body (`struct _report_body_t`)
4.  parent compares this MAC with MAC in the report (`report->mac`)

So as a first step we have verified the MAC of the report and have confirmed the report is authentic. This report does include the `enclave_id` which the attacker might not know. (We cannot guarantee the authenticity of child enclave with just the session key as it is not unique to an enclave) 

>Then what's the point of reportdata at all in your scenario?

Next, we verify the data of the report from child (`reportdata`).
1. Parent compares its generated signed data (session key, report->child enclave_id) and the one from the report (report->enclave_data). This step is verifying the actual `reportdata` blob.

I agree that we don't compare the exact `enclave_id` but the point is we are comparing some unique data which only the child knows, and the adversary cannot tamper with this data.

2. In the current code, we also compare parent mr_enclave with the one in the report (`eport.body.mr_enclave`).

So, to me having `enclave_id` does seem to add value.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 0 of 8 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-MbFzzo16cst_zDBH5uG:b-tcrbf4) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*
> we are comparing some unique data which only the child knows, and the adversary cannot tamper with this data.

No, this statement is wrong. **"Only the child knows"** is wrong -- `enclave_id` is generated by the child and then **sent in the clear** inside the `reportdata` bytes. So the adversary can intercept this `report` from the child, look at `report.reportdata` bytes, and learn the `enclave_id`.

The current code basically did redundant, useless work by generating this `enclave_id` and checking it. It doesn't add any value.

mkow (changes requested):
> Reviewed 5 of 8 files at r1, 6 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/db_process.c, line 89 at r3](https://reviewable.io/reviews/oscarlab/graphene/2380#-MbktYCL2F9lgkjRgoM2:-MbktYCL2F9lgkjRgoM3:b-4egago) ([raw file](https://github.com/oscarlab/graphene/blob/30a628d415bf584ab8f6585181cb326d3aa1104a/Pal/src/host/Linux-SGX/db_process.c#L89)):*
> ```C
> 
>     /* newly established session key between two enclaves is the only secure component here */
>     ret = lib_SHA256Update(&sha, (uint8_t*)session_key, sizeof(*session_key));
> ```

I wonder whether we shouldn't tag the key somehow to ensure that the revealed data (`sha256(session_key)`) can't be reused by the attacker to forge something in TLS. E.g. suffix the hash input with "GRAPHENE_LOCAL_ATTESTATION_CHALLENGE".
___
*[Pal/src/host/Linux-SGX/db_process.c, line 99 at r3](https://reviewable.io/reviews/oscarlab/graphene/2380#-Mbkspbu7H4jIiXcv_TN:-Mbkspbu7H4jIiXcv_TO:bmjma8v) ([raw file](https://github.com/oscarlab/graphene/blob/30a628d415bf584ab8f6585181cb326d3aa1104a/Pal/src/host/Linux-SGX/db_process.c#L99)):*
> ```C
>     return 0;
> fail:
>     log_error("Failed to hash the session key between two enclaves: %d\n", ret);
> ```

I'd drop this message and just `return ret`. This is impossible AFAIK, so we don't need any nice-for-users handling.
___
*[Pal/src/host/Linux-SGX/db_process.c, line 200 at r3](https://reviewable.io/reviews/oscarlab/graphene/2380#-MbktKB83WsxcVEgYi7q:-MbktKB83WsxcVEgYi7r:bojj1ee) ([raw file](https://github.com/oscarlab/graphene/blob/30a628d415bf584ab8f6585181cb326d3aa1104a/Pal/src/host/Linux-SGX/db_process.c#L200)):*
> ```C
> int init_child_process(PAL_HANDLE* parent_handle) {
>     if (g_pal_sec.enclave_flags & PAL_ENCLAVE_INITIALIZED)
>         return false;
> ```

This function returns `int`. Weird that there's no compiler warning :/
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-Mbkvxve97B22vPBDjH5:b-k8h9zj) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*

> we are comparing some unique data which only the child knows, and the adversary cannot tamper with this data.

No, this statement is wrong. **"Only the child knows"** is wrong -- `enclave_id` is generated by the child and then **sent in the clear** inside the `reportdata` bytes. So the adversary can intercept this `report` from the child, look at `report.reportdata` bytes, and learn the `enclave_id`.

The current code basically did redundant, useless work by generating this `enclave_id` and checking it. It doesn't add any value.
</blockquote></details>

I don't have time to engage in this discussion, but I just want to confirm that @dimakuv is right in everything he wrote, the current code is good and sending `enclave_id` doesn't seem useful for anything.
___
*[Pal/src/host/Linux-SGX/pal_security.h, line 10 at r2](https://reviewable.io/reviews/oscarlab/graphene/2380#-Ma-3G808Lq19QN-gEDS:-Ma-3G8184wqz4pg8dXz:bhn1ydi) ([raw file](https://github.com/oscarlab/graphene/blob/9df98a9a6d2a6701bc00ea865c1edc7f367163a7/Pal/src/host/Linux-SGX/pal_security.h#L10)):*
> ```objc
> #include "sgx_arch.h"
> 
> #define PAL_ENCLAVE_INITIALIZED 0x0001ULL
> ```

Declaring bit flags as unsigned is IMO dangerous, as `some_field & ~FLAG` doesn't work as expected if the sizes doesn't match (here they do match, but it's not future-proof IMO).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 6 of 8 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/db_process.c, line 89 at r3](https://reviewable.io/reviews/oscarlab/graphene/2380#-MbktYCL2F9lgkjRgoM2:-Mbl8g_i4tGZwly8i8ky:bc2h37a) ([raw file](https://github.com/oscarlab/graphene/blob/30a628d415bf584ab8f6585181cb326d3aa1104a/Pal/src/host/Linux-SGX/db_process.c#L89)):*

I wonder whether we shouldn't tag the key somehow to ensure that the revealed data (`sha256(session_key)`) can't be reused by the attacker to forge something in TLS. E.g. suffix the hash input with "GRAPHENE_LOCAL_ATTESTATION_CHALLENGE".
</blockquote></details>

Done, but I'm not sure how exactly this helps. Could you explain a bit more?
___
*[Pal/src/host/Linux-SGX/db_process.c, line 99 at r3](https://reviewable.io/reviews/oscarlab/graphene/2380#-Mbkspbu7H4jIiXcv_TN:-Mbl6BUU4iaAwUf15LPm:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/30a628d415bf584ab8f6585181cb326d3aa1104a/Pal/src/host/Linux-SGX/db_process.c#L99)):*

I'd drop this message and just `return ret`. This is impossible AFAIK, so we don't need any nice-for-users handling.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_process.c, line 200 at r3](https://reviewable.io/reviews/oscarlab/graphene/2380#-MbktKB83WsxcVEgYi7q:-MblA1091cvufIx-77rG:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/30a628d415bf584ab8f6585181cb326d3aa1104a/Pal/src/host/Linux-SGX/db_process.c#L200)):*

This function returns `int`. Weird that there's no compiler warning :/
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/pal_security.h, line 10 at r2](https://reviewable.io/reviews/oscarlab/graphene/2380#-Ma-3G808Lq19QN-gEDS:-Mbl5uXo2tMuX5JbzgKj:b35wj2g) ([raw file](https://github.com/oscarlab/graphene/blob/9df98a9a6d2a6701bc00ea865c1edc7f367163a7/Pal/src/host/Linux-SGX/pal_security.h#L10)):*

Declaring bit flags as unsigned is IMO dangerous, as `some_field & ~FLAG` doesn't work as expected if the sizes doesn't match (here they do match, but it's not future-proof IMO).
</blockquote></details>

Done. Good point.

vijaydhanraj (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: 6 of 8 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/db_process.c, line 160 at r4](https://reviewable.io/reviews/oscarlab/graphene/2380#-Mbm5aKd9AeJunkyI1eF:-Mbm5aKd9AeJunkyI1eG:bn2729l) ([raw file](https://github.com/oscarlab/graphene/blob/246138b0991d83ce16630a85c8d85455dd520f4c/Pal/src/host/Linux-SGX/db_process.c#L160)):*
> ```C
>         goto failed;
> 
>     __sgx_mem_aligned sgx_report_data_t sgx_report_data;
> ```

Wondering if we need a mem aligned struct here? The SDM just says that `REPORTDATA` is a `64-byte` data structure. If it is required, should we use 64-byte alignment (`alignas(64)`) instead of 512-byte?

One thing I noted is that the prototype text says it should be `128B aligned` https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux-SGX/sgx_api.h#L26. So, what would be the right one to use?
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 898 at r1](https://reviewable.io/reviews/oscarlab/graphene/2380#-M_u7wP3CCOwIPP24SQY:-Mbm91HNAaV7y1YOIIoP:b-sg6tpp) ([raw file](https://github.com/oscarlab/graphene/blob/e4b7468cbd8b688f4f39599799b52e5109a91294/Pal/src/host/Linux-SGX/enclave_framework.c#L898)):*

I don't have time to engage in this discussion, but I just want to confirm that @dimakuv is right in everything he wrote, the current code is good and sending `enclave_id` doesn't seem useful for anything.
</blockquote></details>

Had an offline chat with @dimakuv and it makes sense that `session key` should be enough. Just that if `session key` is compromised then the communication is compromised. But I don't think we can do anything about it from the hardware perspective. 
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 875 at r3](https://reviewable.io/reviews/oscarlab/graphene/2380#-Mb2aiZ-5YD3YsVWlXnA:-MbmB5YYEIVYXyt0QWas:b-rxvaxh) ([raw file](https://github.com/oscarlab/graphene/blob/30a628d415bf584ab8f6585181cb326d3aa1104a/Pal/src/host/Linux-SGX/enclave_framework.c#L875)):*

FYI: @vijaydhanraj This is where we generate a brand new key for each pair of enclaves. We use Diffie-Hellman to create a shared secret and then use a KDF (currently SHA256, but see https://github.com/oscarlab/graphene/pull/2410) to generate the final session key. Note that the generated session key is unique for each pair of enclaves, so there is no reuse of the key between different enclave pairs.
</blockquote></details>

Got it, will also take a look at #2410.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @vijaydhanraj)

___
*[Pal/src/host/Linux-SGX/db_process.c, line 89 at r3](https://reviewable.io/reviews/oscarlab/graphene/2380#-MbktYCL2F9lgkjRgoM2:-MbmhsMJDUesEYC-UIMd:bpybyt0) ([raw file](https://github.com/oscarlab/graphene/blob/30a628d415bf584ab8f6585181cb326d3aa1104a/Pal/src/host/Linux-SGX/db_process.c#L89)):*

Done, but I'm not sure how exactly this helps. Could you explain a bit more?
</blockquote></details>

I think it's just a "good practice" when doing crypto stuff. It's a safeguard in case some other part of our crypto also hashed the key, but assumed that if key is unknown to the attacker, then the hash is also unknown. I.e. when creating a crypto protocol/API which gets a key from the user you just don't expect that some properties of it are known to the attacker.

mkow (approved):
> Reviewed 6 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2380)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2381, opened 2021-05-17
[LibOS,Pal] Move common lib (`graphene-lib.a`) under `graphene-root/Common`

dimakuv:
## Description of the changes 

Logically, the common library linked in PAL binaries and in the LibOS library is a separate entity in the directory hierarchy. Previously, it was hidden under `graphene-root/Pal` for no good reason.

This PR moves it into its own root directory, and also fixes small inconsistencies (`pal_error.h` is moved into common lib, `pal_debug.h` is merged into `pal.h`, `abort.c` is renamed to `utils.c`).

## How to test this PR? 

All tests must pass, no changes here.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: 0 of 127 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vhvr3AyswPIBRNIdD:-M_vhvr402UVAeLR0rUr:be3iklr) (no related file):*
@woju @mkow @boryspoplawski and maybe @pwmarcz What do you think about this PR and the new directory hierarchy? It is easier to check the changes by browsing my branch: https://github.com/oscarlab/graphene/tree/dimakuv%2Fmv-pal-lib-to-common

If you don't like the name `Common/`, please suggest variants.

---

boryspoplawski (changes requested):
> Reviewed 127 of 127 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv, @mkow, @pwmarcz, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vhvr3AyswPIBRNIdD:-M_vk6c71i67lbPCsS9q:byfmadg) (no related file):*

@woju @mkow @boryspoplawski and maybe @pwmarcz What do you think about this PR and the new directory hierarchy? It is easier to check the changes by browsing my branch: https://github.com/oscarlab/graphene/tree/dimakuv%2Fmv-pal-lib-to-common

If you don't like the name `Common/`, please suggest variants.
</blockquote></details>

At last! Thanks for fixing this.

I like the name, but I wouldn't capitalize it (i.e. `common`). I know the rest of the directories are like that, but I would move away from it.

---
___
*[LibOS/shim/test/regression/Makefile, line 142 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vpDZe1A5laRRI0vk5:-M_vpDZe1A5laRRI0vk6:b-jzbixm) ([raw file](https://github.com/oscarlab/graphene/blob/d765e4ce9832e0039de1b8399ca42377aaa36866/LibOS/shim/test/regression/Makefile#L142)):*
> ```Makefile
> CFLAGS-futex_wake_op = -pthread
> CFLAGS-proc_common = -pthread
> CFLAGS-spinlock += -I$(PALDIR)/../../Common/include -I$(PALDIR)/../../Common/include/arch/$(ARCH) -pthread
> ```

NP: This is now unrelated to PAL, so please just change to appropriate amount of `../` (we use that version in this Makefile anyway).
Not a big deal, meson rewrite should get rid of these.
___
*[Pal/include/pal/pal.h, line 845 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vqoe-30IwdPbSSF15:-M_vqoe-30IwdPbSSF16:br7i95x) ([raw file](https://github.com/oscarlab/graphene/blob/d765e4ce9832e0039de1b8399ca42377aaa36866/Pal/include/pal/pal.h#L845)):*
> ```objc
> void pal_vprintf(const char* fmt, va_list ap) __attribute__((format(printf, 1, 0)));
> 
> void warn(const char* format, ...);
> ```

Please remove this, it's already declared in `assert.h`.
___
*[Pal/include/lib/slabmgr.h, line 357 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vntwB-3XnlfSpy1Fz:-M_vnRyd4pnplZxEoDL5:b-3tuw3j) ([raw file](https://github.com/oscarlab/graphene/blob/8c7ac33f80d403b388248ea7827996cb0cc74019/Pal/include/lib/slabmgr.h#L357)):*
> ```objc
> 
>     if (level >= SLAB_LEVEL) {
>         log_error("Heap corruption detected: invalid heap level %u\n", level);
> ```

Why this change? I know it's not pretty (we use a macro defined by every target, i.e. LibOS and Pal,  in common header), but it was like this before and the ideal fix would be different...
___
*[Pal/include/lib/slabmgr.h, line 393 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vnwK3-uG2d5-Lehgk:-M_vnwK3-uG2d5-Lehgl:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/8c7ac33f80d403b388248ea7827996cb0cc74019/Pal/include/lib/slabmgr.h#L393)):*
> ```objc
>      * so a level of 0 in the header would no longer be a valid level. */
>     if (level >= SLAB_LEVEL) {
>         log_error("Heap corruption detected: invalid heap level %d\n", level);
> ```

ditto

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski, @dimakuv, @mkow, @pwmarcz, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vhvr3AyswPIBRNIdD:-M_vxuLRCXN7_y9Oq_I_:b5q66y2) (no related file):*
> I like the name, but I wouldn't capitalize it (i.e. common). I know the rest of the directories are like that, but I would move away from it.

This is actually @woju's suggestion to capitalize "codebase components". This seems to be one of the frequently used conventions. @woju may explain this better. IIRC, the argument is that capitalized dirs are listed before other, lower-case dirs -- which visually separates the "real code" from the "tools".

---
___
*[Pal/include/lib/slabmgr.h, line 357 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vntwB-3XnlfSpy1Fz:-M_vytoh96PkkZtGGtsk:b-onrlx1) ([raw file](https://github.com/oscarlab/graphene/blob/8c7ac33f80d403b388248ea7827996cb0cc74019/Pal/include/lib/slabmgr.h#L357)):*

Why this change? I know it's not pretty (we use a macro defined by every target, i.e. LibOS and Pal,  in common header), but it was like this before and the ideal fix would be different...
</blockquote></details>

I actually wanted to remove it from here because it's weird to have messages printed from the "standard library". But then I decided to keep at least something, since it's pretty important. And I decided that since we do `abort()`, then it makes sense to use `warn()` -- just like what asserts/FORTIFY_SOURCE in Glibc/gcc do.

But Ok, I'll revert this.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski, @dimakuv, @mkow, @pwmarcz, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vhvr3AyswPIBRNIdD:-M_w8K-X6-tvgDw4SvFf:b-j0f6vx) (no related file):*

> I like the name, but I wouldn't capitalize it (i.e. common). I know the rest of the directories are like that, but I would move away from it.

This is actually @woju's suggestion to capitalize "codebase components". This seems to be one of the frequently used conventions. @woju may explain this better. IIRC, the argument is that capitalized dirs are listed before other, lower-case dirs -- which visually separates the "real code" from the "tools".
</blockquote></details>

We currently also have `Examples`, `Tools`...
@woju I would love to hear reasoning behind it

---

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv, @mkow, @pwmarcz, and @woju)

___
*[Pal/include/lib/slabmgr.h, line 357 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vntwB-3XnlfSpy1Fz:-M_w8dOd8OVyPlHdYh32:bqur9hy) ([raw file](https://github.com/oscarlab/graphene/blob/8c7ac33f80d403b388248ea7827996cb0cc74019/Pal/include/lib/slabmgr.h#L357)):*

I actually wanted to remove it from here because it's weird to have messages printed from the "standard library". But then I decided to keep at least something, since it's pretty important. And I decided that since we do `abort()`, then it makes sense to use `warn()` -- just like what asserts/FORTIFY_SOURCE in Glibc/gcc do.

But Ok, I'll revert this.
</blockquote></details>

Yeah, but I think @mkow wanted to get rid of this `warn` completely.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: 123 of 128 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, @pwmarcz, and @woju)

___
*[LibOS/shim/test/regression/Makefile, line 142 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vpDZe1A5laRRI0vk5:-M_y5stP6mkK5f49IxSN:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d765e4ce9832e0039de1b8399ca42377aaa36866/LibOS/shim/test/regression/Makefile#L142)):*

NP: This is now unrelated to PAL, so please just change to appropriate amount of `../` (we use that version in this Makefile anyway).
Not a big deal, meson rewrite should get rid of these.
</blockquote></details>

Done.
___
*[Pal/include/pal/pal.h, line 845 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vqoe-30IwdPbSSF15:-M_y61eW3xBCEXkN3YN6:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d765e4ce9832e0039de1b8399ca42377aaa36866/Pal/include/pal/pal.h#L845)):*

Please remove this, it's already declared in `assert.h`.
</blockquote></details>

Done.
___
*[Pal/include/lib/slabmgr.h, line 357 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vntwB-3XnlfSpy1Fz:-M_y5Ok49cX8JopVwlKD:b-3rcnly) ([raw file](https://github.com/oscarlab/graphene/blob/8c7ac33f80d403b388248ea7827996cb0cc74019/Pal/include/lib/slabmgr.h#L357)):*

Yeah, but I think @mkow wanted to get rid of this `warn` completely.
</blockquote></details>

Done. Not pretty, but let's keep like this...
___
*[Pal/include/lib/slabmgr.h, line 393 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vnwK3-uG2d5-Lehgk:-M_y5Sa98PQzpumtq3i-:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/8c7ac33f80d403b388248ea7827996cb0cc74019/Pal/include/lib/slabmgr.h#L393)):*

ditto
</blockquote></details>

Done.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: 123 of 129 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vhvr3AyswPIBRNIdD:-M_yFLLN5m7SqMbBIm0-:b-teglf1) (no related file):*
"Common" seems pretty general and not that different from "Tools", but I don't have any better suggestion. I agree that it would help to have a convention that emphasizes project source (I guess `Pal`, `LibOS`, `Common`) from everything else.

>This seems to be one of the frequently used conventions.

I don't think I've seen that, can you give an example? I know of the opposite (Linux capitalizing `Documentation`).

The most frequent convention I see is putting the source one level down, in a nested directory like `<project name>` (so we would have something like `graphene/libos`, `graphene/pal`, `graphene/common`), or even `src`.

(BTW, I would remove the additional level of nesting in `LibOS/shim`, and move patched glibc/gcc to its own directory, but I think that's worth doing regardless of everything else).

Anyway, that can be decided separately, for now I'd just keep `Common`. It's a great cleanup to have a separate component there!

---

mkow (changes requested):
> Reviewed 118 of 127 files at r1, 17 of 17 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vhvr3AyswPIBRNIdD:-M_zge8U1s1jieyRGYO1:bdez1t6) (no related file):*
> (BTW, I would remove the additional level of nesting in LibOS/shim, and move patched glibc/gcc to its own directory, but I think that's worth doing regardless of everything else).

+1 to dropping /shim. Ad glibc, not sure if I understood Paweł, but it should stay under LibOS, probably in a separate directory (something like "libc-patches"?).

Ad "Common" and capitalization - no hard feelings on this, I'm not sure which way is better...

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_z_j2O3x2c4i5tImAu:-M_z_j2O3x2c4i5tImAv:bs75dul) (no related file):*
There are still some leftovers in LICENSE.addendum.txt and LibOS/shim/src/shim_malloc.c (but please grep the whole codebase yourself and doublecheck).

---
___
*[Common/Makefile, line 20 at r2](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_zVBt057a2ZHvCUQ6n:-M_zVBt057a2ZHvCUQ6o:b5k7h0q) ([raw file](https://github.com/oscarlab/graphene/blob/4de7148b62b36e9f864c0d5e8341282bac0535b7/Common/Makefile#L20)):*
> ```Makefile
> sgx-tokens:
> 
> .PHONY: distclean
> ```

You already have an equivalent declaration above
___
*[LibOS/shim/test/regression/syscall.c, line 3 at r2](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_zeUDiAo29S3bnvRss:-M_zeUDiAo29S3bnvRst:b-rlcsvl) ([raw file](https://github.com/oscarlab/graphene/blob/4de7148b62b36e9f864c0d5e8341282bac0535b7/LibOS/shim/test/regression/syscall.c#L3)):*
> ```C
> #define _GNU_SOURCE
> #include <unistd.h>
> #include <sys/syscall.h>
> ```

Please sort
___
*[LibOS/shim/test/regression/syscall.c, line 7 at r2](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_ze_Ij0fkMr6JaHAV6:-M_ze_Ij0fkMr6JaHAV7:b-rppigx) ([raw file](https://github.com/oscarlab/graphene/blob/4de7148b62b36e9f864c0d5e8341282bac0535b7/LibOS/shim/test/regression/syscall.c#L7)):*
> ```C
> int main(int argc, char** argv) {
>     const char buf[] = "Hello world\n";
>     syscall(__NR_write, 1, buf, sizeof(buf) - 1);
> ```

I prefer the old version, we should get rid of the call to the standard library (which probably work only due to `syscall` being a macro?).
___
*[Pal/regression/utils.c, line 37 at r2](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_zfEUx7kqmPA7VvG2e:-M_zfEUx7kqmPA7VvG2f:b-fgwz5l) ([raw file](https://github.com/oscarlab/graphene/blob/4de7148b62b36e9f864c0d5e8341282bac0535b7/Pal/regression/utils.c#L37)):*
> ```C
>     // ENOTRECOVERABLE = 131
>     DkProcessExit(131);
> }
> ```

While we're here, could we add `__stack_chk_fail` here and enable stack cookies for Common/?
___
*[Pal/include/lib/slabmgr.h, line 357 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vntwB-3XnlfSpy1Fz:-M_zWdtW1g3o-0y3xBP6:b7aa0iv) ([raw file](https://github.com/oscarlab/graphene/blob/8c7ac33f80d403b388248ea7827996cb0cc74019/Pal/include/lib/slabmgr.h#L357)):*

Done. Not pretty, but let's keep like this...
</blockquote></details>

Yup, we should get rid of `warn()`.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, @mkow, and @woju)

___
*[LibOS/shim/test/regression/syscall.c, line 7 at r2](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_ze_Ij0fkMr6JaHAV6:-Ma-20V09wexxh7wkERM:bzdn887) ([raw file](https://github.com/oscarlab/graphene/blob/4de7148b62b36e9f864c0d5e8341282bac0535b7/LibOS/shim/test/regression/syscall.c#L7)):*

I prefer the old version, we should get rid of the call to the standard library (which probably work only due to `syscall` being a macro?).
</blockquote></details>

But why? The old version depends on the internal Graphene headers (`<sysdep-arch.h>`), why would we want this?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: 132 of 136 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_z_j2O3x2c4i5tImAu:-Ma-eqsZ-WqdHiDwxFAU:bll3jka) (no related file):*

There are still some leftovers in LICENSE.addendum.txt and LibOS/shim/src/shim_malloc.c (but please grep the whole codebase yourself and doublecheck).
</blockquote></details>

Done. No other places found.

---
___
*[Common/Makefile, line 20 at r2](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_zVBt057a2ZHvCUQ6n:-Ma-exE6B-piBMiK_eM3:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/4de7148b62b36e9f864c0d5e8341282bac0535b7/Common/Makefile#L20)):*

You already have an equivalent declaration above
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/syscall.c, line 3 at r2](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_zeUDiAo29S3bnvRss:-Ma-f4WDEE7OCE04XMNa:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/4de7148b62b36e9f864c0d5e8341282bac0535b7/LibOS/shim/test/regression/syscall.c#L3)):*

Please sort
</blockquote></details>

Done.
___
*[Pal/regression/utils.c, line 37 at r2](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_zfEUx7kqmPA7VvG2e:-Ma-kAdVBLuPfQ7WgxEF:bl01eiq) ([raw file](https://github.com/oscarlab/graphene/blob/4de7148b62b36e9f864c0d5e8341282bac0535b7/Pal/regression/utils.c#L37)):*

While we're here, could we add `__stack_chk_fail` here and enable stack cookies for Common/?
</blockquote></details>

Sorry, can't do. Just tried it now, but we have an Untrusted PAL in our SGX code, and this one uses a different Stack Protector. And this works with "no protection" in the Common library, but if I would enable the Stack Protector in Common, then trusted PAL works correctly but untrusted PAL calls into the Common functions, and everything breaks (because of the mismatch in Stack Protector implementations in untrusted PAL and in the shared Common code). Damn, a single Common library is really troublesome.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @woju)

___
*[LibOS/shim/test/regression/syscall.c, line 7 at r2](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_ze_Ij0fkMr6JaHAV6:-Ma-kpv3D3Yx_dC1bQaR:b-gb9o9s) ([raw file](https://github.com/oscarlab/graphene/blob/4de7148b62b36e9f864c0d5e8341282bac0535b7/LibOS/shim/test/regression/syscall.c#L7)):*

But why? The old version depends on the internal Graphene headers (`<sysdep-arch.h>`), why would we want this?
</blockquote></details>

Aaah, wait, this is LibOS regression, not PAL. I confused the paths.
___
*[Pal/regression/utils.c, line 37 at r2](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_zfEUx7kqmPA7VvG2e:-Ma-l8CV3egXNWIYqG6S:b32vhv) ([raw file](https://github.com/oscarlab/graphene/blob/4de7148b62b36e9f864c0d5e8341282bac0535b7/Pal/regression/utils.c#L37)):*

Sorry, can't do. Just tried it now, but we have an Untrusted PAL in our SGX code, and this one uses a different Stack Protector. And this works with "no protection" in the Common library, but if I would enable the Stack Protector in Common, then trusted PAL works correctly but untrusted PAL calls into the Common functions, and everything breaks (because of the mismatch in Stack Protector implementations in untrusted PAL and in the shared Common code). Damn, a single Common library is really troublesome.
</blockquote></details>

:(

boryspoplawski (changes requested):
> Reviewed 16 of 17 files at r2, 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vhvr3AyswPIBRNIdD:-Ma-Y6eO7NnCDC-enx3l:b7v7rl8) (no related file):*

> (BTW, I would remove the additional level of nesting in LibOS/shim, and move patched glibc/gcc to its own directory, but I think that's worth doing regardless of everything else).

+1 to dropping /shim. Ad glibc, not sure if I understood Paweł, but it should stay under LibOS, probably in a separate directory (something like "libc-patches"?).

Ad "Common" and capitalization - no hard feelings on this, I'm not sure which way is better...
</blockquote></details>

AD capitalization: I would prefer the style from Linux kernel (i.e. "pal", "libos", "common", but "Examples"). I guess it will be easier to **not** do this in this PR, so having "Common" for uniformity sounds like a good idea.

AD dropping "shim": if we keep glibc patches inside "LibOS", then we have to somehow split the source from patches and in such case "shim" sounds like an ok name (only because most of our functions start with `shim_`, not because it's a good name per se). If we would move it to top-level directory, then dropping "shim" is fine.

---
___
*[LibOS/shim/test/regression/Makefile, line 142 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vpDZe1A5laRRI0vk5:-Ma0bq4G5U5AqJICWZpY:b4agk45) ([raw file](https://github.com/oscarlab/graphene/blob/d765e4ce9832e0039de1b8399ca42377aaa36866/LibOS/shim/test/regression/Makefile#L142)):*

Done.
</blockquote></details>

Why `iquote`?
___
*[LibOS/shim/test/regression/syscall.c, line 7 at r3](https://reviewable.io/reviews/oscarlab/graphene/2381#-Ma0cF1JANaj6v-oflyC:-Ma0cF1JANaj6v-oflyD:bt5hkxb) ([raw file](https://github.com/oscarlab/graphene/blob/f92876f28ef4ba9459fd31fbe0d4cc24721d1ba9/LibOS/shim/test/regression/syscall.c#L7)):*
> ```C
> int main(int argc, char** argv) {
>     const char buf[] = "Hello world\n";
>     syscall(__NR_write, 1, buf, sizeof(buf) - 1);
> ```

While you touch this, maybe time to add error checking as well? :)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @woju)

___
*[LibOS/shim/test/regression/Makefile, line 142 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vpDZe1A5laRRI0vk5:-Ma0f759AvZftaMNKEgB:bholylq) ([raw file](https://github.com/oscarlab/graphene/blob/d765e4ce9832e0039de1b8399ca42377aaa36866/LibOS/shim/test/regression/Makefile#L142)):*

Why `iquote`?
</blockquote></details>

I guess we should even use it everywhere for Graphene includes? It ensures that you use doublequotes in `#include`, which is IMO a good convention for non-system includes.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @woju)

___
*[LibOS/shim/test/regression/Makefile, line 142 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vpDZe1A5laRRI0vk5:-Ma0fvQxB_QMcYjgNR5V:b-fvni38) ([raw file](https://github.com/oscarlab/graphene/blob/d765e4ce9832e0039de1b8399ca42377aaa36866/LibOS/shim/test/regression/Makefile#L142)):*

I guess we should even use it everywhere for Graphene includes? It ensures that you use doublequotes in `#include`, which is IMO a good convention for non-system includes.
</blockquote></details>

We should have no system include at all and use `-nostdinc`, but that's a completely different story.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: 135 of 136 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vhvr3AyswPIBRNIdD:-Ma2FODfDaL2YoUvPX3g:bel0j7j) (no related file):*

AD capitalization: I would prefer the style from Linux kernel (i.e. "pal", "libos", "common", but "Examples"). I guess it will be easier to **not** do this in this PR, so having "Common" for uniformity sounds like a good idea.

AD dropping "shim": if we keep glibc patches inside "LibOS", then we have to somehow split the source from patches and in such case "shim" sounds like an ok name (only because most of our functions start with `shim_`, not because it's a good name per se). If we would move it to top-level directory, then dropping "shim" is fine.
</blockquote></details>

I am still waiting for @woju's opinion, but looks like people are generally Ok with `Common` name. So moving to a better naming scheme (capital vs lower-case letters) is for another PR.

---
___
*[LibOS/shim/test/regression/Makefile, line 142 at r1](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vpDZe1A5laRRI0vk5:-Ma2FfKu3-33SZvVZ-TX:b-79zopk) ([raw file](https://github.com/oscarlab/graphene/blob/d765e4ce9832e0039de1b8399ca42377aaa36866/LibOS/shim/test/regression/Makefile#L142)):*

We should have no system include at all and use `-nostdinc`, but that's a completely different story.
</blockquote></details>

@boryspoplawski This is because `spinlock` and `attestation` tests are "normal" Linux applications, but on the other hand they use LibOS-internal headers. I wanted to make sure that "normal" headers (e.g. `<stdin.h>`) are included from host standard headers, and "Graphene" headers (e.g. `"sgx_arch.h"`) are included via `iquote`. Recall that `-I` **prepends to the list of standard headers**, whereas `-iquote` **appends to the list of additional headers**.

I actually hit this because `attestation` includes `assert.h` which should be taken from host headers, but got pulled in from Graphene-internal headers (and build failed because our assert uses Graphene-internal functions). The header was pulled because `-I` prepends to the list of standard headers, whereas `-iquote` does the expected thing. So I decided to replace it everywhere.

In general, I agree that we should simply use `-nostdinc` in internal Graphene sources. But this LibOS regression tests' case is different.
___
*[LibOS/shim/test/regression/syscall.c, line 7 at r3](https://reviewable.io/reviews/oscarlab/graphene/2381#-Ma0cF1JANaj6v-oflyC:-Ma2HxS054ZvmA2uj28g:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f92876f28ef4ba9459fd31fbe0d4cc24721d1ba9/LibOS/shim/test/regression/syscall.c#L7)):*

While you touch this, maybe time to add error checking as well? :)
</blockquote></details>

Done.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @woju)

___
*[LibOS/shim/test/regression/syscall.c, line 11 at r4](https://reviewable.io/reviews/oscarlab/graphene/2381#-Ma3Qs7I8MPFG9pASB-6:-Ma3Qs7R-Gab8ESySMBh:b-cgchix) ([raw file](https://github.com/oscarlab/graphene/blob/36963766f72a8723d3778274355d9cecd8a026f5/LibOS/shim/test/regression/syscall.c#L11)):*
> ```C
>     long ret = syscall(__NR_write, 1, buf, sizeof(buf) - 1);
>     if (ret < 0)
>         err(EXIT_FAILURE, "write syscall");
> ```

You should use `errx()`, as errno is not set.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @woju)

___
*[LibOS/shim/test/regression/syscall.c, line 11 at r4](https://reviewable.io/reviews/oscarlab/graphene/2381#-Ma3Qs7I8MPFG9pASB-6:-Ma3TE9cCqu_PvbBqPbB:b83c2w6) ([raw file](https://github.com/oscarlab/graphene/blob/36963766f72a8723d3778274355d9cecd8a026f5/LibOS/shim/test/regression/syscall.c#L11)):*

You should use `errx()`, as errno is not set.
</blockquote></details>

No, this version is fine. `syscall` sets `errno`.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, and @woju)

___
*[LibOS/shim/test/regression/syscall.c, line 11 at r4](https://reviewable.io/reviews/oscarlab/graphene/2381#-Ma3Qs7I8MPFG9pASB-6:-Ma3TqmN1UmPJz_B6P1e:b-weno57) ([raw file](https://github.com/oscarlab/graphene/blob/36963766f72a8723d3778274355d9cecd8a026f5/LibOS/shim/test/regression/syscall.c#L11)):*

No, this version is fine. `syscall` sets `errno`.
</blockquote></details>

As Borys mentioned, no need to use `errx()`, errno is correctly set by `syscall()` wrapper in Glibc.

dimakuv:
Jenkins, retest Jenkins-20.04 please (`kill11` LTP test failed)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vhvr3AyswPIBRNIdD:-Ma4bXtzDcwtZoE8gYCA:boc7sd0) (no related file):*

The rationale is that capitalized thing are "meta", for things like documentation or README, or Makefile. All other "usual" things should be lowercase. The reason is sorting: many tools (including github's landing: https://github.com/oscarlab/graphene) sort all uppercase before any lowercase, so Documentation would be first (well, third after `.ci` and `.github`). (If you have other locales like `de_DE.UTF-8`, you won't see this in `ls`, because in unicode locales lower and uppercase are not distinguished at all).

`graphene/` convention would be for Python, because the package directory needs to be named exactly the same as import name. This is not true for C, Java and others, so `src/` is a viable option, but not a neccesity (we can just name things `libos/`, `pal/linux` etc. So @boryspoplawski is right that `Documentation` and `Examples` should be upper-cased, but the rest should be lower-cased.

I have no strong opinion here. We can do this later, when we'll be moving other things, although I think this PR would be a good opportunity to move to lowercase `common/`, because it will be one git conflict less. We already have `python` and `tests`. But I don't think this issue should be blocking this PR, we can discuss the optimal layout in some separate issue/RFC.
</blockquote></details>

IMO we should do it, but not in this PR

---

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/test/regression/syscall.c, line 11 at r4](https://reviewable.io/reviews/oscarlab/graphene/2381#-Ma3Qs7I8MPFG9pASB-6:-Ma55f2AD8ljtPdxtM3_:b-2v0flf) ([raw file](https://github.com/oscarlab/graphene/blob/36963766f72a8723d3778274355d9cecd8a026f5/LibOS/shim/test/regression/syscall.c#L11)):*

As Borys mentioned, no need to use `errx()`, errno is correctly set by `syscall()` wrapper in Glibc.
</blockquote></details>

Oh, I thought that `syscall()` is more low-level.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @mkow, @pwmarcz, and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vhvr3AyswPIBRNIdD:-Ma5E57Q39JuQc2zJJN3:biqsjpm) (no related file):*

IMO we should do it, but not in this PR
</blockquote></details>

Thanks, @woju! I remembered it wrong then, it's the opposite.

Let me fix this to `common` (lower-case) in this PR. We should start with something, and this PR is a good point. I'm blocking myself till I'm done with it.

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: 69 of 136 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2381#-M_vhvr3AyswPIBRNIdD:-Ma7SwCm1wMsGpkHVPy4:b720u2b) (no related file):*

Thanks, @woju! I remembered it wrong then, it's the opposite.

Let me fix this to `common` (lower-case) in this PR. We should start with something, and this PR is a good point. I'm blocking myself till I'm done with it.
</blockquote></details>

Done. `Common/` is renamed to `common/` (lower-case).

---

mkow (approved):
> Reviewed 67 of 67 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (approved):
> Reviewed 8 of 8 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

boryspoplawski (approved):
> Reviewed 66 of 67 files at r5, 8 of 8 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2381)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2384, opened 2021-05-19
[Pal/Linux-SGX] Do not use Exitless for clone OCALL

dimakuv:
## Description of the changes 

Exitless feature distinguishes between normal enclave threads and untrusted RPC threads (the latter submit syscall requests on behalf of the former). Enclave threads catch all sync and async signals from the host, including SIGILL. RPC threads block all sync and async signals other than SIGUSR2.

This PR fixes a bug that clone OCALL was considered suitable for Exitless, which meant that the new enclave thread was created on behalf of the RPC tread, and thus inherited its signal mask. Thus, the new enclave thread couldn't handle e.g. SIGILL exceptions (which can happen on legit CPUID and RDTSC instructions), and Graphene terminated the whole process.

## How to test this PR? 

All tests must pass. This bug was found on our Redis example with Exitless enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2384)

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_ocalls.c
+++ b/Pal/src/host/Linux-SGX/enclave_ocalls.c
@@ -854,9 +854,12 @@ int ocall_resume_thread(void* tcs) {
 int ocall_clone_thread(void) {
     int retval = 0;
     void* dummy = NULL;
+    /* FIXME: if there was an EINTR, there may be an untrusted thread left over */
     do {
-        /* FIXME: if there was an EINTR, there may be an untrusted thread left over */
-        retval = sgx_exitless_ocall(OCALL_CLONE_THREAD, dummy);
+        /* clone must happen in the context of current (enclave) thread, cannot use exitless;
+         * in particular, the new (enclave) thread must have the same signal mask as the current
+         * enclave thread (and NOT signal mask of the RPC thread) */
+        retval = sgx_ocall(OCALL_CLONE_THREAD, dummy);
     } while (retval == -EINTR);
     return retval;
 }


mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2384)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2384)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2384#-Ma3ibsDCrV0MmF4Su1A:-Ma3ibsE5mgN99AAQYn-:b6fj6s7) (no related file):*
Actually, why do we block all signals in rpc threads? Doesn't sound like something we should do...
This change is fine anyway, since creating a new enclave is expensive anyway, so exitless doesn't make much sense

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2384)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2384#-Ma3ibsDCrV0MmF4Su1A:-Ma3yhsc1ldsoWOnMLSB:bmkcfqy) (no related file):*

Actually, why do we block all signals in rpc threads? Doesn't sound like something we should do...
This change is fine anyway, since creating a new enclave is expensive anyway, so exitless doesn't make much sense
</blockquote></details>

Why should we keep signals unblocked in RPC threads? What's the point of getting a SIGINT signal in the RPC thread? Or SIGILL in the RPC thread? In the first case, SIGINT should be re-directed to some enclave thread. In the second case, SIGILL should just terminate the whole process, since it's a bug in untrusted RPC code.

By the way, you created a blocking comment. Do you want me to change anything in this PR?

---

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2384)* status: all files reviewed, 1 unresolved discussion (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2384#-Ma3ibsDCrV0MmF4Su1A:-Ma47tjg6T5gnIW_vhPH:b-jsr2ul) (no related file):*
> Why should we keep signals unblocked in RPC threads?

Why should we block them? If a thread gets a signal it's probably for a reason.
 
> What's the point of getting a SIGINT signal in the RPC thread? In the first case, SIGINT should be re-directed to some enclave thread. 

We do not handle SIGINT, so there is no need to block it. Only async signal that's handled and is SIGTERM, which we should block indeed, others are either targeted at the thread directly or not handled at all.

> In the second case, SIGILL should just terminate the whole process, since it's a bug in untrusted RPC code.

Yes, and in what way blocking it helps?

> By the way, you created a blocking comment. Do you want me to change anything in this PR?

Yes, I wanted to understand the situation and possibly fix if there is anything to fix (and I believe there is), but since this is an existing issue, we can postpone fixes for later. IIUC you want to fasttrack this? In such case ignore my comments :)

---

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2384)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @dimakuv)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2384)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2384#-Ma3ibsDCrV0MmF4Su1A:-Ma5BuWj39IIfeOSR1pp:b-vwjsv7) (no related file):*
I don't know, I don't see a compelling reason to unblock all signals in RPC threads but still block SIGTERM (and explicitly unblock SIGUSR2 which is blocked in enclave threads). In other words, we'll still need to do some blocking/unblocking for RPC threads, so why bother?

> Yes, and in what way blocking it helps?

It doesn't but it makes it explicit in the code that there is no signal handler SIGILL for RPC threads. If we wouldn't block, then the host would jump into the Graphene signal handler, and only then terminate the process. For me, "RPC threads catch SIGILL but fail on it" sounds weird -- I don't think it's good readability.

---

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2384)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv:
Jenkins, retest Jenkins-Debug-20.04 please (`fsync02` LTP test timed out, can't be related)

PR #2389, opened 2021-05-20
[Common,Pal,LibOS] Replace `warn/__abort` with `log_always/abort`

dimakuv:
## Description of the changes 

This PR has several commits:

1. **[Common,Pal] Move `debug_map` from common lib to Linux-common PAL**

The `debug_map.{c,h}` files maintain a debug map for GDB, and thus conceptually do not belong to "common host-agnostic functionality". Also, this functionality is only meaningful for PALs and should not be exposed to LibOS. Finally, `debug_map.c` includes `spinlock.h` which in turn includes LibOS-internal headers, thus prevoiusly leading to compilation failures of the LibOS-agnostic common lib `graphene-lib.a`.

2. **[LibOS,Pal] Propagate `IN_PAL/IN_SHIM` macros to common lib build**

This is in preparation for using these macros to build slightly different versions of the common lib for PAL and LibOS.

3. **[Common] Fix uthash to work both in PAL/LibOS and stdlib environments**

Previously, even if uthash was included in a non-Graphene program, it would use Graphene-internal `api.h` which could fail compilation.

4. **[Common,Pal,LibOS] Replace `warn/__abort` with `log_always/abort`**

The common library now expects that the environment specifies two callbacks: `log_always` (previously called `warn`) and `abort`
    (previously called `__abort`). The common library is built slightly differently for different environments:
- For PAL environments (both trusted and untrusted), it expects  `pal_log_always` and `pal_abort` symbols.
- For LibOS, it expects `shim_log_always` and `shim_abort` symbols.
- For all other environments (e.g., stand-alone SGX utilities, PAL  regression tests, etc.) it expects `log_always` and `abort` symbols.

As a side effect, incorrect usages of `__abort()` are modified and some includes are fixed.

## How to test this PR? 

All tests must pass.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2389)


mkow (changes requested):
> Reviewed 23 of 36 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2389)* status: 23 of 36 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBHpqZ1esuAECxl5zP:-MaBHpqZ1esuAECxl5zQ:bdnbr87) (no related file):*
nitpick for the commit messages: `/` shouldn't be inside backticks

---
___
*[common/include/assert.h, line 22 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBLgPD34GoM5vxQgFt:-MaBLgPD34GoM5vxQgFu:bcl8kv7) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/common/include/assert.h#L22)):*
> ```objc
> #elif IN_PAL
> void pal_log_always(const char* format, ...) __attribute__((format(printf, 1, 2)));
> noreturn void pal_abort(void);
> ```

All of this doesn't belong to "assert.h", what if you just want to use logging in some common code? (like I need for the bools-in-TOML)
Including assert.h for `log_*()` seems weird.
___
*[Pal/regression/utils.c, line 24 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBMrNhC7Fdabo6497O:-MaBMrNhC7Fdabo6497P:bcufjmg) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/regression/utils.c#L24)):*
> ```
> Stack Protector
> ```

I don't think this is a proper name
___
*[Pal/regression/utils.c, line 35 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBOAvv1ljAOWYIlv5A:-MaBOAvwBfYsikbXENyx:bpeaw87) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/regression/utils.c#L35)):*
> ```C
> 
> noreturn void abort(void) {
>     log_always("ABORTED\n");
> ```

Why some aborts print a message and some don't? Any reason, or just an unintended discrepancy?
___
*[Pal/src/db_exception.c, line 30 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBN8cS1_64Z7thilmG:-MaBN8cS1_64Z7thilmH:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/db_exception.c#L30)):*
> ```C
> }
> 
> /* The below function is used by Stack Protector's __stack_chk_fail(), _FORTIFY_SOURCE's *_chk()
> ```

ditto
___
*[Pal/src/host/Linux-SGX/sgx_arch.h, line 16 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBNTfX97bC4Atd2sDx:-MaBNTfX97bC4Atd2sDy:bx0js5t) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/sgx_arch.h#L16)):*
> ```objc
> #pragma pack(push, 1)
> 
> #define static_assert _Static_assert
> ```

Why this change? What's wrong in including "assert.h"? Now we're duplicating its functionality here.
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 251 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBNddNEwlLO8aSd0-a:-MaBNddNEwlLO8aSd0-b:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/sgx_exception.c#L251)):*
> ```C
> }
> 
> /* The below function is used by Stack Protector's __stack_chk_fail(), _FORTIFY_SOURCE's *_chk()
> ```

ditto
___
*[Pal/src/host/Linux-SGX/sgx_log.c, line 85 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBNehr9RKZsVenYoNV:-MaBNehr9RKZsVenYoNW:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/sgx_log.c#L85)):*
> ```C
> }
> 
> /* The below function is used by Stack Protector's __stack_chk_fail(), _FORTIFY_SOURCE's *_chk()
> ```

ditto
___
*[Pal/src/host/Linux-SGX/sgx_log.c, line 88 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBOZAoDp9Palz97Vyx:-MaBOZAoDp9Palz97Vyy:bqgzxwm) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/sgx_log.c#L88)):*
> ```C
>  * functions and by assert.h's assert() defined in the common library. Do not use this function in
>  * untrusted-PAL code, instead use urts_log_always() for readability! */
> void pal_log_always(const char* fmt, ...) {
> ```

Quite a confusing clash in the naming (`urts_log_always` vs `pal_log_always`), but I don't have any better idea :/
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files_format.h, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBOjfN2mVJ0zNwfy26:-MaBOjfN2mVJ0zNwfy27:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/protected-files/protected_files_format.h#L15)):*
> ```objc
> #include "protected_files.h"
> 
> #define static_assert _Static_assert
> ```

ditto

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2389)* status: 19 of 37 files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBHpqZ1esuAECxl5zP:-MaD6MApEQoSm_kn4enu:b-vitwox) (no related file):*

nitpick for the commit messages: `/` shouldn't be inside backticks
</blockquote></details>

Do you mean to replace them like this: `IN_PAL/IN_SHIM` -> `IN_PAL`, `IN_SHIM`?

---
___
*[common/include/assert.h, line 22 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBLgPD34GoM5vxQgFt:-MaDAM1E1SwNbWUp49Tw:b-aclhc3) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/common/include/assert.h#L22)):*

All of this doesn't belong to "assert.h", what if you just want to use logging in some common code? (like I need for the bools-in-TOML)
Including assert.h for `log_*()` seems weird.
</blockquote></details>

Done. I wanted to create a separate `callbacks.h` header in a follow-up PR, but you're right, let's just do it now.
___
*[Pal/regression/utils.c, line 24 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBMrNhC7Fdabo6497O:-MaDAfWLD_JqfuVHluBA:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/regression/utils.c#L24)):*

> ```
> Stack Protector
> ```

I don't think this is a proper name
</blockquote></details>

Done.
___
*[Pal/regression/utils.c, line 35 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBOAvv1ljAOWYIlv5A:-MaDBGSz9LOnZhG3gWj1:b-113r25) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/regression/utils.c#L35)):*

Why some aborts print a message and some don't? Any reason, or just an unintended discrepancy?
</blockquote></details>

Done. It was like this before, so yes, historical discrepancy. But now I changed so that aborts do not print a message (it was printed anyway by its callers).
___
*[Pal/src/db_exception.c, line 30 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBN8cS1_64Z7thilmG:-MaDBYmh-8Jc6oBh13ex:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/db_exception.c#L30)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_arch.h, line 16 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBNTfX97bC4Atd2sDx:-MaD7SWYCy8oSaNYKE3I:b-hdsxiv) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/sgx_arch.h#L16)):*

Why this change? What's wrong in including "assert.h"? Now we're duplicating its functionality here.
</blockquote></details>

This header is included both in Graphene-internal sources (and thus should use `"assert.h"`) and in normal-Linux sources (think of RA-TLS, IAS, etc) (and thus should use `<assert.h>`). This means that ideally this header would need:
```
#ifdef USE_STDLIB
#include <assert.h>
#else
#include "assert.h"
#endif
```

I find this code snippet extremely ugly, and we only need `static_assert` which is not even a "real" macro, it's just an alias to compiler's `_Static_assert`.

So to me, it was more readable to do this instead of including assert header. Do you want me to use the "heavy" snippet instead?
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 251 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBNddNEwlLO8aSd0-a:-MaDB_h35EaHopkI3A7_:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/sgx_exception.c#L251)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_log.c, line 85 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBNehr9RKZsVenYoNV:-MaDBaNHA2jJ_mizkaYq:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/sgx_log.c#L85)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_log.c, line 88 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBOZAoDp9Palz97Vyx:-MaD75DcBOG9HJDeg-vM:bwggtch) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/sgx_log.c#L88)):*

Quite a confusing clash in the naming (`urts_log_always` vs `pal_log_always`), but I don't have any better idea :/
</blockquote></details>

I am actually not sure why @pwmarcz went with `urts_log_always` instead of simply `log_always`. I guess for readability, to make it explicit that the message is printed in untrusted part of PAL?

If it would be just `log_always` in the old code, then there wouldn't be this clash (I would simply rename this func to `pal_log_always` and that would be enough).
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files_format.h, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBOjfN2mVJ0zNwfy26:-MaD8M2pCd9gJOOvotyU:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/protected-files/protected_files_format.h#L15)):*

ditto
</blockquote></details>

ditto

mkow (changes requested):
> Reviewed 11 of 36 files at r1, 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2389)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBHpqZ1esuAECxl5zP:-MaE8B0gAj7WA6XPhD6-:b80x6h0) (no related file):*

Do you mean to replace them like this: `IN_PAL/IN_SHIM` -> `IN_PAL`, `IN_SHIM`?
</blockquote></details>

Yup. Because backticks are used to mark literal symbols, so if you include "/" inside it may look like this is a part of the symbol for someone without enough context.

---
___
*[common/include/assert.h, line 12 at r2](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaEWOr1AUfk0xXeljav:-MaEWOr67sxE5DaOc2yC:b-oac8e8) ([raw file](https://github.com/oscarlab/graphene/blob/ce497853865a6d097dbb79f9f1278461165ebbd9/common/include/assert.h#L12)):*
> ```objc
> #include "callbacks.h"
> 
> noreturn void __stack_chk_fail(void);
> ```

Not in callbacks.h?
___
*[common/include/callbacks.h, line 2 at r2](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaEYsoD3kk_UwKQ3tAJ:-MaEYsoD3kk_UwKQ3tAK:bwszfr0) ([raw file](https://github.com/oscarlab/graphene/blob/ce497853865a6d097dbb79f9f1278461165ebbd9/common/include/callbacks.h#L2)):*
> ```objc
> /*
>  * Define a set of callbacks that the common library expects from the environment it is linked into.
> ```

Not "Defines"? Imperative mood sounds weird to me here.
___
*[common/include/callbacks.h, line 6 at r2](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaEZaYS-69hEh-tmslQ:-MaEZaYS-69hEh-tmslR:b-gyhhkr) ([raw file](https://github.com/oscarlab/graphene/blob/ce497853865a6d097dbb79f9f1278461165ebbd9/common/include/callbacks.h#L6)):*
> ```objc
>  * callbacks from PAL, and not-prefixed callbacks from all other environments (e.g., PAL regression
>  * tests and non-Graphene programs). This header aliases the actual callback implementations, i.e.,
>  * `shim_abort()` is aliased as `abort()` for use by the common library.
> ```

I think this description misses the most important information - why do we need different calllback names for LibOS, PAL and outside world.
___
*[Pal/src/host/Linux-SGX/sgx_arch.h, line 16 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBNTfX97bC4Atd2sDx:-MaEWf4D9_4E3RTMDwHp:b-rc188z) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/sgx_arch.h#L16)):*

This header is included both in Graphene-internal sources (and thus should use `"assert.h"`) and in normal-Linux sources (think of RA-TLS, IAS, etc) (and thus should use `<assert.h>`). This means that ideally this header would need:
```
#ifdef USE_STDLIB
#include <assert.h>
#else
#include "assert.h"
#endif
```

I find this code snippet extremely ugly, and we only need `static_assert` which is not even a "real" macro, it's just an alias to compiler's `_Static_assert`.

So to me, it was more readable to do this instead of including assert header. Do you want me to use the "heavy" snippet instead?
</blockquote></details>

Tbh I prefer the snippet, as it makes it explicit that we have two modes (and in a bunch of files you already have this ifdef in place, so you just need to add proper includes there).
___
*[Pal/src/host/Linux-SGX/sgx_log.c, line 88 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBOZAoDp9Palz97Vyx:-MaEXAUl38ZX7sP9-aQt:bucoar6) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/sgx_log.c#L88)):*

I am actually not sure why @pwmarcz went with `urts_log_always` instead of simply `log_always`. I guess for readability, to make it explicit that the message is printed in untrusted part of PAL?

If it would be just `log_always` in the old code, then there wouldn't be this clash (I would simply rename this func to `pal_log_always` and that would be enough).
</blockquote></details>

Hmm, I think there was some reason for this (e.g. something didn't compile due to the linking mess :) ). But probably Paweł remembers better.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2389)* status: 32 of 38 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBHpqZ1esuAECxl5zP:-MaXDq5t8orWoVY2R6FK:b-896fix) (no related file):*

Yup. Because backticks are used to mark literal symbols, so if you include "/" inside it may look like this is a part of the symbol for someone without enough context.
</blockquote></details>

Done.

---
___
*[common/include/assert.h, line 12 at r2](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaEWOr1AUfk0xXeljav:-MaX7rMw1gD8sPqhoowJ:b9b1gsu) ([raw file](https://github.com/oscarlab/graphene/blob/ce497853865a6d097dbb79f9f1278461165ebbd9/common/include/assert.h#L12)):*

Not in callbacks.h?
</blockquote></details>

No, `__stack_chk_fail()` is defined in `common/src/stack_protector.c`. It is not a callback (it is a simple wrapper that uses callbacks: `log_always` and `abort`).
___
*[common/include/callbacks.h, line 2 at r2](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaEYsoD3kk_UwKQ3tAJ:-MaX89MgCJxlfcRki9Rh:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/ce497853865a6d097dbb79f9f1278461165ebbd9/common/include/callbacks.h#L2)):*

Not "Defines"? Imperative mood sounds weird to me here.
</blockquote></details>

Done.
___
*[common/include/callbacks.h, line 6 at r2](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaEZaYS-69hEh-tmslQ:-MaX9H4z0TGmTxsx2HKK:b-ozh7jf) ([raw file](https://github.com/oscarlab/graphene/blob/ce497853865a6d097dbb79f9f1278461165ebbd9/common/include/callbacks.h#L6)):*

I think this description misses the most important information - why do we need different calllback names for LibOS, PAL and outside world.
</blockquote></details>

Done. Note that currently it's not needed but this is a preparation for future PRs.
___
*[Pal/src/host/Linux-SGX/sgx_arch.h, line 16 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBNTfX97bC4Atd2sDx:-MaXC2PSCuf_0BLhfs11:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/sgx_arch.h#L16)):*

Tbh I prefer the snippet, as it makes it explicit that we have two modes (and in a bunch of files you already have this ifdef in place, so you just need to add proper includes there).
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_log.c, line 88 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBOZAoDp9Palz97Vyx:-MaX9W3iB01AKooRY_PE:bvj1l4) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/sgx_log.c#L88)):*

Hmm, I think there was some reason for this (e.g. something didn't compile due to the linking mess :) ). But probably Paweł remembers better.
</blockquote></details>

I will discuss with Pawel, but generally I think that all changes to `log_*` could be unified in a separate PR (for example, I would like to prepare a PR where `log_always` becomes a macro similar to `log_error/log_warning/log_trace`). So, I will probably fix this in the next PR.
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files_format.h, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaBOjfN2mVJ0zNwfy26:-MaXC4jQ04SpbEt-tPRi:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/2634f7aa69e2f9c8ca1dd73e369b20c3211d9fcc/Pal/src/host/Linux-SGX/protected-files/protected_files_format.h#L15)):*

ditto
</blockquote></details>

Done.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2389)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @pwmarcz)

___
*[common/include/assert.h, line 12 at r2](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaEWOr1AUfk0xXeljav:-MaY_VZH4z-n8v7FcjgR:b-xmbflz) ([raw file](https://github.com/oscarlab/graphene/blob/ce497853865a6d097dbb79f9f1278461165ebbd9/common/include/assert.h#L12)):*

No, `__stack_chk_fail()` is defined in `common/src/stack_protector.c`. It is not a callback (it is a simple wrapper that uses callbacks: `log_always` and `abort`).
</blockquote></details>

Actually, why do we even need this declaration here?
Anyways, if its defined in `stack_protector.c`, then it should be declared in `stack_protector.h`, time to get rid of this src/header split chaos.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2389)* status: 36 of 38 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[common/include/assert.h, line 12 at r2](https://reviewable.io/reviews/oscarlab/graphene/2389#-MaEWOr1AUfk0xXeljav:-MaYdfMi3MFH74ST-Ibu:buiedir) ([raw file](https://github.com/oscarlab/graphene/blob/ce497853865a6d097dbb79f9f1278461165ebbd9/common/include/assert.h#L12)):*

Actually, why do we even need this declaration here?
Anyways, if its defined in `stack_protector.c`, then it should be declared in `stack_protector.h`, time to get rid of this src/header split chaos.
</blockquote></details>

Done. Removed completely because `__stack_chk_fail()` is only used implicitly by GCC.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2389)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

mkow (approved):
> Reviewed 1 of 36 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2389)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required) (waiting on @pwmarcz)

PR #2390, opened 2021-05-21
[Pal/Linux-SGX] Use AES-CMAC for key derivation in Protected Files

### LINKED ISSUE #2386: [Pal/Linux-SGX] Wrong crypto in key derivation of Protected Files
For Protected FS, we copied the implementation from Intel SGX SDK:
- https://github.com/oscarlab/graphene/tree/master/Pal/src/host/Linux-SGX/protected-files
- https://github.com/intel/linux-sgx/tree/master/sdk/protected_fs/sgx_tprotected_fs

However, we made one crypto mistake: we use AES-GCM for key derivation, but we must use AES-CMAC. Here are the relevant sources:
- https://github.com/oscarlab/graphene/blob/14812bedc29992e60a910828968cfaf417037975/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L127
- https://github.com/intel/linux-sgx/blob/26c458905b72e66db7ac1feae04b43461ce1b76f/sdk/protected_fs/sgx_tprotected_fs/file_crypto.cpp#L96

This `sgx_rijndael128_cmac_msg()` is an AES-CMAC algorithm. See definition here: https://github.com/intel/linux-sgx/blob/26c458905b72e66db7ac1feae04b43461ce1b76f/psw/ae/aesm_service/source/utils/crypto_cmac128.cpp#L44.

It seems that we'll need to add `lib_AESCMAC` in addition to the currently used `lib_AESGCMEncrypt`:
- https://github.com/oscarlab/graphene/blob/adf6269218dfa80aed276d57121a98e7b13b0f4e/Pal/src/host/Linux-SGX/enclave_pf.c#L108

dimakuv:
## Description of the changes 

Protected FS previously used AES-GCM (with empty output argument, using the MAC as the derived key) as a KDF. This construction lacks any proof of correctness, thus this commit replaces AES-GCM with AES-CMAC as a KDF as per NIST recommendation.

Fixes #2386.

## How to test this PR? 

All tests must pass, especially FS tests (`make pf-test`).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2390)

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_pf.c
+++ b/Pal/src/host/Linux-SGX/enclave_pf.c
@@ -102,6 +102,17 @@ static void cb_debug(const char* msg) {
 }
 #endif
 
+static pf_status_t cb_aes_cmac(const pf_key_t* key, const void* input, size_t input_size,
+                               pf_mac_t* mac) {
+    int ret = lib_AESCMAC((const uint8_t*)key, sizeof(*key), input, input_size, (uint8_t*)mac,
+                          sizeof(*mac));
+    if (ret != 0) {
+        log_error("lib_AESCMAC failed: %d\n", ret);
+        return PF_STATUS_CALLBACK_FAILED;
+    }
+    return PF_STATUS_SUCCESS;
+}
+
 static pf_status_t cb_aes_gcm_encrypt(const pf_key_t* key, const pf_iv_t* iv, const void* aad,
                                       size_t aad_size, const void* input, size_t input_size,
                                       void* output, pf_mac_t* mac) {
@@ -467,8 +478,8 @@ int init_protected_files(void) {
     debug_callback = cb_debug;
 #endif
 
-    pf_set_callbacks(cb_read, cb_write, cb_truncate, cb_aes_gcm_encrypt, cb_aes_gcm_decrypt,
-                     cb_random, debug_callback);
+    pf_set_callbacks(cb_read, cb_write, cb_truncate, cb_aes_cmac, cb_aes_gcm_encrypt,
+                     cb_aes_gcm_decrypt, cb_random, debug_callback);
 
     /* if wrap key is not hard-coded in the manifest, assume that it was received from parent or
      * it will be provisioned after local/remote attestation; otherwise read it from manifest */
--- a/Pal/src/host/Linux-SGX/protected-files/protected_files.c
+++ b/Pal/src/host/Linux-SGX/protected-files/protected_files.c
@@ -36,6 +36,7 @@ static pf_write_f    g_cb_write    = NULL;
 static pf_truncate_f g_cb_truncate = NULL;
 static pf_debug_f    g_cb_debug    = NULL;
 
+static pf_aes_cmac_f        g_cb_aes_cmac        = NULL;
 static pf_aes_gcm_encrypt_f g_cb_aes_gcm_encrypt = NULL;
 static pf_aes_gcm_decrypt_f g_cb_aes_gcm_decrypt = NULL;
 static pf_random_f          g_cb_random          = NULL;
@@ -102,8 +103,15 @@ const char* pf_strerror(int err) {
 // The key derivation function follow recommendations from NIST Special Publication 800-108:
 // Recommendation for Key Derivation Using Pseudorandom Functions
 // https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-108.pdf
-
-// derive a metadata key from user key (if restore is false, the derived key is randomized)
+//
+// This function derives a metadata key in two modes:
+//   - restore == false: derives a per-file random key from user_kdk_key using a random nonce, to
+//                       encrypt the metadata block of the protected file; the nonce is stored in
+//                       plaintext part of the metadata block so that the file can be loaded later
+//                       and decrypted using the same key
+//   - restore == true:  derives a key from user_kdk_key + nonce stored in plaintext part of the
+//                       metadata block, to decrypt the encrypted part of the metadata block (and
+//                       thus "restore" access to the whole protected file)
 static bool ipf_import_metadata_key(pf_context_t* pf, bool restore, pf_key_t* output) {
     kdf_input_t buf = {0};
     pf_status_t status;
@@ -125,8 +133,7 @@ static bool ipf_import_metadata_key(pf_context_t* pf, bool restore, pf_key_t* ou
     // length of output (128 bits)
     buf.output_len = 0x80;
 
-    status = g_cb_aes_gcm_encrypt(&pf->user_kdk_key, &g_empty_iv, &buf, sizeof(buf), NULL, 0, NULL,
-                                  output);
+    status = g_cb_aes_cmac(&pf->user_kdk_key, &buf, sizeof(buf), output);
     if (PF_FAILURE(status)) {
         pf->last_error = status;
         return false;
@@ -1205,12 +1212,13 @@ static file_node_t* ipf_read_mht_node(pf_context_t* pf, uint64_t mht_node_number
 // public API
 
 void pf_set_callbacks(pf_read_f read_f, pf_write_f write_f, pf_truncate_f truncate_f,
-                      pf_aes_gcm_encrypt_f aes_gcm_encrypt_f,
+                      pf_aes_cmac_f aes_cmac_f, pf_aes_gcm_encrypt_f aes_gcm_encrypt_f,
                       pf_aes_gcm_decrypt_f aes_gcm_decrypt_f, pf_random_f random_f,
                       pf_debug_f debug_f) {
     g_cb_read            = read_f;
     g_cb_write           = write_f;
     g_cb_truncate        = truncate_f;
+    g_cb_aes_cmac        = aes_cmac_f;
     g_cb_aes_gcm_encrypt = aes_gcm_encrypt_f;
     g_cb_aes_gcm_decrypt = aes_gcm_decrypt_f;
     g_cb_random          = random_f;
--- a/Pal/src/host/Linux-SGX/protected-files/protected_files.h
+++ b/Pal/src/host/Linux-SGX/protected-files/protected_files.h
@@ -105,6 +105,18 @@ typedef pf_status_t (*pf_truncate_f)(pf_handle_t handle, uint64_t size);
 typedef void (*pf_debug_f)(const char* msg);
 
 /*!
+ * \brief AES-CMAC callback used for key derivation
+ *
+ * \param [in] key AES-GCM key
+ * \param [in] input Plaintext data
+ * \param [in] input_size Size of \a input in bytes
+ * \param [out] mac MAC computed for \a input
+ * \return PF status
+ */
+typedef pf_status_t (*pf_aes_cmac_f)(const pf_key_t* key, const void* input, size_t input_size,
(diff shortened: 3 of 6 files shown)


dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2390)* status: 0 of 6 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 128 at r1](https://reviewable.io/reviews/oscarlab/graphene/2390#-MaDWtcaDgqZx5VGLlIq:-MaDWtcb8D7njk9eqFj8:b-p9xbcl) ([raw file](https://github.com/oscarlab/graphene/blob/75962a812d686e839fabe5d9525e52128a8f1c86/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L128)):*
> ```C
> 
>     status = g_cb_aes_gcm_encrypt(&pf->user_kdk_key, &g_empty_iv, &buf, sizeof(buf), NULL, 0, NULL,
>                                   output);
> ```

FYI: This was the bug.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2390)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2390#-MaEgygpEbmJPu0InIU0:-MaEgygpEbmJPu0InIU1:bovx0is) (no related file):*
> This is cryptographically wrong.

I don't see how is this wrong, but I also don't see how is it correct and overall it looks a bit fishy, probably just an implementation bug. So, the change is definitely good, but please either say in the commit message why the old code was wrong or instead just say that it lacked any proof and we're changing it to a NIST-recommended construction.

---
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 106 at r1](https://reviewable.io/reviews/oscarlab/graphene/2390#-MaEfwB4C8JqQIeuBV8B:-MaEfwB4C8JqQIeuBV8C:b-nv8et1) ([raw file](https://github.com/oscarlab/graphene/blob/d128f893c79e39795061a8032f0c82191969251a/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L106)):*
> ```C
> // https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-108.pdf
> 
> // derive a metadata key from user key (if restore is false, the derived key is randomized)
> ```

While we're here, can we remove this nonsensical `restore==false` mode and just use `ipf_generate_random_key` instead? Will simplify this function considerably.
___
*[Pal/src/host/Linux-SGX/tools/pf_tamper/pf_tamper.c, line 62 at r1](https://reviewable.io/reviews/oscarlab/graphene/2390#-MaEkfw-5B1B60rBnXJv:-MaEkfw0AOH1QiIJ-rhA:b16twav) ([raw file](https://github.com/oscarlab/graphene/blob/d128f893c79e39795061a8032f0c82191969251a/Pal/src/host/Linux-SGX/tools/pf_tamper/pf_tamper.c#L62)):*
> ```C
>     buf.index = 1;
>     strncpy(buf.label, METADATA_KEY_NAME, MAX_LABEL_SIZE);
>     memcpy(&buf.nonce, key_id, sizeof(buf.nonce));
> ```

While we're here, could we use `COPY_ARRAY` to ensure that the types match? The current version looks quite dangerous. I know it's just a test, but this code may be reused later and we may regret this :)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2390)* status: 4 of 6 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2390#-MaEgygpEbmJPu0InIU0:-MaXLargC-Vm_KREVTnf:b-896fix) (no related file):*

> This is cryptographically wrong.

I don't see how is this wrong, but I also don't see how is it correct and overall it looks a bit fishy, probably just an implementation bug. So, the change is definitely good, but please either say in the commit message why the old code was wrong or instead just say that it lacked any proof and we're changing it to a NIST-recommended construction.
</blockquote></details>

Done.

---
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 106 at r1](https://reviewable.io/reviews/oscarlab/graphene/2390#-MaEfwB4C8JqQIeuBV8B:-MaXHip090GJDgyx7CJ_:bnooqpn) ([raw file](https://github.com/oscarlab/graphene/blob/d128f893c79e39795061a8032f0c82191969251a/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L106)):*

While we're here, can we remove this nonsensical `restore==false` mode and just use `ipf_generate_random_key` instead? Will simplify this function considerably.
</blockquote></details>

No, we can't do it. `restore` is essential for the Protected Files functioning.

I updated the comment to explain better what `restore` does. I would rename it to something like `derive_from_key_id` or `use_key_id_as_nonce` but this would be a departure from Intel SGX SDK's names. I dislike Intel SGX SDK names, but at least it is easy to cross-check our implementation against that implementation. So I'd prefer not to rename stuff here.
___
*[Pal/src/host/Linux-SGX/tools/pf_tamper/pf_tamper.c, line 62 at r1](https://reviewable.io/reviews/oscarlab/graphene/2390#-MaEkfw-5B1B60rBnXJv:-MaXKcdZ0laJ0XGZxGwz:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d128f893c79e39795061a8032f0c82191969251a/Pal/src/host/Linux-SGX/tools/pf_tamper/pf_tamper.c#L62)):*

While we're here, could we use `COPY_ARRAY` to ensure that the types match? The current version looks quite dangerous. I know it's just a test, but this code may be reused later and we may regret this :)
</blockquote></details>

Done.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2390)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 106 at r1](https://reviewable.io/reviews/oscarlab/graphene/2390#-MaEfwB4C8JqQIeuBV8B:-MaYYpVF9094UG3DkMXd:b-e4rx0u) ([raw file](https://github.com/oscarlab/graphene/blob/d128f893c79e39795061a8032f0c82191969251a/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L106)):*

No, we can't do it. `restore` is essential for the Protected Files functioning.

I updated the comment to explain better what `restore` does. I would rename it to something like `derive_from_key_id` or `use_key_id_as_nonce` but this would be a departure from Intel SGX SDK's names. I dislike Intel SGX SDK names, but at least it is easy to cross-check our implementation against that implementation. So I'd prefer not to rename stuff here.
</blockquote></details>

But why would you ever call this function with `restore == false`? It just generates a random key in a very weird manner, instead of just calling `ipf_generate_random_key`.
IMO this just obfuscates the whole logic. I want to see in such callsites that it's just generating a completely new, random key, not a weird call for which I need to remember or the operation modes to notice that it's just a `get_random()` call.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2390)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 106 at r1](https://reviewable.io/reviews/oscarlab/graphene/2390#-MaEfwB4C8JqQIeuBV8B:-MaYswfg54iQB1PM7j0J:b-7znpqe) ([raw file](https://github.com/oscarlab/graphene/blob/d128f893c79e39795061a8032f0c82191969251a/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L106)):*

I looked deeper into this logic, and it goes like this:

There is a single METADATA BLOCK in each Protected File. We have a single master key (stored in `pf->user_kdk_key`) and several Protected Files. So we need to derive a "root key" for each Protected File from this master key.

For derivation, we use the algorithm from NIST 800-108. This is what this weird logic does, with label, nonce, length and doing AES-CMAC over this mix.

Now there are two cases we want to cover:
1. Creating the new Protected File or updating it. In this case, we need a new randomized "root key" for the protected file, derived from `user_kdk_key`. We also **need to store the nonce** used in KDF as the `file_metadata.plain_part.metadata_key_id`, so that we can derive the exact same key later (when we re-open the file). This is the case of `restore == false`.
2. Opening the existing Protected File. We need to derive the exact same key as used before, given the same `user_kdk_key`. This is achieved by taking not a random nonce, but a nonce stored in `file_metadata.plain_part.metadata_key_id`. This is the case of `restore == true`.

So I think the part that you missed here is: we create a random key using a `KDF(user_kdk_key, nonce)` and we store `nonce` in the plaintext part of the METADATA BLOCK of the Protected File (for future re-open and decryption).

We can't use `ipf_generate_random_key()` because we need the "root key", which can be derived from some known nonce.
</blockquote></details>

Ah, I missed that part with storing the nonce and thought that it's discarded. Makes much more sense now. Could you explain this briefly in the comment about `restore == false`?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2390)* status: 5 of 6 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 106 at r1](https://reviewable.io/reviews/oscarlab/graphene/2390#-MaEfwB4C8JqQIeuBV8B:-MaZ8xIR-6Ey7zwoIsph:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d128f893c79e39795061a8032f0c82191969251a/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L106)):*

Ah, I missed that part with storing the nonce and thought that it's discarded. Makes much more sense now. Could you explain this briefly in the comment about `restore == false`?
</blockquote></details>

Done.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2390)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

dimakuv:
Jenkins, retest Jenkins-Debug-20.04 please (`clone06` LTP test failed with `read from pipe failed: errno=EINTR(4): Interrupted system call`, probably will be fixed by https://github.com/oscarlab/graphene/pull/2392)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2390)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2392, opened 2021-05-22
[LibOS] Handle automatic syscalls restarting where necessary

boryspoplawski:
## Description of the changes 
Add rest of missing `ERESTART*` returns.

## How to test this PR? 
Added new regression tests, which fails on the current master.
Unfortunately it fails on SGX even with this PR due to pipe encryption layer translating `EINTR` to `EAGAIN` (https://github.com/oscarlab/graphene/blob/master/common/src/crypto/adapters/mbedtls_adapter.c#L314).

### DIFF
--- a/LibOS/shim/src/bookkeep/shim_thread.c
+++ b/LibOS/shim/src/bookkeep/shim_thread.c
@@ -192,7 +192,7 @@ static int init_main_thread(void) {
     unlock(&cur_thread->lock);
 
     int ret = DkEventCreate(&cur_thread->scheduler_event, /*init_signaled=*/false,
-                            /*auto_clear=*/false);
+                            /*auto_clear=*/true);
     if (ret < 0) {
         put_thread(cur_thread);
         return pal_to_unix_errno(ret);;
@@ -304,7 +304,7 @@ struct shim_thread* get_new_thread(void) {
     unlock(&cur_thread->lock);
 
     int ret = DkEventCreate(&thread->scheduler_event, /*init_signaled=*/false,
-                            /*auto_clear=*/false);
+                            /*auto_clear=*/true);
     if (ret < 0) {
         put_thread(thread);
         return NULL;
@@ -641,7 +641,7 @@ BEGIN_RS_FUNC(thread) {
     }
 
     int ret = DkEventCreate(&thread->scheduler_event, /*init_signaled=*/false,
-                            /*auto_clear=*/false);
+                            /*auto_clear=*/true);
     if (ret < 0) {
         return pal_to_unix_errno(ret);
     }
--- a/LibOS/shim/src/sys/shim_epoll.c
+++ b/LibOS/shim/src/sys/shim_epoll.c
@@ -414,6 +414,11 @@ long shim_do_epoll_wait(int epfd, struct __kernel_epoll_event* events, int maxev
         if (error && error != -EAGAIN) {
             unlock(&epoll_hdl->lock);
             put_handle(epoll_hdl);
+            if (error == -EINTR) {
+                /* `epoll_wait` and `epoll_pwait` are not restarted after being interrupted by
+                 * a signal handler. */
+                error = -ERESTARTNOHAND;
+            }
             return error;
         } else if (event_handle_update) {
             /* retry if epoll was updated concurrently (similar to Linux semantics) */
--- a/LibOS/shim/src/sys/shim_getrandom.c
+++ b/LibOS/shim/src/sys/shim_getrandom.c
@@ -27,8 +27,13 @@ long shim_do_getrandom(char* buf, size_t count, unsigned int flags) {
      * but this shouldn't be possible in practice, so we don't care.
      */
     int ret = DkRandomBitsRead(buf, count);
-    if (ret < 0)
-        return pal_to_unix_errno(ret);
+    if (ret < 0) {
+        ret = pal_to_unix_errno(ret);
+        if (ret == -EINTR) {
+            ret = -ERESTARTSYS;
+        }
+        return ret;
+    }
 
     return count;
 }
--- a/LibOS/shim/src/sys/shim_ioctl.c
+++ b/LibOS/shim/src/sys/shim_ioctl.c
@@ -115,5 +115,8 @@ long shim_do_ioctl(unsigned int fd, unsigned int cmd, unsigned long arg) {
     }
 
     put_handle(hdl);
+    if (ret == -EINTR) {
+        ret = -ERESTARTSYS;
+    }
     return ret;
 }
--- a/LibOS/shim/src/sys/shim_open.c
+++ b/LibOS/shim/src/sys/shim_open.c
@@ -54,6 +54,9 @@ long shim_do_read(int fd, void* buf, size_t count) {
 
     int ret = do_handle_read(hdl, buf, count);
     put_handle(hdl);
+    if (ret == -EINTR) {
+        ret = -ERESTARTSYS;
+    }
     return ret;
 }
 
@@ -83,6 +86,9 @@ long shim_do_write(int fd, const void* buf, size_t count) {
 
     int ret = do_handle_write(hdl, buf, count);
     put_handle(hdl);
+    if (ret == -EINTR) {
+        ret = -ERESTARTSYS;
+    }
     return ret;
 }
 
@@ -119,8 +125,13 @@ long shim_do_openat(int dfd, const char* filename, int flags, int mode) {
     }
 
     ret = open_namei(hdl, dir, filename, flags, mode, NULL);
-    if (ret < 0)
+    if (ret < 0) {
+        /* If this was blocking `open` (e.g. on FIFO), it might have returned `-EINTR`. */
+        if (ret == -EINTR) {
+            ret = -ERESTARTSYS;
+        }
         goto out_hdl;
+    }
 
     ret = set_new_fd_handle(hdl, flags & O_CLOEXEC ? FD_CLOEXEC : 0, NULL);
 
--- a/LibOS/shim/src/sys/shim_poll.c
+++ b/LibOS/shim/src/sys/shim_poll.c
@@ -183,6 +183,10 @@ static long _shim_do_poll(struct pollfd* fds, nfds_t nfds, int timeout_ms) {
     if (error == -EAGAIN) {
         /* `poll` returns 0 on timeout. */
         error = 0;
+    } else if (error == -EINTR) {
+        /* `poll`, `ppoll`, `select` and `pselect` are not restarted after being interrupted by
+         * a signal handler. */
+        error = -ERESTARTNOHAND;
     }
     return nrevents ? (long)nrevents : error;
 }
--- a/LibOS/shim/src/sys/shim_socket.c
+++ b/LibOS/shim/src/sys/shim_socket.c
@@ -841,6 +841,11 @@ out:
 
     unlock(&hdl->lock);
     put_handle(hdl);
+    if (ret == -EINTR) {
+        /* TODO: in case of sockets with `SO_RCVTIMEO` or `SO_SNDTIMEO` set we should return
+         * `-ERESTARTNOHAND` here. */
+        ret = -ERESTARTSYS;
+    }
     return ret;
 }
 
@@ -984,6 +989,11 @@ out:
     if (accepted)
         DkObjectClose(accepted);
     unlock(&hdl->lock);
+    if (ret == -EINTR) {
+        /* TODO: in case of sockets with `SO_RCVTIMEO` or `SO_SNDTIMEO` set we should return
+         * `-ERESTARTNOHAND` here. */
+        ret = -ERESTARTSYS;
+    }
     return ret;
 }
 
@@ -1150,6 +1160,11 @@ out_locked:
     unlock(&hdl->lock);
 out:
     put_handle(hdl);
+    if (ret == -EINTR) {
+        /* TODO: in case of sockets with `SO_RCVTIMEO` or `SO_SNDTIMEO` set and once we support
+         * `timeout` argument to `sendmmsg` we should sometimes return `-ERESTARTNOHAND` here. */
+        ret = -ERESTARTSYS;
+    }
     return ret;
 }
 
@@ -1354,6 +1369,7 @@ static ssize_t do_recvmsg(int fd, struct iovec* bufs, size_t nbufs, int flags,
             size_t left_to_read = expected_size - (peek_buffer->end - peek_buffer->start);
             ret = DkStreamRead(pal_hdl, /*offset=*/0, &left_to_read,
                                &peek_buffer->buf[peek_buffer->end], uri, uri ? SOCK_URI_SIZE : 0);
+            /* TODO: shouldn't we call `maybe_epoll_et_trigger` here? */
             if (ret < 0) {
                 ret = ret == -PAL_ERROR_STREAMNOTEXIST ? -ECONNABORTED : pal_to_unix_errno(ret);
                 lock(&hdl->lock);
@@ -1478,6 +1494,11 @@ out_locked:
     free(peek_buffer);
 out:
     put_handle(hdl);
+    if (ret == -EINTR) {
+        /* TODO: in case of sockets with `SO_RCVTIMEO` or `SO_SNDTIMEO` set and once we support
+         * `timeout` argument to `recvmmsg` we should sometimes return `-ERESTARTNOHAND` here. */
+        ret = -ERESTARTSYS;
+    }
     return ret;
 }
 
--- a/LibOS/shim/src/sys/shim_wrappers.c
+++ b/LibOS/shim/src/sys/shim_wrappers.c
@@ -59,6 +59,9 @@ long shim_do_readv(int fd, const struct iovec* vec, int vlen) {
     ret = bytes;
 out:
     put_handle(hdl);
+    if (ret == -EINTR) {
+        ret = -ERESTARTSYS;
+    }
     return ret;
 }
 
@@ -121,5 +124,8 @@ long shim_do_writev(int fd, const struct iovec* vec, int vlen) {
     ret = bytes;
 out:
     put_handle(hdl);
+    if (ret == -EINTR) {
+        ret = -ERESTARTSYS;
+    }
     return ret;
 }
--- a/LibOS/shim/test/regression/.gitignore
+++ b/LibOS/shim/test/regression/.gitignore
@@ -94,6 +94,7 @@
 /spinlock
 /stat_invalid_args
 /syscall
+/syscall_restart
 /sysfs_common
 /tcp_ipv6_v6only
 /tcp_msg_peek
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -83,6 +83,7 @@ c_executables = \
 	spinlock \
 	stat_invalid_args \
 	syscall \
+	syscall_restart \
 	sysfs_common \
 	tcp_ipv6_v6only \
 	tcp_msg_peek \
--- a/LibOS/shim/test/regression/syscall_restart.c
+++ b/LibOS/shim/test/regression/syscall_restart.c
@@ -1,0 +1,67 @@
+#define _GNU_SOURCE
+#include <err.h>
+#include <stdio.h>
+#include <signal.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+static void child(int fd) {
+    /* We need to wait till parent sleeps on `read`, unfortunately there is no way to check that in
+     * Graphene. Hopyfully 100ms is enough. */
+    if (usleep(100 * 1000) < 0) {
+        err(1, "usleep");
+    }
+
+    if (kill(getppid(), SIGCHLD) < 0) {
+        err(1, "kill");
+    }
+
+    /* Let parent notice the signal. */
+    if (usleep(100 * 1000) < 0) {
+        err(1, "usleep");
+    }
+
+    char c = 'a';
+    if (write(fd, &c, 1) != 1) {
+        err(1, "write");
+    }
+}
+
+int main(int argc, char** argv) {
+    setbuf(stdout, NULL);
+    setbuf(stderr, NULL);
+
+    int pfd[2];
+    if (pipe(pfd) < 0) {
+        err(1, "pipe");
+    }
+
+    pid_t p = fork();
+    if (p < 0) {
+        err(1, "fork");
+    } else if (p == 0) {
+        if (close(pfd[0]) < 0) {
+            err(1, "close");
+        }
+        child(pfd[1]);
+        return 0;
+    }
+
+    char c = 0;
+    if (read(pfd[0], &c, 1) != 1) {
+        err(1, "read");
+    }
+
+    int status = 0;
+    if (waitpid(p, &status, 0) < 0) {
+        err(1, "wait");
+    }
+
+    if (!WIFEXITED(status) || WEXITSTATUS(status) != 0) {
+        errx(1, "child died with: %d", status);
+    }
+
+    puts("TEST OK");
+    return 0;
+}
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -543,16 +543,22 @@ class TC_30_Syscall(RegressionTestCase):
         stdout, _ = self.run_binary(['gettimeofday'])
         self.assertIn('TEST OK', stdout)
 
-@unittest.skipUnless(HAS_SGX,
-    'This test is only meaningful on SGX PAL because only SGX catches raw '
-    'syscalls and redirects to Graphene\'s LibOS. If we will add seccomp to '
-    'Linux PAL, then we should allow this test on Linux PAL as well.')
-class TC_31_SyscallSGX(RegressionTestCase):
+class TC_31_Syscall(RegressionTestCase):
+    @unittest.skipUnless(HAS_SGX,
+        'This test is only meaningful on SGX PAL because only SGX catches raw '
+        'syscalls and redirects to Graphene\'s LibOS. If we will add seccomp to '
+        'Linux PAL, then we should allow this test on Linux PAL as well.')
     def test_000_syscall_redirect(self):
         stdout, _ = self.run_binary(['syscall'])
 
         # Syscall Instruction Redirection
         self.assertIn('Hello world', stdout)
+
+    @unittest.skipIf(HAS_SGX, 'currently broken on SGX due to pipe encryption '
+        'layer translating EINTR to EAGAIN')
+    def test_010_syscall_restart(self):
+        stdout, _ = self.run_binary(['syscall_restart'])
+        self.assertIn('TEST OK', stdout)
 
 class TC_40_FileSystem(RegressionTestCase):
     def test_000_proc(self):


mkow (changes requested):
> Reviewed 12 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[LibOS/shim/test/regression/syscall_restart.c, line 4 at r1](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaPDvO_46PHndVuaxRO:-MaPDvO_46PHndVuaxRP:b8u3k7c) ([raw file](https://github.com/oscarlab/graphene/blob/d18709fbf565f8c43116ddbb49ae19adb8fc8177/LibOS/shim/test/regression/syscall_restart.c#L4)):*
> ```C
> #include <err.h>
> #include <stdio.h>
> #include <signal.h>
> ```

please fix sorting
___
*[LibOS/shim/test/regression/syscall_restart.c, line 11 at r1](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaPEFlRDNbXBo1Deh91:-MaPEFlRDNbXBo1Deh92:b-hd0m3i) ([raw file](https://github.com/oscarlab/graphene/blob/d18709fbf565f8c43116ddbb49ae19adb8fc8177/LibOS/shim/test/regression/syscall_restart.c#L11)):*
> ```C
> static void child(int fd) {
>     /* We need to wait till parent sleeps on `read`, unfortunately there is no way to check that in
>      * Graphene. Hopyfully 100ms is enough. */
> ```

`Hopyfully` -> `Hopefully`
___
*[LibOS/shim/test/regression/syscall_restart.c, line 11 at r1](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaPEn4WDlSBJWVWUpDu:-MaPEn4WDlSBJWVWUpDv:b-n7lj82) ([raw file](https://github.com/oscarlab/graphene/blob/d18709fbf565f8c43116ddbb49ae19adb8fc8177/LibOS/shim/test/regression/syscall_restart.c#L11)):*
> ```C
> static void child(int fd) {
>     /* We need to wait till parent sleeps on `read`, unfortunately there is no way to check that in
>      * Graphene. Hopyfully 100ms is enough. */
> ```

This comment sounds more worrying than it really is, could you add that in the worst case of losing the race the test will just succeed without really testing the restart handling? (i.e. there's no risk of random failures due to this)
___
*[LibOS/shim/test/regression/syscall_restart.c, line 16 at r1](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaPEW5Z2SKXVKxSH41c:-MaPEW5Z2SKXVKxSH41d:b4ko8c5) ([raw file](https://github.com/oscarlab/graphene/blob/d18709fbf565f8c43116ddbb49ae19adb8fc8177/LibOS/shim/test/regression/syscall_restart.c#L16)):*
> ```C
>     }
> 
>     if (kill(getppid(), SIGCHLD) < 0) {
> ```

Could we also test a handler with `SA_RESTART` in this test? Or maybe it's not implemented correctly yet?
___
*[LibOS/shim/test/regression/syscall_restart.c, line 62 at r1](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaPFD3B4QiQqJU7sLYw:-MaPFD3C6kP-07Jjbo83:b-hiez4x) ([raw file](https://github.com/oscarlab/graphene/blob/d18709fbf565f8c43116ddbb49ae19adb8fc8177/LibOS/shim/test/regression/syscall_restart.c#L62)):*
> ```C
> 
>     if (!WIFEXITED(status) || WEXITSTATUS(status) != 0) {
>         errx(1, "child died with: %d", status);
> ```

Shouldn't you wrap `status` in some macro to get a meaningful value?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: 10 of 12 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/test/regression/syscall_restart.c, line 4 at r1](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaPDvO_46PHndVuaxRO:-MaQOwEEADk_5FkChhGB:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d18709fbf565f8c43116ddbb49ae19adb8fc8177/LibOS/shim/test/regression/syscall_restart.c#L4)):*

please fix sorting
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/syscall_restart.c, line 11 at r1](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaPEFlRDNbXBo1Deh91:-MaQP0cLClSxocLOtkUK:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d18709fbf565f8c43116ddbb49ae19adb8fc8177/LibOS/shim/test/regression/syscall_restart.c#L11)):*

`Hopyfully` -> `Hopefully`
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/syscall_restart.c, line 11 at r1](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaPEn4WDlSBJWVWUpDu:-MaQPgP63JeYHP_liQ3h:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d18709fbf565f8c43116ddbb49ae19adb8fc8177/LibOS/shim/test/regression/syscall_restart.c#L11)):*

This comment sounds more worrying than it really is, could you add that in the worst case of losing the race the test will just succeed without really testing the restart handling? (i.e. there's no risk of random failures due to this)
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/syscall_restart.c, line 16 at r1](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaPEW5Z2SKXVKxSH41c:-MaQRyb23zHE8QOQGSDC:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d18709fbf565f8c43116ddbb49ae19adb8fc8177/LibOS/shim/test/regression/syscall_restart.c#L16)):*

Could we also test a handler with `SA_RESTART` in this test? Or maybe it's not implemented correctly yet?
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/syscall_restart.c, line 62 at r1](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaPFD3B4QiQqJU7sLYw:-MaQS-tMDrtMKVr-jDF3:b-fz3d8m) ([raw file](https://github.com/oscarlab/graphene/blob/d18709fbf565f8c43116ddbb49ae19adb8fc8177/LibOS/shim/test/regression/syscall_restart.c#L62)):*

Shouldn't you wrap `status` in some macro to get a meaningful value?
</blockquote></details>

Wast of time, if this triggers then we have a bug in Graphene - we can investigate only then.
Besides, this value can be anything, died from signal, weirdly stopped, exited with non-zero exit code - this would be a big switch printing different stuff depending on value. IMO it's not needed here - you get all info in logs, you can parse it later.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/test/regression/syscall_restart.c, line 62 at r1](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaPFD3B4QiQqJU7sLYw:-MaQaPBZ45FIs-97WI6s:b-nec2kw) ([raw file](https://github.com/oscarlab/graphene/blob/d18709fbf565f8c43116ddbb49ae19adb8fc8177/LibOS/shim/test/regression/syscall_restart.c#L62)):*

Wast of time, if this triggers then we have a bug in Graphene - we can investigate only then.
Besides, this value can be anything, died from signal, weirdly stopped, exited with non-zero exit code - this would be a big switch printing different stuff depending on value. IMO it's not needed here - you get all info in logs, you can parse it later.
</blockquote></details>

Ok, makes sense.
___
*[LibOS/shim/test/regression/test_libos.py, line 567 at r2](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaQ_kY5BsJ2hrzntYG2:-MaQ_kY6457a5gNE996r:b-5fmsud) ([raw file](https://github.com/oscarlab/graphene/blob/59107fd2d180ea7b210574685d8d758d9526b788/LibOS/shim/test/regression/test_libos.py#L567)):*
> ```Python
> Got: P
> TEST 2 OK
> """
> ```

Hmm, won't this test fail if we lose the race, after you asserted the point where the handler printed the message? I think we have to split this into separate `if`s somehow. Or maybe remove the print from the handler, but set a global variable and check it in the C code?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/test/regression/test_libos.py, line 567 at r2](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaQ_kY5BsJ2hrzntYG2:-MaTyJ8I2Mu6BslDo4bi:b-3vl1yu) ([raw file](https://github.com/oscarlab/graphene/blob/59107fd2d180ea7b210574685d8d758d9526b788/LibOS/shim/test/regression/test_libos.py#L567)):*

Hmm, won't this test fail if we lose the race, after you asserted the point where the handler printed the message? I think we have to split this into separate `if`s somehow. Or maybe remove the print from the handler, but set a global variable and check it in the C code?
</blockquote></details>

The race with window artificially decreased by sleeps? No.
Unfortunately there is another (rather hidden) race here: we have no guarantees that `kill -> write` sequence won't be reordered, mainly due to signals being handled in another thread (ipc worker thread receiving the message, appending the signal and waking up the appropriate thread). I'm not sure what guarantees normal Linux gives w.r.t. to such ordering. Probably none?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/test/regression/test_libos.py, line 567 at r2](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaQ_kY5BsJ2hrzntYG2:-MaVLMqpAy4sKpNIhZQ2:brsbd1x) ([raw file](https://github.com/oscarlab/graphene/blob/59107fd2d180ea7b210574685d8d758d9526b788/LibOS/shim/test/regression/test_libos.py#L567)):*

The race with window artificially decreased by sleeps? No.
Unfortunately there is another (rather hidden) race here: we have no guarantees that `kill -> write` sequence won't be reordered, mainly due to signals being handled in another thread (ipc worker thread receiving the message, appending the signal and waking up the appropriate thread). I'm not sure what guarantees normal Linux gives w.r.t. to such ordering. Probably none?
</blockquote></details>

As per a chat on priv: Borys will either loose the checks in the test or add this guarantee to Graphene (by ACKing IPC kill messages).

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: 11 of 12 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/test/regression/test_libos.py, line 567 at r2](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaQ_kY5BsJ2hrzntYG2:-MaVSwG3CaveNutBoY4H:bea0off) ([raw file](https://github.com/oscarlab/graphene/blob/59107fd2d180ea7b210574685d8d758d9526b788/LibOS/shim/test/regression/test_libos.py#L567)):*

As per a chat on priv: Borys will either loose the checks in the test or add this guarantee to Graphene (by ACKing IPC kill messages).
</blockquote></details>

Just loosened the test.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (changes requested):
> Reviewed 10 of 12 files at r1, 1 of 2 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/src/bookkeep/shim_thread.c, line 195 at r3](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXb407BTvO7oFILpEA:-MaXb407BTvO7oFILpEB:bl3euvi) ([raw file](https://github.com/oscarlab/graphene/blob/368ede712a443ea47c45836bdd6b5880b27e8233/LibOS/shim/src/bookkeep/shim_thread.c#L195)):*
> ```C
> 
>     int ret = DkEventCreate(&cur_thread->scheduler_event, /*init_signaled=*/false,
>                             /*auto_clear=*/true);
> ```

Why is this change? Doesn't look related to syscall restarts.
___
*[LibOS/shim/src/sys/shim_socket.c, line 1372 at r3](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXc9FzBouOgys_OkEq:-MaXc9FzBouOgys_OkEr:b-9mmkuf) ([raw file](https://github.com/oscarlab/graphene/blob/368ede712a443ea47c45836bdd6b5880b27e8233/LibOS/shim/src/sys/shim_socket.c#L1372)):*
> ```C
>             ret = DkStreamRead(pal_hdl, /*offset=*/0, &left_to_read,
>                                &peek_buffer->buf[peek_buffer->end], uri, uri ? SOCK_URI_SIZE : 0);
>             /* TODO: shouldn't we call `maybe_epoll_et_trigger` here? */
> ```

Yeah, true, looks like we should. Not blocking, probably for another PR.
___
*[LibOS/shim/test/regression/syscall_restart.c, line 1 at r3](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXekgS9Lcyk11BC4nx:-MaXekgS9Lcyk11BC4ny:boy2huf) ([raw file](https://github.com/oscarlab/graphene/blob/368ede712a443ea47c45836bdd6b5880b27e8233/LibOS/shim/test/regression/syscall_restart.c#L1)):*
> ```C
> #define _GNU_SOURCE
> ```

Could you add a top-level comment to this test? Something along those lines that: the child process first sends a signal to the parent and then writes one char to the pipe, whereas the parent first reads from the pipe (and expects to read one char) and then waits on the child. This tests the syscall-restart capability of the `read()` syscall, in two flavors: when the parent process doesn't have a handler (and `read()` syscall is restarted by default) and when the parent process has an `SA_RESTART` handler.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXu3PO2uUDXzVTB2BS:-MaXu3PP3BtRgS14nwll:bbxrvqk) (no related file):*
> Unfortunately it fails on SGX even with this PR due to pipe encryption layer translating EINTR to EAGAIN

I looked into this, and it's not good. mbedTLS doesn't distinguish between EINTR and EAGAIN/EWOULDBLOCK; mbedTLS only has `MBEDTLS_ERR_SSL_WANT_READ/WRITE`. So we technically can't do anything better than the current line: https://github.com/oscarlab/graphene/blob/75962a812d686e839fabe5d9525e52128a8f1c86/common/src/crypto/adapters/mbedtls_adapter.c#L291

However, we could try to side-step this by changing like this
```
        case MBEDTLS_ERR_SSL_WANT_READ:
        case MBEDTLS_ERR_SSL_WANT_WRITE:
            return -PAL_ERROR_TRYAGAIN;

        case MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS:
        case MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS:
            return -PAL_ERROR_INTERRUPTED; // this is the change
```

And then changing our callbacks:
```
static int recv_cb(void* ctx, uint8_t* buf, size_t buf_size) {
    ...
    if (ret < 0) {
        if (ret == -EINTR)
            return MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS;
        if (ret == -EAGAIN || ret == -EWOULDBLOCK)
            return MBEDTLS_ERR_SSL_WANT_READ;
        if (ret == -EPIPE)
            return MBEDTLS_ERR_NET_CONN_RESET;
        return MBEDTLS_ERR_NET_RECV_FAILED;
    }

    return ret;
}
```

(same for `send_cb`)

This is not perfect because mbedTLS API doesn't mention that send/recv callbacks can return anything other than `MBEDTLS_ERR_SSL_WANT_READ/MBEDTLS_ERR_NET_RECV_FAILED/MBEDTLS_ERR_NET_CONN_RESET`. But I looked into the current mbedTLS code, it simply propagates the error up to the caller (which is `lib_SSLRead/Write` in our Graphene case). So we can abuse `MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS` for this, should be fine.

@boryspoplawski What do you think about this workaround? Could you add it here? Or should I create a separate PR with this?

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXu3PO2uUDXzVTB2BS:-MaY-Qk75oO3zAa8zLs-:bc863j1) (no related file):*

> Unfortunately it fails on SGX even with this PR due to pipe encryption layer translating EINTR to EAGAIN

I looked into this, and it's not good. mbedTLS doesn't distinguish between EINTR and EAGAIN/EWOULDBLOCK; mbedTLS only has `MBEDTLS_ERR_SSL_WANT_READ/WRITE`. So we technically can't do anything better than the current line: https://github.com/oscarlab/graphene/blob/75962a812d686e839fabe5d9525e52128a8f1c86/common/src/crypto/adapters/mbedtls_adapter.c#L291

However, we could try to side-step this by changing like this
```
        case MBEDTLS_ERR_SSL_WANT_READ:
        case MBEDTLS_ERR_SSL_WANT_WRITE:
            return -PAL_ERROR_TRYAGAIN;

        case MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS:
        case MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS:
            return -PAL_ERROR_INTERRUPTED; // this is the change
```

And then changing our callbacks:
```
static int recv_cb(void* ctx, uint8_t* buf, size_t buf_size) {
    ...
    if (ret < 0) {
        if (ret == -EINTR)
            return MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS;
        if (ret == -EAGAIN || ret == -EWOULDBLOCK)
            return MBEDTLS_ERR_SSL_WANT_READ;
        if (ret == -EPIPE)
            return MBEDTLS_ERR_NET_CONN_RESET;
        return MBEDTLS_ERR_NET_RECV_FAILED;
    }

    return ret;
}
```

(same for `send_cb`)

This is not perfect because mbedTLS API doesn't mention that send/recv callbacks can return anything other than `MBEDTLS_ERR_SSL_WANT_READ/MBEDTLS_ERR_NET_RECV_FAILED/MBEDTLS_ERR_NET_CONN_RESET`. But I looked into the current mbedTLS code, it simply propagates the error up to the caller (which is `lib_SSLRead/Write` in our Graphene case). So we can abuse `MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS` for this, should be fine.

@boryspoplawski What do you think about this workaround? Could you add it here? Or should I create a separate PR with this?
</blockquote></details>

Sorry, a small typo in the first listing: `MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS` should still return `PAL_ERROR_TRYAGAIN`.

---

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: 11 of 12 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXu3PO2uUDXzVTB2BS:-MaYa61NC96Aa-Urlfst:b-h2ufn4) (no related file):*
I think it should be the other way around: `ASYNC_IN_PROGRESS` should translate to `TRYAGAIN` and `WANT_READ` to `EINTR`.
What are these names supposed to mean anyway? Does mbedtls has any description of what each error means?

> Could you add it here?

Yes, I will do, once we settle on what to do exactly.

---
___
*[LibOS/shim/src/bookkeep/shim_thread.c, line 195 at r3](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXb407BTvO7oFILpEA:-MaYac-68pt3-rq_aXdd:b-rvmnfx) ([raw file](https://github.com/oscarlab/graphene/blob/368ede712a443ea47c45836bdd6b5880b27e8233/LibOS/shim/src/bookkeep/shim_thread.c#L195)):*

Why is this change? Doesn't look related to syscall restarts.
</blockquote></details>

Ah, yes. This was actually a bug: previously we looped around `thread_wait` without resetting the event, which basically turned it into active wait (busy loop).
___
*[LibOS/shim/test/regression/syscall_restart.c, line 1 at r3](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXekgS9Lcyk11BC4nx:-MaYb18C-rQy9WCpssRk:b-cl8j0s) ([raw file](https://github.com/oscarlab/graphene/blob/368ede712a443ea47c45836bdd6b5880b27e8233/LibOS/shim/test/regression/syscall_restart.c#L1)):*

Could you add a top-level comment to this test? Something along those lines that: the child process first sends a signal to the parent and then writes one char to the pipe, whereas the parent first reads from the pipe (and expects to read one char) and then waits on the child. This tests the syscall-restart capability of the `read()` syscall, in two flavors: when the parent process doesn't have a handler (and `read()` syscall is restarted by default) and when the parent process has an `SA_RESTART` handler.
</blockquote></details>

I've skipped the first part, since it just describes code line by line, but added the second part.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXu3PO2uUDXzVTB2BS:-MaYfTxVEJzxKe7IZmoR:b-l4whe7) (no related file):*

I think it should be the other way around: `ASYNC_IN_PROGRESS` should translate to `TRYAGAIN` and `WANT_READ` to `EINTR`.
What are these names supposed to mean anyway? Does mbedtls has any description of what each error means?

> Could you add it here?

Yes, I will do, once we settle on what to do exactly.
</blockquote></details>

- https://tls.mbed.org/api/ssl_8h.html#ac06b19b96b643090c55f19b6c28b10c4 -- descriptions of macros
- https://tls.mbed.org/api/ssl_8h.html#aa2c29eeb1deaf5ad9f01a7515006ede5 -- description of `mbedtls_ssl_read` (it uses our callback for an underlying host-level read)
- https://tls.mbed.org/api/ssl_8h.html#ac06b19b96b643090c55f19b6c28b10c4 -- description of the callback

---
___
*[LibOS/shim/src/bookkeep/shim_thread.c, line 195 at r3](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXb407BTvO7oFILpEA:-MaYg9320oHhmmmnMpoG:b-uowl23) ([raw file](https://github.com/oscarlab/graphene/blob/368ede712a443ea47c45836bdd6b5880b27e8233/LibOS/shim/src/bookkeep/shim_thread.c#L195)):*

Ah, yes. This was actually a bug: previously we looped around `thread_wait` without resetting the event, which basically turned it into active wait (busy loop).
</blockquote></details>

OK. I'd prefer to split this in a separate commit, but not blocking.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXu3PO2uUDXzVTB2BS:-MaZ4XHa6JskP0lGO1Bz:b-sbc8ii) (no related file):*

Thanks.
Ok, from reading the sources it turns out that indeed `EINTR` and `EAGAIN` are indistinguishable in mbedtls: https://github.com/ARMmbed/mbedtls/blob/development/include/mbedtls/ssl.h#L544
When I think about it, it actually makes sense: it should follow from the context: non-blocking IO fails with `EAGAIN` and never with `EINTR`, blocking the other way around.
Now this gets tricky: we need to somehow know if the fd is blocking or non-blocking, which I think we cannot know on mbedtls wrappers layer, only on LibOS layer. But we don't want to return mbedtls errors to LibOS (or do we?)...
Any ideas?
</blockquote></details>

But what about my workaround? We can still distinguish between EINTR and EAGAIN at the level of `ocall_read()`/`ocall_write()`, and then propagate it to LibOS via the following transformations:
- `EINTR` -> `MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS` -> (hope that mbedTLS simply propagates to our PAL code) -> `PAL_ERROR_INTERRUPTED` -> LibOS
- `EAGAIN` -> `MBEDTLS_ERR_SSL_WANT_READ` -> `PAL_ERROR_TRYAGAIN` -> LibOS

This should work at least in our current mbedTLS version (2.26.0), by manually inspecting the code.

---

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXu3PO2uUDXzVTB2BS:-MaZFXTe69CiAZ5rf9Yq:b-xv6y5t) (no related file):*

But what about my workaround? We can still distinguish between EINTR and EAGAIN at the level of `ocall_read()`/`ocall_write()`, and then propagate it to LibOS via the following transformations:
- `EINTR` -> `MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS` -> (hope that mbedTLS simply propagates to our PAL code) -> `PAL_ERROR_INTERRUPTED` -> LibOS
- `EAGAIN` -> `MBEDTLS_ERR_SSL_WANT_READ` -> `PAL_ERROR_TRYAGAIN` -> LibOS

This should work at least in our current mbedTLS version (2.26.0), by manually inspecting the code.
</blockquote></details>

That's the problem: it's just a workaround that would currently work without any guarantees it doesn't brake if we update mbedtls. It basically reuses existing error code for another meaning, hoping that the original is not used. I was hoping for a more robust solution.
Also note that we do not get mbedtls erros in PAL, they are translated already on `mbedtls_adapter` layer.

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXu3PO2uUDXzVTB2BS:-MaZVUlC8aKXg0c1Mo0F:b-7nbcmd) (no related file):*
> Also note that we do not get mbedtls erros in PAL, they are translated already on mbedtls_adapter layer.

Yes, I know this. In the code snippet above I also suggested to change that mbedtls-to-pal error code translation function. 

> It basically reuses existing error code for another meaning, hoping that the original is not used. I was hoping for a more robust solution.

Yeah, I understand. Not sure how to make it robustly (in a simple way).

---

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: 9 of 17 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXu3PO2uUDXzVTB2BS:-Maa5NTQ06kFEjPmnq6u:buba9cj) (no related file):*

> Also note that we do not get mbedtls erros in PAL, they are translated already on mbedtls_adapter layer.

Yes, I know this. In the code snippet above I also suggested to change that mbedtls-to-pal error code translation function. 

> It basically reuses existing error code for another meaning, hoping that the original is not used. I was hoping for a more robust solution.

Yeah, I understand. Not sure how to make it robustly (in a simple way).
</blockquote></details>

I've found a better way. Turns out we know whether handle is blocking or not on each callsite to `DkSteramSecureSomething`

---

dimakuv (changes requested):
> Reviewed 8 of 8 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2392#-MaXu3PO2uUDXzVTB2BS:-MabG72lDfENVLTQft6x:b-uflez1) (no related file):*

I've found a better way. Turns out we know whether handle is blocking or not on each callsite to `DkSteramSecureSomething`
</blockquote></details>

OK, awesome!

---
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1238 at r5](https://reviewable.io/reviews/oscarlab/graphene/2392#-MabFlSP81fIGTRmi_iJ:-MabFlSQ780I0C3JdkM6:bjz40pf) ([raw file](https://github.com/oscarlab/graphene/blob/6eb3b8e33203a8f43646dac8d1c58f15416806fc/Pal/src/host/Linux-SGX/enclave_framework.c#L1238)):*
> ```C
>          * split it back. */
>         return -PAL_ERROR_INTERRUPTED;
>     }
> ```

Do you want to add an opposite case here as well? `if (!is_blocking && ret == -PAL_ERROR_INTERRUPTED)  return -PAL_ERROR_TRYAGAIN`. This shouldn't happen in benign case, but may happen due to bugs in our OCALL SGX code or due to a malicious host.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1248 at r5](https://reviewable.io/reviews/oscarlab/graphene/2392#-MabG3s10zgkT2S6HRUw:-MabG3s2BS7C9fRRXIav:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/6eb3b8e33203a8f43646dac8d1c58f15416806fc/Pal/src/host/Linux-SGX/enclave_framework.c#L1248)):*
> ```C
>         /* See the explanation in `_DkStreamSecureRead`. */
>         return -PAL_ERROR_INTERRUPTED;
>     }
> ```

ditto

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1238 at r5](https://reviewable.io/reviews/oscarlab/graphene/2392#-MabFlSP81fIGTRmi_iJ:-MacXblLEVrbB5Gor3zX:bxxkglq) ([raw file](https://github.com/oscarlab/graphene/blob/6eb3b8e33203a8f43646dac8d1c58f15416806fc/Pal/src/host/Linux-SGX/enclave_framework.c#L1238)):*

Do you want to add an opposite case here as well? `if (!is_blocking && ret == -PAL_ERROR_INTERRUPTED)  return -PAL_ERROR_TRYAGAIN`. This shouldn't happen in benign case, but may happen due to bugs in our OCALL SGX code or due to a malicious host.
</blockquote></details>

No, that would be incorrect. `EAGAIN` means that there is no data to read, so we cannot have spurious returns like that (e.g. combined with EPOLLET that could cause deadlocks in correct apps).
Besides our mbedTLS wrappers do not return `PAL_ERROR_INTERRUPTED`.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1248 at r5](https://reviewable.io/reviews/oscarlab/graphene/2392#-MabG3s10zgkT2S6HRUw:-MacY-I_9pce1JFoSjM_:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/6eb3b8e33203a8f43646dac8d1c58f15416806fc/Pal/src/host/Linux-SGX/enclave_framework.c#L1248)):*

ditto
</blockquote></details>

ditto

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 1238 at r5](https://reviewable.io/reviews/oscarlab/graphene/2392#-MabFlSP81fIGTRmi_iJ:-MacZ-Ka-HjP_P12VA0a:bmctigb) ([raw file](https://github.com/oscarlab/graphene/blob/6eb3b8e33203a8f43646dac8d1c58f15416806fc/Pal/src/host/Linux-SGX/enclave_framework.c#L1238)):*

No, that would be incorrect. `EAGAIN` means that there is no data to read, so we cannot have spurious returns like that (e.g. combined with EPOLLET that could cause deadlocks in correct apps).
Besides our mbedTLS wrappers do not return `PAL_ERROR_INTERRUPTED`.
</blockquote></details>

OK, true.

mkow (dismissed):
> Reviewed 8 of 8 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv (approved):
> Reviewed 1 of 8 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

mkow (approved):
> Reviewed 1 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2392)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2402, opened 2021-05-27
[LibOS] Split `shim_mount` and `shim_fs`

pwmarcz:
(I think this finally unblocks the "mount semantics" refactoring! I'm checking that right now, but it's going to be a separate PR.)

This change replaces most usages of `shim_mount` with `shim_fs`. Graphene uses the `shim_mount` structure to represent a mounted filesystem, but in most cases we're only interested in the function table.

The main reason for this change is special treatment of sockets and FIFOs: before, we replaced the `fs` field (actually `shim_mount`) for their dentries, so that they are served by a different filesystem. However, that loses information about which part of the tree the dentry actually is in.

The new solution is to store both `shim_mount` and `shim_fs` for a dentry. This follows Linux, which stores both `super_block` and `inode_iperations` for an inode, and replaces the latter in case of sockets and FIFOs. This will allow me to refactor the way mounted filesystems are represented.

As a bonus, this change removes some hacks like unwarranted usage of `*_builtin_fs` and storing filesystem type separately for checkpointing purposes.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -122,8 +122,15 @@ DEFINE_LISTP(shim_dentry);
 struct shim_dentry {
     int state; /* flags for managing state */
 
-    struct shim_mount* fs;     /* this dentry's mounted fs */
-    struct shim_qstr name;     /* caching the file's name. */
+    /* File name, maximum of NAME_MAX characters. By convention, the root has an empty name. */
+    struct shim_qstr name;
+
+    /* Mounted filesystem this dentry belongs to */
+    struct shim_mount* mount;
+
+    /* Filesystem to use for operations on this file: this is usually `mount->fs`, but can be
+     * different in case of special files (such as named pipes or sockets). */
+    struct shim_fs* fs;
 
     struct shim_dentry* parent;
     size_t nchildren;
@@ -216,22 +223,26 @@ struct shim_d_ops {
 #define NAME_MAX 255   /* filename length, NOT including null terminator */
 #define PATH_MAX 4096  /* path size, including null terminator */
 
+struct shim_fs {
+    /* Null-terminated, used in manifest and for uniquely identifying a filesystem. */
+    char name[8];
+    struct shim_fs_ops* fs_ops;
+    struct shim_d_ops* d_ops;
+};
+
 DEFINE_LIST(shim_mount);
 struct shim_mount {
-    char type[8];  // Null-terminated.
+    struct shim_fs* fs;
 
     /*
-     * TODO: this field currently functions as both the mountpoint AND the filesystem root. Instead,
-     * `mount_point` should be the dentry from the parent filesystem, and `root` should be the
-     * filesystem root.
+     * TODO: this field currently functions as both the mountpoint AND the mount root. Instead,
+     * `mount_point` should be the dentry from the parent filesystem, and `root` should be the mount
+     * root.
      */
     struct shim_dentry* mount_point;
 
     struct shim_qstr path;
     struct shim_qstr uri;
-
-    struct shim_fs_ops* fs_ops;
-    struct shim_d_ops* d_ops;
 
     /* TODO: this is unused right now (see `mount_point`) */
     struct shim_dentry* root;
@@ -277,18 +288,11 @@ int init_mount(void);
 int mount_fs(const char* mount_type, const char* mount_uri, const char* mount_point,
              struct shim_dentry* parent, struct shim_dentry** dentp, bool make_ancestor);
 int unmount_fs(const char* mount_point);
-int search_builtin_fs(const char* type, struct shim_mount** fs);
 
 void get_mount(struct shim_mount* mount);
 void put_mount(struct shim_mount* mount);
 
 struct shim_mount* find_mount_from_uri(const char* uri);
-
-static inline void set_handle_fs(struct shim_handle* hdl, struct shim_mount* fs) {
-    get_mount(fs);
-    hdl->fs = fs;
-    memcpy(hdl->fs_type, fs->type, sizeof(hdl->fs_type));
-}
 
 int walk_mounts(int (*walk)(struct shim_mount* mount, void* arg), void* arg);
 
@@ -569,7 +573,7 @@ ino_t dentry_ino(struct shim_dentry* dent);
  * \brief Allocate and initialize a new dentry
  *
  * \param parent the parent node, or NULL if this is supposed to be the dentry root
- * \param fs the filesystem the dentry is under, or NULL
+ * \param mount the mount the dentry is under
  * \param name name of the new dentry
  * \param name_len length of the name
  *
@@ -577,16 +581,16 @@ ino_t dentry_ino(struct shim_dentry* dent);
  *
  * The caller should hold `g_dcache_lock`.
  *
- * The function will initialize the following fields: `fs` (if provided), `name`, and
- * parent/children links.
+ * The function will initialize the following fields: `mount` and `fs` (if `mount` provided),
+ * `name`, and parent/children links.
  *
  * The reference count of the returned dentry will be 2 if `parent` was provided, 1 otherwise.
  *
- * The `fs` parameter should typically be `parent->fs`, but is passed explicitly to support
- * initializing the root dentry of a newly mounted filesystem. If `fs` is NULL, the resulting
- * dentry's filesystem will be left as NULL.
+ * The `mount` parameter should typically be `parent->mount`, but is passed explicitly to support
+ * initializing the root dentry of a newly mounted filesystem. The `fs` field will be initialized to
+ * `mount->fs`, but you can later change it to support special files.
  */
-struct shim_dentry* get_new_dentry(struct shim_mount* fs, struct shim_dentry* parent,
+struct shim_dentry* get_new_dentry(struct shim_mount* mount, struct shim_dentry* parent,
                                    const char* name, size_t name_len);
 
 /*!
@@ -641,12 +645,18 @@ extern struct shim_d_ops str_d_ops;
 extern struct shim_fs_ops tmp_fs_ops;
 extern struct shim_d_ops tmp_d_ops;
 
-extern struct shim_mount chroot_builtin_fs;
-extern struct shim_mount pipe_builtin_fs;
-extern struct shim_mount fifo_builtin_fs;
-extern struct shim_mount socket_builtin_fs;
-extern struct shim_mount epoll_builtin_fs;
-extern struct shim_mount eventfd_builtin_fs;
+extern struct shim_fs chroot_builtin_fs;
+extern struct shim_fs proc_builtin_fs;
+extern struct shim_fs dev_builtin_fs;
+extern struct shim_fs sys_builtin_fs;
+extern struct shim_fs tmp_builtin_fs;
+extern struct shim_fs pipe_builtin_fs;
+extern struct shim_fs fifo_builtin_fs;
+extern struct shim_fs socket_builtin_fs;
+extern struct shim_fs epoll_builtin_fs;
+extern struct shim_fs eventfd_builtin_fs;
+
+struct shim_fs* find_fs(const char* name);
 
 /* pseudo file systems (separate treatment since they don't have associated dentries) */
 #define DIR_RX_MODE  0555
--- a/LibOS/shim/include/shim_handle.h
+++ b/LibOS/shim/include/shim_handle.h
@@ -291,7 +291,7 @@ struct shim_epoll_handle {
     LISTP_TYPE(shim_epoll_item) fds;
 };
 
-struct shim_mount;
+struct shim_fs;
 struct shim_qstr;
 struct shim_dentry;
 
@@ -303,8 +303,7 @@ struct shim_handle {
 
     REFTYPE ref_count;
 
-    char fs_type[8];
-    struct shim_mount* fs;
+    struct shim_fs* fs;
     struct shim_dentry* dentry;
 
     /* If this handle is registered for any epoll handle, this list contains
--- a/LibOS/shim/src/bookkeep/shim_handle.c
+++ b/LibOS/shim/src/bookkeep/shim_handle.c
@@ -45,7 +45,7 @@ static int init_tty_handle(struct shim_handle* hdl, bool write) {
     if (ret < 0)
         return ret;
 
-    set_handle_fs(hdl, dent->fs);
+    hdl->fs = dent->fs;
     hdl->dentry = dent;
     return 0;
 }
@@ -73,15 +73,27 @@ static inline int init_exec_handle(void) {
     exec->flags    = O_RDONLY;
     exec->acc_mode = MAY_READ;
 
-    struct shim_mount* fs = find_mount_from_uri(g_pal_control->executable);
-    if (fs) {
-        const char* p = g_pal_control->executable + fs->uri.len;
+    /*
+     * Find a mounted filesystem for an executable, and initialize the handle (dentry and fs).
+     *
+     * TODO: The below code is broken, and should be rewritten:
+     *
+     * - Instead of manually constructing the handle, use the proper function for opening a file
+     *   (`open_namei`).
+     *
+     * - The treatment of relative paths is wrong, and really only makes sense for the default case
+     *   when the root mount is "file:." and executable is "file:<executable>". We should probably
+     *   convert both of these to absolute paths first.
+     */
+    struct shim_mount* mount = find_mount_from_uri(g_pal_control->executable);
+    if (mount) {
+        const char* p = g_pal_control->executable + mount->uri.len;
         /* Lookup for `g_pal_control->executable` needs to be done under a given mount point which
          * requires a relative path name. OTOH the one in manifest file can be absolute path. */
         while (*p == '/') {
             p++;
         }
-        int ret = path_lookupat(fs->mount_point, p, LOOKUP_FOLLOW, &exec->dentry);
+        int ret = path_lookupat(mount->mount_point, p, LOOKUP_FOLLOW, &exec->dentry);
         if (ret < 0) {
             log_error("init_exec_handle: cannot find executable in filesystem: %d\n", ret);
             return ret;
@@ -91,15 +103,20 @@ static inline int init_exec_handle(void) {
             log_error("init_exec_handle: executable is not a regular file\n");
             return -EINVAL;
         }
-        if (exec->dentry->fs != fs) {
+        if (exec->dentry->mount != mount) {
             log_error("init_exec_handle: cannot find executable in filesystem, "
                       "it seems to be shadowed by a mount\n");
             return -EINVAL;
         }
-        set_handle_fs(exec, fs);
-        put_mount(fs);
+        exec->fs = mount->fs;
+        put_mount(mount);
     } else {
-        set_handle_fs(exec, &chroot_builtin_fs);
+        /*
+         * TODO: This is for cases where the above lookup fails, e.g. `host_root_fs` regression test
+         * (which mounts "file:/" as root). We construct a dentry-less handle in such cases, but
+         * once the lookup is fixed, we should return an error instead.
+         */
+        exec->fs = &chroot_builtin_fs;
     }
 
     lock(&g_process.fs_lock);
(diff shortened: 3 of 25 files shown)


dimakuv (changes requested):
> Reviewed 25 of 25 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_fs.h, line 593 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0PgAeBoQgAqRJ8X38:-Mb0PgAeBoQgAqRJ8X39:b-iwm87e) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/include/shim_fs.h#L593)):*
> ```objc
>  *
>  * If the `mount` parameter is provided, the resulting dentry's `mount` field will be initialized to
>  * it, and the `fs` field will initialized to `mount->fs` (but you can override `fs` later for
> ```

`will be initialized`
___
*[LibOS/shim/include/shim_fs.h, line 594 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0PiBX-UTeRClxwiT8:-Mb0PiBX-UTeRClxwiT9:b-jkx3s3) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/include/shim_fs.h#L594)):*
> ```objc
>  * If the `mount` parameter is provided, the resulting dentry's `mount` field will be initialized to
>  * it, and the `fs` field will initialized to `mount->fs` (but you can override `fs` later for
>  * special files). Otherwise, both `mount` and `fs` parameters will be left as NULL.
> ```

What's the point of having `dentry.mount` and `dentry.fs` as NULL? Is it used in Graphene? Can we add an assert that `mount != NULL`?
___
*[LibOS/shim/src/bookkeep/shim_handle.c, line 102 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0R24UB_4qdIWPSu1M:-Mb0R24UB_4qdIWPSu1N:bwn55wt) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/bookkeep/shim_handle.c#L102)):*
> ```C
>         put_mount(mount);
>     } else {
>         exec->fs = &chroot_builtin_fs;
> ```

Actually, what is the point of this hack? If we end up here, then the executable is not really mounted, so it is like a "pseudo" file. How is this useful?
___
*[LibOS/shim/src/bookkeep/shim_handle.c, line 655 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0RcV51zmFCDl5ryLR:-Mb0RcV51zmFCDl5ryLS:bh0ibil) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/bookkeep/shim_handle.c#L655)):*
> ```C
> 
>         lock(&hdl->lock);
>         struct shim_fs* fs = hdl->fs;
> ```

Can we remove this `fs` alias and use `hdl->fs` in the below two lines of code?
___
*[LibOS/shim/src/fs/shim_fs.c, line 615 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0T3bN79pDzZBzJvb_:-Mb0T3bN79pDzZBzJvba:b-mjcm73) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/fs/shim_fs.c#L615)):*
> ```C
>  * Note that checkpointing the `shim_fs` structure copies it, instead of using a pointer to
>  * corresponding global object on the remote side. This does not waste too much memory (because each
>  * global object is only copied once), but it means that `shim_fs` objects cannot be compared by
> ```

Btw, did you check that `each global object is only copied once` is actually true? I know we have this `GET_FROM_CP_MAP(obj)` that avoids duplication of already-checkpointed objects (based on pointer value), but I never checked if this macro actually works as expected.
___
*[LibOS/shim/src/fs/pipe/fs.c, line 264 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0ULyQ1w3ElZESOIhP:-Mb0ULyQ1w3ElZESOIhQ:bdb29q1) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/fs/pipe/fs.c#L264)):*
> ```C
> 
> struct shim_fs fifo_builtin_fs = {
>     .name   = "fifo",
> ```

Why do we have hard-coded `fifo` instead of macro like `URI_TYPE_FIFO`?

Or maybe the question should be the other way around: why do we have `URI_TYPE_PIPE` instead of hard-coded `pipe`?

pwmarcz (commented):
Rebased to fix typo in a commit message: `inode_iperations` -> `inode_operations`.

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402#-:-Mb1E9Bz5OMZ3bnr5gib:b-ey8kaw)* status: 20 of 25 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[LibOS/shim/include/shim_fs.h, line 593 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0PgAeBoQgAqRJ8X38:-Mb14agkEmeBXADSmx_k:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/include/shim_fs.h#L593)):*

`will be initialized`
</blockquote></details>

Fixed.
___
*[LibOS/shim/include/shim_fs.h, line 594 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0PiBX-UTeRClxwiT8:-Mb15Jbl8gshMpze7jSi:b-kgxrqz) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/include/shim_fs.h#L594)):*

What's the point of having `dentry.mount` and `dentry.fs` as NULL? Is it used in Graphene? Can we add an assert that `mount != NULL`?
</blockquote></details>

There's the `g_dentry_root`, which is initialized as "mount-less" dentry. But it does not use `get_new_dentry()`.
___
*[LibOS/shim/src/bookkeep/shim_handle.c, line 102 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0R24UB_4qdIWPSu1M:-Mb17oWD03A2JCEoxL5O:bfvodoj) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/bookkeep/shim_handle.c#L102)):*

Actually, what is the point of this hack? If we end up here, then the executable is not really mounted, so it is like a "pseudo" file. How is this useful?
</blockquote></details>

I removed it and I got an answer immediately by running the tests:

See the `host_root_fs` test. It mounts `/` as root, but specifies the executable as relative URI same as other tests (`file:host_root_fs`). So we're not able to determine where `host_root_fs` is in the mounted filesystem, and execute this line, initializing a dentry-less handle.

Apparently this code for finding a mount works mostly for the common case of mounting `file:.` as root, and specifying executable in relative URI.

I think a proper fix would be to deal with absolute URIs, both for executable and for mounted filesystems. The manifest may still specify to mount `file:.`, but internally we translate it to absolute path the manifest itself (probably supplied by PAL). Does that make sense?

(I'd like to leave a TODO / explanation here after we agree on something).
___
*[LibOS/shim/src/bookkeep/shim_handle.c, line 655 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0RcV51zmFCDl5ryLR:-Mb19dTB-XY5d8F154wr:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/bookkeep/shim_handle.c#L655)):*

Can we remove this `fs` alias and use `hdl->fs` in the below two lines of code?
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/shim_fs.c, line 615 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0T3bN79pDzZBzJvb_:-Mb1ByA87yV2zpsQepYm:b-gyo1x7) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/fs/shim_fs.c#L615)):*

Btw, did you check that `each global object is only copied once` is actually true? I know we have this `GET_FROM_CP_MAP(obj)` that avoids duplication of already-checkpointed objects (based on pointer value), but I never checked if this macro actually works as expected.
</blockquote></details>

Yes, I added debug prints to both branches in this function and they confirm that:

```
[P26031:T1:getdents] fs chroot
[P26031:T1:getdents] fs existing chroot
[P26031:T1:getdents] fs proc
[P26031:T1:getdents] fs existing proc
[P26031:T1:getdents] fs dev
[P26031:T1:getdents] fs existing dev
[P26031:T1:getdents] fs existing chroot
[P26031:T1:getdents] fs existing chroot
[P26031:T1:getdents] fs sys
[P26031:T1:getdents] fs existing sys
[P26031:T1:getdents] fs existing chroot
[P26031:T1:getdents] fs existing chroot
[P26031:T1:getdents] fs existing chroot
[P26031:T1:getdents] fs existing chroot
```
___
*[LibOS/shim/src/fs/pipe/fs.c, line 264 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0ULyQ1w3ElZESOIhP:-Mb1DXCICZKzKadzkunO:b-7rk464) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/fs/pipe/fs.c#L264)):*

Why do we have hard-coded `fifo` instead of macro like `URI_TYPE_FIFO`?

Or maybe the question should be the other way around: why do we have `URI_TYPE_PIPE` instead of hard-coded `pipe`?
</blockquote></details>

Yeah, looks inconsistent. I'd rather hardcode the name, it seems like a misuse of `URI_TYPE_*`.

(Also I found one for eventfd and changed it).

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/src/bookkeep/shim_handle.c, line 102 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0R24UB_4qdIWPSu1M:-Mb1N4dr90VFJ9Ps2kzy:b-dkcje4) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/bookkeep/shim_handle.c#L102)):*

I removed it and I got an answer immediately by running the tests:

See the `host_root_fs` test. It mounts `/` as root, but specifies the executable as relative URI same as other tests (`file:host_root_fs`). So we're not able to determine where `host_root_fs` is in the mounted filesystem, and execute this line, initializing a dentry-less handle.

Apparently this code for finding a mount works mostly for the common case of mounting `file:.` as root, and specifying executable in relative URI.

I think a proper fix would be to deal with absolute URIs, both for executable and for mounted filesystems. The manifest may still specify to mount `file:.`, but internally we translate it to absolute path the manifest itself (probably supplied by PAL). Does that make sense?

(I'd like to leave a TODO / explanation here after we agree on something).
</blockquote></details>

I didn't understand your proposal for a fix. Could you maybe describe on the example of `host_root_fs`? Here is the current `host_root_fs.manifest.template` for ease of discussion (I retained only the relevant lines):
```
libos.entrypoint = "file:host_root_fs"

fs.root.type = "chroot"
fs.root.path = "/"
fs.root.uri = "file:/"

sgx.trusted_files.host_root_fs = "file:host_root_fs"
```

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[LibOS/shim/src/bookkeep/shim_handle.c, line 102 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0R24UB_4qdIWPSu1M:-Mb1XH772wB4LmoDrdXm:bedx17h) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/bookkeep/shim_handle.c#L102)):*

I didn't understand your proposal for a fix. Could you maybe describe on the example of `host_root_fs`? Here is the current `host_root_fs.manifest.template` for ease of discussion (I retained only the relevant lines):
```
libos.entrypoint = "file:host_root_fs"

fs.root.type = "chroot"
fs.root.path = "/"
fs.root.uri = "file:/"

sgx.trusted_files.host_root_fs = "file:host_root_fs"
```
</blockquote></details>

OK. Following that example, let's say we run this from `/home/user/graphene/LibOS/shim/test/regression`, and all relative paths should be resolved relative to this directory.

PAL would determine that path during bootstrapping, and pass it to LibOS as something like `g_pal_control->host_cwd_uri = "file:/home/user/graphene/LibOS/shim/test/regression"`.

Then:
```c
const char* exec_uri = g_pal_control->executable;
// exec_uri = "file:host_root_fs"

exec_uri = to_abspath(g_pal_control->host_cwd_uri, exec_uri);
// exec_uri = "file:/home/user/graphene/LibOS/shim/test/regression"

struct shim_mount* mount;
const char* exec_rel_path;
find_mount(exec_uri, &mount, &exec_rel_path);
// mount = <chroot mount for />
// mount->uri = "file:/"
// exec_rel_path = "home/user/graphene/LibOS/shim/test/regression"

struct shim_dentry* dent;
path_lookupat(mount->mount_point, exec_rel_path, ..., &dent);
```

The same code would work in the "regular" case, i.e. when `fs.root.uri = "file:.", because we would also convert `mount->uri` to be an absolute. It would also work for other combinations of mount URI / mount path / executable URI that are probably broken right now.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[LibOS/shim/src/bookkeep/shim_handle.c, line 102 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0R24UB_4qdIWPSu1M:-Mb1aVPx8whAYwY00DBL:b87r0kx) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/bookkeep/shim_handle.c#L102)):*
>The same code would work in the "regular" case, i.e. when `fs.root.uri = "file:."`, because we would also convert `mount->uri` to be an absolute path. 

Sorry about the typo, can't fix on Reviewable.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[LibOS/shim/src/bookkeep/shim_handle.c, line 102 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0R24UB_4qdIWPSu1M:-Mb2RpN4EvDKq5ENGutU:b-ezokot) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/bookkeep/shim_handle.c#L102)):*

>The same code would work in the "regular" case, i.e. when `fs.root.uri = "file:."`, because we would also convert `mount->uri` to be an absolute path. 

Sorry about the typo, can't fix on Reviewable.
</blockquote></details>

Understood. Yes, this proposal looks good to me.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)* status: 24 of 25 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/src/bookkeep/shim_handle.c, line 102 at r1](https://reviewable.io/reviews/oscarlab/graphene/2402#-Mb0R24UB_4qdIWPSu1M:-Mb62E8Q5btdnyCCPH26:b-4lwxd9) ([raw file](https://github.com/oscarlab/graphene/blob/a40082fc58823ccdd1401e33663f425a52f36445/LibOS/shim/src/bookkeep/shim_handle.c#L102)):*

Understood. Yes, this proposal looks good to me.
</blockquote></details>

I added a comment to this code to explain the situation.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 20 of 25 files at r1, 4 of 5 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 473 at r3](https://reviewable.io/reviews/oscarlab/graphene/2402#-MbC-0ii5vIld_Ltognh:-MbC-0ii5vIld_Ltogni:b-48nv97) ([raw file](https://github.com/oscarlab/graphene/blob/8fad86c04f65027800d4dc9bf27f85a70747637d/LibOS/shim/src/fs/shim_dcache.c#L473)):*
> ```C
>         new_dent->state &= ~DENTRY_LISTED;
> 
>         if (new_dent->type != S_IFIFO) {
> ```

How does this `type` field correspond to the underlying FS implementation on our side? Seems like the information in these fields overlap and may be hard to always keep them in sync?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/src/fs/shim_dcache.c, line 473 at r3](https://reviewable.io/reviews/oscarlab/graphene/2402#-MbC-0ii5vIld_Ltognh:-MbL6xPF6_rSmkk6aic2:bgwd71s) ([raw file](https://github.com/oscarlab/graphene/blob/8fad86c04f65027800d4dc9bf27f85a70747637d/LibOS/shim/src/fs/shim_dcache.c#L473)):*

How does this `type` field correspond to the underlying FS implementation on our side? Seems like the information in these fields overlap and may be hard to always keep them in sync?
</blockquote></details>

Currently, yes, unfortunately individual filesystems (mostly `chroot`) also keep some type information, and do this messy synchronization.

My goal (in a few PRs...) is to remove these extra sources of information, and always write to `type`. This already happened for file permissions.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

dimakuv (approved):
> Reviewed 12 of 12 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> Reviewed 12 of 12 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2402)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2403, opened 2021-05-27
[LibOS] Change mount semantics

pwmarcz:
## Description of the changes 

This changes the mount operation to create and attach new dentry instead of replacing an existing one. See the commit message for details.

## How to test this PR? 

I believe the current tests are enough.

Here is an example of the new dentry tree. `M` is a mountpoint dentry, with the actual new filesystem root attached under it.

```
[P67026:T1:busybox] [       V..  3] 040700 drwx M /
[P67026:T1:busybox] [chroot V.. 26] 040700 drwx *   /
[P67026:T1:busybox] [chroot V.S  2] 100000 ---- M     proc
[P67026:T1:busybox] [  proc V..  1] 040555 dr-x *       proc/
[P67026:T1:busybox] [chroot V.S  2] 100000 ---- M     dev
[P67026:T1:busybox] [   dev V..  2] 040555 dr-x *       dev/
[P67026:T1:busybox] [   dev V..  2] 020666 crw- M         tty
[P67026:T1:busybox] [chroot V..  3] 020600 crw- *           tty
[P67026:T1:busybox] [chroot V.S  2] 100000 ---- M     sys
[P67026:T1:busybox] [   sys V..  1] 040555 dr-x *       sys/
[P67026:T1:busybox] [chroot V.S  2] 100000 ---- M     lib
[P67026:T1:busybox] [chroot V..  6] 040700 drwx *       lib/
[P67026:T1:busybox] [chroot V.S  2] 100000 ---- M         x86_64-linux-gnu
[P67026:T1:busybox] [chroot V..  1] 040700 drwx *           x86_64-linux-gnu/
[P67026:T1:busybox] [chroot V..  3] 100700 -rwx           ld-linux-x86-64.so.2
[P67026:T1:busybox] [chroot V..  2] 100700 -rwx           libm.so.6
[P67026:T1:busybox] [chroot V..  2] 100700 -rwx           libresolv.so.2
[P67026:T1:busybox] [chroot V..  2] 100700 -rwx           libc.so.6
[P67026:T1:busybox] [chroot V.S  2] 100000 ---- M     etc
[P67026:T1:busybox] [chroot V..  2] 040700 drwx *       etc/
[P67026:T1:busybox] [chroot V..  1] 000000 ?---  -        ld.so.preload
[P67026:T1:busybox] [chroot V..  2] 100700 -rwx       busybox
[P67026:T1:busybox] [chroot V..  1] 100600 -rw-       .gitignore
[P67026:T1:busybox] [chroot V..  1] 100600 -rw-       busybox.tar.bz2
[P67026:T1:busybox] [chroot V..  1] 100600 -rw-       README.md
[P67026:T1:busybox] [chroot V..  1] 040700 drwx       src/
[P67026:T1:busybox] [chroot V..  1] 100600 -rw-       busybox.manifest
[P67026:T1:busybox] [chroot V..  1] 100600 -rw-       busybox.manifest.template
[P67026:T1:busybox] [chroot V..  1] 100600 -rw-       Makefile
```

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -91,9 +91,7 @@ struct shim_fs_ops {
 
 #define DENTRY_VALID       0x0001 /* this dentry is verified to be valid */
 #define DENTRY_NEGATIVE    0x0002 /* recently deleted or inaccessible */
-#define DENTRY_RECENTLY    0x0004 /* recently used */
 #define DENTRY_PERSIST     0x0008 /* added as a persistent dentry */
-#define DENTRY_MOUNTPOINT  0x0040 /* this dentry is a mount point */
 #define DENTRY_ISLINK      0x0080 /* this dentry is a link */
 #define DENTRY_ISDIRECTORY 0x0100 /* this dentry is a directory */
 #define DENTRY_LOCKED      0x0200 /* locked by mountpoints at children */
@@ -132,12 +130,17 @@ struct shim_dentry {
      * different in case of special files (such as named pipes or sockets). */
     struct shim_fs* fs;
 
+    /* Parent of this dentry, but only within the same mount. If you need the dentry one level up,
+     * regardless of mounts (i.e. `..`), you should use `dentry_up()` instead. */
     struct shim_dentry* parent;
+
     size_t nchildren;
     LISTP_TYPE(shim_dentry) children; /* These children and siblings link */
     LIST_TYPE(shim_dentry) siblings;
 
-    struct shim_mount* mounted;
+    /* Filesystem mounted under this dentry. If set, this dentry is a mountpoint: filesystem
+     * operations should use `attached_mount->root` instead of this dentry. */
+    struct shim_mount* attached_mount;
 
     /* file type: S_IFREG, S_IFDIR, S_IFLNK etc. */
     mode_t type;
@@ -234,17 +237,11 @@ DEFINE_LIST(shim_mount);
 struct shim_mount {
     struct shim_fs* fs;
 
-    /*
-     * TODO: this field currently functions as both the mountpoint AND the mount root. Instead,
-     * `mount_point` should be the dentry from the parent filesystem, and `root` should be the mount
-     * root.
-     */
     struct shim_dentry* mount_point;
 
     struct shim_qstr path;
     struct shim_qstr uri;
 
-    /* TODO: this is unused right now (see `mount_point`) */
     struct shim_dentry* root;
 
     void* data;
@@ -285,9 +282,33 @@ int init_mount_root(void);
 int init_mount(void);
 
 /* file system operations */
-int mount_fs(const char* mount_type, const char* mount_uri, const char* mount_point,
-             struct shim_dentry* parent, struct shim_dentry** dentp, bool make_ancestor);
-int unmount_fs(const char* mount_point);
+
+/*!
+ * \brief Mount a new filesystem
+ *
+ * \param type Filesystem type (currently defined in `mountable_fs` in `shim_fs.c`)
+ * \param uri PAL URI to mount, or NULL if not applicable
+ * \param mount_path Path to the mountpoint
+ *
+ * Creates a new `shim_mount` structure (mounted filesystem) and attaches to the dentry under
+ * `mount_path`. That means (assuming the dentry is called `mount_point`):
+ *
+ * - `mount_point->attached_mount` is the new filesystem,
+ * - `mount_point->attached_mount->root` is the dentry of new filesystem's root.
+ *
+ * Subsequent lookups for `mount_path` and paths starting with `mount_path` will retrieve the new
+ * filesystem's root, not the mountpoint.
+ *
+ * As a result, multiple mount operations for the same path will create a chain (mount1 -> root1 ->
+ * mount2 -> root2 ...), effectively stacking the mounts and ensuring only the last one is visible.
+ *
+ * The function will ensure that the mountpoint exists: new dentries marked with DENTRY_SYNTHETIC
+ * will be created, if necessary. This is a departure from Unix mount, necessary to implement
+ * Graphene manifest semantics.
+ *
+ * TODO: On failure, this function does not clean the synthetic nodes it just created.
+ */
+int mount_fs(const char* type, const char* uri, const char* mount_path);
 
 void get_mount(struct shim_mount* mount);
 void put_mount(struct shim_mount* mount);
@@ -500,6 +521,19 @@ void get_dentry(struct shim_dentry* dent);
 void put_dentry(struct shim_dentry* dent);
 
 /*!
+ * \brief Get the dentry one level up
+ *
+ * \param dent the dentry
+ *
+ * \return the dentry one level up, or NULL if one does not exist
+ *
+ * Computes the dentry pointed to by ".." from the current one, unless the current dentry is at
+ * global root. Unlike the `parent` field, this traverses mounted filesystems (i.e. works also for a
+ * root dentry of a mount).
+ */
+struct shim_dentry* dentry_up(struct shim_dentry* dent);
+
+/*!
  * \brief Garbage-collect a dentry, if possible
  *
  * \param dentry the dentry (has to have a parent)
@@ -539,8 +573,8 @@ void dentry_gc(struct shim_dentry* dent);
  * This function computes an absolute path for dentry, allocating a new buffer for it. The path
  * should later be freed using `free`.
  *
- * An absolute path is a combination of all names up to the root (not including the root, which by
- * convention has an empty name), separated by `/`, and beginning with `/`.
+ * An absolute path is a combination of all names up to the global root (not including the root,
+ * which by convention has an empty name), separated by `/`, and beginning with `/`.
  */
 int dentry_abs_path(struct shim_dentry* dent, char** path, size_t* size);
 
@@ -556,8 +590,8 @@ int dentry_abs_path(struct shim_dentry* dent, char** path, size_t* size);
  * This function computes a relative path for dentry, allocating a new buffer for it. The path
  * should later be freed using `free`.
  *
- * A relative path is a combination of all names up to the mountpoint (not including the
- * mountpoint), separated by `/`. A relative path never begins with `/`.
+ * A relative path is a combination of all names up to the root of the dentry's filesystem (not
+ * including the root), separated by `/`. A relative path never begins with `/`.
  */
 int dentry_rel_path(struct shim_dentry* dent, char** path, size_t* size);
 
@@ -572,8 +606,8 @@ ino_t dentry_ino(struct shim_dentry* dent);
 /*!
  * \brief Allocate and initialize a new dentry
  *
- * \param parent the parent node, or NULL if this is supposed to be the dentry root
  * \param mount the mount the dentry is under
+ * \param parent the parent node, or NULL if this is supposed to be the mount root
  * \param name name of the new dentry
  * \param name_len length of the name
  *
@@ -608,14 +642,9 @@ struct shim_dentry* get_new_dentry(struct shim_mount* mount, struct shim_dentry*
  */
 struct shim_dentry* lookup_dcache(struct shim_dentry* parent, const char* name, size_t name_len);
 
-/* This function recursively deletes and frees all dentries under root
- *
- * XXX: Current code doesn't do a free..
- */
-int __del_dentry_tree(struct shim_dentry* root);
-
 /*
- * Returns true if `anc` is an ancestor of `dent`.
+ * Returns true if `anc` is an ancestor of `dent`. Both dentries need to be within the same mounted
+ * filesystem.
  */
 bool dentry_is_ancestor(struct shim_dentry* anc, struct shim_dentry* dent);
 
--- a/LibOS/shim/src/fs/shim_dcache.c
+++ b/LibOS/shim/src/fs/shim_dcache.c
@@ -114,8 +114,8 @@ static void free_dentry(struct shim_dentry* dent) {
     assert(LISTP_EMPTY(&dent->children));
     assert(LIST_EMPTY(dent, siblings));
 
-    if (dent->mounted) {
-        put_mount(dent->mounted);
+    if (dent->attached_mount) {
+        put_mount(dent->attached_mount);
     }
 
     /* XXX: We are leaking `data` field here. This field seems to have different meaning for
@@ -198,6 +198,13 @@ struct shim_dentry* get_new_dentry(struct shim_mount* mount, struct shim_dentry*
     return dent;
 }
 
+struct shim_dentry* dentry_up(struct shim_dentry* dent) {
+    while (!dent->parent && dent->mount) {
+        dent = dent->mount->mount_point;
+    }
+    return dent->parent;
+}
+
 struct shim_dentry* lookup_dcache(struct shim_dentry* parent, const char* name, size_t name_len) {
     assert(locked(&g_dcache_lock));
 
@@ -217,35 +224,9 @@ struct shim_dentry* lookup_dcache(struct shim_dentry* parent, const char* name, 
     return NULL;
 }
 
-/* This function recursively removes children and drops the reference count
- * under root (but not the root itself).
- *
- * For memory-constrained systems (arguably SGX enclaves), there is a
- * legitimate concern that this could overflow the stack, as a path can have
- * as many as 4096 characters, leading to as many as 2048 stack frames.  It
- * may be preferable to rewrite this using tail recursion or allocating a
- * structure on the heap to track progress.
- */
-int __del_dentry_tree(struct shim_dentry* root) {
-    assert(locked(&g_dcache_lock));
-
-    struct shim_dentry *cursor, *n;
-
-    LISTP_FOR_EACH_ENTRY_SAFE(cursor, n, &root->children, siblings) {
-        // Recur if this is a non-empty directory
-        if (!LISTP_EMPTY(&cursor->children))
-            __del_dentry_tree(cursor);
-
-        LISTP_DEL_INIT(cursor, &root->children, siblings);
-        cursor->parent = NULL;
-        root->nchildren--;
-        put_dentry(cursor);
-    }
-
-    return 0;
-}
-
 bool dentry_is_ancestor(struct shim_dentry* anc, struct shim_dentry* dent) {
+    assert(anc->mount == dent->mount);
+
     while (dent) {
         if (dent == anc) {
             return true;
@@ -266,9 +247,9 @@ static size_t dentry_path_size(struct shim_dentry* dent, bool relative) {
     /* initial size is 1 for null terminator */
     size_t size = 1;
 
-    while (dent->parent) {
-        /* If the path is relative, also finish when encountering a mountpoint */
-        if (relative && (dent->state & DENTRY_MOUNTPOINT))
+    while (true) {
+        struct shim_dentry* up = relative ? dent->parent : dentry_up(dent);
+        if (!up)
             break;
 
         /* Add '/' after name, except the first one */
@@ -279,7 +260,7 @@ static size_t dentry_path_size(struct shim_dentry* dent, bool relative) {
         /* Add name */
         size += dent->name.len;
 
-        dent = dent->parent;
+        dent = up;
     }
 
     /* Add beginning '/' if absolute path */
@@ -301,9 +282,9 @@ static char* dentry_path_into_buf(struct shim_dentry* dent, bool relative, char*
     buf[pos] = '\0';
 
     /* Add names, starting from the last one, until we encounter root */
-    while (dent->parent) {
-        /* If the path is relative, also finish when encountering a mountpoint */
-        if (relative && (dent->state & DENTRY_MOUNTPOINT))
+    while (true) {
+        struct shim_dentry* up = relative ? dent->parent : dentry_up(dent);
+        if (!up)
             break;
 
         /* Add '/' after name, except the first one */
@@ -321,7 +302,7 @@ static char* dentry_path_into_buf(struct shim_dentry* dent, bool relative, char*
         pos -= dent->name.len;
         memcpy(&buf[pos], qstrgetstr(&dent->name), dent->name.len);
 
-        dent = dent->parent;
+        dent = up;
     }
 
     /* Add beginning '/' if absolute path */
@@ -417,7 +398,13 @@ static void dump_dentry(struct shim_dentry* dent, unsigned int level) {
 
     dump_dentry_mode(&buf, dent->type, dent->perm);
 
-    DUMP_FLAG(DENTRY_MOUNTPOINT, "*", " ");
+    if (dent->attached_mount) {
+        buf_puts(&buf, "M");
+    } else if (!dent->parent) {
+        buf_puts(&buf, "*");
+    } else {
+        buf_puts(&buf, " ");
+    }
     DUMP_FLAG(DENTRY_NEGATIVE, "-", " ");
 
     for (unsigned int i = 0; i < level; i++)
@@ -428,9 +415,14 @@ static void dump_dentry(struct shim_dentry* dent, unsigned int level) {
     DUMP_FLAG(DENTRY_ISLINK, " -> ", "");
     buf_flush(&buf);
 
-    struct shim_dentry* child;
-    LISTP_FOR_EACH_ENTRY(child, &dent->children, siblings) {
-        dump_dentry(child, level + 1);
+    if (dent->attached_mount) {
+        struct shim_dentry* root = dent->attached_mount->root;
+        dump_dentry(root, level + 1);
+    } else {
+        struct shim_dentry* child;
+        LISTP_FOR_EACH_ENTRY(child, &dent->children, siblings) {
+            dump_dentry(child, level + 1);
+        }
     }
 }
 
@@ -486,8 +478,8 @@ BEGIN_CP_FUNC(dentry) {
         if (dent->parent)
             DO_CP_MEMBER(dentry, dent, new_dent, parent);
 
-        if (dent->mounted)
-            DO_CP_MEMBER(mount, dent, new_dent, mounted);
+        if (dent->attached_mount)
+            DO_CP_MEMBER(mount, dent, new_dent, attached_mount);
 
         unlock(&dent->lock);
         ADD_CP_FUNC_ENTRY(off);
@@ -509,7 +501,7 @@ BEGIN_RS_FUNC(dentry) {
     CP_REBASE(dent->mount);
     CP_REBASE(dent->fs);
     CP_REBASE(dent->parent);
-    CP_REBASE(dent->mounted);
+    CP_REBASE(dent->attached_mount);
 
     if (!create_lock(&dent->lock)) {
         return -ENOMEM;
@@ -528,8 +520,8 @@ BEGIN_RS_FUNC(dentry) {
         LISTP_ADD_TAIL(dent, &dent->parent->children, siblings);
     }
 
-    if (dent->mounted) {
-        get_mount(dent->mounted);
+    if (dent->attached_mount) {
+        get_mount(dent->attached_mount);
     }
 }
 END_RS_FUNC(dentry)
--- a/LibOS/shim/src/fs/shim_fs.c
+++ b/LibOS/shim/src/fs/shim_fs.c
@@ -65,9 +65,13 @@ static struct shim_mount* alloc_mount(void) {
     return get_mem_obj_from_mgr_enlarge(mount_mgr, size_align_up(MOUNT_MGR_ALLOC));
 }
 
+static void free_mount(struct shim_mount* mount) {
+    free_mem_obj_to_mgr(mount_mgr, mount);
+}
+
 static bool mount_migrated = false;
 
-static int __mount_root(struct shim_dentry** root) {
+static int __mount_root(void) {
     int ret = 0;
     char* fs_root_type = NULL;
     char* fs_root_uri  = NULL;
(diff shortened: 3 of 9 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2403#-MbepTOb3R8dPddTLhcv:-MbepTOb3R8dPddTLhcw:b-e9comn) (no related file):*
> [P67026:T1:busybox] [chroot V..  1] 000000 ?---  -        ld.so.preload

Unrelated, but what does this mean?

---
___
*[LibOS/shim/include/shim_fs.h, line 134 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-MbepfW1CWZZ8d1LrmzD:-MbepfW1CWZZ8d1LrmzE:b8tzxu6) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/include/shim_fs.h#L134)):*
> ```objc
> 
>     /* Parent of this dentry, within the same filesystem. Use `dentry_up` to get the dentry above
>      * (going across filesystems, if necessary). */
> ```

The comment is confusing. Do you mean that `Don't use this field directly, instead use dentry_up...`? Or do you mean that `if parent == NULL, then use dentry_up to go across filesystems (if possible)`?

UPDATE: It looks like both are used in the code. `dentry->parent` is used for relative paths, `dentry_up(dentry)` is used for absolute paths. Maybe add something like this in your comment?
___
*[LibOS/shim/include/shim_fs.h, line 143 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-MbetpOB2NfWguarFvFX:-MbetpOC8GqO16MKwXlX:b-f1fhax) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/include/shim_fs.h#L143)):*
> ```objc
>     /* Filesystem mounted under this dentry. If set, this dentry is a mountpoint: filesystem
>      * operations should use `mounted->fs->root` instead of this dentry. */
>     struct shim_mount* mounted;
> ```

I find this name unreadable. First, I think it must be a boolean (due to the `-ed` ending). Then, I always confuse `dent->mount` and `dent->mounted`. Basically, `mount` means "my mount point I'm under" and `mounted` means "I am dentry where new mount point is attached".

Could we rename `mounted` to something like `attached_mount`?
___
*[LibOS/shim/include/shim_fs.h, line 304 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-MbeqtGQ2w7x4mys3byR:-MbeqtGR1ihRZDG9PncS:b-7dilqa) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/include/shim_fs.h#L304)):*
> ```objc
>  *
>  * If `make_synthetic` is true, the function will also ensure that the mountpoint exists: new
>  * dentries marked with DENTRY_SYNTHETIC will be created, if necessary. This is is a departure from
> ```

`is is a` -> `is a`
___
*[LibOS/shim/include/shim_fs.h, line 310 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-MberHL26aPXmnagpUu9:-MberHL26aPXmnagpUuA:bi82qds) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/include/shim_fs.h#L310)):*
> ```objc
>  * `make_synthetic` is true).
>  */
> int mount_fs(const char* type, const char* uri, const char* mount_path, bool make_synthetic);
> ```

What is the case when the user wants `make_synthetic = false`? Why do we need such functionality (apart from the global root, but there this flag is not even checked)?
___
*[LibOS/shim/include/shim_fs.h, line 609 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mberq-6DDb0xOIX5QYi:-Mberq-6DDb0xOIX5QYj:b-xfkq9) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/include/shim_fs.h#L609)):*
> ```objc
>  *
>  * \param parent the parent node, or NULL if this is supposed to be the mount root
>  * \param mount the mount the dentry is under
> ```

Please re-order `parent` and `mount`
___
*[LibOS/shim/src/fs/shim_fs.c, line 132 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mbeuc9F2AzgJrrLn5E6:-Mbeuc9F2AzgJrrLn5E7:b-ejy2dq) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/src/fs/shim_fs.c#L132)):*
> ```C
> 
>     log_debug("Mounting terminal device /dev/tty under /dev\n");
>     if ((ret = mount_fs("chroot", URI_PREFIX_DEV "tty", "/dev/tty", /*make_synthetic=*/true)) < 0) {
> ```

BTW, we'll need to fix this weirdness: `/dev/tty` is of chroot file type, but mounted (separately!) under `/dev`.
___
*[LibOS/shim/src/fs/shim_namei.c, line 140 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mbf-MsG200fgx9hSNPh:-Mbf-MsG200fgx9hSNPi:bl0irhn) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/src/fs/shim_namei.c#L140)):*
> ```C
>     if (!up)
>         up = g_dentry_root;
>     ret = do_path_lookupat(up, qstrgetstr(&link_target), flags, found, link_depth);
> ```

Unrelated question but how does this work? Why is the link target searched under the parent dir?
___
*[LibOS/shim/src/fs/shim_namei.c, line 364 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mbf345m2V3kliyuEXif:-Mbf345m2V3kliyuEXig:bm5u75) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/src/fs/shim_namei.c#L364)):*
> ```C
>         .link_depth = link_depth,
>     };
>     dent = NULL;
> ```

What's the point in this assignment?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: 5 of 9 files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2403#-MbepTOb3R8dPddTLhcv:-MbfZ0dy3a6JqTipk3VP:b-wp61og) (no related file):*

> [P67026:T1:busybox] [chroot V..  1] 000000 ?---  -        ld.so.preload

Unrelated, but what does this mean?
</blockquote></details>

This is a negative dentry: the code tried to open `/etc/ld.so.preload` recently, and Graphene confirmed that it does not exist.

---
___
*[LibOS/shim/include/shim_fs.h, line 134 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-MbepfW1CWZZ8d1LrmzD:-MbfZnHvEtjWNFJa8RGh:b-a3u8vt) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/include/shim_fs.h#L134)):*

The comment is confusing. Do you mean that `Don't use this field directly, instead use dentry_up...`? Or do you mean that `if parent == NULL, then use dentry_up to go across filesystems (if possible)`?

UPDATE: It looks like both are used in the code. `dentry->parent` is used for relative paths, `dentry_up(dentry)` is used for absolute paths. Maybe add something like this in your comment?
</blockquote></details>

How about: "Parent of this dentry, but only within the same mount. If you need the dentry one level up, regardless of mounts (i.e. `..`), you should use `dentry_up()` instead."
___
*[LibOS/shim/include/shim_fs.h, line 143 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-MbetpOB2NfWguarFvFX:-MbfahZc1ugaLnR6UTOn:b-2bcrqf) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/include/shim_fs.h#L143)):*

I find this name unreadable. First, I think it must be a boolean (due to the `-ed` ending). Then, I always confuse `dent->mount` and `dent->mounted`. Basically, `mount` means "my mount point I'm under" and `mounted` means "I am dentry where new mount point is attached".

Could we rename `mounted` to something like `attached_mount`?
</blockquote></details>

Sounds good. Changed to `attached_mount`.
___
*[LibOS/shim/include/shim_fs.h, line 304 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-MbeqtGQ2w7x4mys3byR:-MbfavLK3u9DxAGGVi-8:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/include/shim_fs.h#L304)):*

`is is a` -> `is a`
</blockquote></details>

Fixed.
___
*[LibOS/shim/include/shim_fs.h, line 310 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-MberHL26aPXmnagpUu9:-MbfaxIq9JFIEohvewM7:b-2v9f6m) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/include/shim_fs.h#L310)):*

What is the case when the user wants `make_synthetic = false`? Why do we need such functionality (apart from the global root, but there this flag is not even checked)?
</blockquote></details>

I kept it around because `make_synthetic = false` gives the "standard" mount behaviour that we might want if we implement mount in Graphene.

But now that you point it out, yeah, I'd rather simplify now. Removed.
___
*[LibOS/shim/include/shim_fs.h, line 609 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mberq-6DDb0xOIX5QYi:-MbfbzYjALGngCWgskSO:by7biuu) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/include/shim_fs.h#L609)):*

Please re-order `parent` and `mount`
</blockquote></details>

Oh, right. Fixed.
___
*[LibOS/shim/src/fs/shim_namei.c, line 140 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mbf-MsG200fgx9hSNPh:-Mbfc9KHBNLT_WDNdVii:b-yi6bzm) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/src/fs/shim_namei.c#L140)):*

Unrelated question but how does this work? Why is the link target searched under the parent dir?
</blockquote></details>

If the target is relative, it's relative to the directory the link is in. For instance, here we want `/foo/bar/link` to work, so we need to search the target under `/foo/bar`:

```
/
  foo/
    bar/
      link -> ../target.txt
    target.txt
```
___
*[LibOS/shim/src/fs/shim_namei.c, line 364 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mbf345m2V3kliyuEXif:-MbfdWFNEG03sxKx11Ca:beil37y) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/src/fs/shim_namei.c#L364)):*

What's the point in this assignment?
</blockquote></details>

The error path... but yeah, it's not used there. Removed.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_fs.h, line 134 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-MbepfW1CWZZ8d1LrmzD:-MbfogZZ1fKbJM-4ozuK:bo1b3jj) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/include/shim_fs.h#L134)):*

How about: "Parent of this dentry, but only within the same mount. If you need the dentry one level up, regardless of mounts (i.e. `..`), you should use `dentry_up()` instead."
</blockquote></details>

Yes, this new text is better.
___
*[LibOS/shim/src/fs/shim_namei.c, line 140 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mbf-MsG200fgx9hSNPh:-MbfpW128M3P_wEEjttt:b-nzb79) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/src/fs/shim_namei.c#L140)):*

If the target is relative, it's relative to the directory the link is in. For instance, here we want `/foo/bar/link` to work, so we need to search the target under `/foo/bar`:

```
/
  foo/
    bar/
      link -> ../target.txt
    target.txt
```

</blockquote></details>

OK, now I see that this is only for relative link targets. And absolute link targets submitted to `do_path_lookupat()` force that function to ignore the `up` argument.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/include/shim_fs.h, line 134 at r1](https://reviewable.io/reviews/oscarlab/graphene/2403#-MbepfW1CWZZ8d1LrmzD:-MbftXut8aHFutS1xl8w:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/37d75123bbfb0a36fd10cac6f1b4a02c5a1c0ed3/LibOS/shim/include/shim_fs.h#L134)):*

Yes, this new text is better.
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 5 of 9 files at r1, 3 of 4 files at r2, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc1aqzJ0xJ3QoPzyaez:-Mc1aqzK-u-CxG8Qhkzb:b-ckbrxy) (no related file):*
Not sure if this is a known issue, but I l played a bit with a shell in busybox on this PR and one FS still seems to be broken:
```
/ # ls /sys
ls: sys/devices: Permission denied
```

---
___
*[LibOS/shim/include/shim_fs.h, line 293 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc0P-yP3AetZx7AaE6m:-Mc0P-yQ-pPhxHDwy5OS:b-g0mtw9) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/include/shim_fs.h#L293)):*
> ```
> converts the dentry pointed to by `mount_path` to a mountpoint
> ```

So, this means that the mounts are not stackable, like in Linux?
___
*[LibOS/shim/include/shim_fs.h, line 306 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc0O-Na5e9xEBmeSptk:-Mc0O-Nb5t60jh_29B84:b-xnujiz) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/include/shim_fs.h#L306)):*
> ```objc
>  * Graphene manifest semantics.
>  *
>  * TODO: On failure, this function does not clean the synthetic nodes it just created.
> ```

How hard is it to fix this? It's just a leftover you forgot to fix before submitting the PR, or something which needs more changes than just a few deallocations?
___
*[LibOS/shim/src/fs/shim_fs.c, line 408 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc1dNk_6hwGzkN7moPA:-Mc1dNka0rTWGqvvj3k_:b-9g8kpl) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/src/fs/shim_fs.c#L408)):*
> ```C
>     mount->mount_point = mount_point;
>     get_dentry(mount_point);
>     mount_point->attached_mount = mount;
> ```

How do you actually free this circular reference? (`mount` <-> `mount_point`) Which of these pointers is non-owning/weak-ref?
___
*[LibOS/shim/src/fs/shim_namei.c, line 261 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc6YlCS7w5CMxabC7lK:-Mc6YFRt7dozVRKd_hmX:bxis8ga) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/src/fs/shim_namei.c#L261)):*
> ```C
>         } else if (!is_final || !(lookup->flags & LOOKUP_CREATE)) {
>             return -ENOENT;
>         }
> ```

What if we don't get caught in any of the if-else branches? It has to be negative, without `LOOKUP_MAKE_SYNTHETIC`, with `is_final` and with `LOOKUP_CREATE`. So, shouldn't we create it?

// This question may not make any sense, I admit I got lost in all the logic and IFs here ;)
___
*[LibOS/shim/src/fs/shim_namei.c, line 670 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc6cpM736GJ2NYWhQSQ:-Mc6cpM736GJ2NYWhQSR:byn8sge) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/src/fs/shim_namei.c#L670)):*
> ```C
>             if (ret != -EACCES) {
>                 /* Fail on underlying lookup errors, except -EACCES (for which we will just ignore the
>                  * file). The lookup might fail with -EACCES for host symlinks pointing to inaccessible
> ```

please rewrap

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: 8 of 9 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc1aqzJ0xJ3QoPzyaez:-Mc8oX-J8FhiYEbcX-Le:ban28gt) (no related file):*

Not sure if this is a known issue, but I l played a bit with a shell in busybox on this PR and one FS still seems to be broken:
```
/ # ls /sys
ls: sys/devices: Permission denied
```
</blockquote></details>

Seems broken on master as well. It looks like the regression tests checks only specific paths (such as `/sys/devices/system/cpu`), and the intermediate ones do not work :(

I'm working on a pseudo-FS rewrite now. I'll make sure the tests are stronger.

---
___
*[LibOS/shim/include/shim_fs.h, line 293 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc0P-yP3AetZx7AaE6m:-Mc8pfHx3CT2Hg-LeW8d:b-4b3l0c) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/include/shim_fs.h#L293)):*

> ```
> converts the dentry pointed to by `mount_path` to a mountpoint
> ```

So, this means that the mounts are not stackable, like in Linux?
</blockquote></details>

They are. You can mount several filesystems under the same path:

```
mount_fs("chroot" , "file:.");                   // mount1 
mount_fs("chroot" , "file:/foo", "/mnt/x");      // mount2
mount_fs("chroot" , "file:/bar", "/mnt/x");      // mount3
```

This will attach...
* `mount1` at `g_dentry_root`,
* `mount2` at the `mount1` dentry for `mnt/x`,
* `mount3` at the root dentry of `mount2`.

This is not how Linux implements it (AFAICT Linux stores mount information separately, and looks it up in a hash table), but the BSD family does more or less that (a vnode has a "mounted here" pointer to a mount).
___
*[LibOS/shim/include/shim_fs.h, line 306 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc0O-Na5e9xEBmeSptk:-Mc8ws-5AO28oUCnmuR5:b-snzoln) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/include/shim_fs.h#L306)):*

How hard is it to fix this? It's just a leftover you forgot to fix before submitting the PR, or something which needs more changes than just a few deallocations?
</blockquote></details>

The problem I see is we don't track which synthetic nodes were created by us, and which were created before.

I thought about recursively removing all synthetic nodes that seem unused (i.e. are empty directories), but decided against doing it now, mostly because it's not very useful: right now we mount only on Graphene startup (so errors will be fatal), and if we implement Unix `mount()`, it will not create synthetic nodes anyway.
___
*[LibOS/shim/src/fs/shim_fs.c, line 408 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc1dNk_6hwGzkN7moPA:-Mc8yc7Q62xoxeVytevO:bptrrhg) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/src/fs/shim_fs.c#L408)):*

How do you actually free this circular reference? (`mount` <-> `mount_point`) Which of these pointers is non-owning/weak-ref?
</blockquote></details>

If we wanted to have some sort of cleanup, then it would probably treat the mount as a child of mountpoint (i.e. dentries and mounts form a tree). So `mount.mount_point` would be the weak reference.
___
*[LibOS/shim/src/fs/shim_namei.c, line 261 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc6YlCS7w5CMxabC7lK:-Mc93Y1U05GM1_lDYNGJ:b-gm1bvw) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/src/fs/shim_namei.c#L261)):*

What if we don't get caught in any of the if-else branches? It has to be negative, without `LOOKUP_MAKE_SYNTHETIC`, with `is_final` and with `LOOKUP_CREATE`. So, shouldn't we create it?

// This question may not make any sense, I admit I got lost in all the logic and IFs here ;)
</blockquote></details>

`LOOKUP_CREATE` actually means "I want to create a file", so we proceed with a negative dentry.

I rewrote this condition slightly, I think it's much easier to read when flipped. And I added some explanations to these checks.
___
*[LibOS/shim/src/fs/shim_namei.c, line 670 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc6cpM736GJ2NYWhQSQ:-Mc95_c2BV9FigIBxlSl:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/src/fs/shim_namei.c#L670)):*

please rewrap
</blockquote></details>

Done.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: all files reviewed, 6 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc1aqzJ0xJ3QoPzyaez:-Mc9VMi02ddqhtHq9EBX:bkqjmh0) (no related file):*

Seems broken on master as well. It looks like the regression tests checks only specific paths (such as `/sys/devices/system/cpu`), and the intermediate ones do not work :(

I'm working on a pseudo-FS rewrite now. I'll make sure the tests are stronger.
</blockquote></details>

Ok, resolving for now then.

---
___
*[LibOS/shim/include/shim_fs.h, line 293 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc0P-yP3AetZx7AaE6m:-Mc9VcbA1iovvMQboUsL:b-ykkj70) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/include/shim_fs.h#L293)):*

They are. You can mount several filesystems under the same path:

```
mount_fs("chroot" , "file:.");                   // mount1 
mount_fs("chroot" , "file:/foo", "/mnt/x");      // mount2
mount_fs("chroot" , "file:/bar", "/mnt/x");      // mount3
```

This will attach...
* `mount1` at `g_dentry_root`,
* `mount2` at the `mount1` dentry for `mnt/x`,
* `mount3` at the root dentry of `mount2`.

This is not how Linux implements it (AFAICT Linux stores mount information separately, and looks it up in a hash table), but the BSD family does more or less that (a vnode has a "mounted here" pointer to a mount).
</blockquote></details>

Hmm, but the comment says that "this function converts the dentry pointed to by `mount_path`", how does this work when there's already a mountpoint there? I think I'm just confused by this comment, it reads like the directory was just overridden by the mounpoint :)
___
*[LibOS/shim/include/shim_fs.h, line 306 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc0O-Na5e9xEBmeSptk:-Mc9W-XM8NQVV_FdZsCH:b5w0bmo) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/include/shim_fs.h#L306)):*

The problem I see is we don't track which synthetic nodes were created by us, and which were created before.

I thought about recursively removing all synthetic nodes that seem unused (i.e. are empty directories), but decided against doing it now, mostly because it's not very useful: right now we mount only on Graphene startup (so errors will be fatal), and if we implement Unix `mount()`, it will not create synthetic nodes anyway.
</blockquote></details>

Ah, right, then it's not a big deal.
___
*[LibOS/shim/src/fs/shim_namei.c, line 261 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc6YlCS7w5CMxabC7lK:-Mc9WJBJB4gltE9ljBuT:be33irv) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/src/fs/shim_namei.c#L261)):*

`LOOKUP_CREATE` actually means "I want to create a file", so we proceed with a negative dentry.

I rewrote this condition slightly, I think it's much easier to read when flipped. And I added some explanations to these checks.
</blockquote></details>

Ah, I see. It's much more readable now, thanks!

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/include/shim_fs.h, line 293 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc0P-yP3AetZx7AaE6m:-Mc9p_yc3c0IpmG2q32_:b-rfy9fz) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/include/shim_fs.h#L293)):*

Hmm, but the comment says that "this function converts the dentry pointed to by `mount_path`", how does this work when there's already a mountpoint there? I think I'm just confused by this comment, it reads like the directory was just overridden by the mounpoint :)
</blockquote></details>

Right, it's better to talk about attaching, not converting. I reworded the comment, and explained stacked mounts - if you have a better idea how to make it more clear, please suggest wording.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/include/shim_fs.h, line 293 at r4](https://reviewable.io/reviews/oscarlab/graphene/2403#-Mc0P-yP3AetZx7AaE6m:-Mc9sEx636yD_Ju-kQhm:b-90rp) ([raw file](https://github.com/oscarlab/graphene/blob/98aa1255146c0d6d74da03d94a5fcc92c4fd2c52/LibOS/shim/include/shim_fs.h#L293)):*

Right, it's better to talk about attaching, not converting. I reworded the comment, and explained stacked mounts - if you have a better idea how to make it more clear, please suggest wording.
</blockquote></details>

Yeah, much better now! :)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2403)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2405, opened 2021-05-27
[LibOS] Avoid the comparison operations on the pid of netlink messages in glibc2.31

bigdata-memory:
## Description of the changes
For the details check commit message and comments in code.
In addition, I will add patch for 2.33 later since I did patch for 2.27 but it got removed in the latest revision. thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2405)

### DIFF
--- a/LibOS/glibc-patches/glibc-2.31.patch
+++ b/LibOS/glibc-patches/glibc-2.31.patch
@@ -70,6 +70,51 @@ index 07334c63d81cb3b0dca848aa28c6a80292059bfb..2093438c050443920da0fcafe3083d72
  
    if (highest < count) {
      printf "*** errlist.c count %d vs Versions sys_errlist@%s count %d\n", \
+diff --git a/sysdeps/unix/sysv/linux/check_pf.c b/sysdeps/unix/sysv/linux/check_pf.c
+index 5a00cddeb122f86e53f38476e391c7341183ef8d..7860aeb464a2611a3244b3b04ac49706ab070e77 100644
+--- a/sysdeps/unix/sysv/linux/check_pf.c
++++ b/sysdeps/unix/sysv/linux/check_pf.c
+@@ -180,7 +180,7 @@ make_request (int fd, pid_t pid)
+ 	   NLMSG_OK (nlmh, (size_t) read_len);
+ 	   nlmh = (struct nlmsghdr *) NLMSG_NEXT (nlmh, read_len))
+ 	{
+-	  if (nladdr.nl_pid != 0 || (pid_t) nlmh->nlmsg_pid != pid
++	  if (nladdr.nl_pid != 0
+ 	      || nlmh->nlmsg_seq != req.nlh.nlmsg_seq)
+ 	    continue;
+ 
+diff --git a/sysdeps/unix/sysv/linux/ifaddrs.c b/sysdeps/unix/sysv/linux/ifaddrs.c
+index ccacbfdb1694efb6ce83f8aa7b2300dca6dc4f6d..ad14eefc2ef62d5ec4844d3f7b4e09b7fde0965d 100644
+--- a/sysdeps/unix/sysv/linux/ifaddrs.c
++++ b/sysdeps/unix/sysv/linux/ifaddrs.c
+@@ -187,8 +187,7 @@ __netlink_request (struct netlink_handle *h, int type)
+ 	   NLMSG_OK (nlmh, remaining_len);
+ 	   nlmh = (struct nlmsghdr *) NLMSG_NEXT (nlmh, remaining_len))
+ 	{
+-	  if ((pid_t) nlmh->nlmsg_pid != h->pid
+-	      || nlmh->nlmsg_seq != h->seq)
++	  if (nlmh->nlmsg_seq != h->seq)
+ 	    continue;
+ 
+ 	  ++count;
+@@ -367,7 +366,7 @@ getifaddrs_internal (struct ifaddrs **ifap)
+       for (nlh = nlp->nlh; NLMSG_OK (nlh, size); nlh = NLMSG_NEXT (nlh, size))
+ 	{
+ 	  /* Check if the message is what we want.  */
+-	  if ((pid_t) nlh->nlmsg_pid != nh.pid || nlh->nlmsg_seq != nlp->seq)
++	  if (nlh->nlmsg_seq != nlp->seq)
+ 	    continue;
+ 
+ 	  /* If the dump got interrupted, we can't rely on the results
+@@ -450,7 +449,7 @@ getifaddrs_internal (struct ifaddrs **ifap)
+ 	  int ifa_index = 0;
+ 
+ 	  /* Check if the message is the one we want */
+-	  if ((pid_t) nlh->nlmsg_pid != nh.pid || nlh->nlmsg_seq != nlp->seq)
++	  if (nlh->nlmsg_seq != nlp->seq)
+ 	    continue;
+ 
+ 	  if (nlh->nlmsg_type == NLMSG_DONE)
 diff --git a/sysdeps/unix/sysv/linux/x86_64/____longjmp_chk.S b/sysdeps/unix/sysv/linux/x86_64/____longjmp_chk.S
 index cbce00832cfbb3ab56c27b4887170dbc54480671..ac5d209ccfabd89db49eb1ea1283ec316c85a68d 100644
 --- a/sysdeps/unix/sysv/linux/x86_64/____longjmp_chk.S


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2405)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @bigdata-memory)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2405#-Mb0HT9A4iXeOGQVkyjE:-Mb0HT9A4iXeOGQVkyjF:bog21qf) (no related file):*
Could you explain what "netlink message carried pid" mean exactly? At which point and by whom this "message carried PID" is added?

It looks like it can be indeed possible that this PID is added by the host Linux in every Netlink message, and then the issue is real and must be solved somehow (maybe using your PR, though I don't like patching Glibc). But I'd like to learn more about this problem. Any links / references are welcome.

---

bigdata-memory:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2405#-Mb0HT9A4iXeOGQVkyjE:-Mb8N61mEcNCppYe8U45:b227rkk) (no related file):*

Could you explain what "netlink message carried pid" mean exactly? At which point and by whom this "message carried PID" is added?

It looks like it can be indeed possible that this PID is added by the host Linux in every Netlink message, and then the issue is real and must be solved somehow (maybe using your PR, though I don't like patching Glibc). But I'd like to learn more about this problem. Any links / references are welcome.
</blockquote></details>

Please checkout the following link for your reference. Thanks.
"For a user-space process,  nl_pid is usually the PID of the process owning the destination socket"
https://www.man7.org/linux/man-pages/man7/netlink.7.html

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2405)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @bigdata-memory)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2405#-Mb0HT9A4iXeOGQVkyjE:-MbAXc7_3rXd6ifATQ2E:b-8ze52b) (no related file):*

Please checkout the following link for your reference. Thanks.
"For a user-space process,  nl_pid is usually the PID of the process owning the destination socket"
https://www.man7.org/linux/man-pages/man7/netlink.7.html
</blockquote></details>

I read about netlink sockets now. I don't think your PR is correct.

First, Graphene doesn't have support for netlink sockets currently. So this PR is currently unusable anyway.

I assume that you implemented support for netlink sockets in your own fork of Graphene. It would be interesting to check it out.

Second, `sockaddr_nl.nl_pid` is NOT set up by the Linux kernel -- it is set up by the application, and this object is standardized. So what should actually happen is this:
- The application on top of Graphene should continue using Graphene-specific (virtual) PIDs.
- Graphene must intercept netlink-socket operations and emulate them.
- As part of this emulation, Graphene must rewire `nl_pid` from Graphene-specific (virtual) PIDs to Linux-host (actual) PIDs.

This way there is no need to modify Glibc sources, and this PR is not needed.

---

bigdata-memory:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2405#-Mb0HT9A4iXeOGQVkyjE:-Mbbgdr_3R1gfTnYHlKb:bdgjjai) (no related file):*

I read about netlink sockets now. I don't think your PR is correct.

First, Graphene doesn't have support for netlink sockets currently. So this PR is currently unusable anyway.

I assume that you implemented support for netlink sockets in your own fork of Graphene. It would be interesting to check it out.

Second, `sockaddr_nl.nl_pid` is NOT set up by the Linux kernel -- it is set up by the application, and this object is standardized. So what should actually happen is this:
- The application on top of Graphene should continue using Graphene-specific (virtual) PIDs.
- Graphene must intercept netlink-socket operations and emulate them.
- As part of this emulation, Graphene must rewire `nl_pid` from Graphene-specific (virtual) PIDs to Linux-host (actual) PIDs.

This way there is no need to modify Glibc sources, and this PR is not needed.
</blockquote></details>

Essentially, it is the fake PID that makes this patch necessary for the Netlink protocol enabling  in our implementation, as you suggested, do we have any GSGX internal mechanism that rewires the fake PID to real PID ? Thanks.

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2405)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @bigdata-memory)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2405#-Mb0HT9A4iXeOGQVkyjE:-MbeIGHb1kpaSL4F6GTB:bu3gf1w) (no related file):*

Essentially, it is the fake PID that makes this patch necessary for the Netlink protocol enabling  in our implementation, as you suggested, do we have any GSGX internal mechanism that rewires the fake PID to real PID ? Thanks.
</blockquote></details>

You must go through the LibOS and PAL layers to rewire the Graphene-internal (fake) PID to a Linux-host (real) PID.

For example, in case of pipes, there is something like this:
- https://github.com/oscarlab/graphene/blob/master/LibOS/shim/src/sys/shim_pipe.c#L134 -- this is where the `shim_handle` object for the pipe gets its fake PID.
- https://github.com/oscarlab/graphene/blob/master/LibOS/shim/src/fs/pipe/fs.c#L32 -- this is where the `shim_handle` object for the pipe is requested to perform a Linux-host read (still using the fake PID). At this point, LibOS calls into PAL's `DkStreamRead(hdl->pal_handle)` (via the PAL handle object).
- https://github.com/oscarlab/graphene/blob/master/Pal/src/host/Linux/db_pipes.c#L290 -- this is where `DkStreamRead(hdl->pal_handle)` ends up. As you can see, the PAL handle object contains a field `handle->pipe.fd` -- this is the Linux-host (real) PID. This is used to perform an actual host-level system call `read()`.

So you can always get to a real PID from the LibOS object using something like this: `hdl->pal_handle->pipe.fd`. And you can get the LibOS object `hdl` using something like this: `get_fd_handle(fake-pid)`.

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2405)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @bigdata-memory)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2405#-Mb0HT9A4iXeOGQVkyjE:-MbjBuRPDlxxa6aF1Omz:b2ns36o) (no related file):*

I got a little bit confused, the `fd` stands for `file descriptor`, how it could be the value of `pid` ?
Did you mean the `handle->pipe.fd` is a `fake fd` or it is identical to`the real pid` ? Thanks.
</blockquote></details>

This was an example for pipes and FDs, to illustrate the concept. But you're right, I should have just given a PID example:

1. At this LibOS level, there is a global `g_process.pid` -- this is the fake PID.
2. At the PAL level, there is a global `g_pal_control.process_id` -- this is the real PID. You can also get it using the PAL-internal function `_DkGetProcessId()`.

---

bigdata-memory (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2405)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2405#-Mb0HT9A4iXeOGQVkyjE:-Mbl5ormDoANP_Xfm7dx:b-13z701) (no related file):*

This was an example for pipes and FDs, to illustrate the concept. But you're right, I should have just given a PID example:

1. At this LibOS level, there is a global `g_process.pid` -- this is the fake PID.
2. At the PAL level, there is a global `g_pal_control.process_id` -- this is the real PID. You can also get it using the PAL-internal function `_DkGetProcessId()`.
</blockquote></details>

The assembly of `netlink protocol packets` is happened in the code of `glibc`. it could become quite ugly to intercept those `packets` in shim layer and decode/encode it as middle man just for replacing the `fake pid` with `real pid`, I saw that the `glibc` already got patched in multiple places for various purposes e.g. `syscall`,  why we can not patch it for netlink purpose ? it works and straightforward to implement, thanks.

---

bigdata-memory:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2405#-Mb0HT9A4iXeOGQVkyjE:-McAWcID6dAV_gCYGfvj:b-f4lf7b) (no related file):*

The assembly of `netlink protocol packets` is happened in the code of `glibc`. it could become quite ugly to intercept those `packets` in shim layer and decode/encode it as middle man just for replacing the `fake pid` with `real pid`, I saw that the `glibc` already got patched in multiple places for various purposes e.g. `syscall`,  why we can not patch it for netlink purpose ? it works and straightforward to implement, thanks.
</blockquote></details>

In addition, to keep the integrity of code logics of glibc, why not provision the `real PID` to glibc in first place ? Thanks.

---

bigdata-memory (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2405)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2405#-Mb0HT9A4iXeOGQVkyjE:-MdY04lKAjbFdYqHmKJD:bopdpnw) (no related file):*

In addition, to keep the integrity of code logics of glibc, why not provision the `real PID` to glibc in first place ? Thanks.
</blockquote></details>

@dimakuv any further suggestion on it ? thanks.

---

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2405)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @bigdata-memory)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2405#-Mb0HT9A4iXeOGQVkyjE:-MdprJDa45FnMVDtb6ls:bxmlafo) (no related file):*

@dimakuv any further suggestion on it ? thanks.
</blockquote></details>

Graphene is a Library OS with a focus on security (protecting the app running inside Graphene from the malicious host OS). To achieve this, Graphene re-implements almost all relevant subsystems of the host OS and adds corresponding metadata information (metadata on files, sockets, pipes, threads, processes, memory management, etc). Graphene tries very hard to verify whatever the host OS does, and fails when it finds inconsistencies.

For this reason, Graphene implements what you call `fake pids`. Graphene wants to virtualize PIDs, that's why Graphene doesn't reuse `real pids` from the (possibly malicious) host OS.

This patch worsens security guarantees of Graphene by allowing the application and Glibc to operate on PIDs fully controlled by the possibly malicious host OS. As such, this patch cannot be accepted into Graphene.

---

bigdata-memory:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2405#-Mb0HT9A4iXeOGQVkyjE:-MduOIzW0wqShPuu6oo9:bf4kdjq) (no related file):*

Graphene is a Library OS with a focus on security (protecting the app running inside Graphene from the malicious host OS). To achieve this, Graphene re-implements almost all relevant subsystems of the host OS and adds corresponding metadata information (metadata on files, sockets, pipes, threads, processes, memory management, etc). Graphene tries very hard to verify whatever the host OS does, and fails when it finds inconsistencies.

For this reason, Graphene implements what you call `fake pids`. Graphene wants to virtualize PIDs, that's why Graphene doesn't reuse `real pids` from the (possibly malicious) host OS.

This patch worsens security guarantees of Graphene by allowing the application and Glibc to operate on PIDs fully controlled by the possibly malicious host OS. As such, this patch cannot be accepted into Graphene.
</blockquote></details>

My recent proposal is to provision the `real PID` to custom Glibc which is part of Graphene LibOS as it compiled, measured and loaded with Graphene LibOS & trusted PAL, so if the LibOS/trusted PAL can get the `real PID', why Glibc cannot ? by the way, the pid usage is agnostic to application which actually doesn't need to know `PID`.

---

PR #2411, opened 2021-05-31
Meson: Pal/Linux-SGX

woju:
## Description of the changes 

Mesonification of Linux-SGX.

SGX' `aesm.proto` was moved to a common directory for two reasons: it will be merged with python's `aesm.proto` and `protoc-c` is capricious about paths.

## How to test this PR? 

Check compilation. Check different SGX drivers.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2411)


woju:
wrong branch name, missing uid prefix

PR #2422, opened 2021-06-04
[LibOS] Add rt_sigtimedwait syscall

dimakuv:
## Description of the changes 

This commit also fixes a data race when a child process terminates during `kill()` in the parent process (in this case, kill must become a no-op and succeed) -- this race was found in `sigtimedwait01` LTP test.

## How to test this PR? 

Several LTP tests pertaining to signal handling are enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[LibOS/shim/include/shim_signal.h, line 147 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMD_rd5lTXMNN1ie3R:-MbMD_rd5lTXMNN1ie3S:b-o4913x) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/include/shim_signal.h#L147)):*
> ```objc
>  * \param[out] signal  Pointer to signal, filled with signal info.
>  *
>  * Checks whether there is a pending non-blocked signal among forced signals, normal per-thread and
> ```

I don't think this should handle forced signals. They are by definition forced and should be always delivered to the app, skipping any signal masks and whatnot.
___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 128 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMFGLHD5b8i_8PgBil:-MbMFGLI1e0HwczBcaE8:b-87phlf) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/bookkeep/shim_signal.c#L128)):*
> ```C
>  * support injecting only 1 instance of 1 signal only once, as this feature is meant only for
>  * graceful termination of the user application. Note that the only host-injected signal currently
>  * supported is SIGTERM; see also pop_allowed_signal().
> ```

Please backtick-ify the symbol name (function).
___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 604 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMFpmzCLKMxOCefB7x:-MbMFpmzCLKMxOCefB7y:bn58rwt) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/bookkeep/shim_signal.c#L604)):*
> ```C
>     assert(current);
> 
>     lock(&current->lock);
> ```

Why did you move locking?
Update: I see, you look at current signal mask in else-if clause. I don't like this version: the old one was lockless in the common case of no signals.
___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 608 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMJZr40b8mAtMHRcqG:-MbMJZr40b8mAtMHRcqH:b-qc8bax) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/bookkeep/shim_signal.c#L608)):*
> ```C
>     if (have_forced_signal()) {
>         /* forced signals are not artifically generated by kill()/raise()/etc. (i.e., they are
>          * generated due to hardware exception) and thus cannot be blocked, so no mask check */
> ```

I wouldn't add such comment: it's currently the case, but in principle they could be anything (basically just a way to skip signal mask checks). E.g. at some point we might need to have Graphene (kernel) forced signals.
Please either remove or elaborate.
___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 610 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMHtlT9w6rCB_l4wOI:-MbMHtlT9w6rCB_l4wOJ:bka51jb) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/bookkeep/shim_signal.c#L610)):*
> ```C
>          * generated due to hardware exception) and thus cannot be blocked, so no mask check */
>         get_forced_signal(signal);
>     } else if (__atomic_load_n(&current->pending_signals, __ATOMIC_ACQUIRE) ||
> ```

Do we have any official rules on boolean operator placement? I can't find anything in our docs, google-style allows for both version (and I always used to place them at the beginning).
___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 611 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMGCFH0gocL__qzvvh:-MbMGCFH0gocL__qzvvi:b-v4cx1i) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/bookkeep/shim_signal.c#L611)):*
> ```C
>         get_forced_signal(signal);
>     } else if (__atomic_load_n(&current->pending_signals, __ATOMIC_ACQUIRE) ||
>                      __atomic_load_n(&g_process_pending_signals_cnt, __ATOMIC_ACQUIRE)) {
> ```

Please align.
___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 72 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMN_PDCb7d5y-JeCN0:-MbMN_PE6MpwixOEJ7xV:bl2v3sn) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/ipc/shim_ipc_pid.c#L72)):*
> ```C
>              * thread (thus marking it as a zombie) and sending a KILL in this thread, so we have to
>              * assume here that the destination child process indeed terminated during this KILL. */
>             bool dest_process_is_child = false;
> ```

Please create another function `is_child_process_or_zombie` (or something) inside `bookkeep/shim_process.c`
___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 94 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMNpIM6pQMlmmxjPQp:-MbMNpINDon3Wy4MPNl-:bmpg638) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/ipc/shim_ipc_pid.c#L94)):*
> ```C
>             if (dest_process_is_child) {
>                 log_debug("IPC send to terminated child process %u is dropped\n", dest_pid);
>                 return 0;
> ```

I don't like this: the malicious host can now skip message (signal) delivery and nothing will notice that. That's the case in the old code too, but if we change `send_ipc_message` above to also wait for confirmation, it wouldn't be then.
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 260 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMTHGG4lo8eudqu-vW:-MbMTHGHAUufPzd_Byio:b-j27phn) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L260)):*
> ```C
>     clear_illegal_signals(&unblocked_mask);
> 
>     /* Note that the user of rt_sigtimedwait() is supposed to block the signals in `allowed` set via
> ```

Please backtickify symbol names.
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 269 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMTX_K04Xq-UsKyQ2X:-MbMTX_K04Xq-UsKyQ2Y:bvn9m0s) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L269)):*
> ```C
>     lock(&current->lock);
>     get_sig_mask(current, &old);
>     __sigandset(&mask, &unblocked_mask, &old);
> ```

Isn't this equivalent to `signotset(&mask, &old, allowed_ptr)`? Why all the shenanigans with full set?
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 275 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMX9IY2cnT1QSsE-aU:-MbMX9IZ2xojqJ6L-297:by8gcfa) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L275)):*
> ```C
>     int thread_wait_res = -EINTR;
>     thread_prepare_wait();
>     if (!have_pending_signals()) {
> ```

This should be `while`, shouldn't it?
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 276 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMQjdA8ZBlZBgLqv5p:-MbMQjdBAfUlXbJFwEa9:b-q03xml) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L276)):*
> ```C
>     thread_prepare_wait();
>     if (!have_pending_signals()) {
>         /* FIXME: timeout is not updated for consequtive thread_wait invocations */
> ```

Wait, `thread_wait` does update the timeout, you just don't use it...
Look at functions in `shim_sleep.c` on how to do it.
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 291 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMYwNxBawotjZb8Ie3:-MbMYwNy1C4foOYDqjz9:b-td1k93) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L291)):*
> ```C
>         ret = signal.siginfo.si_signo;
>     } else {
>         ret = (thread_wait_res == -ETIMEDOUT ? -EAGAIN : -ERESTARTNOHAND);
> ```

After some thinking: this should be `-EINTR`. Seems Linux allows for spurious EINTRs in this case (if we race with another thread to handle this signal).

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: 4 of 11 files reviewed, 13 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/include/shim_signal.h, line 51 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbaT1TT-h60A7nExVZU:-MbaT1TT-h60A7nExVZV:b4gkcfh) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/include/shim_signal.h#L51)):*
> ```objc
>         int __cnt = _SIGSET_NWORDS;        \
>         const __sigset_t* __set = (set);   \
>         int __ret = __set->__val[--__cnt]; \
> ```

FYI: This was a bug. This made real-time signals (above 32 bits in the 64-bit variable) to disappear, which broke some LTP tests.
___
*[LibOS/shim/include/shim_signal.h, line 147 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMD_rd5lTXMNN1ie3R:-Mb_b4da0H4Vqdjngh4Q:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/include/shim_signal.h#L147)):*

I don't think this should handle forced signals. They are by definition forced and should be always delivered to the app, skipping any signal masks and whatnot.
</blockquote></details>

Done.
___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 128 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMFGLHD5b8i_8PgBil:-Mb_b9cPAJlzxIcaiXS_:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/bookkeep/shim_signal.c#L128)):*

Please backtick-ify the symbol name (function).
</blockquote></details>

Done.
___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 604 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMFpmzCLKMxOCefB7x:-Mb_c7Y71dOJRraahR3k:b-leuscl) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/bookkeep/shim_signal.c#L604)):*

Why did you move locking?
Update: I see, you look at current signal mask in else-if clause. I don't like this version: the old one was lockless in the common case of no signals.
</blockquote></details>

Done. You are right, that was a bad "optimization".
___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 608 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMJZr40b8mAtMHRcqG:-Mb_ayWrBvZqZTSdb5Ee:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/bookkeep/shim_signal.c#L608)):*

I wouldn't add such comment: it's currently the case, but in principle they could be anything (basically just a way to skip signal mask checks). E.g. at some point we might need to have Graphene (kernel) forced signals.
Please either remove or elaborate.
</blockquote></details>

Done.
___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 610 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMHtlT9w6rCB_l4wOI:-Mb_bMTx1p_mvlDi5IdA:bw5zy3u) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/bookkeep/shim_signal.c#L610)):*

Do we have any official rules on boolean operator placement? I can't find anything in our docs, google-style allows for both version (and I always used to place them at the beginning).
</blockquote></details>

Done. I don't think we have any rules, but I changed as you requested.
___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 611 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMGCFH0gocL__qzvvh:-Mb_bPy35N5tx4tlWtT3:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/bookkeep/shim_signal.c#L611)):*

Please align.
</blockquote></details>

Done.
___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 72 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMN_PDCb7d5y-JeCN0:-Mb_e6z9B7xGe2sUZHwK:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/ipc/shim_ipc_pid.c#L72)):*

Please create another function `is_child_process_or_zombie` (or something) inside `bookkeep/shim_process.c`
</blockquote></details>

Done.
___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 94 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMNpIM6pQMlmmxjPQp:-Mb_e9k556Th7uf94hfb:b-r1uh7t) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/ipc/shim_ipc_pid.c#L94)):*

I don't like this: the malicious host can now skip message (signal) delivery and nothing will notice that. That's the case in the old code too, but if we change `send_ipc_message` above to also wait for confirmation, it wouldn't be then.
</blockquote></details>

I added a more specific check on `EACCES` and `EPIPE`. Other than that, I'm not sure what we can do. Retry on error a couple times, with some timeout (assuming that `g_process.zombies` is updated in the meantime)?

If you want to see a TODO here, then should it be something like `TODO: Malicious host could drop the message, and this code would hide this failure: maybe add a retry?`
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 260 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMTHGG4lo8eudqu-vW:-Mb_f9Rz1GC5XLJHRVPX:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L260)):*

Please backtickify symbol names.
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 269 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMTX_K04Xq-UsKyQ2X:-Mb_fKe48YpK3NUAuoIX:bth16vn) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L269)):*

Isn't this equivalent to `signotset(&mask, &old, allowed_ptr)`? Why all the shenanigans with full set?
</blockquote></details>

But we need `unblocked_mask` below, in `pop_allowed_signal(&unblocked_mask, &signal)`. And since we need to define `unblocked_mask` anyway, why not reuse it here.

And there seems to be no helper macro to simply revert the `allowed` mask (to get `unblocked_mask`). So I added the full-set variable `all_blocked`.
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 275 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMX9IY2cnT1QSsE-aU:-Mb_fic6CpSOBzloBuQH:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L275)):*

This should be `while`, shouldn't it?
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 276 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMQjdA8ZBlZBgLqv5p:-Mb_hMx256zeMjMp4RHK:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L276)):*

Wait, `thread_wait` does update the timeout, you just don't use it...
Look at functions in `shim_sleep.c` on how to do it.
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 291 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMYwNxBawotjZb8Ie3:-Mb_fpRb2HrMwF_wZxNb:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L291)):*

After some thinking: this should be `-EINTR`. Seems Linux allows for spurious EINTRs in this case (if we race with another thread to handle this signal).
</blockquote></details>

Done.

boryspoplawski (changes requested):
> Reviewed 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/include/shim_process.h, line 88 at r2](https://reviewable.io/reviews/oscarlab/graphene/2422#-Mbap1et9VJCzGxoI5Ic:-Mbap1eu0X-8aNbi9ICn:b8guyge) ([raw file](https://github.com/oscarlab/graphene/blob/0a30acd75e7fee01f32e3f2651c8f9c27e846575/LibOS/shim/include/shim_process.h#L88)):*
> ```objc
> bool mark_child_exited_by_pid(IDTYPE pid, IDTYPE child_uid, int exit_code, int signal);
> 
> /* Returns whether a process is found in the children list or zombie list of `g_process` */
> ```

NP: dot at the end.
Actually, please use doxygen syntax.
___
*[LibOS/shim/include/shim_signal.h, line 51 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbaT1TT-h60A7nExVZU:-MbardIX1KaoIX6yeMBB:b-xp9530) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/include/shim_signal.h#L51)):*

FYI: This was a bug. This made real-time signals (above 32 bits in the 64-bit variable) to disappear, which broke some LTP tests.
</blockquote></details>

Oh boy.
___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 597 at r2](https://reviewable.io/reviews/oscarlab/graphene/2422#-Mbb3MH72IOkUxuFYkVs:-Mbb3MH72IOkUxuFYkVt:bicnlt1) ([raw file](https://github.com/oscarlab/graphene/blob/0a30acd75e7fee01f32e3f2651c8f9c27e846575/LibOS/shim/src/bookkeep/shim_signal.c#L597)):*
> ```C
> }
> 
> void pop_allowed_signal(__sigset_t* mask, struct shim_signal* signal) {
> ```

Sorry for late notice, but what `allowed` means in this context? I don't like it, it's more like `unmasked` or `unblocked`. Please change name (my suggestions are: `pop_signal`, `pop_unmasked_signal`)
___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 94 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMNpIM6pQMlmmxjPQp:-Mbawsvm0GG6uztEO6Ce:b-e8qw8n) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/ipc/shim_ipc_pid.c#L94)):*

I added a more specific check on `EACCES` and `EPIPE`. Other than that, I'm not sure what we can do. Retry on error a couple times, with some timeout (assuming that `g_process.zombies` is updated in the meantime)?

If you want to see a TODO here, then should it be something like `TODO: Malicious host could drop the message, and this code would hide this failure: maybe add a retry?`
</blockquote></details>

Checking specific errors make little sense: malicious host can spoof some of them. Besides I think currently SGX PAL returns `EACCES` in most cases anyway.
I don't like adding a `TODO` here - there is not much legacy code around that, so the situation won't get better.
I too don't see a good solution to this problem, so maybe really retry a couple of times (probably with small sleeps)? This is a very rare case and the chances of spurious error would be small: if the process is dead already, then it must have sent the death notification message (or simply disconnect) already -> we just need to wait for IPC worker to handle it. Also, IMO spurious fail would be better than hiding the error.
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 269 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMTX_K04Xq-UsKyQ2X:-Mbb582H8ENql7c9R0cv:b-rd2r7c) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L269)):*

But we need `unblocked_mask` below, in `pop_allowed_signal(&unblocked_mask, &signal)`. And since we need to define `unblocked_mask` anyway, why not reuse it here.

And there seems to be no helper macro to simply revert the `allowed` mask (to get `unblocked_mask`). So I added the full-set variable `all_blocked`.
</blockquote></details>

But this `unblocked_mask` seems kind of unnatural and leads to confusions. E.g. you do `clear_illegal_signals(&unblocked_mask)` which is wrong: you pop signals not present in that mask, so if you clear `SIGKILL` then you will be able to pop it and return from this function. You should do `clear_illegal_signals(&x)` where `x` is a local copy of `allowed_ptr`.
Maybe move calculation of `unblocked_mask` to the place where it's actually needed? `signotset(&mask, &old, allowed_ptr)` is obvious (clear `allowed` from `old`)  whereas `__sigandset(&mask, &unblocked_mask, &old)` is not - get intersection of unblocked and currently blocked?!
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 229 at r2](https://reviewable.io/reviews/oscarlab/graphene/2422#-Mbb4CDm0br7kDXN0-Yc:-Mbb4CDm0br7kDXN0-Yd:bgnan7i) ([raw file](https://github.com/oscarlab/graphene/blob/0a30acd75e7fee01f32e3f2651c8f9c27e846575/LibOS/shim/src/sys/shim_sigaction.c#L229)):*
> ```C
> }
> 
> long shim_do_rt_sigtimedwait(const __sigset_t* allowed_ptr, siginfo_t* info,
> ```

ditto (`allowed`)
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 258 at r2](https://reviewable.io/reviews/oscarlab/graphene/2422#-Mbb5FKh-E3sABF3dbvx:-Mbb5FKi7vHAuKHc3PkE:byrq2m7) ([raw file](https://github.com/oscarlab/graphene/blob/0a30acd75e7fee01f32e3f2651c8f9c27e846575/LibOS/shim/src/sys/shim_sigaction.c#L258)):*
> ```C
>     __sigfillset(&all_blocked);
>     __signotset(&unblocked_mask, &all_blocked, allowed_ptr);
>     clear_illegal_signals(&unblocked_mask);
> ```

see below

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 597 at r2](https://reviewable.io/reviews/oscarlab/graphene/2422#-Mbb3MH72IOkUxuFYkVs:-MbbjNJ2C4O3sp8zuciS:b-1pj49p) ([raw file](https://github.com/oscarlab/graphene/blob/0a30acd75e7fee01f32e3f2651c8f9c27e846575/LibOS/shim/src/bookkeep/shim_signal.c#L597)):*

Sorry for late notice, but what `allowed` means in this context? I don't like it, it's more like `unmasked` or `unblocked`. Please change name (my suggestions are: `pop_signal`, `pop_unmasked_signal`)
</blockquote></details>

I will change to `pop_unblocked_signal()`.

Am I the only one who constantly confuses what mask means (it is a mask of blocked signals, which is contrary to everything else in Linux and our codebase). Really hard for me to remember, is there any mnemonic for this?
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 269 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMTX_K04Xq-UsKyQ2X:-Mbbhjiy4AU0n1113lhS:bcgbknu) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L269)):*
> You should do `clear_illegal_signals(&x)` where `x` is a local copy of `allowed_ptr`.

Oops, you're right.

> whereas `__sigandset(&mask, &unblocked_mask, &old)` is not - get intersection of unblocked and currently blocked?!

This is actually an intersection of the "newly blocked" (with only user-requested signals unblocked) and "currently blocked". And given that `mask` means that unset bits == unblocked and set bits == blocked, this intersection is "unblock all signals requested by the user in this syscall and all previously unblocked signals".

I guess my `unblocked_mask` name is confusing. Probably I should just call it `new`.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: 4 of 11 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/include/shim_process.h, line 88 at r2](https://reviewable.io/reviews/oscarlab/graphene/2422#-Mbap1et9VJCzGxoI5Ic:-MbeRt2j6oGOYCK2HuRt:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/0a30acd75e7fee01f32e3f2651c8f9c27e846575/LibOS/shim/include/shim_process.h#L88)):*

NP: dot at the end.
Actually, please use doxygen syntax.
</blockquote></details>

Done.
___
*[LibOS/shim/src/bookkeep/shim_signal.c, line 597 at r2](https://reviewable.io/reviews/oscarlab/graphene/2422#-Mbb3MH72IOkUxuFYkVs:-MbeSDpH-c2pwFIGx5m9:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/0a30acd75e7fee01f32e3f2651c8f9c27e846575/LibOS/shim/src/bookkeep/shim_signal.c#L597)):*

I will change to `pop_unblocked_signal()`.

Am I the only one who constantly confuses what mask means (it is a mask of blocked signals, which is contrary to everything else in Linux and our codebase). Really hard for me to remember, is there any mnemonic for this?
</blockquote></details>

Done.
___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 94 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMNpIM6pQMlmmxjPQp:-MbeV8_J-wW-gq-ex6tJ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/ipc/shim_ipc_pid.c#L94)):*

Checking specific errors make little sense: malicious host can spoof some of them. Besides I think currently SGX PAL returns `EACCES` in most cases anyway.
I don't like adding a `TODO` here - there is not much legacy code around that, so the situation won't get better.
I too don't see a good solution to this problem, so maybe really retry a couple of times (probably with small sleeps)? This is a very rare case and the chances of spurious error would be small: if the process is dead already, then it must have sent the death notification message (or simply disconnect) already -> we just need to wait for IPC worker to handle it. Also, IMO spurious fail would be better than hiding the error.
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 269 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMTX_K04Xq-UsKyQ2X:-MbeZqrdACx3H8kieoo8:beone7j) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/sys/shim_sigaction.c#L269)):*

> You should do `clear_illegal_signals(&x)` where `x` is a local copy of `allowed_ptr`.

Oops, you're right.

> whereas `__sigandset(&mask, &unblocked_mask, &old)` is not - get intersection of unblocked and currently blocked?!

This is actually an intersection of the "newly blocked" (with only user-requested signals unblocked) and "currently blocked". And given that `mask` means that unset bits == unblocked and set bits == blocked, this intersection is "unblock all signals requested by the user in this syscall and all previously unblocked signals".

I guess my `unblocked_mask` name is confusing. Probably I should just call it `new`.
</blockquote></details>

Done. Indeed, your proposed variant is more readable.
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 229 at r2](https://reviewable.io/reviews/oscarlab/graphene/2422#-Mbb4CDm0br7kDXN0-Yc:-MbeVXQ04FSXTjtx7uwa:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/0a30acd75e7fee01f32e3f2651c8f9c27e846575/LibOS/shim/src/sys/shim_sigaction.c#L229)):*

ditto (`allowed`)
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_sigaction.c, line 258 at r2](https://reviewable.io/reviews/oscarlab/graphene/2422#-Mbb5FKh-E3sABF3dbvx:-MbeYX1U1xXNgw0kv2lo:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/0a30acd75e7fee01f32e3f2651c8f9c27e846575/LibOS/shim/src/sys/shim_sigaction.c#L258)):*

see below
</blockquote></details>

Done.

boryspoplawski (changes requested):
> Reviewed 6 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: 10 of 11 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 94 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMNpIM6pQMlmmxjPQp:-MbfSK6R2d_-NAFvs9S2:bt8u0l8) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/ipc/shim_ipc_pid.c#L94)):*

Done.
</blockquote></details>

Now the malicious host can spoof error value in the first 2 `send_ipc_msg` and then return `0` in 3rd, which will effectively send 3 signals. The looping should happen around `is_zombie_process` check only.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: 10 of 11 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 94 at r1](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbMNpIM6pQMlmmxjPQp:-MbfnQCIDEOY4JMPVKBS:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/d4b320e8aef5dc4f79b95554266e6fd12a878f01/LibOS/shim/src/ipc/shim_ipc_pid.c#L94)):*

Now the malicious host can spoof error value in the first 2 `send_ipc_msg` and then return `0` in 3rd, which will effectively send 3 signals. The looping should happen around `is_zombie_process` check only.
</blockquote></details>

Done.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 71 at r4](https://reviewable.io/reviews/oscarlab/graphene/2422#-Mbfq9O97LS_H_FIoxA2:-Mbfq9O97LS_H_FIoxA3:book9ns) ([raw file](https://github.com/oscarlab/graphene/blob/65584a9072e50755d29f61f1ff3f9be5b3f1a647/LibOS/shim/src/ipc/shim_ipc_pid.c#L71)):*
> ```C
>             int wait_iter = 3;
>             while (wait_iter--) {
>                 if (is_zombie_process(dest_pid)) {
> ```

BTW, this won't work if the zombie is not our child, but I guess no application should use that anyway. Maybe add a comment about that somewhere?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: 10 of 11 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 71 at r4](https://reviewable.io/reviews/oscarlab/graphene/2422#-Mbfq9O97LS_H_FIoxA2:-Mbfqj2M8rQN_EL0z_UP:bqikwt) ([raw file](https://github.com/oscarlab/graphene/blob/65584a9072e50755d29f61f1ff3f9be5b3f1a647/LibOS/shim/src/ipc/shim_ipc_pid.c#L71)):*

BTW, this won't work if the zombie is not our child, but I guess no application should use that anyway. Maybe add a comment about that somewhere?
</blockquote></details>

Done. I had a small comment on this above (`...became a zombie (if it was our child)...`), but the comment was buried and a bit misleading. Added more sentences.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 83 at r5](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbfxyW9799j37WNZ6GQ:-MbfxcU70aTT2yTmW0lj:b2zp3n4) ([raw file](https://github.com/oscarlab/graphene/blob/461c8b9d933ca39e8e78c29c1da3ca2323285430/LibOS/shim/src/ipc/shim_ipc_pid.c#L83)):*
> ```C
>                      * a KILL in this thread, so we sleep for a bit and check for zombie again. */
>                     uint64_t timeout_us = 10000;
>                     thread_wait(&timeout_us, /*ignore_pending_signals=*/false);
> ```

Just noticed: we should ignore pending signals here.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: 10 of 11 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 83 at r5](https://reviewable.io/reviews/oscarlab/graphene/2422#-MbfxyW9799j37WNZ6GQ:-Mbg8vfp2G6-jMfXgt68:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/461c8b9d933ca39e8e78c29c1da3ca2323285430/LibOS/shim/src/ipc/shim_ipc_pid.c#L83)):*

Just noticed: we should ignore pending signals here.
</blockquote></details>

Done.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2422)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2433, opened 2021-06-09
[LibOS] Remove IPC msg_with_ack and introduce message responses

boryspoplawski:
## Description of the changes 
Function `ipc_send_msg_and_get_response` is introduced, which sends the probided IPC message and waits for a response to it. The response could be anything and is returned by this function - it is the caller responsibility to interpret it.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)

### DIFF
--- a/LibOS/shim/include/shim_ipc.h
+++ b/LibOS/shim/include/shim_ipc.h
@@ -7,6 +7,8 @@
 #ifndef SHIM_IPC_H_
 #define SHIM_IPC_H_
 
+#include <stdint.h>
+
 #include "avl_tree.h"
 #include "pal.h"
 #include "shim_defs.h"
@@ -15,27 +17,19 @@
 #include "shim_thread.h"
 #include "shim_types.h"
 
-/* if callback func returns RESPONSE_CALLBACK, send response msg even if callback succeeded */
-#define RESPONSE_CALLBACK 1
-
 #define RANGE_SIZE 32
 
 enum {
     IPC_MSG_RESP = 0,
     IPC_MSG_CONNBACK,      /*!< Request for establishing a connection to the sender. */
-    IPC_MSG_DUMMY,         /*!< Dummy callback, wakes up the thread waiting for this response. */
     IPC_MSG_CHILDEXIT,
     IPC_MSG_LEASE,
-    IPC_MSG_OFFER,
     IPC_MSG_SUBLEASE,
     IPC_MSG_QUERY,
     IPC_MSG_QUERYALL,
-    IPC_MSG_ANSWER,
     IPC_MSG_PID_KILL,
     IPC_MSG_PID_GETSTATUS,
-    IPC_MSG_PID_RETSTATUS,
     IPC_MSG_PID_GETMETA,
-    IPC_MSG_PID_RETMETA,
     IPC_MSG_SYNC_REQUEST_UPGRADE,
     IPC_MSG_SYNC_REQUEST_DOWNGRADE,
     IPC_MSG_SYNC_REQUEST_CLOSE,
@@ -81,7 +75,8 @@ int connect_to_process(IDTYPE dest);
  *
  * \param dest vmid of the destination process
  *
- * If there is no outgoing connection to \p dest, does nothing.
+ * If there is no outgoing connection to \p dest, does nothing. If any thread waits for a response
+ * to a message sent to \p dest, it is woken up and notified about the disconnect.
  */
 void remove_outgoing_ipc_connection(IDTYPE dest);
 /*!
@@ -91,67 +86,53 @@ void remove_outgoing_ipc_connection(IDTYPE dest);
  * broadcast IPC messages.
  */
 int request_leader_connect_back(void);
+/*!
+ * \brief Callback for an IPC connection request
+ *
+ * Parameters as per IPC callback interface.
+ * See also: #request_leader_connect_back
+ */
+int ipc_connect_back_callback(IDTYPE src, void* data, uint64_t seq);
 
-/* TODO: add `__attribute__((packed))` once `struct shim_ipc_msg` is removed.
- * Currently we might send some padding bytes, but that's not a problem since we only send those
- * between Graphene processes, which are all considered equally trusted. */
 struct ipc_msg_header {
     size_t size;
-    unsigned long seq;
+    uint64_t seq;
     unsigned char code;
-};
+} __attribute__((packed));
 
 struct shim_ipc_msg {
     struct ipc_msg_header header;
     char data[];
 };
-/*
- * XXX: Currently required by:
- * - `msg_add_range` in `LibOS/shim/src/ipc/shim_ipc_ranges.c`,
- */
-static_assert(offsetof(struct shim_ipc_msg, data) % 8 == 0, "Currently proper alignment is required");
-
-struct shim_ipc_msg_with_ack {
-    struct avl_tree_node node;
-    struct shim_thread* thread;
-    int retval;
-    void* private;
-    struct shim_ipc_msg msg;
-};
-
-struct shim_ipc_resp {
-    int retval;
-} __attribute__((packed));
 
 static inline size_t get_ipc_msg_size(size_t payload) {
     return sizeof(struct shim_ipc_msg) + payload;
 }
-static inline size_t get_ipc_msg_with_ack_size(size_t payload) {
-    return sizeof(struct shim_ipc_msg_with_ack) + payload;
-}
 
-void init_ipc_msg(struct shim_ipc_msg* msg, int code, size_t size);
-void init_ipc_msg_with_ack(struct shim_ipc_msg_with_ack* msg, int code, size_t size);
+void init_ipc_msg(struct shim_ipc_msg* msg, unsigned char code, size_t size);
+void init_ipc_response(struct shim_ipc_msg* msg, uint64_t seq, size_t size);
 
 /*!
  * \brief Send an IPC message
  *
- * \param msg message to send
  * \param dest vmid of the destination process
+ * \param msg message to send
  */
-int send_ipc_message(struct shim_ipc_msg* msg, IDTYPE dest);
+int ipc_send_message(IDTYPE dest, struct shim_ipc_msg* msg);
 /*!
- * \brief Send an IPC message and wait for response
+ * \brief Send an IPC message and wait for a response
  *
- * \param msg message to send
  * \param dest vmid of the destination process
- * \param[out] seq upon return contains sequence number of this message
+ * \param msg message to send
+ * \param[out] resp upon successful return contains a pointer to the response
  *
  * Send an IPC message to the \p dest process and wait for a response. An unique number is assigned
  * before sending the message and this thread will wait for a response IPC message, which contains
- * the same sequence number.
+ * the same sequence number. If this function succeeds, \p resp will contain pointer to the response
+ * data, which should be freed using `free` function. If \p resp is NULL, the response will be
+ * discarded, but still awaited for.
  */
-int send_ipc_message_with_ack(struct shim_ipc_msg_with_ack* msg, IDTYPE dest, unsigned long* seq);
+int ipc_send_msg_and_get_response(IDTYPE dest, struct shim_ipc_msg* msg, void** resp);
 /*!
  * \brief Broadcast an IPC message
  *
@@ -160,36 +141,22 @@ int send_ipc_message_with_ack(struct shim_ipc_msg_with_ack* msg, IDTYPE dest, un
  *
  * Send an IPC message \p msg to all known (connected) processes except for \p exclude_vmid.
  */
-int broadcast_ipc(struct shim_ipc_msg* msg, IDTYPE exclude_vmid);
-/*!
- * \brief Callback for a dummy message, just wakes the waiting thread
- *
- * \param src unused (just for interface conformance)
- * \param data unused (just for interface conformance)
- * \param seq sequence number of the message
- */
-int ipc_dummy_callback(IDTYPE src, void* data, unsigned long seq);
+int ipc_broadcast(struct shim_ipc_msg* msg, IDTYPE exclude_vmid);
 
 /*!
  * \brief Handle a response to a previously sent message
  *
+ * \param src ID of sender
+ * \param data body of the response
  * \param seq sequence number of the original message
- * \param callback callback to be called on the original message
- * \param data passed as the second argument to \p callback
  *
- * Searches for a `shim_ipc_msg_with_ack` sent previously to \p src, which has \p seq sequential
- * number. Then calls \p callback with the found message (or `NULL` if no message was found) as
- * the first argument and \p data as the second.
+ * Searches for a thread waiting for a response to a message previously sent to \p src with
+ * the sequence number \p seq. If such thread is found, it is woken up and \p data is passed to it
+ * (returned in `resp` argument of #ipc_send_msg_and_get_response).
+ * This function always takes the ownership of \p data, the caller of this function should never
+ * free it!
  */
-void ipc_msg_response_handle(unsigned long seq,
-                             void (*callback)(struct shim_ipc_msg_with_ack*, void*), void* data);
-/*!
- * \brief Wake up the thread awaiting for a response to \p req_msg
- *
- * \param req_msg original message which got the response
- * \param data unused (just to conform to #ipc_msg_response_handle callbacks interface)
- */
-void wake_req_msg_thread(struct shim_ipc_msg_with_ack* req_msg, void* data);
+int ipc_response_callback(IDTYPE src, void* data, uint64_t seq);
 
 /* common functions for pid & sysv namespaces */
 int add_ipc_subrange(IDTYPE idx, IDTYPE owner);
@@ -219,20 +186,17 @@ struct shim_ipc_cld_exit {
 } __attribute__((packed));
 
 int ipc_cld_exit_send(unsigned int exitcode, unsigned int term_signal);
-int ipc_cld_exit_callback(IDTYPE src, void* data, unsigned long seq);
+int ipc_cld_exit_callback(IDTYPE src, void* data, uint64_t seq);
 void ipc_child_disconnect_callback(IDTYPE vmid);
 
 int ipc_lease_send(void);
-int ipc_lease_callback(IDTYPE src, void* data, unsigned long seq);
+int ipc_lease_callback(IDTYPE src, void* data, uint64_t seq);
 
 /* OFFER: offer a range of IDs */
 struct shim_ipc_offer {
     IDTYPE base;
     IDTYPE size;
 } __attribute__((packed));
-
-int ipc_offer_send(IDTYPE dest, IDTYPE base, IDTYPE size, unsigned long seq);
-int ipc_offer_callback(IDTYPE src, void* data, unsigned long seq);
 
 /* SUBLEASE: lease a range of IDs */
 struct shim_ipc_sublease {
@@ -241,7 +205,7 @@ struct shim_ipc_sublease {
 } __attribute__((packed));
 
 int ipc_sublease_send(IDTYPE tenant, IDTYPE idx);
-int ipc_sublease_callback(IDTYPE src, void* data, unsigned long seq);
+int ipc_sublease_callback(IDTYPE src, void* data, uint64_t seq);
 
 /* QUERY: query for a certain ID */
 struct shim_ipc_query {
@@ -249,19 +213,17 @@ struct shim_ipc_query {
 } __attribute__((packed));
 
 int ipc_query_send(IDTYPE idx);
-int ipc_query_callback(IDTYPE src, void* data, unsigned long seq);
+int ipc_query_callback(IDTYPE src, void* data, uint64_t seq);
 
 /* QUERYALL: query for all IDs */
 int ipc_queryall_send(void);
-int ipc_queryall_callback(IDTYPE src, void* data, unsigned long seq);
+int ipc_queryall_callback(IDTYPE src, void* data, uint64_t seq);
 
 /* ANSWER: reply to the query with my offered IDs */
 struct shim_ipc_answer {
     size_t answers_cnt;
     struct ipc_ns_offered answers[];
 } __attribute__((packed));
-
-int ipc_answer_callback(IDTYPE src, void* data, unsigned long seq);
 
 /* PID_KILL: send signal to certain pid */
 struct shim_ipc_pid_kill {
@@ -275,11 +237,11 @@ int ipc_kill_process(IDTYPE sender, IDTYPE target, int sig);
 int ipc_kill_thread(IDTYPE sender, IDTYPE dest_pid, IDTYPE target, int sig);
 int ipc_kill_pgroup(IDTYPE sender, IDTYPE pgid, int sig);
 int ipc_kill_all(IDTYPE sender, int sig);
-int ipc_pid_kill_callback(IDTYPE src, void* data, unsigned long seq);
+int ipc_pid_kill_callback(IDTYPE src, void* data, uint64_t seq);
 
 /* PID_GETSTATUS: check if certain pid(s) exists */
 struct shim_ipc_pid_getstatus {
-    int npids;
+    size_t npids;
     IDTYPE pids[];
 };
 
@@ -289,17 +251,14 @@ struct pid_status {
     IDTYPE pgid;
 } __attribute__((packed));
 
-int ipc_pid_getstatus_send(IDTYPE dest, int npids, IDTYPE* pids, struct pid_status** status);
-int ipc_pid_getstatus_callback(IDTYPE src, void* data, unsigned long seq);
-
 /* PID_RETSTATUS: return status of pid(s) */
 struct shim_ipc_pid_retstatus {
-    int nstatus;
+    size_t count;
     struct pid_status status[];
 } __attribute__((packed));
 
-int ipc_pid_retstatus_send(IDTYPE dest, int nstatus, struct pid_status* status, unsigned long seq);
-int ipc_pid_retstatus_callback(IDTYPE src, void* data, unsigned long seq);
+int ipc_pid_getstatus(IDTYPE dest, int npids, IDTYPE* pids, struct shim_ipc_pid_retstatus** status);
+int ipc_pid_getstatus_callback(IDTYPE src, void* data, uint64_t seq);
 
 int get_all_pid_status(struct pid_status** status);
 
@@ -309,20 +268,15 @@ struct shim_ipc_pid_getmeta {
     enum pid_meta_code code;
 } __attribute__((packed));
 
-int ipc_pid_getmeta_send(IDTYPE pid, enum pid_meta_code code, void** data);
-int ipc_pid_getmeta_callback(IDTYPE src, void* data, unsigned long seq);
-
 /* PID_RETMETA: return metadata of certain pid */
 struct shim_ipc_pid_retmeta {
-    IDTYPE pid;
-    enum pid_meta_code code;
-    int datasize;
+    size_t datasize;
+    int ret_val;
     char data[];
 } __attribute__((packed));
 
-int ipc_pid_retmeta_send(IDTYPE dest, IDTYPE pid, enum pid_meta_code code, const void* data,
-                         int datasize, unsigned long seq);
-int ipc_pid_retmeta_callback(IDTYPE src, void* data, unsigned long seq);
+int ipc_pid_getmeta(IDTYPE pid, enum pid_meta_code code, struct shim_ipc_pid_retmeta** data);
+int ipc_pid_getmeta_callback(IDTYPE src, void* data, uint64_t seq);
 
 /* SYNC_REQUEST_*, SYNC_CONFIRM_ */
 struct shim_ipc_sync {
--- a/LibOS/shim/src/fs/proc/ipc-thread.c
+++ b/LibOS/shim/src/fs/proc/ipc-thread.c
@@ -73,7 +73,7 @@ static int find_ipc_thread_link(const char* name, struct shim_qstr* link,
 
     struct shim_dentry* dent = NULL;
     enum pid_meta_code ipc_code;
-    void* ipc_data = NULL;
+    struct shim_ipc_pid_retmeta* ipc_data = NULL;
 
     if (!memcmp(next, "root", next_len)) {
         ipc_code = PID_META_ROOT;
@@ -93,12 +93,12 @@ static int find_ipc_thread_link(const char* name, struct shim_qstr* link,
     ret = -ENOENT;
     goto out;
 do_ipc:
-    ret = ipc_pid_getmeta_send(pid, ipc_code, &ipc_data);
+    ret = ipc_pid_getmeta(pid, ipc_code, &ipc_data);
     if (ret < 0)
         goto out;
 
     if (link)
-        qstrsetstr(link, (char*)ipc_data, strlen((char*)ipc_data));
+        qstrsetstr(link, ipc_data->data, ipc_data->datasize);
 
 out:
     if (dent)
--- a/LibOS/shim/src/ipc/shim_ipc.c
+++ b/LibOS/shim/src/ipc/shim_ipc.c
@@ -1,16 +1,17 @@
 /* SPDX-License-Identifier: LGPL-3.0-or-later */
-/* Copyright (C) 2014 Stony Brook University
- * Copyright (C) 2021 Intel Corporation
+/* Copyright (C) 2021 Intel Corporation
  *                    Borys Popławski <borysp@invisiblethingslab.com>
  */
 
 /*
  * This file provides functions for dealing with outgoing IPC connections, mainly sending IPC
- * messages (`shim_ipc_msg` and `shim_ipc_msg_with_ack`).
+ * messages.
  */
 
 #include <stdbool.h>
+#include <stdint.h>
 
+#include "api.h"
 #include "assert.h"
 #include "avl_tree.h"
 #include "pal.h"
@@ -42,14 +43,22 @@ static bool ipc_connection_cmp(struct avl_tree_node* _a, struct avl_tree_node* _
 static struct avl_tree g_ipc_connections = { .cmp = ipc_connection_cmp };
 static struct shim_lock g_ipc_connections_lock;
 
-static bool msg_with_ack_cmp(struct avl_tree_node* _a, struct avl_tree_node* _b) {
-    struct shim_ipc_msg_with_ack* a = container_of(_a, struct shim_ipc_msg_with_ack, node);
-    struct shim_ipc_msg_with_ack* b = container_of(_b, struct shim_ipc_msg_with_ack, node);
-    return a->msg.header.seq <= b->msg.header.seq;
+struct ipc_msg_waiter {
+    struct avl_tree_node node;
+    PAL_HANDLE event;
+    uint64_t seq;
+    IDTYPE dest;
+    void* response_data;
+};
+
+static bool ipc_msg_waiter_cmp(struct avl_tree_node* _a, struct avl_tree_node* _b) {
+    struct ipc_msg_waiter* a = container_of(_a, struct ipc_msg_waiter, node);
+    struct ipc_msg_waiter* b = container_of(_b, struct ipc_msg_waiter, node);
+    return a->seq <= b->seq;
 }
 
-static struct avl_tree g_msg_with_ack_tree = { .cmp = msg_with_ack_cmp };
-static struct shim_lock g_msg_with_ack_tree_lock;
+static struct avl_tree g_msg_waiters_tree = { .cmp = ipc_msg_waiter_cmp };
+static struct shim_lock g_msg_waiters_tree_lock;
 
 IDTYPE g_self_vmid;
 struct shim_ipc_ids g_process_ipc_ids;
@@ -58,7 +67,7 @@ int init_ipc(void) {
     if (!create_lock(&g_ipc_connections_lock)) {
         return -ENOMEM;
     }
-    if (!create_lock(&g_msg_with_ack_tree_lock)) {
+    if (!create_lock(&g_msg_waiters_tree_lock)) {
         return -ENOMEM;
     }
 
@@ -175,27 +184,41 @@ void remove_outgoing_ipc_connection(IDTYPE dest) {
         _remove_ipc_connection(conn);
     }
     unlock(&g_ipc_connections_lock);
+
+    lock(&g_msg_waiters_tree_lock);
+    struct avl_tree_node* node = avl_tree_first(&g_msg_waiters_tree);
+    /* Usually there are no or very few waiters, so linear loop should be fine. If this becomes
+     * a problem for some reason, we can make `g_msg_waiters_tree` use `dest` as a part of the key
+     * and search for matching entries here. */
+    while (node) {
+        struct ipc_msg_waiter* waiter = container_of(node, struct ipc_msg_waiter, node);
+        if (waiter->dest == dest) {
+            waiter->response_data = NULL;
+            DkEventSet(waiter->event);
+            log_debug("Woke up a thread waiting for a message from a disconnected process\n");
+        }
+        node = avl_tree_next(node);
+    }
+    unlock(&g_msg_waiters_tree_lock);
 }
 
-void init_ipc_msg(struct shim_ipc_msg* msg, int code, size_t size) {
-    msg->header.size = get_ipc_msg_size(size);
-    msg->header.seq = 0;
-    msg->header.code = code;
+void init_ipc_msg(struct shim_ipc_msg* msg, unsigned char code, size_t size) {
+    SET_UNALIGNED(msg->header.size, get_ipc_msg_size(size));
+    SET_UNALIGNED(msg->header.seq, 0ul);
+    SET_UNALIGNED(msg->header.code, code);
 }
 
-void init_ipc_msg_with_ack(struct shim_ipc_msg_with_ack* msg, int code, size_t size) {
-    init_ipc_msg(&msg->msg, code, size);
-    msg->thread = NULL;
-    msg->retval = 0;
-    msg->private = NULL;
+void init_ipc_response(struct shim_ipc_msg* msg, uint64_t seq, size_t size) {
+    init_ipc_msg(msg, IPC_MSG_RESP, size);
+    SET_UNALIGNED(msg->header.seq, seq);
 }
 
-static int send_ipc_message_to_conn(struct shim_ipc_msg* msg, struct shim_ipc_connection* conn) {
+static int ipc_send_message_to_conn(struct shim_ipc_connection* conn, struct shim_ipc_msg* msg) {
     log_debug("Sending ipc message to %u\n", conn->vmid);
 
     lock(&conn->lock);
 
-    int ret = write_exact(conn->handle, msg,  msg->header.size);
+    int ret = write_exact(conn->handle, msg,  GET_UNALIGNED(msg->header.size));
     if (ret < 0) {
         log_error("Failed to send IPC msg to %u: %d\n", conn->vmid, ret);
         unlock(&conn->lock);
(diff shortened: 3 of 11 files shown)


boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: 0 of 7 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2433#-MblLPv4-uuK2wxLMMzm:-MblLPv4-uuK2wxLMMzn:b12dbiy) (no related file):*
TODO: add some debug logging in appropriate places.

---

boryspoplawski:
Jenkins, retest this please (just making sure)

boryspoplawski:
Jenkins, retest this please (just making sure)

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: 0 of 7 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[LibOS/shim/src/ipc/shim_ipc_worker.c, line 190 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbloPd_1pk2tBdfzx9Y:-MbloPd_1pk2tBdfzx9Z:bxi5yfe) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/src/ipc/shim_ipc_worker.c#L190)):*
> ```C
>         }
> 
>         /* TODO: fix this interface - ownership passing. */
> ```

Currently `msg_data` is passed to the callbacks, but only `IPS_MSG_RESP` takes the ownership, which is a bit ugly to deal with.
I see 3 options:
1) Leave it as it is (and have this if here).
2) Always pass ownership to callbacks - we would need to free it on all return paths in all callbacks - I very much dislike this approach.
3) Pass ownership only if the callback was successful i.e. returned `0`. We would still need to free `msg_data` in callbacks, but only on success paths - still cumbersome.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: 0 of 7 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbn2P1794wpcnX48MDZ:-Mbn2P184NTvoEu3aoq4:b-yh8j21) (no related file):*
@pwmarcz you might want to take a look at these changes, since you're probably going to use them soon ;)

---

mkow (changes requested):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

___
*[LibOS/shim/include/shim_ipc.h, line 81 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbmnW8L0gG9f7f3NwUX:-MbmnW8M0l_XUbuN0c_M:bu56mrg) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/include/shim_ipc.h#L81)):*
> ```objc
> int request_leader_connect_back(void);
> /*!
>  * \brief Callback for a IPC connection request
> ```

`a` -> `an`
___
*[LibOS/shim/include/shim_ipc.h, line 92 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbmoytC0jughDvTsr6I:-MbmoytC0jughDvTsr6J:b-tjmqt6) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/include/shim_ipc.h#L92)):*
> ```objc
>     unsigned long seq;
>     unsigned char code;
> } __attribute__((packed));
> ```

I'd add a comment that this is legal only assuming that the fields are sorted by decreasing alignment.
___
*[LibOS/shim/include/shim_ipc.h, line 250 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbmtboECDruLf1PGE7G:-MbmtboECDruLf1PGE7H:b-f7lo4y) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/include/shim_ipc.h#L250)):*
> ```
> int npids
> ```

Should be `size_t`.
___
*[LibOS/shim/src/ipc/shim_ipc.c, line 279 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbmx1kL8oaMPwgFurWK:-Mbmx1kL8oaMPwgFurWL:behf40d) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/src/ipc/shim_ipc.c#L279)):*
> ```
> unsigned long seq
> ```

I'm looking at this each time I see an IPC-related PR and I'm wondering whether this is a good type. Shouldn't we use `uint64_t` for it? On 32-bit it may potentially be overflowable under weird circumstances...
___
*[LibOS/shim/src/ipc/shim_ipc_worker.c, line 190 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbloPd_1pk2tBdfzx9Y:-Mbn2vWc9z39jf5Oc_qP:bg7xz2w) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/src/ipc/shim_ipc_worker.c#L190)):*

Currently `msg_data` is passed to the callbacks, but only `IPS_MSG_RESP` takes the ownership, which is a bit ugly to deal with.
I see 3 options:
1) Leave it as it is (and have this if here).
2) Always pass ownership to callbacks - we would need to free it on all return paths in all callbacks - I very much dislike this approach.
3) Pass ownership only if the callback was successful i.e. returned `0`. We would still need to free `msg_data` in callbacks, but only on success paths - still cumbersome.
</blockquote></details>

3. seems to be the least evil to me.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: 4 of 8 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/include/shim_ipc.h, line 81 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbmnW8L0gG9f7f3NwUX:-Mbn5HmS4r9HOJatn6nG:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/include/shim_ipc.h#L81)):*

`a` -> `an`
</blockquote></details>

Done.
___
*[LibOS/shim/include/shim_ipc.h, line 92 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbmoytC0jughDvTsr6I:-Mbn5ObA9l2TwgAMF_bP:b3u9ecy) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/include/shim_ipc.h#L92)):*

I'd add a comment that this is legal only assuming that the fields are sorted by decreasing alignment.
</blockquote></details>

Even then it's not legal. I have a TODO to change usages of this fields into memcpy, but seems I didn't add it to the code/reviewable.

Update: after recent changes this was trivial to fix, so stashed changes into this PR.
___
*[LibOS/shim/include/shim_ipc.h, line 250 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbmtboECDruLf1PGE7G:-Mbn5aKC-9qCcfsDxGYN:b-vrz457) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/include/shim_ipc.h#L250)):*

> ```
> int npids
> ```

Should be `size_t`.
</blockquote></details>

Yes it should. But it's an old interface, I didn't change it in this PR. It's waiting for its turn :)
___
*[LibOS/shim/src/ipc/shim_ipc.c, line 279 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbmx1kL8oaMPwgFurWK:-Mbn5xSA1m_VvXzg9WTp:b-9zu9s2) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/src/ipc/shim_ipc.c#L279)):*

> ```
> unsigned long seq
> ```

I'm looking at this each time I see an IPC-related PR and I'm wondering whether this is a good type. Shouldn't we use `uint64_t` for it? On 32-bit it may potentially be overflowable under weird circumstances...
</blockquote></details>

Actually after this PR the overflow should not matter, as long as there are no 2 messages with the same `seq` at the same time, but that would mean 2**32 concurrent messages going out of 1 thread, which seems unlikely.
Changed it to start from `0` and removed the check below.
___
*[LibOS/shim/src/ipc/shim_ipc_worker.c, line 190 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbloPd_1pk2tBdfzx9Y:-Mbn9ALOE12LnvDaZ5EQ:b-2rwu0b) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/src/ipc/shim_ipc_worker.c#L190)):*

3. seems to be the least evil to me.
</blockquote></details>

TBH: I'm leaning towards 1, beacuase `IPC_MSG_RESP` is pretty special. Any new callback that is created should not be interested in keeping this `msg_data` for itself (i.e. requiring ownership passing). So it's not like the list will grow and we would need to add more ifs here. Also such new callback wouldn't need to worry about whether and when to free `msg_data`.

My original impression was that the current version is super ugly and needs to be somehow fixed, but after thinking a bit about it, I don't se a better way (i.e. I think all other proposed, including 3, are even worse).

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[common/include/api.h, line 110 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbnOTH25M-8BPQjTrC6:-MbnOTH25M-8BPQjTrC7:b-r1xmmm) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/common/include/api.h#L110)):*
> ```objc
> 
> #define SET_UNALIGNED(a, b) ({                  \
>     __typeof__(b) _b = (b);                     \
> ```

Why having anti-double-eval on `b`, but not on `a`? Both occur more than once in this macro.
___
*[LibOS/shim/include/shim_ipc.h, line 92 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbmoytC0jughDvTsr6I:-MbnOgO5-XxE7TA7yhsr:b-wdj8zi) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/include/shim_ipc.h#L92)):*

Even then it's not legal. I have a TODO to change usages of this fields into memcpy, but seems I didn't add it to the code/reviewable.

Update: after recent changes this was trivial to fix, so stashed changes into this PR.
</blockquote></details>

Nice, thanks!
___
*[LibOS/shim/src/ipc/shim_ipc.c, line 279 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbmx1kL8oaMPwgFurWK:-MbnP7a1DtUH6LZdrhz9:b16w16) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/src/ipc/shim_ipc.c#L279)):*

Actually after this PR the overflow should not matter, as long as there are no 2 messages with the same `seq` at the same time, but that would mean 2**32 concurrent messages going out of 1 thread, which seems unlikely.
Changed it to start from `0` and removed the check below.
</blockquote></details>

Yeah, I was thinking the same previously, but can't there be a case where you stall the first message somehow (e.g. by freezing its recipient process) and then if the rest 2**32-1 would be addressed to other processes they would then get answered quickly? This way the messages wouldn't need to be concurrent.
___
*[LibOS/shim/src/ipc/shim_ipc_worker.c, line 190 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbloPd_1pk2tBdfzx9Y:-MbnQLL34FBXPSPFF3RQ:b-rqqs1y) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/src/ipc/shim_ipc_worker.c#L190)):*

TBH: I'm leaning towards 1, beacuase `IPC_MSG_RESP` is pretty special. Any new callback that is created should not be interested in keeping this `msg_data` for itself (i.e. requiring ownership passing). So it's not like the list will grow and we would need to add more ifs here. Also such new callback wouldn't need to worry about whether and when to free `msg_data`.

My original impression was that the current version is super ugly and needs to be somehow fixed, but after thinking a bit about it, I don't se a better way (i.e. I think all other proposed, including 3, are even worse).
</blockquote></details>

I'm fine with this, I'm relying on you in this area of Graphene code :)

Also, I thought a bit about alternative solutions, maybe something around returning by arguments could make this ownership passing more explicit? E.g. something like: you get this data buffer through `type**` arg, and if you want to take ownership then you NULL-ify it. But I'm not sure if it's better, sounds like a pretty uncommon convention (but proper ownership control in most C codebases is also an uncommon thing :) ).

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[common/include/api.h, line 110 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbnOTH25M-8BPQjTrC6:-MbnUW1lD7BCIdlyGoFv:bbm93uj) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/common/include/api.h#L110)):*
This has nothing to do with double-eval, it's to support rvalues in `b` (e.g. immediates or function call results).
Note we take an address of `_b`.

> Both occur more than once in this macro.

But other usage places do not evaluate them, there is only one evaluation for each (unless they are VLAs, but then static_assert should fail).
___
*[LibOS/shim/src/ipc/shim_ipc.c, line 279 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbmx1kL8oaMPwgFurWK:-MbnXEcOCo19U00M-ckM:b-qvuy6) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/src/ipc/shim_ipc.c#L279)):*

Yeah, I was thinking the same previously, but can't there be a case where you stall the first message somehow (e.g. by freezing its recipient process) and then if the rest 2**32-1 would be addressed to other processes they would then get answered quickly? This way the messages wouldn't need to be concurrent.
</blockquote></details>

That's true, and with careful ordering of the overlap and wakeup we could get a type confusion (on response data).
I'll restore the previous version (with 0 acting as a invalid value) and change it to 64bits.
___
*[LibOS/shim/src/ipc/shim_ipc_worker.c, line 190 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbloPd_1pk2tBdfzx9Y:-MbnW-hV2CJe5E802vs-:b-90fpya) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/src/ipc/shim_ipc_worker.c#L190)):*
Ah, yes, that was the original idea (to have `void**` and NULL-ify it if we took the ownership of the pointer), the problem is that the interface would change (not a problem per se) an just as with 3 - we really only want to do that in response handling (`IPC_MSG_RESP`), which means other places would have to use `*data`. Maybe this is better than special casing like now?

> sounds like a pretty uncommon convention

Not sure if it's or it's not, but I've definitely used it in Graphene before :)

Let me try that tomorrow, I'll see if it looks better.

dimakuv (changes requested):
> Reviewed 4 of 7 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 23 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/include/shim_ipc.h, line 118 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mboqkv8E71L04snGHbt:-Mboqkv8E71L04snGHbu:b-z6yyiq) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/include/shim_ipc.h#L118)):*
> ```objc
>  * \param dest vmid of the destination process
>  * \param msg message to send
>  * \param[out] resp upon successful return contains poitner to the response
> ```

`pointer`
___
*[LibOS/shim/include/shim_ipc.h, line 126 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mboqtsy4cqwPLnoFm8d:-Mboqtsy4cqwPLnoFm8e:brmk5h7) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/include/shim_ipc.h#L126)):*
> ```objc
>  * discarded, but still awaited for.
>  */
> int ipc_send_msg_and_get_response(IDTYPE dst, struct shim_ipc_msg* msg, void** resp);
> ```

`dst` -- typo? You have `dest` in all other places (including comments for this func).
___
*[LibOS/shim/include/shim_ipc.h, line 135 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MboqYfZ-knUXUgL1UJx:-MboqYfZ-knUXUgL1UJy:bh4br3j) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/include/shim_ipc.h#L135)):*
> ```objc
>  * Send an IPC message \p msg to all known (connected) processes except for \p exclude_vmid.
>  */
> int broadcast_ipc(struct shim_ipc_msg* msg, IDTYPE exclude_vmid);
> ```

Since we're changing all function names to have a prefix `ipc_`, maybe also change this to `ipc_broadcast`?
___
*[LibOS/shim/include/shim_ipc.h, line 145 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MborsaJ17pW8wjSv5_6:-MborsaJ17pW8wjSv5_7:b6eyjzf) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/include/shim_ipc.h#L145)):*
> ```objc
>  *
>  * Searches for a thread waiting for a response to a message previously sent to \p src with
>  * the sequence number \p seq. If such thread is found, it is woken up and \p data is passed to it.
> ```

What does it mean `data is passed to it`? Previously you called a callback, with data as one argument. Could you make it clearer in this comment, how does the woken thread "detects" this data?

I haven't seen to implementation, but I'm assuming that the thread-to-awake must sleep in the `ipc_send_msg_and_get_response()` function, and the `data` is found in the `resp` object there.
___
*[LibOS/shim/include/shim_ipc.h, line 189 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbotJw-CDJ3_Fz_yfV6:-MbotJw-CDJ3_Fz_yfV7:b8fo8vx) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/include/shim_ipc.h#L189)):*
> ```objc
>     IDTYPE base;
>     IDTYPE size;
> } __attribute__((packed));
> ```

Where does this OFFER message belong to? Is it for `ipc_lease_send`? Could you move it above that function then? (And also add a quick comment.)
___
*[LibOS/shim/include/shim_ipc.h, line 250 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbou3ab-ylaRwnSPSyl:-Mbou3ab-ylaRwnSPSym:brb1xj9) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/include/shim_ipc.h#L250)):*
> ```objc
> } __attribute__((packed));
> 
> int ipc_pid_getstatus(IDTYPE dest, int npids, IDTYPE* pids, struct shim_ipc_pid_retstatus** status);
> ```

Please change to `size_t npids` since you've done it above.
___
*[LibOS/shim/src/ipc/shim_ipc.c, line 2 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MboupVp-HTDHQDORJ9b:-MboupVq7b-rGYLWwcUj:b-60gj3c) ([raw file](https://github.com/oscarlab/graphene/blob/f0604438cda00379b3ea4c67a1a1c2a55a6f0416/LibOS/shim/src/ipc/shim_ipc.c#L2)):*
> ```C
> /* Copyright (C) 2014 Stony Brook University
> ```

Is this correct from a legal perspective?
___
*[LibOS/shim/src/ipc/shim_ipc.c, line 190 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbovPtbBCbFn9dQrGAy:-MbovPtbBCbFn9dQrGAz:b-4u5wdk) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/src/ipc/shim_ipc.c#L190)):*
> ```C
>     SET_UNALIGNED(msg->header.size, get_ipc_msg_size(size));
>     SET_UNALIGNED(msg->header.seq, 0ul);
>     SET_UNALIGNED(msg->header.code, code);
> ```

What is the point of `SET_UNALIGNED`?
___
*[LibOS/shim/src/ipc/shim_ipc.c, line 232 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbow9VC9mTO8GJlH7i-:-Mbow9VC9mTO8GJlH7i0:b6dz94k) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/src/ipc/shim_ipc.c#L232)):*
> ```C
>     int ret = 0;
>     do {
>         ret = pal_to_unix_errno(DkEventWait(waiter->event, NULL));
> ```

Please add `/*timeout=*/`
___
*[LibOS/shim/src/ipc/shim_ipc.c, line 248 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbowirSCAAecFnn92iI:-MbowirSCAAecFnn92iJ:bsa1p3f) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/src/ipc/shim_ipc.c#L248)):*
> ```C
>         .response_data = NULL,
>     };
>     int ret = DkEventCreate(&waiter.event, /*init_signaled=*/false, /*auto_clear=*/false);
> ```

What was wrong with using `thread_wait()`? Is it to support "send ipc message during sighandler" cases (where the normal-execution flow was probably already thread-waiting)?

Using a dedicated event is of course better, but won't it introduce performance overheads? For each message send, we now need to create an event, which is not fast at all in SGX case: https://github.com/oscarlab/graphene/blob/c781b2ab9081bc5725270a0344206aca28fbdc3e/Pal/src/host/Linux-SGX/db_events.c#L26
___
*[LibOS/shim/src/ipc/shim_ipc.c, line 268 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mboxov92m_NRsyww8Dn:-MboxovA23YGkWOfs6Yk:b54dsvn) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/src/ipc/shim_ipc.c#L268)):*
> ```C
> 
>     if (resp) {
>         *resp = waiter.response_data;
> ```

Maybe add an assert that `response_data` contains something? I would also add a comment about the passing of ownership here.
___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 152 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbozVb1BV7NDoG0eo2k:-MbozVb28EqD4Udepaf7:b-vverrb) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/src/ipc/shim_ipc_pid.c#L152)):*
> ```C
> }
> 
> int ipc_pid_getstatus(IDTYPE dest, int npids, IDTYPE* pids,
> ```

`size_t npids`
___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 237 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbozxG83l5zBySxBTjz:-MbozxG83l5zBySxBTk-:b-mwpcc) ([raw file](https://github.com/oscarlab/graphene/blob/f0604438cda00379b3ea4c67a1a1c2a55a6f0416/LibOS/shim/src/ipc/shim_ipc_pid.c#L237)):*
> ```C
>                   status[0].pid);
>     else
>         log_debug("ipc send to %u: IPC_MSG_PID_RETSTATUS(0, [])\n", dest);
> ```

Don't you want to keep this log message? Looks useful to know that the response was sent.
___
*[LibOS/shim/src/ipc/shim_ipc_pid.c, line 415 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbp-3SDEZpxjX8lj8fu:-Mbp-3SDEZpxjX8lj8fv:b-mwpcc) ([raw file](https://github.com/oscarlab/graphene/blob/f0604438cda00379b3ea4c67a1a1c2a55a6f0416/LibOS/shim/src/ipc/shim_ipc_pid.c#L415)):*
> ```C
> 
>     log_debug("ipc send to %u: IPC_MSG_PID_RETMETA(%d, %s, %d)\n", dest, pid,
>               pid_meta_code_str[code], datasize);
> ```

Don't you want to keep this log message? Looks useful to know that the response was sent.
___
*[LibOS/shim/src/ipc/shim_ipc_ranges.c, line 495 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbp-PR090V_ElsZXr-m:-Mbp-PR090V_ElsZXr-n:b-lpyh5l) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/src/ipc/shim_ipc_ranges.c#L495)):*
> ```C
>         return ret;
>     }
> ```

Maybe add a log message that you received a response with an offer?
___
*[LibOS/shim/src/ipc/shim_ipc_ranges.c, line 529 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbp4hT1AV6z9lUzU3a8:-Mbp4c60DKCHk8OgNVG3:b-cmj732) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/src/ipc/shim_ipc_ranges.c#L529)):*
> ```C
> 
>     memcpy(&msg->data, &msgin, sizeof(msgin));
> ```

Log message on sending the response?
___
*[LibOS/shim/src/ipc/shim_ipc_ranges.c, line 592 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbp4l0I0BpQmROHCpvN:-Mbp4l0I0BpQmROHCpvO:b-cmj732) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/src/ipc/shim_ipc_ranges.c#L592)):*
> ```C
> 
>     memcpy(&msg->data, &offer, sizeof(offer));
> ```

Log message on sending the response?
___
*[LibOS/shim/src/ipc/shim_ipc_ranges.c, line 609 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbp4snxDcdmpWTvHDvA:-Mbp4snxDcdmpWTvHDvB:b-d3hauy) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/src/ipc/shim_ipc_ranges.c#L609)):*
> Log message on sending the response?
___
*[LibOS/shim/src/ipc/shim_ipc_worker.c, line 190 at r1](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbloPd_1pk2tBdfzx9Y:-Mbp5NIB0IwecqGSNMv_:b-afankd) ([raw file](https://github.com/oscarlab/graphene/blob/93bf7fb6ad920cbe2398e10e5ff8b5a162c4865a/LibOS/shim/src/ipc/shim_ipc_worker.c#L190)):*

Ah, yes, that was the original idea (to have `void**` and NULL-ify it if we took the ownership of the pointer), the problem is that the interface would change (not a problem per se) an just as with 3 - we really only want to do that in response handling (`IPC_MSG_RESP`), which means other places would have to use `*data`. Maybe this is better than special casing like now?

> sounds like a pretty uncommon convention

Not sure if it's or it's not, but I've definitely used it in Graphene before :)

Let me try that tomorrow, I'll see if it looks better.
</blockquote></details>

I think option 1 is fine. But I don't like the `|| ret != 0` part in this IF. Why not just pass the ownership of `msg_data` to `ipc_response_callback()` completely, and it will be its responsibility to `free(msg_data)` even on failures?
___
*[LibOS/shim/src/ipc/shim_ipc_worker.c, line 188 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbp5yqTD8PDWjML6RTU:-Mbp5yqTD8PDWjML6RTV:b-g5rj54) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/src/ipc/shim_ipc_worker.c#L188)):*
> ```C
>         } else {
>             log_error(LOG_PREFIX "received unknown IPC msg type: %u\n", msg_code);
>             ret = -EINVAL;
> ```

Maybe just `continue;` here so we don't deal with this special case of `ret == -EINVAL` below?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 22 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[common/include/api.h, line 110 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbnOTH25M-8BPQjTrC6:-MbpI9vNCziAkSfDbR9e:b-7vo826) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/common/include/api.h#L110)):*

This has nothing to do with double-eval, it's to support rvalues in `b` (e.g. immediates or function call results).
Note we take an address of `_b`.

> Both occur more than once in this macro.

But other usage places do not evaluate them, there is only one evaluation for each (unless they are VLAs, but then static_assert should fail).
</blockquote></details>

Ah, I see.
___
*[LibOS/shim/src/ipc/shim_ipc.c, line 2 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MboupVp-HTDHQDORJ9b:-MbpIa3E7UQSVjv-r1rq:bdcm9qt) ([raw file](https://github.com/oscarlab/graphene/blob/f0604438cda00379b3ea4c67a1a1c2a55a6f0416/LibOS/shim/src/ipc/shim_ipc.c#L2)):*

Is this correct from a legal perspective?
</blockquote></details>

AFAIK at this point Borys have rewritten all the lines in this file.
___
*[LibOS/shim/src/ipc/shim_ipc.c, line 190 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbovPtbBCbFn9dQrGAy:-MbpIoGqBuEGsG_luo2A:b-rciir1) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/src/ipc/shim_ipc.c#L190)):*

What is the point of `SET_UNALIGNED`?
</blockquote></details>

`msg->header` is packed, so accessing its fields other than by using `memcpy()` is illegal.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 21 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/src/ipc/shim_ipc.c, line 290 at r3](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbsIO3RDdI7pjUpPrHh:-MbsIO3RDdI7pjUpPrHi:bbz2ye2) ([raw file](https://github.com/oscarlab/graphene/blob/80973a0169658d6cb2be3120c7e36fcc625d2606/LibOS/shim/src/ipc/shim_ipc.c#L290)):*
> ```C
>     if (!seq) {
>         log_error("Got an IPC response without a sequence number\n");
>         return -EINVAL;
> ```

I think `data` leaks here? Not sure if this error is fatal though, if it is then it may not be a problem.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: 7 of 8 files reviewed, 22 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/src/ipc/shim_ipc.c, line 290 at r3](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbsIO3RDdI7pjUpPrHh:-MbsanRcE6Z_B_O_3AkK:b-h5d02v) ([raw file](https://github.com/oscarlab/graphene/blob/80973a0169658d6cb2be3120c7e36fcc625d2606/LibOS/shim/src/ipc/shim_ipc.c#L290)):*

I think `data` leaks here? Not sure if this error is fatal though, if it is then it may not be a problem.
</blockquote></details>

Oops, right, stashed too many unrelated changes in one fixup and missed this.
___
*[LibOS/shim/src/ipc/shim_ipc_worker.c, line 186 at r3](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbscrkED6IckKh64HNB:-MbscrkFAWR337RDFeYw:blrzfnm) ([raw file](https://github.com/oscarlab/graphene/blob/80973a0169658d6cb2be3120c7e36fcc625d2606/LibOS/shim/src/ipc/shim_ipc_worker.c#L186)):*
> ```C
>         if (msg_code < ARRAY_SIZE(ipc_callbacks) && ipc_callbacks[msg_code]) {
>             ret = ipc_callbacks[msg_code](conn->vmid, msg_data, msg_seq);
>             if (ret < 0) {
> ```

TBH I have no idea what to do here...
Should this be a fatal error?

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 21 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/src/ipc/shim_ipc_worker.c, line 186 at r3](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbscrkED6IckKh64HNB:-MbuUsfF1hNoLuSpt_1F:b-7sju7h) ([raw file](https://github.com/oscarlab/graphene/blob/80973a0169658d6cb2be3120c7e36fcc625d2606/LibOS/shim/src/ipc/shim_ipc_worker.c#L186)):*

TBH I have no idea what to do here...
Should this be a fatal error?
</blockquote></details>

I'd say so, but maybe I'm missing a case where it can legitimately fail.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 20 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/src/ipc/shim_ipc_worker.c, line 186 at r3](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbscrkED6IckKh64HNB:-Mc8rme078fiRMyONol1:btmddzw) ([raw file](https://github.com/oscarlab/graphene/blob/80973a0169658d6cb2be3120c7e36fcc625d2606/LibOS/shim/src/ipc/shim_ipc_worker.c#L186)):*

I'd say so, but maybe I'm missing a case where it can legitimately fail.
</blockquote></details>

Turns out it can, but I think it should not be the case, I'll move some thing around and make this fatal.

dimakuv (approved):
> Reviewed 5 of 6 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/include/shim_ipc.h, line 250 at r2](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mbou3ab-ylaRwnSPSyl:-Mc93ZTJ6H4Jz5HZRZRw:b32w0x) ([raw file](https://github.com/oscarlab/graphene/blob/1286e0415eb1f06e0854011826f296ef75910c87/LibOS/shim/include/shim_ipc.h#L250)):*

ditto (in this PR I've only changes what was necessary)
</blockquote></details>

OK

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: 3 of 10 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/src/ipc/shim_ipc_worker.c, line 186 at r3](https://reviewable.io/reviews/oscarlab/graphene/2433#-MbscrkED6IckKh64HNB:-Mc9Yg4w1sbutANnSzo1:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/80973a0169658d6cb2be3120c7e36fcc625d2606/LibOS/shim/src/ipc/shim_ipc_worker.c#L186)):*

Turns out it can, but I think it should not be the case, I'll move some thing around and make this fatal.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/sighandler_reset.c, line 42 at r5](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mc9Yklc5NYa3xNB6jL7:-Mc9Yklc5NYa3xNB6jL8:b-udvjhp) ([raw file](https://github.com/oscarlab/graphene/blob/f0604438cda00379b3ea4c67a1a1c2a55a6f0416/LibOS/shim/test/regression/sighandler_reset.c#L42)):*
> ```C
>     }
> 
>     wait(NULL);
> ```

FYI: this returned `-EINTR`, so I fixed everything.

dimakuv (approved):
> Reviewed 7 of 7 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 7 of 7 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/src/ipc/shim_ipc.c, line 187 at r5](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mc9_ONoDOKpZNzOQACF:-Mc9_ONoDOKpZNzOQACG:bdqy03k) ([raw file](https://github.com/oscarlab/graphene/blob/f68853da4c9aec4b2f857603ef5e4565032865a4/LibOS/shim/src/ipc/shim_ipc.c#L187)):*
> ```C
>     }
>     unlock(&g_ipc_connections_lock);
> ```

What if a new connection to `dest` gets created here? Is this possible? If not, please add a comment explaining why.

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/src/ipc/shim_ipc.c, line 187 at r5](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mc9_ONoDOKpZNzOQACF:-Mc9fXuF45RUZ5mvgObm:b-yskwcr) ([raw file](https://github.com/oscarlab/graphene/blob/f68853da4c9aec4b2f857603ef5e4565032865a4/LibOS/shim/src/ipc/shim_ipc.c#L187)):*

What if a new connection to `dest` gets created here? Is this possible? If not, please add a comment explaining why.
</blockquote></details>

Well, nothing? It is possible.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[LibOS/shim/src/ipc/shim_ipc.c, line 187 at r5](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mc9_ONoDOKpZNzOQACF:-Mc9gmtt2T53deIgtyuo:bem48n1) ([raw file](https://github.com/oscarlab/graphene/blob/f68853da4c9aec4b2f857603ef5e4565032865a4/LibOS/shim/src/ipc/shim_ipc.c#L187)):*

Well, nothing? It is possible.
</blockquote></details>

And the code below will delete the response from it, sounds wrong?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/src/ipc/shim_ipc.c, line 187 at r5](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mc9_ONoDOKpZNzOQACF:-Mc9hR6T80TLF06dNXXt:b-4eyti5) ([raw file](https://github.com/oscarlab/graphene/blob/f68853da4c9aec4b2f857603ef5e4565032865a4/LibOS/shim/src/ipc/shim_ipc.c#L187)):*

And the code below will delete the response from it, sounds wrong?
</blockquote></details>

If there is a response waiter, that means we had a connection that was removed above - in such case there was an error, so removing it is fine.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/ipc/shim_ipc.c, line 187 at r5](https://reviewable.io/reviews/oscarlab/graphene/2433#-Mc9_ONoDOKpZNzOQACF:-Mc9mYLL2QLGREgmS1ww:bz29wzx) ([raw file](https://github.com/oscarlab/graphene/blob/f68853da4c9aec4b2f857603ef5e4565032865a4/LibOS/shim/src/ipc/shim_ipc.c#L187)):*

If there is a response waiter, that means we had a connection that was removed above - in such case there was an error, so removing it is fine.
</blockquote></details>

Discussed on priv, seems legit, resolving.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2433)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2446, opened 2021-06-15
[Pal/Linux-SGX] Disallow creating unknown files

dimakuv:
## Description of the changes 

Previously, Linux-SGX PAL had a bug of creating any files from within the enclave, even though such files were not listed as allowed or trusted files/paths. This commit fixes this bug, and also disallows opening a trusted file for write/append (because this doesn't make sense and write operations are already disallowed for trusted files).

## How to test this PR? 

The `file_check_policy` LibOS regression test is augmented to test this bug fix.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2446)

PR #2450, opened 2021-06-16
[Pal/Linux-SGX] Refactor `file_open` and `load_trusted_file` functions

dimakuv:
## Description of the changes 

This PR moves all possible checks (file must be in a set of protected, trusted, allowed files from the manifest; trusted files cannot be created; etc.) before actual opening of the file in `file_open`. As a side effect, this fixes a bug when app wants to open an unknown file for write/append -- previously, Graphene would open such file, possibly truncate it and only then return error.

Also, this PR refactors `load_trusted_file` into `get_trusted_or_allowed_file` and `load_trusted_or_allowed_file`.

~~Must be rebased on top of #2446.~~

Originally found by @mkow.

## How to test this PR? 

All tests must succeed. Especially `file_check_policy` LibOS regression tests.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -54,21 +54,22 @@ confidence=
 
 enable=*
 disable=
-    bad-option-value,  # for compatibility between different pylint versions
     bad-continuation,
+    bad-option-value,  # for compatibility between different pylint versions
     c-extension-no-member,
+    fixme,
+    invalid-name,
     missing-docstring,
     missing-function-docstring,
     missing-module-docstring,
-    invalid-name,
-    fixme,
+    no-self-use,
     raise-missing-from,
-    too-many-instance-attributes,
-    too-many-branches,
-    too-many-statements,
     too-few-public-methods,
+    too-many-branches,
+    too-many-instance-attributes,
+    too-many-lines,
     too-many-public-methods,
-    no-self-use
+    too-many-statements
 
 
 [REPORTS]
--- a/LibOS/shim/test/regression/.gitignore
+++ b/LibOS/shim/test/regression/.gitignore
@@ -1,6 +1,9 @@
 /*.manifest
 /*.xml
 
+/nonexisting_testfile
+/testfile
+
 /.cache
 /abort
 /abort_multithread
@@ -102,7 +105,6 @@
 /sysfs_common
 /tcp_ipv6_v6only
 /tcp_msg_peek
-/testfile
 /tmp
 /udp
 /unix
--- a/LibOS/shim/test/regression/file_check_policy.c
+++ b/LibOS/shim/test/regression/file_check_policy.c
@@ -1,3 +1,4 @@
+#include <err.h>
 #include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -5,19 +6,29 @@
 
 int main(int argc, char** argv) {
     if (argc != 3) {
-        fprintf(stderr, "Usage: %s read|append <filename>\n", argv[0]);
+        fprintf(stderr, "Usage: %s read|write|append <filename>\n", argv[0]);
         return 1;
     }
 
     /* we use append instead of write simply to not overwrite the file */
-    FILE* fp = fopen(argv[2], argv[1][0] == 'r' ? "r" : "a");
+    const char* mode;
+    if (!strcmp(argv[1], "read")) {
+        mode = "r";
+    } else if (!strcmp(argv[1], "write")) {
+        mode = "r+"; // without file creation
+    } else if (!strcmp(argv[1], "append")) {
+        mode = "a";
+    } else {
+        errx(1, "invalid cmdline argument");
+    }
+    FILE* fp = fopen(argv[2], mode);
     if (!fp) {
         perror("fopen failed");
         return 2;
     }
 
-    int reti = fclose(fp);
-    if (reti) {
+    int ret = fclose(fp);
+    if (ret) {
         perror("fclose failed");
         return 3;
     }
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -272,32 +272,78 @@ class TC_03_FileCheckPolicy(RegressionTestCase):
         self.assertIn('file_check_policy succeeded', stdout)
 
     def test_001_strict_fail(self):
-        with self.expect_returncode(2):
+        try:
             self.run_binary(['file_check_policy_strict', 'read', 'unknown_testfile'])
+            self.fail('expected to return nonzero')
+        except subprocess.CalledProcessError as e:
+            self.assertEqual(e.returncode, 2)
+            stderr = e.stderr.decode()
+            self.assertIn('Disallowing access to file \'unknown_testfile\'', stderr)
 
     def test_002_strict_fail_create(self):
-        with self.expect_returncode(2):
+        if os.path.exists('nonexisting_testfile'):
+            os.remove('nonexisting_testfile')
+        try:
             # this tests a previous bug in Graphene that allowed creating unknown files
-            self.run_binary(['file_check_policy_strict', 'append', 'unknown_testfile'])
+            self.run_binary(['file_check_policy_strict', 'append', 'nonexisting_testfile'])
+            self.fail('expected to return nonzero')
+        except subprocess.CalledProcessError as e:
+            self.assertEqual(e.returncode, 2)
+            stderr = e.stderr.decode()
+            self.assertIn('Disallowing access to file \'nonexisting_testfile\'', stderr)
+            if os.path.exists('nonexisting_testfile'):
+                self.fail('test created a file unexpectedly')
 
-    def test_003_allow_all_but_log_unknown(self):
+    def test_003_strict_fail_write(self):
+        try:
+            # writing to trusted files should not be possible
+            self.run_binary(['file_check_policy_strict', 'write', 'trusted_testfile'])
+            self.fail('expected to return nonzero')
+        except subprocess.CalledProcessError as e:
+            self.assertEqual(e.returncode, 2)
+            stderr = e.stderr.decode()
+            self.assertIn('Disallowing create/write/append to a trusted file \'trusted_testfile\'',
+                          stderr)
+
+    def test_004_allow_all_but_log_unknown(self):
         stdout, stderr = self.run_binary(['file_check_policy_allow_all_but_log', 'read',
                                           'unknown_testfile'])
-        self.assertIn('Allowing access to an unknown file due to file_check_policy settings: '
-                      'file:unknown_testfile', stderr)
+        self.assertIn('Allowing access to unknown file \'unknown_testfile\' due to '
+                      'file_check_policy settings.', stderr)
         self.assertIn('file_check_policy succeeded', stdout)
 
-    def test_004_allow_all_but_log_trusted(self):
+    def test_005_allow_all_but_log_trusted(self):
         stdout, stderr = self.run_binary(['file_check_policy_allow_all_but_log', 'read',
                                           'trusted_testfile'])
-        self.assertNotIn('Allowing access to an unknown file due to file_check_policy settings: '
-                         'file:trusted_testfile', stderr)
+        self.assertNotIn('Allowing access to unknown file \'trusted_testfile\' due to '
+                         'file_check_policy settings.', stderr)
         self.assertIn('file_check_policy succeeded', stdout)
 
-    def test_005_allow_all_but_log_trusted_create_fail(self):
-        with self.expect_returncode(2):
+    def test_006_allow_all_but_log_trusted_create_fail(self):
+        try:
             # this fails because modifying trusted files is prohibited
             self.run_binary(['file_check_policy_allow_all_but_log', 'append', 'trusted_testfile'])
+            self.fail('expected to return nonzero')
+        except subprocess.CalledProcessError as e:
+            self.assertEqual(e.returncode, 2)
+            stderr = e.stderr.decode()
+            self.assertIn('Disallowing create/write/append to a trusted file \'trusted_testfile\'',
+                          stderr)
+
+    def test_007_allow_all_but_log_unknown_create(self):
+        if os.path.exists('nonexisting_testfile'):
+            os.remove('nonexisting_testfile')
+        try:
+            stdout, stderr = self.run_binary(['file_check_policy_allow_all_but_log', 'append',
+                                              'nonexisting_testfile'])
+            self.assertIn('Allowing access to unknown file \'nonexisting_testfile\' due to '
+                          'file_check_policy settings.', stderr)
+            self.assertIn('file_check_policy succeeded', stdout)
+            if not os.path.exists('nonexisting_testfile'):
+                self.fail('test did not create a file')
+        finally:
+            os.remove('nonexisting_testfile')
+
 
 @unittest.skipUnless(HAS_SGX,
     'These tests are only meaningful on SGX PAL because only SGX supports attestation.')
--- a/Pal/regression/File.manifest
+++ b/Pal/regression/File.manifest
@@ -6,7 +6,7 @@ fs.mount.root.uri = "file:"
 
 sgx.nonpie_binary = true
 
-sgx.trusted_files.tmp1 = "file:File"
-sgx.trusted_files.tmp2 = "file:../regression/File"
+sgx.allowed_files.tmp1 = "file:File"
+sgx.allowed_files.tmp2 = "file:../regression/File"
 sgx.allowed_files.tmp3 = "file:file_nonexist.tmp"
 sgx.allowed_files.tmp4 = "file:file_delete.tmp"
--- a/Pal/regression/normalize_path.c
+++ b/Pal/regression/normalize_path.c
@@ -54,8 +54,8 @@ static int run_test(void) {
             return 1;
         }
 
-        if (strlen(buf) != size) {
-            print_err(func_name, i, "returned wrong size: %zu\n", size);
+        if (size != strlen(buf) + 1) {
+            print_err(func_name, i, "returned wrong size: %lu\n", size);
             return 1;
         }
 
--- a/Pal/src/host/Linux-SGX/db_files.c
+++ b/Pal/src/host/Linux-SGX/db_files.c
@@ -9,8 +9,11 @@
 #include <asm/stat.h>
 #include <linux/fs.h>
 #include <linux/types.h>
+#include <sys/types.h>
 
 #include "api.h"
+#include "enclave_pf.h"
+#include "enclave_tf.h"
 #include "pal.h"
 #include "pal_defs.h"
 #include "pal_error.h"
@@ -28,125 +31,189 @@
  * GBs in size, and a pread OCALL could fail with -ENOMEM, so we cap to reasonably small size) */
 #define MAX_READ_SIZE (PRESET_PAGESIZE * 1024 * 32)
 
-/* 'open' operation for file streams */
-static int file_open(PAL_HANDLE* handle, const char* type, const char* uri, int access, int share,
-                     int create, int options) {
+static int file_open(PAL_HANDLE* handle, const char* type, const char* uri, int pal_access,
+                     int pal_share, int pal_create, int pal_options) {
+    int ret;
+    int fd = -1;
+    PAL_HANDLE hdl = NULL;
+    bool do_create = (pal_create & PAL_CREATE_ALWAYS) || (pal_create & PAL_CREATE_TRY);
+
+    struct stat st;
+    int flags = PAL_ACCESS_TO_LINUX_OPEN(pal_access) |
+                PAL_CREATE_TO_LINUX_OPEN(pal_create) |
+                PAL_OPTION_TO_LINUX_OPEN(pal_options);
+
     if (strcmp(type, URI_TYPE_FILE))
         return -PAL_ERROR_INVAL;
 
-    /* prepare the file handle */
-    size_t uri_size = strlen(uri) + 1;
-    PAL_HANDLE hdl = calloc(1, HANDLE_SIZE(file) + uri_size);
-    if (!hdl)
+    /* normalize uri into normpath */
+    size_t normpath_size = strlen(uri) + 1;
+    char* normpath = malloc(normpath_size);
+    if (!normpath)
         return -PAL_ERROR_NOMEM;
+
+    ret = get_norm_path(uri, normpath, &normpath_size);
+    if (ret < 0) {
+        log_warning("Could not normalize path (%s): %s", uri, pal_strerror(ret));
+        free(normpath);
+        return -PAL_ERROR_DENIED;
+    }
+
+    /* create file PAL handle with path string placed at the end of this handle object */
+    hdl = calloc(1, HANDLE_SIZE(file) + normpath_size);
+    if (!hdl) {
+        free(normpath);
+        return -PAL_ERROR_NOMEM;
+    }
 
     SET_HANDLE_TYPE(hdl, file);
     HANDLE_HDR(hdl)->flags |= RFD(0) | WFD(0);
-    char* path = (void*)hdl + HANDLE_SIZE(file);
-    int ret;
-    if ((ret = get_norm_path(uri, path, &uri_size)) < 0) {
-        log_warning("Could not normalize path (%s): %s", uri, pal_strerror(ret));
-        free(hdl);
-        return ret;
-    }
-    hdl->file.realpath = (PAL_STR)path;
 
-    struct protected_file* pf = get_protected_file(path);
-    struct stat st;
-    /* whether to re-initialize the PF */
-    bool pf_create = (create & PAL_CREATE_ALWAYS) || (create & PAL_CREATE_TRY);
+    memcpy((char*)hdl + HANDLE_SIZE(file), normpath, normpath_size);
+    hdl->file.realpath = (PAL_STR)hdl + HANDLE_SIZE(file);
 
-    /* try to do the real open */
-    int fd = ocall_open(uri, PAL_ACCESS_TO_LINUX_OPEN(access)  |
-                             PAL_CREATE_TO_LINUX_OPEN(create)  |
-                             PAL_OPTION_TO_LINUX_OPEN(options),
-                        share);
-    if (fd < 0) {
-        ret = unix_to_pal_error(fd);
-        goto out;
+    free(normpath); /* was copied into the file PAL handle object, not needed anymore */
+    normpath = NULL;
+
+    struct protected_file* pf = get_protected_file(hdl->file.realpath);
+    struct trusted_file* tf   = get_trusted_or_allowed_file(hdl->file.realpath);
+
+    if (!pf && !tf && get_file_check_policy() != FILE_CHECK_POLICY_ALLOW_ALL_BUT_LOG) {
+        log_error("Disallowing access to file '%s'; file is not protected, trusted or allowed.",
+                  hdl->file.realpath);
+        ret = -PAL_ERROR_DENIED;
+        goto fail;
     }
 
-    hdl->file.fd = fd;
+    if (!pf && !tf) {
+        log_always("Allowing access to unknown file '%s' due to file_check_policy settings.",
+                   hdl->file.realpath);
 
-    /* check if the file is seekable and get real file size */
-    ret = ocall_fstat(fd, &st);
-    if (ret < 0) {
-        log_warning("file_open(%s): fstat failed: %d", path, ret);
-        ret = unix_to_pal_error(ret);
-        goto out;
+        fd = ocall_open(uri, flags, pal_share);
+        if (fd < 0) {
+            ret = unix_to_pal_error(fd);
+            goto fail;
+        }
+
+        ret = ocall_fstat(fd, &st);
+        if (ret < 0) {
+            ret = unix_to_pal_error(ret);
+            goto fail;
+        }
+
+        hdl->file.fd = fd;
+        hdl->file.seekable = !S_ISFIFO(st.st_mode);
+        hdl->file.total = st.st_size;
+
+        *handle = hdl;
+        return 0;
     }
-
-    hdl->file.seekable = !S_ISFIFO(st.st_mode);
 
     if (pf) {
+        pf_lock();
+
         pf_file_mode_t pf_mode = 0;
-        if ((access & PAL_ACCESS_RDWR) == PAL_ACCESS_RDWR)
+        if (pal_access & PAL_ACCESS_RDWR)
             pf_mode = PF_FILE_MODE_READ | PF_FILE_MODE_WRITE;
-        else if ((access & PAL_ACCESS_WRONLY) == PAL_ACCESS_WRONLY)
+        else if (pal_access & PAL_ACCESS_WRONLY)
             pf_mode = PF_FILE_MODE_WRITE;
         else
             pf_mode = PF_FILE_MODE_READ;
 
-        /* disallow opening more than one writable handle to a PF */
-        if (pf_mode & PF_FILE_MODE_WRITE) {
-            if (pf->writable_fd >= 0) {
-                log_warning("file_open(%s): disallowing concurrent writable handle", path);
-                ret = -PAL_ERROR_DENIED;
-                goto out;
-            }
+        if ((pf_mode & PF_FILE_MODE_WRITE) && pf->writable_fd >= 0) {
+            log_warning("file_open(%s): disallowing concurrent writable handle",
+                        hdl->file.realpath);
+            ret = -PAL_ERROR_DENIED;
+            goto fail_pf_unlock;
         }
 
-        ret = -PAL_ERROR_DENIED;
-
-        /* the protected files should be regular files (seekable) */
-        if (!hdl->file.seekable) {
-            log_warning("file_open(%s): disallowing non-seekable file handle", path);
-            goto out;
+        fd = ocall_open(uri, flags, pal_share);
+        if (fd < 0) {
+            ret = unix_to_pal_error(fd);
+            goto fail_pf_unlock;
         }
 
-        pf = load_protected_file(path, (int*)&hdl->file.fd, st.st_size, pf_mode, pf_create, pf);
-        if (pf) {
-            pf->refcount++;
-            if (pf_mode & PF_FILE_MODE_WRITE) {
-                pf->writable_fd = fd;
-            }
-        } else {
-            log_warning("load_protected_file(%s, %d) failed", path, hdl->file.fd);
-            goto out;
-        }
-    } else {
-        sgx_chunk_hash_t* chunk_hashes;
-        uint64_t total;
-        void* umem;
-        ret = load_trusted_file(hdl, &chunk_hashes, &total, create, &umem);
+        ret = ocall_fstat(fd, &st);
         if (ret < 0) {
-            log_warning("Accessing file:%s is denied (%s). This file is not trusted or allowed."
-                        " Trusted files should be regular files (seekable).", hdl->file.realpath,
-                        pal_strerror(ret));
-            goto out;
+            ret = unix_to_pal_error(ret);
+            goto fail_pf_unlock;
         }
 
-        if (chunk_hashes && total) {
-            assert(umem);
+        hdl->file.fd = fd;
+        hdl->file.seekable = !S_ISFIFO(st.st_mode);
+
+        pf = load_protected_file(hdl->file.realpath, (int*)&hdl->file.fd, st.st_size, pf_mode,
+                                 do_create, pf);
+        if (!pf) {
+            log_warning("load_protected_file(%s, %d) failed", hdl->file.realpath, hdl->file.fd);
+            ret = -PAL_ERROR_DENIED;
+            goto fail_pf_unlock;
         }
 
-        hdl->file.chunk_hashes = (PAL_PTR)chunk_hashes;
-        hdl->file.total = total;
-        hdl->file.umem  = umem;
+        if (pf_mode & PF_FILE_MODE_WRITE) {
+            pf->writable_fd = fd;
+        }
+
+        pf->refcount++;
+        pf_unlock();
+
+        *handle = hdl;
+        return 0;
     }
+
+    assert(tf); /* at this point, we want to open a trusted or allowed file */
+
+    if (!tf->allowed && (do_create
+                         || (pal_access & PAL_ACCESS_RDWR)
+                         || (pal_access & PAL_ACCESS_WRONLY))) {
+        log_error("Disallowing create/write/append to a trusted file '%s'", hdl->file.realpath);
+        ret = -PAL_ERROR_DENIED;
+        goto fail;
+    }
+
+    fd = ocall_open(uri, flags, pal_share);
+    if (fd < 0) {
+        ret = unix_to_pal_error(fd);
+        goto fail;
+    }
+
+    ret = ocall_fstat(fd, &st);
+    if (ret < 0) {
+        ret = unix_to_pal_error(ret);
+        goto fail;
+    }
+
+    hdl->file.fd = fd;
+    hdl->file.seekable = !S_ISFIFO(st.st_mode);
+    hdl->file.total = st.st_size;
+
+    sgx_chunk_hash_t* chunk_hashes;
+    uint64_t total;
+    void* umem;
+    ret = load_trusted_or_allowed_file(tf, hdl, do_create, &chunk_hashes, &total, &umem);
+    if (ret < 0) {
+        log_warning("load_trusted_or_allowed_file(%s, %d) failed", hdl->file.realpath,
+                    hdl->file.fd);
+        goto fail;
+    }
+
+    hdl->file.chunk_hashes = (PAL_PTR)chunk_hashes;
+    hdl->file.total = total;
+    hdl->file.umem  = umem;
 
     *handle = hdl;
-    ret = 0;
+    return 0;
 
-out:
-    if (ret != 0) {
-        if (pf && pf->context && pf->refcount == 0)
-            unload_protected_file(pf);
+fail_pf_unlock:
+    pf_unlock();
+fail:
+    if (pf && pf->context && pf->refcount == 0)
+        unload_protected_file(pf);
 
-        free(hdl);
-        if (fd >= 0)
-            ocall_close(fd);
-    }
+    if (fd >= 0)
+        ocall_close(fd);
+
+    free(hdl);
     return ret;
 }
 
@@ -266,10 +333,12 @@ static int pf_file_close(struct protected_file* pf, PAL_HANDLE handle) {
         return -PAL_ERROR_INVAL;
     }
 
-    pf->refcount--;
-
+    pf_lock();
     if (pf->writable_fd == fd)
         pf->writable_fd = -1;
+    pf_unlock();
+
+    pf->refcount--;
 
     if (pf->refcount == 0)
         return unload_protected_file(pf);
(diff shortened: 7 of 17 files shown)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: 0 of 5 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[LibOS/shim/test/regression/test_libos.py, line 276 at r1](https://reviewable.io/reviews/oscarlab/graphene/2450#-McKXbQ-1C8ItMZwJ2IF:-McKXbQ-1C8ItMZwJ2IG:b-b9z1qp) ([raw file](https://github.com/oscarlab/graphene/blob/c9a3be6e6661daca0041bbe69485b53019aec709/LibOS/shim/test/regression/test_libos.py#L276)):*
> ```Python
>     def test_001_strict_fail(self):
>         with self.expect_returncode(2):
>             self.run_binary(['file_check_policy_strict', 'read', 'unknown_testfile'])
> ```

Could we also assert that the file wasn't created after running the binary?
___
*[LibOS/shim/test/regression/test_libos.py, line 281 at r1](https://reviewable.io/reviews/oscarlab/graphene/2450#-McKY11L5zWlKLVf0N7f:-McKY11L5zWlKLVf0N7g:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/c9a3be6e6661daca0041bbe69485b53019aec709/LibOS/shim/test/regression/test_libos.py#L281)):*
> ```Python
>         with self.expect_returncode(2):
>             # this tests a previous bug in Graphene that allowed creating unknown files
>             self.run_binary(['file_check_policy_strict', 'append', 'unknown_testfile'])
> ```

ditto

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: 0 of 6 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/test/regression/test_libos.py, line 276 at r1](https://reviewable.io/reviews/oscarlab/graphene/2450#-McKXbQ-1C8ItMZwJ2IF:-McO2BHyD84IECvyByy0:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/c9a3be6e6661daca0041bbe69485b53019aec709/LibOS/shim/test/regression/test_libos.py#L276)):*

Could we also assert that the file wasn't created after running the binary?
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/test_libos.py, line 281 at r1](https://reviewable.io/reviews/oscarlab/graphene/2450#-McKY11L5zWlKLVf0N7f:-McO2BeS9hYi6f343yWT:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/c9a3be6e6661daca0041bbe69485b53019aec709/LibOS/shim/test/regression/test_libos.py#L281)):*

ditto
</blockquote></details>

Done.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: 0 of 6 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mcj425C9mrU_56yvuaV:-Mcj425C9mrU_56yvuaW:ba20xk9) (no related file):*
Rebased on top of #2446.

---

mkow (changes requested):
> Reviewed 4 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: 4 of 6 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[LibOS/shim/test/regression/test_libos.py, line 276 at r1](https://reviewable.io/reviews/oscarlab/graphene/2450#-McKXbQ-1C8ItMZwJ2IF:-MgWVZPl0qFbtO81b1zA:b-iixhq0) ([raw file](https://github.com/oscarlab/graphene/blob/c9a3be6e6661daca0041bbe69485b53019aec709/LibOS/shim/test/regression/test_libos.py#L276)):*

Done.
</blockquote></details>

Not done? You added this check only to the other test.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 119 at r2](https://reviewable.io/reviews/oscarlab/graphene/2450#-MgWYnhy6QFiHIYPas8i:-MgWYnhy6QFiHIYPas8j:bdubq3g) ([raw file](https://github.com/oscarlab/graphene/blob/fe0006b2aade49b8ecb1fcb4df53ef3241aeb258/Pal/src/host/Linux-SGX/db_files.c#L119)):*
> ```
> pf->writable_fd >= 0
> ```

Isn't this a race?
___
*[Pal/src/host/Linux-SGX/db_files.c, line 160 at r2](https://reviewable.io/reviews/oscarlab/graphene/2450#-MgW_JBz6x_IVAKfzUln:-MgW_JBz6x_IVAKfzUlo:bwf5k7x) ([raw file](https://github.com/oscarlab/graphene/blob/fe0006b2aade49b8ecb1fcb4df53ef3241aeb258/Pal/src/host/Linux-SGX/db_files.c#L160)):*
> ```C
>     assert(tf); /* at this point, we want to open a trusted or allowed file */
> 
>     if (create && !tf->allowed) {
> ```

`create` is not a bool, but a bitmask (which should actually be an enum)
___
*[Pal/src/host/Linux-SGX/db_files.c, line 162 at r2](https://reviewable.io/reviews/oscarlab/graphene/2450#-MgWZTZ6CesunS_Zrcvy:-MgWZTZ6CesunS_Zrcvz:bl8chhn) ([raw file](https://github.com/oscarlab/graphene/blob/fe0006b2aade49b8ecb1fcb4df53ef3241aeb258/Pal/src/host/Linux-SGX/db_files.c#L162)):*
> ```C
>     if (create && !tf->allowed) {
>         log_error("file_open(%s): disallowing create/write/append on trusted file.\n",
>                 hdl->file.realpath);
> ```

wrong alignment
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 198 at r2](https://reviewable.io/reviews/oscarlab/graphene/2450#-MgWdHpt24Fdvk2Y7msB:-MgWdHpt24Fdvk2Y7msC:b-tsrp6h) ([raw file](https://github.com/oscarlab/graphene/blob/6ba10c60b94a066d516bb9121c31280c0ba69468/Pal/src/host/Linux-SGX/enclave_framework.c#L198)):*
> ```C
>  * file contents. */
> DEFINE_LIST(trusted_file);
> struct trusted_file {
> ```

These declarations were placed in a seemingly random file, but you moved them to just another random file? Could we create a dedicated header for trusted/protected/allowed files handling?

I see that you added a TODO about .c files, but headers can be cleaned up now, without the need to move the implementation. With this PR we'll move all the declarations twice.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: 3 of 12 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[LibOS/shim/test/regression/test_libos.py, line 276 at r1](https://reviewable.io/reviews/oscarlab/graphene/2450#-McKXbQ-1C8ItMZwJ2IF:-Mge3myz9oZoDXmbL4QE:b-yl4dxr) ([raw file](https://github.com/oscarlab/graphene/blob/c9a3be6e6661daca0041bbe69485b53019aec709/LibOS/shim/test/regression/test_libos.py#L276)):*

Not done? You added this check only to the other test.
</blockquote></details>

But we can't check that `unknown_testfile` wasn't created because his file is always there (it is already created on the host). Here it is: https://github.com/oscarlab/graphene/blob/master/LibOS/shim/test/regression/unknown_testfile
___
*[Pal/src/host/Linux-SGX/db_files.c, line 119 at r2](https://reviewable.io/reviews/oscarlab/graphene/2450#-MgWYnhy6QFiHIYPas8i:-MgeBim2AXUBn5S1pcrp:b-wtzrcb) ([raw file](https://github.com/oscarlab/graphene/blob/fe0006b2aade49b8ecb1fcb4df53ef3241aeb258/Pal/src/host/Linux-SGX/db_files.c#L119)):*

> ```
> pf->writable_fd >= 0
> ```

Isn't this a race?
</blockquote></details>

Done. Added one global lock/unlock here; not perfect but shouldn't affect performance that much.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 160 at r2](https://reviewable.io/reviews/oscarlab/graphene/2450#-MgW_JBz6x_IVAKfzUln:-Mge4qI_9QRRg5JCJEcw:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/fe0006b2aade49b8ecb1fcb4df53ef3241aeb258/Pal/src/host/Linux-SGX/db_files.c#L160)):*

`create` is not a bool, but a bitmask (which should actually be an enum)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 162 at r2](https://reviewable.io/reviews/oscarlab/graphene/2450#-MgWZTZ6CesunS_Zrcvy:-Mge4sFw7tn8qti-vXqT:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/fe0006b2aade49b8ecb1fcb4df53ef3241aeb258/Pal/src/host/Linux-SGX/db_files.c#L162)):*

wrong alignment
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 198 at r2](https://reviewable.io/reviews/oscarlab/graphene/2450#-MgWdHpt24Fdvk2Y7msB:-Mge9lIL-V5MR-T5JcgO:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6ba10c60b94a066d516bb9121c31280c0ba69468/Pal/src/host/Linux-SGX/enclave_framework.c#L198)):*

These declarations were placed in a seemingly random file, but you moved them to just another random file? Could we create a dedicated header for trusted/protected/allowed files handling?

I see that you added a TODO about .c files, but headers can be cleaned up now, without the need to move the implementation. With this PR we'll move all the declarations twice.
</blockquote></details>

Done.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: 2 of 12 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2450#-MgpYySR-lDi1F_FLLI9:-MgpYySS0BOJjwtxZSHz:b-5bii6h) (no related file):*
FYI: Rebased to current master.

---
___
*[Pal/src/host/Linux-SGX/db_files.c, line 77 at r3](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mgp3eHE9lbZEmVJMNpH:-MgpZBYQ4JWQL4-VjMjc:b-x58fgu) ([raw file](https://github.com/oscarlab/graphene/blob/2fc74db83c4d76ece54bf8de869b9490326e9821/Pal/src/host/Linux-SGX/db_files.c#L77)):*

I know this is unrelated to this PR, but locking here still looks wrong... I don't see any refcounting on `pf` (it should be done in `get_protected_file()` while it's still holding the lock), right now the file may just disappear in the middle of this function? (e.g. before we do `pf_lock()` below).
</blockquote></details>

Yeah, refcounting is quite broken on Protected Files. I had this PR to at least partially fix it: https://github.com/oscarlab/graphene/pull/2372, but this is also quite stale now.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 338 at r3](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mgp50jpAw_BHCTtAkVl:-MgpZSNv-fEabJma6wwn:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/2fc74db83c4d76ece54bf8de869b9490326e9821/Pal/src/host/Linux-SGX/db_files.c#L338)):*

same problem as above, but here it looks easier to solve - just do this before decrementing the refcount
</blockquote></details>

Done.

mkow (changes requested):
> Reviewed 1 of 9 files at r3, 8 of 9 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/db_files.c, line 77 at r3](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mgp3eHE9lbZEmVJMNpH:-Mgq-O728M3zsC73qAiM:b-p8j4c1) ([raw file](https://github.com/oscarlab/graphene/blob/2fc74db83c4d76ece54bf8de869b9490326e9821/Pal/src/host/Linux-SGX/db_files.c#L77)):*

Yeah, refcounting is quite broken on Protected Files. I had this PR to at least partially fix it: https://github.com/oscarlab/graphene/pull/2372, but this is also quite stale now.
</blockquote></details>

Right, resolving then.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 62 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mgq7jCI4Y0zzvVPw3Lc:-Mgq7jCI4Y0zzvVPw3Ld:bcxpj8c) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L62)):*
> ```C
> 
>     /* create file PAL handle with path string placed at the end of this handle object */
>     hdl = calloc(1, HANDLE_SIZE(file) + normpath_size + 1);
> ```

this +1 is not needed anymore
___
*[Pal/src/host/Linux-SGX/db_files.c, line 71 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mgq7xyNC-s5RSNwYzGA:-Mgq7xyNC-s5RSNwYzGB:b5qc6r1) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L71)):*
> ```C
>     HANDLE_HDR(hdl)->flags |= RFD(0) | WFD(0);
> 
>     memcpy((char*)hdl + HANDLE_SIZE(file), normpath, normpath_size + 1);
> ```

ditto (and this time it's OOB read on heap - we really need ASan in CI :) CC: @pwmarcz)
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 254 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh1-esADtwZyrPIYCTn:-Mh1-esADtwZyrPIYCTo:bib1bd9) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_framework.c#L254)):*
> ```
> if (create)
> ```

This isn't a boolean. This variable (which appears across the whole trusted/allowed files code) is extremely confusing...
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 18 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh12rb57KEhZwyP0EJj:-Mh12rb57KEhZwyP0EJk:bb9xhcg) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L18)):*
> ```objc
>  * Perhaps confusingly, but `struct trusted_file` describes not only "sgx.trusted_files" but also
>  * "sgx.allowed_files". For allowed files, `allowed = true`, `chunk_hashes = NULL`, and `uri` can be
>  * not only a file but also a directory. TODO: Perhaps split "allowed_files" in a separate file.
> ```

`in a separate file` -> `into a separate header` (the original version was a bit confusing for me when reading it the first time)
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 21 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh133-90IySQAYJfIJS:-Mh133-90IySQAYJfIJT:b-m9f4su) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L21)):*
> ```objc
>  */
> 
> /* TODO: Move trusted/allowed files implementation in separate file `enclave_tf.c` */
> ```

`in` -> `into a` + parens around the filename name
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 56 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh14FpC0DtWfa7zOw0-:-Mh14FpC0DtWfa7zOw00:b-timizo) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L56)):*
> ```objc
>     uint64_t size;
>     bool allowed;
>     sgx_file_hash_t file_hash;      /* hash over the whole file, must be the same as in manifest */
> ```

it was here before, but let's fix it now: "must be the same as in manifest" -> "retrieved from the manifest"
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 76 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh1-VfN8pQldJ_gXPsJ:-Mh1-VfN8pQldJ_gXPsK:b-791eli) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L76)):*
> ```
> * \param create            whether this file is newly created
> ```

Sounds like a description of a bool, but this is actually a bitmask/enum.
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 79 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh16GU-0RB9AGpaJFDY:-Mh16GU094ksboEnKStP:bvluix0) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L79)):*
> ```
> file is loaded
> ```

"is" -> "was" (it's a ret-by-arg value)
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 102 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh16pIzCCHhnbjxG6hC:-Mh16pIzCCHhnbjxG6hD:bjkhnya) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L102)):*
> ```objc
>  * \return 0 on success, negative error code on failure
>  *
>  * If needed, regions at either the beginning or the end of the copied regions are copied into a
> ```

This comment is weird, I'd just drop it. Why are beginnings and ends special in any way?

pwmarcz (changes requested):
> Reviewed 2 of 9 files at r3, 8 of 9 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Pal/src/host/Linux-SGX/db_files.c, line 71 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mgq7xyNC-s5RSNwYzGA:-MhDXiDKApqCvTibdZ3s:b-vvzb9a) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L71)):*

ditto (and this time it's OOB read on heap - we really need ASan in CI :) CC: @pwmarcz)
</blockquote></details>

Acknowledged :)
___
*[Pal/src/host/Linux-SGX/db_files.c, line 117 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhDYtpz41phDyvc8-aX:-MhDYtpz41phDyvc8-aY:b75pfio) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L117)):*
> ```C
> 
>         pf_file_mode_t pf_mode = 0;
>         if ((access & PAL_ACCESS_RDWR) == PAL_ACCESS_RDWR)
> ```

This compare looks confusing and unnecessary, and also this is the only place where we do it. Can you change to `if (access & PAL_ACCESS_RDWR)` and `if (access & PAL_ACCESS_WRONLY)`?
___
*[Pal/src/host/Linux-SGX/db_files.c, line 167 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhD_lXO8npcEzy1z6l-:-MhD_lXO8npcEzy1z6l0:bignfw4) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L167)):*
> ```C
>     assert(tf); /* at this point, we want to open a trusted or allowed file */
> 
>     if ((create & PAL_CREATE_ALWAYS || create & PAL_CREATE_TRY) && !tf->allowed) {
> ```

This looks correct wrt precedence, but can we have parentheses around `create & PAL_CREATE_ALWAYS` and `create & PAL_CREATE_TRY` anyway? It would be easier to read.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 167 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhDavYh5hFGmLj0uw8N:-MhDavYh5hFGmLj0uw8O:b-gm4nq7) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L167)):*
>Disallowing create/write/append of trusted file '%s'

But the check seems to be only for create options. If I want to open the file for writing (without `O_CREAT`), it will pass.

And what about this case, actually? The documentation says that trusted files cannot be opened for writing, but I can't find any check for it, and I don't see tests either.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 340 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhDe6vA2RmiPoBRjV3A:-MhDe6vA2RmiPoBRjV3B:bit34sn) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L340)):*
> ```C
>     pf_unlock();
> 
>     pf->refcount--;
> ```

Looks like we sometimes access or even change `pf->refcount` while not under a lock, without using any atomics. Isn't this wrong?

Also, maybe this is a good moment  to introduce `get_protected_file` and `put_protected_file` operations?
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 207 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhDfolD4A_AJzImZ_sZ:-MhDfolD4A_AJzImZ_s_:bv407se) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_framework.c#L207)):*
> ```C
>         return true;
>     }
>     if (tf_path[tf_path_len - 1] == '/' || path[tf_path_len] == '/') {
> ```

This took me a while. How about being a little more explicit?

```c
if (tf_path[tf_path_len] == '/') {
    /* tf path is a subpath of `path` (with slash), e.g. "foo/" and "foo/bar" */
    return true;
}
if (path[tf_path_len] == '/') {
    /* tf path is a subpath of `path` (without slash), e.g. "foo" and "foo/bar" */
    return true;
}
```
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 254 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh1-esADtwZyrPIYCTn:-MhDhO8vAzumgkDezJK8:b-b5u1ox) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_framework.c#L254)):*

> ```
> if (create)
> ```

This isn't a boolean. This variable (which appears across the whole trusted/allowed files code) is extremely confusing...
</blockquote></details>

In protected files it's a boolean, and it looks like here it can be a boolean as well. I would rename the boolean parameters to something like `is_create` though, to distinguish from flags.
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 18 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh12rb57KEhZwyP0EJj:-MhDiziq385H7DEqGDik:bjiu5uw) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L18)):*

`in a separate file` -> `into a separate header` (the original version was a bit confusing for me when reading it the first time)
</blockquote></details>

Or, a separate struct?

mkow (commented):
FYI: @dimakuv is on vacation, I'm taking over this PR (with his explicit approval).

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450#-:-MhJ8DAD-MpEBwQcryzF:b-pj54rf)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

mkow (commented):
> Reviewed 6 of 8 files at r6, 2 of 2 files at r7, 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhTWu2JBoKAebjwjy9M:-MhTWu2JBoKAebjwjy9N:bxddtei) (no related file):*
TODO: update commit message to include all the changes + co-authored-by

---
___
*[Pal/src/host/Linux-SGX/db_files.c, line 62 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mgq7jCI4Y0zzvVPw3Lc:-MhT_s9E2zJjsXIY3iOw:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L62)):*

this +1 is not needed anymore
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 71 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mgq7xyNC-s5RSNwYzGA:-MhT_t7j63fCf5kRjBU0:b-qpwnm9) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L71)):*

Acknowledged :)
</blockquote></details>

Fixed. @pwmarcz: Please verify my recent changes carefully, especially w.r.t off-by-ones :)
___
*[Pal/src/host/Linux-SGX/db_files.c, line 117 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhDYtpz41phDyvc8-aX:-MhT_wAIEa6xyrAw3QYh:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L117)):*

This compare looks confusing and unnecessary, and also this is the only place where we do it. Can you change to `if (access & PAL_ACCESS_RDWR)` and `if (access & PAL_ACCESS_WRONLY)`?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 167 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhD_lXO8npcEzy1z6l-:-MhTa0aAAjuctGQ62exi:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L167)):*

This looks correct wrt precedence, but can we have parentheses around `create & PAL_CREATE_ALWAYS` and `create & PAL_CREATE_TRY` anyway? It would be easier to read.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 167 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhDavYh5hFGmLj0uw8N:-MhTa25A85bJwGivdJWn:b-n40job) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L167)):*

>Disallowing create/write/append of trusted file '%s'

But the check seems to be only for create options. If I want to open the file for writing (without `O_CREAT`), it will pass.

And what about this case, actually? The documentation says that trusted files cannot be opened for writing, but I can't find any check for it, and I don't see tests either.
</blockquote></details>

Done + added the missing test.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 340 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhDe6vA2RmiPoBRjV3A:-MhR0STb4Ez083EaUmhK:bhb1cne) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L340)):*

Looks like we sometimes access or even change `pf->refcount` while not under a lock, without using any atomics. Isn't this wrong?

Also, maybe this is a good moment  to introduce `get_protected_file` and `put_protected_file` operations?
</blockquote></details>

There's a PR from Dmitrii which already tries to fix this problem (#2372), but no one had time so far to review it.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 207 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhDfolD4A_AJzImZ_sZ:-MhR22ETEbGfl5ySVPeT:b-inr51g) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_framework.c#L207)):*

This took me a while. How about being a little more explicit?

```c
if (tf_path[tf_path_len] == '/') {
    /* tf path is a subpath of `path` (with slash), e.g. "foo/" and "foo/bar" */
    return true;
}
if (path[tf_path_len] == '/') {
    /* tf path is a subpath of `path` (without slash), e.g. "foo" and "foo/bar" */
    return true;
}
```
</blockquote></details>

Done (modulo missing `- 1`)
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 254 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh1-esADtwZyrPIYCTn:-MhR2N4n5CM9nPabT4pf:brb7jez) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_framework.c#L254)):*

In protected files it's a boolean, and it looks like here it can be a boolean as well. I would rename the boolean parameters to something like `is_create` though, to distinguish from flags.
</blockquote></details>

I went with leaving `create` here, but changing the bitmask vars to `pal_create`. What do you think about this version?
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 567 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhTb3223NVjMRbYPpLi:-MhTb3223NVjMRbYPpLj:b4xsvgt) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_framework.c#L567)):*
> ```C
>     int ret;
> 
>     const size_t norm_uri_size = URI_MAX;
> ```

FYI: I refactored this function to drop unnecessary `URI_MAX` limitation (I was just going through all `get_norm_path()` callsites).
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 18 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh12rb57KEhZwyP0EJj:-MhTbKHL47Qyk42Q1wEb:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L18)):*

Or, a separate struct?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 21 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh133-90IySQAYJfIJS:-MhTbLBv7Tr-UTsPeSQV:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L21)):*

`in` -> `into a` + parens around the filename name
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 56 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh14FpC0DtWfa7zOw0-:-MhTbNih6AehOGhWQglG:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L56)):*

it was here before, but let's fix it now: "must be the same as in manifest" -> "retrieved from the manifest"
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 76 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh1-VfN8pQldJ_gXPsJ:-MhTbP3Y4oKAZQOq3BwM:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L76)):*

> ```
> * \param create            whether this file is newly created
> ```

Sounds like a description of a bool, but this is actually a bitmask/enum.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 79 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh16GU-0RB9AGpaJFDY:-MhTbPvBDKfzaRH2BZVG:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L79)):*

> ```
> file is loaded
> ```

"is" -> "was" (it's a ret-by-arg value)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_tf.h, line 102 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh16pIzCCHhnbjxG6hC:-MhTbRZx5R00pIFAsrUd:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_tf.h#L102)):*

This comment is weird, I'd just drop it. Why are beginnings and ends special in any way?
</blockquote></details>

Done.

pwmarcz (changes requested):
> Reviewed 6 of 8 files at r6, 1 of 2 files at r7, 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhTjTh5480TP1ONPjQp:-MhTjTh5480TP1ONPjQq:b-xkh5pj) (no related file):*
`C:  1, 0: Too many lines in module (1005/1000) (too-many-lines)`

I suggest disabling this in our `.pylintrc` :)

---
___
*[common/src/path.c, line 144 at r8](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhTg1gbEKypr59Y9yUq:-MhTg1gbEKypr59Y9yUr:b-3ocjfy) ([raw file](https://github.com/oscarlab/graphene/blob/c688636eb55e0cedf4c82eec54956cd88ee58208/common/src/path.c#L144)):*
> ```C
>  * After returning it holds number of bytes actually written to it (including the trailing '\0').
>  */
> int get_base_name(const char* path, char* buf, size_t* inout_size) {
> ```

I can't find any usages of `get_base_name`... Delete?

Looks like I removed last ones in 995dac4b1188daff9a253374782b55e4b6cdbfd5, when I made the code look at parsed path (dentries) instead.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 340 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhDe6vA2RmiPoBRjV3A:-MhTghZA3Oso41TmRTVQ:btmqwi7) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/db_files.c#L340)):*

There's a PR from Dmitrii which already tries to fix this problem (#2372), but no one had time so far to review it.
</blockquote></details>

OK, I can take a look later, perhaps when @dimakuv will be back.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 254 at r5](https://reviewable.io/reviews/oscarlab/graphene/2450#-Mh1-esADtwZyrPIYCTn:-MhThX21AAKUKE1n1TKo:bq9f41m) ([raw file](https://github.com/oscarlab/graphene/blob/dd7dfb1c7eea33c73f1771fbdc4b8c551d394c9c/Pal/src/host/Linux-SGX/enclave_framework.c#L254)):*

I went with leaving `create` here, but changing the bitmask vars to `pal_create`. What do you think about this version?
</blockquote></details>

Looks clear enough.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 576 at r8](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhThqlj6kcxG7sXmFoR:-MhThqlj6kcxG7sXmFoS:b6mn5av) ([raw file](https://github.com/oscarlab/graphene/blob/c688636eb55e0cedf4c82eec54956cd88ee58208/Pal/src/host/Linux-SGX/enclave_framework.c#L576)):*
> ```C
>     }
> 
>     const size_t norm_uri_size = URI_PREFIX_FILE_LEN + strlen(uri) + 1;
> ```

Wait, `uri` already starts with the prefix, aren't you counting it twice?
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 356 at r8](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhTf_2z74Dv9PIF9uIZ:-MhTf_2z74Dv9PIF9uI_:blfb0oo) ([raw file](https://github.com/oscarlab/graphene/blob/c688636eb55e0cedf4c82eec54956cd88ee58208/Pal/src/host/Linux-SGX/enclave_pf.c#L356)):*
> ```C
>         goto out;
> 
>     size_t len = URI_MAX;
> ```

Can you also convert this callsite like you did the others? (Allocate `strlen + 1` bytes, rename variable to `size`).

mkow (commented):
> Reviewed 3 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhTjTh5480TP1ONPjQp:-MhTk_TK35g2q8td-8d8:b-my6kka) (no related file):*

`C:  1, 0: Too many lines in module (1005/1000) (too-many-lines)`

I suggest disabling this in our `.pylintrc` :)
</blockquote></details>

I'm getting more and more annoyed by it :) Done.

---
___
*[common/src/path.c, line 144 at r8](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhTg1gbEKypr59Y9yUq:-MhTkyFP2UYZjHezBQ23:bzg4pen) ([raw file](https://github.com/oscarlab/graphene/blob/c688636eb55e0cedf4c82eec54956cd88ee58208/common/src/path.c#L144)):*

I can't find any usages of `get_base_name`... Delete?

Looks like I removed last ones in 995dac4b1188daff9a253374782b55e4b6cdbfd5, when I made the code look at parsed path (dentries) instead.
</blockquote></details>

Yup, I noticed that it's currently unused, but sounds like something potentially useful and already well tested, so I wanted to leave it so that we won't reimplement it in a moment. But if you think it won't ever be needed then I can remove it.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 576 at r8](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhThqlj6kcxG7sXmFoR:-MhTlf3kER9tN01hA2An:benrmxv) ([raw file](https://github.com/oscarlab/graphene/blob/c688636eb55e0cedf4c82eec54956cd88ee58208/Pal/src/host/Linux-SGX/enclave_framework.c#L576)):*

Wait, `uri` already starts with the prefix, aren't you counting it twice?
</blockquote></details>

Oh, right! Fixed.
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 356 at r8](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhTf_2z74Dv9PIF9uIZ:-MhTq93S1Yt82X6Dmcbk:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/c688636eb55e0cedf4c82eec54956cd88ee58208/Pal/src/host/Linux-SGX/enclave_pf.c#L356)):*

Can you also convert this callsite like you did the others? (Allocate `strlen + 1` bytes, rename variable to `size`).
</blockquote></details>

Done.

pwmarcz (approved):
> Reviewed 3 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[common/src/path.c, line 144 at r8](https://reviewable.io/reviews/oscarlab/graphene/2450#-MhTg1gbEKypr59Y9yUq:-MhTuPIM1rI3yXf9xu_K:b7aovdi) ([raw file](https://github.com/oscarlab/graphene/blob/c688636eb55e0cedf4c82eec54956cd88ee58208/common/src/path.c#L144)):*

Yup, I noticed that it's currently unused, but sounds like something potentially useful and already well tested, so I wanted to leave it so that we won't reimplement it in a moment. But if you think it won't ever be needed then I can remove it.
</blockquote></details>

OK, let's keep it.

mkow (approved):
> Reviewed 1 of 1 files at r10, 1 of 1 files at r11.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: :shipit: complete! all files reviewed, all discussions resolved

pwmarcz (approved):
> Reviewed 1 of 1 files at r10, 1 of 1 files at r11.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2450)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2452, opened 2021-06-16
[Pal/Linux-SGX] Perform `SHA256(g_x || g_y)` for SGX local attestation

dimakuv:
## Description of the changes 

Previously, Graphene bound the identity of the peer enclave to the SHA256 hash of generated session key -- `SHA256(K_e)` during Diffie-Hellman key exchange and SGX local attestation. However, the standard ISO KE protocol (after which we model our Graphene protocol) uses DH public keys instead of the session key -- `SIG_B(g_x, g_y, A)` and `SIG_A(g_y, g_x, B)` (notice the reversed order -- this is to deflect interleaving attacks). 

This PR changes Graphene's protocol implementation to more closely follow the ISO KE protocol.

This PR also adds more documentation on Graphene's protocol details.

UPDATE: Added memory scrubbing and reverse order in the SHA256 hashes to conform to the ISO KE protocol. Rebased as I need to change the commit message as well.

## How to test this PR? 

All tests must pass. You can manually check (via e.g. `strace`) that all IPC is encrypted.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2452)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_process.c
+++ b/Pal/src/host/Linux-SGX/db_process.c
@@ -27,30 +27,69 @@
 #include "protected-files/protected_files.h"
 #include "spinlock.h"
 
-#define LOCAL_ATTESTATION_CHALLENGE_STR "GRAPHENE_LOCAL_ATTESTATION_CHALLENGE"
-
 /*
- * For SGX, the creation of a child process requires a clean enclave and a secure channel
- * between the parent and child processes (enclaves). The establishment of the secure
- * channel must be resilient to a host-level, root-privilege adversary. Such an adversary
- * can either create arbitrary enclaves, or intercept the handshake protocol between the
- * parent and child enclaves to launch a man-in-the-middle attack.
+ * For SGX, the creation of a child process requires a clean enclave and a secure channel between
+ * the parent and child processes (enclaves). The establishment of the secure channel must be
+ * resilient to a host-level, root-privilege adversary. Such an adversary can either create
+ * arbitrary enclaves, or intercept the handshake protocol between the parent and child enclaves to
+ * launch a man-in-the-middle attack.
+ *
+ * The below protocol combines unauthenticated Diffie-Hellman Key Exchange (DHKE) with the SGX local
+ * attestation (this combination provides authenticated DHKE). The protocol is a modified ISO KE
+ * protocol (as described in Section 4 of the "SIGMA: the `SIGn-and-MAc` Approach to Authenticated
+ * Diffie-Hellman and its Use in the IKE Protocols" paper from Hugo Krawczyk). The ISO KE protocol
+ * does not support the "identity protection" feature; this feature is however useless in the SGX
+ * local attestation threat model (malicious host knows identities of all running SGX enclaves
+ * anyway). The ISO KE protocol is proven to be secure and minimal.
+ *
+ * One superficial difference of the below protocol from the original ISO KE protocol is the number
+ * of steps involved. Our protocol uses 5 steps (send DH's g_x, receive DH's g_y, send SGX target
+ * info with identity of enclave A, receive SGX report of enclave B, send SGX report of enclave A).
+ * Note here that SGX report of an enclave contains also the identity of the enclave. The original
+ * ISO KE protocol uses 3 steps (send DH's g_x and identity of A, receive DH's g_y and identity of B
+ * and signature of B over {g_x, g_y, A}, send signature of A over {g_y, g_x, B}). Looking at the
+ * steps, it is clear that our protocol could be optimized to send 3 messages instead of 5 messages
+ * and would result in the original ISO KE protocol. We implement our protocol in 5 steps purely for
+ * readability.
+ *
+ * Another difference of the below protocol from the original ISO KE protocol is the notion of
+ * "identity". In SGX local attestation, identity of an enclave is a combination of its
+ * measurements: mrenclave, attributes, configsvn, etc. This enclave identity is enclosed in the SGX
+ * targetinfo struct as well as in the SGX report. Thus, we use SGX targetinfo and SGX report as
+ * identities.
+ *
+ * One more difference is that instead of calculating SHA256(g_x || g_y) for enclave A and
+ * SHA256(g_y || g_x) for enclave B, our protocol calculates SHA256(K_e || tag1) for enclave A and
+ * SHA256(K_e || tag2) for enclave B. This still introduces assymetry in the protocol (desired to
+ * prevent reflection and interleaving attacks), but is more robust to future modifications.
+ *
+ * Final difference is CMAC used in SGX local attestation for SGX report validation, instead of the
+ * signatures in the original ISO KE protocol. This CMAC however provides the same security
+ * guarantees because it is generated over the SGX report that contains the target identity of the
+ * peer enclave, thus it corresponds to the "directed signature from B to A" in ISO KE.
+ *
+ * Note that Graphene does *not* use the SIGMA-like protocol (used in e.g. Intel SGX SDK). SIGMA
+ * family of protocols is an enhancement over ISO KE to add the "identity protection" feature, which
+ * is irrelevant for SGX local attestation. Thus, Graphene chooses a simpler protocol.
  *
  * Prerequisites of a secure channel:
+ *
  * (1) A session key needs to be shared only between the parent and child enclaves.
  *
- *       See the implementation in _DkStreamKeyExchange().
- *       When initializing an RPC stream, both ends of the stream needs to use
- *       Diffie-Hellman to exchange a session key. The key will be used to both identify
- *       the connection (to prevent man-in-the-middle attack) and for future encryption.
+ *       See the implementation in _DkStreamKeyExchange(). When initializing a secure stream, both
+ *       ends of the stream needs to use Diffie-Hellman to establish a session key. The hashes over
+ *       the established DH session key are used to identify the connection (via SHA256(K_e || tag1)
+ *       for parent enclave A and via SHA256(K_e || tag2) for child enclave B to prevent
+ *       man-in-the-middle/interleaving attacks) as well as to derive other session keys, e.g., to
+ *       encrypt pipes for secure IPC.
  *
  * (2) Both the parent and child enclaves need to be proven by the Intel CPU.
  *
- *       See the implementation in _DkStreamReportRequest() and _DkStreamReportRespond().
- *       The two ends of the RPC stream need to exchange local attestation reports
- *       signed by the Intel CPUs to prove themselves to be running inside enclaves
- *       on the same platform. The local attestation reports contain no secret information
- *       and can be verified cryptographically, and can be sent on an unencrypted channel.
+ *       See the implementation in _DkStreamReportRequest() and _DkStreamReportRespond(). The two
+ *       ends of the stream need to exchange SGX local attestation reports signed by the Intel CPUs
+ *       to prove themselves to be running inside enclaves on the same platform. The local
+ *       attestation reports contain no secret information and can be verified cryptographically,
+ *       and can be sent on an unencrypted channel.
  *
  *       The flow of local attestation is as follows:
  *         - Parent: Send targetinfo(Parent) to Child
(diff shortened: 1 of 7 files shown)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2452)* status: 0 of 5 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mg3Iv6IBkbOWnjiRCo-:-Mg3Iv6IBkbOWnjiRCo0:b-4cj8ez) (no related file):*
> Previously, Graphene bound the identity of the peer enclave to the
> SHA256 hash of generated session key

What's an "enclave identity"?

You also say:

> Thus, we use SGX targetinfo and SGX report as identities.

which seems to clash with the meaning of "identity" you used in the commit message.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mg3KliE4SEwMiqeVJo1:-Mg3KliE4SEwMiqeVJo2:bld2rjj) (no related file):*
Was there any problem with using `hash(exchanged_key)`? I know that leaking `hash(aes_key)` is a bad idea, but instead I'd just tag the key before hashing, so that we won't leak anything useful.

The solution from this PR seems brittle in practice and makes the code more complex - we need to ensure that we grabbed the whole public context from the crypto library and include it in the hash, otherwise it will be completely insecure. E.g. in the future we may add support for random P and G and forget to add both or one of them to the hash here. Everything will continue to work flawlessly, except that it will be MITMable.

---
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 986 at r2](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mg3LVb_9jj6NgcTCvYy:-Mg3LVb_9jj6NgcTCvYz:b1xwmx) ([raw file](https://github.com/oscarlab/graphene/blob/0302591e93e817bff5f5689a661af77f83422e38/Pal/src/host/Linux-SGX/enclave_framework.c#L986)):*
> ```C
> out:
>     if (ret < 0) {
>         memset(key, 0, sizeof(*key)); /* scrub session key on failure */
> ```

This is bad, `memset` can be elided by the compiler. I'd either not scrub at all (as this is a rather questionable mitigation) or do this properly. See `erase_memory()` in protected files.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 993 at r2](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mg3M0CJ9cxUYo0vhRw2:-Mg3M0CJ9cxUYo0vhRw3:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/0302591e93e817bff5f5689a661af77f83422e38/Pal/src/host/Linux-SGX/enclave_framework.c#L993)):*
> ```C
>     memset(&secret, 0, sizeof(secret));
>     memset(&my_public, 0, sizeof(my_public));
>     memset(&peer_public, 0, sizeof(peer_public));
> ```

ditto

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2452)* status: 0 of 7 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mg3Iv6IBkbOWnjiRCo-:-Mg4dcKk8r2iy9QE16ML:b-sa9h6v) (no related file):*

> Previously, Graphene bound the identity of the peer enclave to the
> SHA256 hash of generated session key

What's an "enclave identity"?

You also say:

> Thus, we use SGX targetinfo and SGX report as identities.

which seems to clash with the meaning of "identity" you used in the commit message.

</blockquote></details>

No, there is no clash. But I guess the wording "bound the identity of the peer enclave to the SHA256 hash of generated session key" is not very clear.

Enclave identity = SGX targetinfo or SGX report (both these structs contain MRENCLAVE + ATTRIBUTES + couplemorefields which constitute the enclave identity).

Now what I meant to say by "bound the identity of the peer enclave to the SHA256 hash of generated session key". Graphene needs to tie together the DH session and the SGX local attestation flow. For this, Graphene needs to put some collaterals from the DH session into the SGX reports. Previously, Graphene put the "SHA256 hash of generated session key" DH collateral into the SGX report. With this commit, Graphene puts the "SHA256 hash of the concatenation of public components" DH collateral into the SGX report.

I can rephrase "Graphene bound the identity of the peer enclave to the SHA256 hash of generated session key" to something like this:
```
To tie together the DH session and the SGX local attestation flow,
Graphene needs to put some collaterals from the DH session into
the SGX reports. Previously, Graphene put the SHA256 hash of
generated session key -- `SHA256(K_e)` -- as a DH collateral in both
SGX reports (from enclave A to enclave B, and vice versa).

 However, the standard ISO KE protocol ...
```

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mg3KliE4SEwMiqeVJo1:-Mg4hPR80mzQkb9fj_D7:bso4b8i) (no related file):*
> Was there any problem with using `hash(exchanged_key)`? I know that leaking `hash(aes_key)` is a bad idea, but instead I'd just tag the key before hashing, so that we won't leak anything useful.

The real issue is the symmetry of the protocol if both parties would use `hash(exchanged_key || tag)` (i.e., even with some tag added). This opens the door for reflection attacks and interleaving attacks. I am not aware of any particular attack on just using `hash(exchanged_key || tag)` but the general idea would be that the attacker could just take this `hash(exchanged_key || tag)` and replay it back to the good enclave. When we add the assemtry `SHA256(g_x || g_y)` for enclave A, and `SHA256(g_y || g_x)` for enclave B, we force the generation of a hash each time anew (the attacker cannot re-use the received hash).

Btw, I don't know whether the following scheme is logically equivalent: `hash(exchanged_key || tag1)` for enclave A and `hash(exchanged_key || tag2)` for enclave B. This seems to add assymetry, still tie into the DH collaterals, and precludes the attacker of re-using the hash as-is. But in this case, there must be a specific ordering: enclave A must know it should use `tag1` and enclave B must know it should use `tag2`. Anyway, this seems like a weird construct, and I've never seen crypto algorithms use something like this.

This is what the https://people.scs.carleton.ca/~paulv/papers/sts-final.pdf paper says (page 18):
> 2. Asymmetry in a protocol is desirable.
Symmetries in a protocol should be used with caution, due to both the possibility of
reflection attacks, and attacks in which responses from one party can be re-used within a
protocol.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mg4m4hZ5biSJ-HPYivC:-Mg4m4h_0m2pWaffWLBl:bjkjery) (no related file):*
> The solution from this PR seems brittle in practice and makes the code more complex - we need to ensure that we grabbed the whole public context from the crypto library and include it in the hash, otherwise it will be completely insecure. E.g. in the future we may add support for random P and G and forget to add both or one of them to the hash here.

I don't understand your point. We may add support for random P and G? Are you refering to `DH_SIZE` and that it can be different for different MODP groups?

I agree that this PR's solution is more complex, but we need to add assymetry in the protocol. I was advised to add it by crypto reviewers, and I also believe it makes the protocol more secure.

---
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 986 at r2](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mg3LVb_9jj6NgcTCvYy:-Mg4q26s0a2Oqnwu7AyW:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/0302591e93e817bff5f5689a661af77f83422e38/Pal/src/host/Linux-SGX/enclave_framework.c#L986)):*

This is bad, `memset` can be elided by the compiler. I'd either not scrub at all (as this is a rather questionable mitigation) or do this properly. See `erase_memory()` in protected files.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 993 at r2](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mg3M0CJ9cxUYo0vhRw2:-Mg4q2_0E5VNuCl0tnFE:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/0302591e93e817bff5f5689a661af77f83422e38/Pal/src/host/Linux-SGX/enclave_framework.c#L993)):*

ditto
</blockquote></details>

Done.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2452)* status: 0 of 7 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mg3Iv6IBkbOWnjiRCo-:-MiWdLJl8mtyCNwIKUft:b70rn9m) (no related file):*

The new message proposal looks better.

btw. "collaterals" used in this context sounds weird, is this actually the term people use for this thing?
</blockquote></details>

Done. As discussed offline, reworded to not have "collaterals" at all.

Had to rebase to incorporate this change to the commit message.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mg3KliE4SEwMiqeVJo1:-MiWdXDz33ntHCI6kDeO:b-896fix) (no related file):*

> But in this case, there must be a specific ordering: enclave A must know it should use `tag1` and enclave B must know it should use `tag2`.

In our code this would be quite trivial: `_DkProcessCreate` would call `hash_session_key()` with tag1, and `init_child_process()` would call `hash_session_key()` with tag2. It adds asymmetry without the risks of missing something from the hash.
</blockquote></details>

Done.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mg4m4hZ5biSJ-HPYivC:-MiWdZaH6iYilg52MRm2:b-896fix) (no related file):*

> We may add support for random P and G? Are you refering to DH_SIZE and that it can be different for different MODP groups?

I mean literally any change to the DH we do, e.g. allowing different P and G constants (could be random or predefined, and sent together with the exchange). Or, if we allow more DH sizes than only the currently hardcoded one (e.g. at some point we may want to increase the DH size, but have some grace period when we support both).

All of them will work, but will silently turn the protocol into a completely insecure construction (as the binding won't be full anymore). I.e. `hash_over_public_dh_components()` may miss some components in the hash, and there's no easy way to detect this with regression tests. And there's no such problem if we hash the final DH result like we do now.
</blockquote></details>

Done.

---

mkow (changes requested):
> Reviewed 2 of 5 files at r1, 2 of 3 files at r3, 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2452)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 914 at r4](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mi_wfXK5k4BwomNuEm5:-Mi_wfXLCztBKCmO1ocY:b-9qb7co) ([raw file](https://github.com/oscarlab/graphene/blob/b7070e39f377c2ba970fe9abe088711a7836986d/Pal/src/host/Linux-SGX/enclave_framework.c#L914)):*
> ```C
> 
>     int64_t total;
>     int64_t bytes;
> ```

These are used only inside the loops, could you define them there? This way it's obvious that they aren't used outside them.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 941 at r4](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mia0l8FDzoCiYt-8VxS:-Mia0l8FDzoCiYt-8VxT:be5syxw) ([raw file](https://github.com/oscarlab/graphene/blob/b7070e39f377c2ba970fe9abe088711a7836986d/Pal/src/host/Linux-SGX/enclave_framework.c#L941)):*
> ```C
> 
>     for (bytes = 0, total = 0; total < (int64_t)sizeof(peer_public); total += bytes) {
>         bytes = _DkStreamRead(stream, 0, sizeof(peer_public) - total, peer_public + total, NULL, 0);
> ```

It was like that before, but isn't this missing the handling of EOF?
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 983 at r4](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mia1xYtCT-Jvp4Ng-cK:-Mia1xYtCT-Jvp4Ng-cL:be0cz36) ([raw file](https://github.com/oscarlab/graphene/blob/b7070e39f377c2ba970fe9abe088711a7836986d/Pal/src/host/Linux-SGX/enclave_framework.c#L983)):*
> ```C
> out:
>     if (ret < 0) {
>         erase_memory(out_key, sizeof(*out_key)); /* scrub session key on failure */
> ```

Why? On failure this key won't be used for anything.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2452)* status: 6 of 7 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2452#-Miafg_V6JRCnBT_1-ST:-Miafg_V6JRCnBT_1-SU:b-qfogp) (no related file):*
FYI: Two Sanitizers Jenkins pipelines failed with this error:
```
ERROR: /var/lib/jenkins/workspace/graphene-direct-sanitizers@script/.ci/linux-direct-sanitizers.jenkinsfile not found
```

This is because this PR is not yet rebased to the latest master and thus doesn't have the corresponding Jenkinsfiles. Ignore these two failures.

---
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 914 at r4](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mi_wfXK5k4BwomNuEm5:-MiaSXmD6OFyjb9bW_VZ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b7070e39f377c2ba970fe9abe088711a7836986d/Pal/src/host/Linux-SGX/enclave_framework.c#L914)):*

These are used only inside the loops, could you define them there? This way it's obvious that they aren't used outside them.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 941 at r4](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mia0l8FDzoCiYt-8VxS:-MiaUTP03_IO-zskhewX:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b7070e39f377c2ba970fe9abe088711a7836986d/Pal/src/host/Linux-SGX/enclave_framework.c#L941)):*

It was like that before, but isn't this missing the handling of EOF?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 983 at r4](https://reviewable.io/reviews/oscarlab/graphene/2452#-Mia1xYtCT-Jvp4Ng-cK:-MiaUdtU8bIXMHLVvKLL:b35wj2g) ([raw file](https://github.com/oscarlab/graphene/blob/b7070e39f377c2ba970fe9abe088711a7836986d/Pal/src/host/Linux-SGX/enclave_framework.c#L983)):*

Why? On failure this key won't be used for anything.
</blockquote></details>

Done. Good point.

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2452)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (approved):
> Reviewed 6 of 6 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2452)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv:
Jenkins, retest Jenkins-SGX-18.04-apps please (Jenkins got stuck, probably due to Woju's unrelated changes to Jenkins)

PR #2463, opened 2021-06-21
Update building.rst

codingwithblackcoffee:
## Description of the changes 
The default path , Pal/src/host/Linux-SGX/signer/enclave-key.pem, was not mentioned in the command, hence not generating the key in the desired location.

## How to test this PR? 

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2463)

### DIFF
--- a/Documentation/building.rst
+++ b/Documentation/building.rst
@@ -151,7 +151,7 @@ versions of the SDK and driver, download and install them from:
 A 3072-bit RSA private key (PEM format) is required for signing the manifest.
 If you don't have a private key, create it with the following command::
 
-   openssl genrsa -3 -out enclave-key.pem 3072
+   openssl genrsa -3 -out Pal/src/host/Linux-SGX/signer/enclave-key.pem/enclave-key.pem 3072
 
 You can either place the generated enclave key in the default path,
 :file:`Pal/src/host/Linux-SGX/signer/enclave-key.pem`, or specify the key's


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2463)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @codingwithblackcoffee)

___
*[Documentation/building.rst, line 157 at r1](https://reviewable.io/reviews/oscarlab/graphene/2463#-MchYKlvBHveOIJEJwEt:-MchYKlvBHveOIJEJwEu:bqeodlc) ([raw file](https://github.com/oscarlab/graphene/blob/64e0cf03bcc040b30b987a816f1cb458b50a5b86/Documentation/building.rst#L157)):*
> ```reStructuredText
> 
> You can either place the generated enclave key in the default path,
> :file:`Pal/src/host/Linux-SGX/signer/enclave-key.pem`, or specify the key's
> ```

But we already mention this path here. So I'm not sure we need your proposed change.

To give more context: the idea is that the developer would ideally create a key under his/her own directory, and then specify its path to Graphene via `SGX_SIGNER_KEY`. Putting the key under `Pal/src/host/Linux-SGX/signer/` is more of a quick hack to make Graphene work. This is not a good idea for production usages.

So I would close this PR as it doesn't seem to add value.

Issue #2419, opened 2021-06-03
Stress-ng test suite causing memory fault and heap corruption failure in Graphene


jinengandhi-intel:
## Description of the problem
While trying to enable the different stressors with Graphene I am seeing some failures Memory fault, Heap corruption and other failures from Graphene. 
 
Failure 1: Testing procfs.
 
Run 1:
$ graphene-direct stress-ng --procfs 8 --timeout 40s
error: Using insecure argv source. Graphene will continue application execution, but this configuration must not be used in production!
error: Forwarding host environment variables to the app is enabled. Graphene will continue application execution, but this configuration must not be used in production!
stress-ng: info:  [1] dispatching hogs: 8 procfs
stress-ng: error: [1] glob on regex "/sys/devices/system/cpu/cpu0/cache/index[0-9]*" failed: 1
stress-ng: info:  [1] cache allocate: using built-in defaults as unable to determine cache details
[P15695:T2:stress-ng] error: Internal memory fault at 0x100000013 (IP = +0x2d234, VMID = 15695, TID = 2)
[P15697:i1:stress-ng] error: IPC worker: unexpected event (4) on exit handle
[P15711:i1:stress-ng] error: Internal memory fault at 0x100000014 (IP = +0x39e67, VMID = 15711, TID = 0)
[P15718:i1:stress-ng] error: Internal memory fault at 0x100000014 (IP = +0x39e67, VMID = 15718, TID = 0)
[P15725:i1:stress-ng] error: Internal memory fault at 0x100000014 (IP = +0x39e67, VMID = 15725, TID = 0)
[P15732:i1:stress-ng] error: Internal memory fault at 0x100000014 (IP = +0x39e67, VMID = 15732, TID = 0)
[P15739:i1:stress-ng] error: Internal memory fault at 0x100000014 (IP = +0x39e67, VMID = 15739, TID = 0)
stress-ng: info:  [1] unsuccessful run completed in 40.06s
 
Run 2:
$ graphene-direct stress-ng --procfs 8 --timeout 40s
error: Using insecure argv source. Graphene will continue application execution, but this configuration must not be used in production!
error: Forwarding host environment variables to the app is enabled. Graphene will continue application execution, but this configuration must not be used in production!
stress-ng: info:  [1] dispatching hogs: 8 procfs
stress-ng: error: [1] glob on regex "/sys/devices/system/cpu/cpu0/cache/index[0-9]*" failed: 1
[...]

PR #2498, opened 2021-07-02
[LibOS] Fix poll() crashing on pseudo-files

pwmarcz:
Found when investigating `stress-ng` (#2419).

## Description of the changes 

The code assumed that all handles (other than TYPE_FILE and TYPE_DEV) had an associated PAL handle, which caused a crash when polling files from /proc. This change adds support for polling TYPE_STR, and also adds a check for PAL handle in all cases.

(I think poll deserves a bigger rewrite, this is just a quick fix).

## How to test this PR? 

* The procfs stress test (`stress-ng --procfs 8 --timeout 40s`) stopped crashing for me (see #2419).
* I rewrote `poll` regression test to be more thorough, it checks several types of files now.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2498)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -706,5 +706,6 @@ ssize_t str_write(struct shim_handle* hdl, const void* buf, size_t count);
 off_t str_seek(struct shim_handle* hdl, off_t offset, int whence);
 int str_flush(struct shim_handle* hdl);
 int str_truncate(struct shim_handle* hdl, off_t len);
+off_t str_poll(struct shim_handle* hdl, int poll_type);
 
 #endif /* _SHIM_FS_H_ */
--- a/LibOS/shim/src/fs/shim_fs_pseudo.c
+++ b/LibOS/shim/src/fs/shim_fs_pseudo.c
@@ -442,18 +442,16 @@ static int pseudo_close(struct shim_handle* hdl) {
     }
 }
 
-/* TODO: add support for polling TYPE_STR handles; currently `shim_do_poll` doesn't call this for
- * anything else than TYPE_STR and TYPE_DEV */
 static off_t pseudo_poll(struct shim_handle* hdl, int poll_type) {
-    if (poll_type == FS_POLL_SZ)
-        return 0;
-
     assert(hdl->dentry);
     struct pseudo_node* node = pseudo_find(hdl->dentry);
     if (!node)
         return -ENOENT;
     switch (node->type) {
         case PSEUDO_DEV: {
+            if (poll_type == FS_POLL_SZ)
+                return 0;
+
             off_t ret = 0;
             if ((poll_type & FS_POLL_RD) && node->dev.dev_ops.read)
                 ret |= FS_POLL_RD;
@@ -461,6 +459,9 @@ static off_t pseudo_poll(struct shim_handle* hdl, int poll_type) {
                 ret |= FS_POLL_WR;
             return ret;
         }
+        case PSEUDO_STR:
+            return str_poll(hdl, poll_type);
+
         default:
             return -ENOSYS;
     }
--- a/LibOS/shim/src/fs/str/fs.c
+++ b/LibOS/shim/src/fs/str/fs.c
@@ -258,6 +258,30 @@ int str_truncate(struct shim_handle* hdl, off_t len) {
     return ret;
 }
 
+off_t str_poll(struct shim_handle* hdl, int poll_type) {
+    assert(hdl->type == TYPE_STR);
+
+    struct shim_str_handle* strhdl = &hdl->info.str;
+    struct shim_str_data* data = strhdl->data;
+    assert(data);
+
+    if (poll_type == FS_POLL_SZ)
+        return data->len;
+
+    off_t ret = 0;
+    if (poll_type & FS_POLL_RD) {
+        if (data->len > 0) {
+            assert(data->str);
+            if (!strhdl->ptr || strhdl->ptr < (data->str + data->len))
+                ret |= FS_POLL_RD;
+        }
+    }
+    if (poll_type & FS_POLL_WR)
+        ret |= FS_POLL_WR;
+
+    return ret;
+}
+
 struct shim_fs_ops str_fs_ops = {
     .close    = &str_close,
     .read     = &str_read,
@@ -265,6 +289,7 @@ struct shim_fs_ops str_fs_ops = {
     .seek     = &str_seek,
     .flush    = &str_flush,
     .truncate = &str_truncate,
+    .poll     = &str_poll,
 };
 
 struct shim_d_ops str_d_ops = {
--- a/LibOS/shim/src/sys/shim_poll.c
+++ b/LibOS/shim/src/sys/shim_poll.c
@@ -102,9 +102,12 @@ static long _shim_do_poll(struct pollfd* fds, nfds_t nfds, int timeout_ms) {
             continue;
         }
 
-        if (hdl->type == TYPE_FILE || hdl->type == TYPE_DEV) {
-            /* Files and devs are special cases: their poll is emulated at LibOS level; do not
-             * include them in handles-to-poll array but instead use handle-specific callback. */
+        if (hdl->type == TYPE_FILE || hdl->type == TYPE_DEV || hdl->type == TYPE_STR) {
+            /* Files, devs and strings are special cases: their poll is emulated at LibOS level; do
+             * not include them in handles-to-poll array but instead use handle-specific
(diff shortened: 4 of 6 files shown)


dimakuv (commented):
> Reviewed 2 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2498)* status: 2 of 6 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (changes requested):
> Reviewed 4 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2498)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 446 at r1](https://reviewable.io/reviews/oscarlab/graphene/2498#-MdpeJw_9nrOPJDwCCxj:-MdpeJw_9nrOPJDwCCxk:bb93wac) ([raw file](https://github.com/oscarlab/graphene/blob/c25a0f4074254e4179895573f5f49a28871823c7/LibOS/shim/src/fs/shim_fs_pseudo.c#L446)):*
> ```C
> 
> /* TODO: add support for polling TYPE_STR handles; currently `shim_do_poll` doesn't call this for
>  * anything else than TYPE_STR and TYPE_DEV */
> ```

Btw, this comment was wrong? I see that `shim_do_poll()` calls `hdl->fs->fs_ops->poll()` on TYPE_FILE and TYPE_DEV. There was no mention of TYPE_STR. 
___
*[LibOS/shim/src/fs/str/fs.c, line 280 at r1](https://reviewable.io/reviews/oscarlab/graphene/2498#-MdplGVo2mBDiKZISSgV:-MdplGVo2mBDiKZISSgW:bclqx8o) ([raw file](https://github.com/oscarlab/graphene/blob/13106ef9b9ab447338ca783e043cf4664040e65b/LibOS/shim/src/fs/str/fs.c#L280)):*
> ```C
>     }
>     if (poll_type & FS_POLL_WR)
>         ret |= FS_POLL_WR;
> ```

Can we really always write in a pseudo-str-file? Some files are read-only (e.g. `/dev/attestation/report`), don't we want to check whether the permissions allow writing?
___
*[LibOS/shim/src/sys/shim_poll.c, line 110 at r1](https://reviewable.io/reviews/oscarlab/graphene/2498#-Mdplr_h2gfN07E7BkcG:-Mdplr_h2gfN07E7BkcH:b-synayz) ([raw file](https://github.com/oscarlab/graphene/blob/13106ef9b9ab447338ca783e043cf4664040e65b/LibOS/shim/src/sys/shim_poll.c#L110)):*
> ```C
>              * callback.
>              *
>              * TODO: we probably should use the poll() callback in all cases. */
> ```

What do you mean by this comment?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2498)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

___
*[LibOS/shim/src/fs/shim_fs_pseudo.c, line 446 at r1](https://reviewable.io/reviews/oscarlab/graphene/2498#-MdpeJw_9nrOPJDwCCxj:-Mdq5KxC7Z4hmlEW-omC:b-bfitf1) ([raw file](https://github.com/oscarlab/graphene/blob/c25a0f4074254e4179895573f5f49a28871823c7/LibOS/shim/src/fs/shim_fs_pseudo.c#L446)):*

Btw, this comment was wrong? I see that `shim_do_poll()` calls `hdl->fs->fs_ops->poll()` on TYPE_FILE and TYPE_DEV. There was no mention of TYPE_STR. 
</blockquote></details>

Yes, that was supposed to be "anything else than `TYPE_FILE` and `TYPE_DEV`".
___
*[LibOS/shim/src/fs/str/fs.c, line 280 at r1](https://reviewable.io/reviews/oscarlab/graphene/2498#-MdplGVo2mBDiKZISSgV:-Mdq5YDUDOB1sd-qcLdE:bosbg9u) ([raw file](https://github.com/oscarlab/graphene/blob/13106ef9b9ab447338ca783e043cf4664040e65b/LibOS/shim/src/fs/str/fs.c#L280)):*

Can we really always write in a pseudo-str-file? Some files are read-only (e.g. `/dev/attestation/report`), don't we want to check whether the permissions allow writing?
</blockquote></details>

We check file permissions when setting handle mode, and the caller of `poll` callback (such as `shim_do_poll`) checks the handle mode. I don't think the `poll` callback itself should do that - the check is the same regardless of handle.

(This does allow a loophole where a user `chmod`s the file and then writes to it. I just checked and Linux allows you to `chmod` a file like `/proc/meminfo` to be writable, and even open it, but fails with `EIO` on the actual `write` call... I don't want to bother with replicating that now, but I'm planning to rewrite the `str_*` filesystem so maybe that will be the time.)
___
*[LibOS/shim/src/sys/shim_poll.c, line 110 at r1](https://reviewable.io/reviews/oscarlab/graphene/2498#-Mdplr_h2gfN07E7BkcG:-Mdq8JTBBXJL8uv91yYa:b-njehb1) ([raw file](https://github.com/oscarlab/graphene/blob/13106ef9b9ab447338ca783e043cf4664040e65b/LibOS/shim/src/sys/shim_poll.c#L110)):*

What do you mean by this comment?
</blockquote></details>

This function collects PAL handles and uses `DkStreamsWaitEvents` on them, disregarding the fact that most of filesystems actually have a `poll` callback defined. It calls the `poll` callback only for some specific handle types. That seems weird to me: what are the other `poll` callbacks even for?

I don't want to touch that right now, though, it looks like bigger change and I would need to understand our implementation of poll/epoll/select better. Right now I want to fix this one thing with STR handles.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2498)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[LibOS/shim/src/fs/str/fs.c, line 280 at r1](https://reviewable.io/reviews/oscarlab/graphene/2498#-MdplGVo2mBDiKZISSgV:-MdqB3EHD6iJFvJODWP2:bs4ixp3) ([raw file](https://github.com/oscarlab/graphene/blob/13106ef9b9ab447338ca783e043cf4664040e65b/LibOS/shim/src/fs/str/fs.c#L280)):*

We check file permissions when setting handle mode, and the caller of `poll` callback (such as `shim_do_poll`) checks the handle mode. I don't think the `poll` callback itself should do that - the check is the same regardless of handle.

(This does allow a loophole where a user `chmod`s the file and then writes to it. I just checked and Linux allows you to `chmod` a file like `/proc/meminfo` to be writable, and even open it, but fails with `EIO` on the actual `write` call... I don't want to bother with replicating that now, but I'm planning to rewrite the `str_*` filesystem so maybe that will be the time.)
</blockquote></details>

OK, fair enough.
___
*[LibOS/shim/src/sys/shim_poll.c, line 110 at r1](https://reviewable.io/reviews/oscarlab/graphene/2498#-Mdplr_h2gfN07E7BkcG:-MdqBAp0BUFBb8MfL_Rb:b-4qgiru) ([raw file](https://github.com/oscarlab/graphene/blob/13106ef9b9ab447338ca783e043cf4664040e65b/LibOS/shim/src/sys/shim_poll.c#L110)):*

This function collects PAL handles and uses `DkStreamsWaitEvents` on them, disregarding the fact that most of filesystems actually have a `poll` callback defined. It calls the `poll` callback only for some specific handle types. That seems weird to me: what are the other `poll` callbacks even for?

I don't want to touch that right now, though, it looks like bigger change and I would need to understand our implementation of poll/epoll/select better. Right now I want to fix this one thing with STR handles.
</blockquote></details>

OK. The idea of a `poll` callback was to have a fall-back for those handle types that do not have a corresonding PAL handle (= emulated completely inside LibOS, without the help of the host OS). But let's not touch this now.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2498)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2498)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2499, opened 2021-07-02
[LibOS] Do not reuse dentries after removing a file (fixes mknod crash)

pwmarcz:
Found while investigating #2419.

## Description of the changes 

This change ensures that dentries for removed files are not reused. Instead of marking a dentry as negative, we mark it as invalid, and make sure that a new one is created during next lookup.

This fixes a crash where we create a named pipe, remove it, and then reuse the same object for a normal file. More generally, it ensures that filesystems always receive a dentry in a consistent state.

Several unused dentry flags are also removed.

## How to test this PR? 

* Should fix the mknod crash in #2419 
   * Note that the mknod stress test still fails, because our mknod doesn't create sockets, and hit EMFILE from hst on creating too many named pipes). However, it should not crash anymore.
* I added a short snippet to the end of `mkfifo` that crashes in same manner on master.
* To verify the operations on dentries (e.g. an entry is invalidated and removed later), I added `dump_dcache(NULL);` to `shim_exit.c`, and ran a few sanity checks with `graphene-direct python -c "<script>"`:
  * `import os; open('x', 'w').close(); os.unlink('x')` (should show an invalid dentry)
  * `...; os.path.exists('x')` (should show a new, negative dentry, with no mode etc. set)
  * `...; os.readdir(".")` (should show no dentry, due to lookup triggering `dentry_gc()`)

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2499)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -169,10 +169,7 @@ struct shim_d_ops {
     /* this is to check file type and access, returning the stat.st_mode */
     int (*mode)(struct shim_dentry* dent, mode_t* mode);
 
-    /*
-     * Detach internal data from dentry. Called before the dentry is destroyed.
-     * TODO: some filesystems (`str`, `tmpfs`) abuse this callback for reference counting.
-     */
+    /* detach internal data from dentry */
     int (*dput)(struct shim_dentry* dent);
 
     /* create a dentry inside a directory */
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -169,7 +169,10 @@ struct shim_d_ops {
     /* this is to check file type and access, returning the stat.st_mode */
     int (*mode)(struct shim_dentry* dent, mode_t* mode);
 
-    /* detach internal data from dentry */
+    /*
+     * Detach internal data from dentry. Called before the dentry is destroyed.
+     * TODO: some filesystems (`str`, `tmpfs`) abuse this callback for reference counting.
+     */
     int (*dput)(struct shim_dentry* dent);
 
     /* create a dentry inside a directory */
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -91,14 +91,9 @@ struct shim_fs_ops {
 
 #define DENTRY_VALID       0x0001 /* this dentry is verified to be valid */
 #define DENTRY_NEGATIVE    0x0002 /* recently deleted or inaccessible */
-#define DENTRY_PERSIST     0x0008 /* added as a persistent dentry */
 #define DENTRY_ISLINK      0x0080 /* this dentry is a link */
 #define DENTRY_ISDIRECTORY 0x0100 /* this dentry is a directory */
-#define DENTRY_LOCKED      0x0200 /* locked by mountpoints at children */
 /* These flags are not used */
-//#define DENTRY_REACHABLE    0x0400  /* permission checked to be reachable */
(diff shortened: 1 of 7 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2499)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/src/sys/shim_file.c, line 39 at r1](https://reviewable.io/reviews/oscarlab/graphene/2499#-MdpqLPgEQCRp0ER004n:-MdpqLPgEQCRp0ER004o:bu8pb2j) ([raw file](https://github.com/oscarlab/graphene/blob/824ab13178d45b15b091b516094795092d9a6f00/LibOS/shim/src/sys/shim_file.c#L39)):*
> ```C
>     struct shim_dentry* dir = NULL;
>     struct shim_dentry* dent = NULL;
>     int ret = 0;
> ```

Can we remove `= 0` now, so that the compiler may catch errors on unassigned-ret paths?
___
*[LibOS/shim/test/regression/mkfifo.c, line 112 at r1](https://reviewable.io/reviews/oscarlab/graphene/2499#-MdpqqP42luOXOiVCe4W:-MdpqqP42luOXOiVCe4X:b-pu01n4) ([raw file](https://github.com/oscarlab/graphene/blob/824ab13178d45b15b091b516094795092d9a6f00/LibOS/shim/test/regression/mkfifo.c#L112)):*
> ```C
>         }
>     }
> ```

Can we add `TEST OK` here, just for sanity?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2499)* status: 0 of 7 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2499#-MdqOjC721zSmJjHw5ll:-MdqOjC721zSmJjHw5lm:b-4a1aqd) (no related file):*
Update: Unfortunately the FS code doesn't seem to be ready for this fix :( It conflicts with reference counting in strfs/tmpfs.

I could proceed without adding `dput`, but it will make the memory leak (when deleting a dentry) worse, as we'll keep creating and destroying new dentries.

I will prepare a quick fix for this issue with mknod specifically.

---
___
*[LibOS/shim/src/sys/shim_file.c, line 39 at r1](https://reviewable.io/reviews/oscarlab/graphene/2499#-MdpqLPgEQCRp0ER004n:-MdqLfcrDKoqu039lHam:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/824ab13178d45b15b091b516094795092d9a6f00/LibOS/shim/src/sys/shim_file.c#L39)):*

Can we remove `= 0` now, so that the compiler may catch errors on unassigned-ret paths?
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/mkfifo.c, line 112 at r1](https://reviewable.io/reviews/oscarlab/graphene/2499#-MdpqqP42luOXOiVCe4W:-MdqM2gMBnP7xr9uQB03:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/824ab13178d45b15b091b516094795092d9a6f00/LibOS/shim/test/regression/mkfifo.c#L112)):*

Can we add `TEST OK` here, just for sanity?
</blockquote></details>

Done.

PR #2505, opened 2021-07-05
[LibOS] Fix a crash after reusing dentry touched by `mknod`

pwmarcz:
Found while investigating #2419. I tried to make a proper fix in PR #2499, but that seems to be blocked by conflict with tmpfs/strfs code. Merging #2499 would probably make an existing memory leak worse, so I would like to take care of tmpfs and strfs first. In the mean time, here is the workaround for the actual crash.

I ticketed the proper fix as #2507 and intend to work on it.

## Description of the changes 

This fixes a crash where we create a named pipe, remove it, and then try to create a normal file with the same name. Unfortunately, fixing the underlying problem (dentry reuse) is more problematic, so this is only a workaround for that specific case.

## How to test this PR? 

* Should fix the mknod crash in #2419 
   * Note that the mknod stress test still fails, because our mknod doesn't create sockets, and (when these failures are ignored) hits EMFILE from host on creating too many named pipes. However, it should not cause Graphene to crash anymore.
* I added a short snippet to the end of `mkfifo` that crashes in same manner on master.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2505)

### DIFF
--- a/LibOS/shim/src/fs/shim_namei.c
+++ b/LibOS/shim/src/fs/shim_namei.c
@@ -58,6 +58,24 @@ static struct shim_dentry* lookup_dcache_or_create(struct shim_dentry* parent, c
     return dent;
 }
 
+/*
+ * HACK: Reset dentry state, in case this is a dentry for which we changed `fs` earlier.
+ *
+ * This is a workaround for a situation in which we create a special node (named pipe or socket) and
+ * then delete or rename it. Currently that marks the old dentry as `DENTRY_NEGATIVE` and allows it
+ * to be reused later, despite changed `fs` field.
+ *
+ * Ideally, we should always start with a fresh dentry, which would ensure all fields are in a valid
+ * state. However, that requires fixing the memory leak of `dent->data` first (see `free_dentry()`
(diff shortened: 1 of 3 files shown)


dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2505)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2505)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2505)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2505)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2521, opened 2021-07-08
[Pal/Linux-SGX] Fix renaming issue with protected files

svenkata9:
## Description of the changes 
With certain workloads involving Federated Learning, the framework (Tensorflow, for example) creates a protected file and renames it as part of the flow. This does not work currently with protected files because every protected file maintains the path name in its metadata (stored when the file was first created) and which is checked during open to make sure the incoming path is the same as what is already stored in the metadata for this file.

This PR is to address this issue by handling rename for protected files. A normal rename call does not need the file to be open in RW mode, but since the metadata needs to be updated with the new path, we pass a special `PAL_OPTION_RENAME` flag to `DkStreamOpen` during rename and open the protected file in RW mode with required permissions, update the metadata with the new path and flush it. Then the flow goes through the one for normal file renaming.

There is a new regression test in LibOS `pf_rename` to test this

## How to test this PR? 
The newly added regression test in LibOS `pf_rename` should pass.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)


Pal/src/host/Linux-SGX/protected-files/protected_files.c
```diff
@@ -1329,6 +1343,29 @@ pf_status_t pf_flush(pf_context_t* pf) {
     return PF_STATUS_SUCCESS;
 }
 
+pf_status_t pf_rename(pf_context_t* pf, const char* new_path) {
+    if (!g_initialized)
```
duanbing, line 1347: Could we rename a protected file out of protected_files?  I think we should check the new path if it has the same common path as the old file,  where the common path is under the protected_files.

svenkata9: I think if you are trying to put it in a different path, it will fail primarily because the destination file or path is not (white-)listed in the manifest

duanbing: thanks, @svenkata9 . I will check that.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: all files reviewed, 22 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @duanbing and @svenkata9)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me93GxK7fp4pLDupNwC:-Me93GxL17FJVxAqvOnA:bshcp40) (no related file):*
@svenkata9 Please don't use the merge flow for PRs. Please use the rebase flow (`git rebase master`), but even this is discouraged: for any fixes, you should use the fixup flow (`git fixup`). Only in rare circumstances, the rebase flow is accepted in Graphene PR lifecycle.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me93Y5S8URDBN3oS-EC:-Me93Y5TE0QXixUh4YqE:b-wmfj13) (no related file):*
The text in the PR description is great. Please add all this text to the commit message. (Or I can do it during final rebase, as you wish.) If you'll add the text yourself, please try to remove some unnecessary phrases (like the mention of TensorFlow -- it doesn't matter for understanding the idea of this PR).

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9CvEC0OUWJK0l1JmE:-Me9CvEC0OUWJK0l1JmF:b-hbp6pk) (no related file):*
Could we have some (small) discussion on the caveats of renaming Protected Files? From my perspective, since `rename` operation is initiated by the in-enclave application, and we have some checks on the allowed renames in this PR, there is nothing wrong with renaming. If the application did a wrong rename, then it's the problem of an application, not Graphene.

I know @mkow was doubtful.

---
___
*[LibOS/shim/test/regression/manifest.template, line 37 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me94AzW1oP5qFz0GZmn:-Me94AzW1oP5qFz0GZmo:b-6ox2tp) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/manifest.template#L37)):*
> ```
> 
> sgx.protected_files_key = "ffeeddccbbaa99887766554433221100"
> sgx.protected_files.pftmp = "file:pftmp"
> ```

Since this is a directory, can you make it more explicit by using `file:pftmp/` (`/` at the end)?
___
*[LibOS/shim/test/regression/pf_rename.c, line 6 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me99uX92U_zKE-RDEKT:-Me99uX92U_zKE-RDEKU:b-q8bm5n) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L6)):*
> ```C
> #include <fcntl.h>
> #include <unistd.h>
> #include <sys/stat.h>
> ```

Please sort these includes.
___
*[LibOS/shim/test/regression/pf_rename.c, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me99wKa3vZF57Z5JvC5:-Me99wKa3vZF57Z5JvC6:b-od9g60) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L9)):*
> ```C
> 
> int main(void)
> {
> ```

Please move this `{` to the previous line
___
*[LibOS/shim/test/regression/pf_rename.c, line 12 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9A0UP4r3dE7R7bZKV:-Me9A0UP4r3dE7R7bZKW:bcagkk0) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L12)):*
> ```C
>     char buf[15];
>     char input_text[] = "Hello, world!";
>     int fd = open("./pftmp/foo.txt", O_RDWR | O_CREAT, S_IRWXU);
> ```

Why do you have `./` everywhere in this test? I suggest to remove it, just for readability.
___
*[LibOS/shim/test/regression/pf_rename.c, line 12 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9AW4X5e2hVDPWTLYa:-Me9AW4X5e2hVDPWTLYb:b9787u0) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L12)):*
> ```C
>     char buf[15];
>     char input_text[] = "Hello, world!";
>     int fd = open("./pftmp/foo.txt", O_RDWR | O_CREAT, S_IRWXU);
> ```

Why do you have an Execute permission on this file? Let's replace with `S_IRUSR | S_IWUSR`.
___
*[LibOS/shim/test/regression/pf_rename.c, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9AiJT7vjVeilXDvyS:-Me9AiJT7vjVeilXDvyT:b-wpzh6p) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L15)):*
> ```C
>     if (fd < 0) {
>         printf("Cannot create file ./pftmp/foo.txt\n");
>         goto out;
> ```

You cannot `close(fd)` if you didn't open it.
___
*[LibOS/shim/test/regression/pf_rename.c, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9At2p5z1X3anhpEHU:-Me9At2p5z1X3anhpEHV:btg82y5) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L15)):*
> ```C
>     if (fd < 0) {
>         printf("Cannot create file ./pftmp/foo.txt\n");
>         goto out;
> ```

Actually, these error handlings are all weird. You `return 0` even on failures. I suggest to replace all these with `err(1, "open(pftmp/foo.txt)")` and similar. Note that `err()` already prints the error message and exits with error code `1`, so this is very simple to use.

Of course, you'll need to add `close(fd)` on all error paths where it is necessary.
___
*[LibOS/shim/test/regression/pf_rename.c, line 18 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9BS7x10LlRWXL4ZNu:-Me9BS7x10LlRWXL4ZNv:b-95hwfn) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L18)):*
> ```C
>     }
> 
>     int num_bytes = write(fd, input_text, strlen(input_text));
> ```

I would prefer +1 here.
___
*[LibOS/shim/test/regression/pf_rename.c, line 23 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9BsI-DDAonkA43Ncx:-Me9BsI-DDAonkA43Ncy:bgbm78s) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L23)):*
> ```C
>         goto out;
>     }
> ```

You're supposed to `close(fd)` before doing a rename.
___
*[LibOS/shim/test/regression/pf_rename.c, line 24 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9BZRF5qV6Ufg2BCK3:-Me9BZRF5qV6Ufg2BCK4:bozaxvj) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L24)):*
> ```C
>     }
> 
>     rename("./pftmp/foo.txt", "./pftmp/bar.txt");
> ```

You need to have a return value and check it. `int ret = rename(); if (ret < 0) {error}`
___
*[LibOS/shim/test/regression/pf_rename.c, line 37 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9C4fF-xxbUEeHrg5U:-Me9C4fF-xxbUEeHrg5V:b-lg5jz9) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L37)):*
> ```C
>         goto out;
>     }
> ```

Please also add `buf[sizeof(buf) - 1] = '\0';` just for sanity.
___
*[LibOS/shim/test/regression/pf_rename.c, line 38 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9CTzZE7nSAwMFCGGX:-Me9CTzZE7nSAwMFCGGY:b-44p95z) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L38)):*
> ```C
>     }
> 
>     if (strcmp(input_text, buf)) {
> ```

Let's replace with `strncmp(input_text, buf, sizeof(input_text))`, just to be safe.
___
*[LibOS/shim/test/regression/pf_rename.c, line 47 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9Cg8t5nQ29eBIjCsB:-Me9Cg8t5nQ29eBIjCsC:bd40ak2) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L47)):*
> ```C
> out:
>     close(fd);
>     return 0;
> ```

You shouldn't return success (`0`) on failures.
___
*[Pal/include/host/Linux-common/pal_flags_conv.h, line 55 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me95WKM4MbIMowW92l_:-Me95WKM4MbIMowW92la:b-9iwzx8) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/include/host/Linux-common/pal_flags_conv.h#L55)):*
> ```objc
>     return (options & PAL_OPTION_CLOEXEC  ? O_CLOEXEC  : 0) |
>            (options & PAL_OPTION_NONBLOCK ? O_NONBLOCK : 0) |
>            (options & PAL_OPTION_RENAME   ? O_RDWR     : 0);
> ```

Actually, can we revert this part? The `RENAME` -> `RDWR` convertion is very specific to Protected Files, but here you're changing the code that is common for all kinds of files. See my other comment.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 65 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me95sEr9aWX7lVv0CFG:-Me95sEr9aWX7lVv0CFH:byfndol) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/src/host/Linux-SGX/db_files.c#L65)):*
> ```C
>      */
>     if (pf && (options & PAL_OPTION_RENAME))
>         share = PAL_SHARE_OWNER_R | PAL_SHARE_OWNER_W;
> ```

I would also add `O_RDWR` here, in this special case of protected files (and remove from the common `PAL_OPTION_TO_LINUX_OPEN` function, see my other comment).

In other words, I would refactor like this:
```
    int flags = PAL_ACCESS_TO_LINUX_OPEN(access)  |
                PAL_CREATE_TO_LINUX_OPEN(create)  |
                PAL_OPTION_TO_LINUX_OPEN(options);

    /* if it is a protected file ... */
    if (pf && (options & PAL_OPTION_RENAME)) {
        share = PAL_SHARE_OWNER_R | PAL_SHARE_OWNER_W;
        flags |= O_RDWR;
    }

    options &= ~PAL_OPTION_RENAME; /* not needed anymore, reset for sanity */

    int fd = ocall_open(uri, flags, share);
    ...

```
___
*[Pal/src/host/Linux-SGX/db_files.c, line 742 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me98U4T1JQr3ZvDESTH:-Me98U4U0-2PzCJFY2DR:b1ihaa3) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/src/host/Linux-SGX/db_files.c#L742)):*
> ```C
> 
>         if (get_norm_path(uri, new_path, &uri_size) < 0) {
>             log_warning("Could not normalize path (%s)", uri);
> ```

You should also add `free(tmp);` before returning here.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 745 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me981oqBIJ74CE0X1Le:-Me981oqBIJ74CE0X1Lf:b-7q91w6) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/src/host/Linux-SGX/db_files.c#L745)):*
> ```C
>             return -PAL_ERROR_DENIED;
>         }
> ```

As mentioned in the other comment, we should actually check that `new_path` is one of the allowed Protected Files paths. So we should add something like this before `pf_rename`:
```
    if (!get_protected_file(new_path)) {
            log_warning("New path is not allowed for protected files (%s)", new_path);
            return -PAL_ERROR_DENIED;
    }
```
___
*[Pal/src/host/Linux-SGX/db_files.c, line 751 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me98Ygb-iupOVAiCgQ8:-Me98Ygb-iupOVAiCgQ9:b1ihaa3) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/src/host/Linux-SGX/db_files.c#L751)):*
> ```C
> 
>         if (PF_FAILURE(pf_ret)) {
>             log_error("pf_rename failed: %s", pf_strerror(pf_ret));
> ```

You should also add `free(tmp);` before returning here.
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 1347 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me4lVabffQf1J0Q_4mu-r1-1347:-Me98ucG324QDHdlDzrR:bdyyvd2) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L1347)):*
<details><summary><i>Previously, duanbing (Bing) wrote…</i></summary><blockquote>

thanks, @svenkata9 . I will check that. 
</blockquote></details>

This is a valid comment. See my other comment with the suggestion how to enforce it.

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: 6 of 11 files reviewed, 22 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, @mkow, and @svenkata9)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me93GxK7fp4pLDupNwC:-Me9iFmKDiMW7y3IwWCM:b-pbdqz3) (no related file):*

@svenkata9 Please don't use the merge flow for PRs. Please use the rebase flow (`git rebase master`), but even this is discouraged: for any fixes, you should use the fixup flow (`git fixup`). Only in rare circumstances, the rebase flow is accepted in Graphene PR lifecycle.
</blockquote></details>

Agreed.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me93Y5S8URDBN3oS-EC:-Me9iIjVA0bqALR2UYK9:b-ler7fn) (no related file):*

The text in the PR description is great. Please add all this text to the commit message. (Or I can do it during final rebase, as you wish.) If you'll add the text yourself, please try to remove some unnecessary phrases (like the mention of TensorFlow -- it doesn't matter for understanding the idea of this PR).
</blockquote></details>

Thanks. Could you update when you do the final rebase?

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9CvEC0OUWJK0l1JmE:-Me9iNwX9Nuoowf16B8P:bds25c6) (no related file):*

Could we have some (small) discussion on the caveats of renaming Protected Files? From my perspective, since `rename` operation is initiated by the in-enclave application, and we have some checks on the allowed renames in this PR, there is nothing wrong with renaming. If the application did a wrong rename, then it's the problem of an application, not Graphene.

I know @mkow was doubtful.
</blockquote></details>

We definitely can have the discussion to be on the same page. My thought process was the same as yours because the workload itself is triggering the rename during inside Graphene-SGX.

---
___
*[LibOS/shim/test/regression/manifest.template, line 37 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me94AzW1oP5qFz0GZmn:-Me9l22oEbyU5G-PZgBd:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/manifest.template#L37)):*

Since this is a directory, can you make it more explicit by using `file:pftmp/` (`/` at the end)?
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/pf_rename.c, line 6 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me99uX92U_zKE-RDEKT:-Me9l3hF0ovnf0mIZYX_:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L6)):*

Please sort these includes.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/pf_rename.c, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me99wKa3vZF57Z5JvC5:-Me9l67I8CJZXWC-eznN:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L9)):*

Please move this `{` to the previous line
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/pf_rename.c, line 12 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9A0UP4r3dE7R7bZKV:-Me9l75mERpjlKeExNb9:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L12)):*

Why do you have `./` everywhere in this test? I suggest to remove it, just for readability.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/pf_rename.c, line 12 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9AW4X5e2hVDPWTLYa:-Me9l7lN471U9HKXwfum:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L12)):*

Why do you have an Execute permission on this file? Let's replace with `S_IRUSR | S_IWUSR`.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/pf_rename.c, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9At2p5z1X3anhpEHU:-Me9lDCk119O_1ip9L5i:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L15)):*

Actually, these error handlings are all weird. You `return 0` even on failures. I suggest to replace all these with `err(1, "open(pftmp/foo.txt)")` and similar. Note that `err()` already prints the error message and exits with error code `1`, so this is very simple to use.

Of course, you'll need to add `close(fd)` on all error paths where it is necessary.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/pf_rename.c, line 18 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9BS7x10LlRWXL4ZNu:-Me9lEEFA38R7YGR4_Ml:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L18)):*

I would prefer +1 here.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/pf_rename.c, line 23 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9BsI-DDAonkA43Ncx:-Me9lF8t4c3Iu97OejBI:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L23)):*

You're supposed to `close(fd)` before doing a rename.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/pf_rename.c, line 24 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9BZRF5qV6Ufg2BCK3:-Me9lYQV3cLHbsah5gtd:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L24)):*

You need to have a return value and check it. `int ret = rename(); if (ret < 0) {error}`
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/pf_rename.c, line 37 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9C4fF-xxbUEeHrg5U:-Me9mJAy7ffDqXjY3mZT:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L37)):*

Please also add `buf[sizeof(buf) - 1] = '\0';` just for sanity.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/pf_rename.c, line 38 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9CTzZE7nSAwMFCGGX:-Me9mK4fCUP8pAFteYVh:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L38)):*

Let's replace with `strncmp(input_text, buf, sizeof(input_text))`, just to be safe.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/pf_rename.c, line 47 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9Cg8t5nQ29eBIjCsB:-Me9n7MM7GEwtx4H-L0z:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L47)):*

You shouldn't return success (`0`) on failures.
</blockquote></details>

Done.
___
*[Pal/include/host/Linux-common/pal_flags_conv.h, line 55 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me95WKM4MbIMowW92l_:-Me9n9uB95YS-wzi0IhK:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/include/host/Linux-common/pal_flags_conv.h#L55)):*

Actually, can we revert this part? The `RENAME` -> `RDWR` convertion is very specific to Protected Files, but here you're changing the code that is common for all kinds of files. See my other comment.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 65 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me95sEr9aWX7lVv0CFG:-Me9nF6P6vZiYiWIx8hg:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/src/host/Linux-SGX/db_files.c#L65)):*

I would also add `O_RDWR` here, in this special case of protected files (and remove from the common `PAL_OPTION_TO_LINUX_OPEN` function, see my other comment).

In other words, I would refactor like this:
```
    int flags = PAL_ACCESS_TO_LINUX_OPEN(access)  |
                PAL_CREATE_TO_LINUX_OPEN(create)  |
                PAL_OPTION_TO_LINUX_OPEN(options);

    /* if it is a protected file ... */
    if (pf && (options & PAL_OPTION_RENAME)) {
        share = PAL_SHARE_OWNER_R | PAL_SHARE_OWNER_W;
        flags |= O_RDWR;
    }

    options &= ~PAL_OPTION_RENAME; /* not needed anymore, reset for sanity */

    int fd = ocall_open(uri, flags, share);
    ...

```
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 742 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me98U4T1JQr3ZvDESTH:-Me9nI8u7Rnn_HyRhqdN:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/src/host/Linux-SGX/db_files.c#L742)):*

You should also add `free(tmp);` before returning here.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 745 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me981oqBIJ74CE0X1Le:-Me9nIce3fpL4PsGrDbS:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/src/host/Linux-SGX/db_files.c#L745)):*

As mentioned in the other comment, we should actually check that `new_path` is one of the allowed Protected Files paths. So we should add something like this before `pf_rename`:
```
    if (!get_protected_file(new_path)) {
            log_warning("New path is not allowed for protected files (%s)", new_path);
            return -PAL_ERROR_DENIED;
    }
```
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 751 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me98Ygb-iupOVAiCgQ8:-Me9nJzX5rVyTMdiuEpi:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/src/host/Linux-SGX/db_files.c#L751)):*

You should also add `free(tmp);` before returning here.
</blockquote></details>

Done.

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: 6 of 11 files reviewed, 22 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, and @mkow)

___
*[LibOS/shim/test/regression/pf_rename.c, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9AiJT7vjVeilXDvyS:-Me9nU3P6UhVQyoKsMqk:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L15)):*

You cannot `close(fd)` if you didn't open it.
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/pf_rename.c, line 15 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9At2p5z1X3anhpEHU:-Me9nUrQ5fDL2JpE4xnD:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/LibOS/shim/test/regression/pf_rename.c#L15)):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Done.

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: 6 of 11 files reviewed, 22 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, and @mkow)

___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 1347 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me4lVabffQf1J0Q_4mu-r1-1347:-Me9nYB4AVmqxzyxEzVf:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L1347)):*

This is a valid comment. See my other comment with the suggestion how to enforce it.
</blockquote></details>

Done.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: all files reviewed, 11 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @duanbing, @mkow, and @svenkata9)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me93Y5S8URDBN3oS-EC:-MeAml0yA6IEq-TYAxAX:b-louncl) (no related file):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

Thanks. Could you update when you do the final rebase?
</blockquote></details>

Yes, will do.

---
___
*[LibOS/shim/test/regression/pf_rename.c, line 43 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAnKHl3oxobEs2mygt:-MeAnKHl3oxobEs2mygu:b-2q7wk1) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/LibOS/shim/test/regression/pf_rename.c#L43)):*
> ```C
>         err(1, "File content mismatching");
>     }
> ```

Please also `close(fd)` here (since we opened and read from it, but haven't yet closed).
___
*[Pal/src/host/Linux-SGX/db_files.c, line 67 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAoUhHAtUXzW1X-c57:-MeAoUhHAtUXzW1X-c58:b4ul0q1) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L67)):*
> ```C
>      * the metadata in the file, so open with RDWR mode with necessary share permissions.
>      * For normal files, reset this option.
>      */
> ```

Personal preference, but could we move this `*/` to the end of the previous line?
___
*[Pal/src/host/Linux-SGX/db_files.c, line 742 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAoi7r6hsK8LGhnuBB:-MeAoi7r6hsK8LGhnuBC:b6z5opo) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L742)):*
> ```C
>     if (pf) {
>         size_t uri_size = strlen(uri) + 1;
>         char* new_path = (char*)calloc(1, uri_size);
> ```

You should error out on not-allocated `new_path`, like this:
```
char* new_path = (char*)calloc(1, uri_size);
if (!new_path) {
    free(tmp);
    return -PAL_ERROR_NOMEM;
}
```
___
*[Pal/src/host/Linux-SGX/db_files.c, line 745 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAnmIQ8EiezroJKxQf:-MeAnmIQ8EiezroJKxQg:b-9lmbra) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L745)):*
> ```C
> 
>         if (get_norm_path(uri, new_path, &uri_size) < 0) {
>             log_error("Could not normalize path (%s)", uri);
> ```

We use `log_warning()` for all non-fatal errors, and this one is not fatal (because we don't kill the whole process, we just return the error to the application, and the app decides what to do). Please use `log_warning`.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 746 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAow843sGnHAArPhYv:-MeAow843sGnHAArPhYw:b-qmv59e) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L746)):*
> ```C
>         if (get_norm_path(uri, new_path, &uri_size) < 0) {
>             log_error("Could not normalize path (%s)", uri);
>             free(tmp);
> ```

We also need to `free(new_path)`
___
*[Pal/src/host/Linux-SGX/db_files.c, line 751 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAo0leApovOQNLkgTj:-MeAo0leApovOQNLkgTk:b5uc73g) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L751)):*
> ```C
> 
>         if (!get_protected_file(new_path)) {
>             log_error("New path is disallowed for protected files (%s)", new_path);
> ```

ditto (please use `log_warning`)
___
*[Pal/src/host/Linux-SGX/db_files.c, line 752 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAp0pb4Uhdrw2zONaX:-MeAp0pb4Uhdrw2zONaY:b-qmv59e) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L752)):*
> ```C
>         if (!get_protected_file(new_path)) {
>             log_error("New path is disallowed for protected files (%s)", new_path);
>             free(tmp);
> ```

We also need to `free(new_path)`
___
*[Pal/src/host/Linux-SGX/db_files.c, line 761 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAoGmt8HmeS8ymjF2W:-MeAoGmt8HmeS8ymjF2X:b5uc73g) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L761)):*
> ```C
> 
>         if (PF_FAILURE(pf_ret)) {
>             log_error("pf_rename failed: %s", pf_strerror(pf_ret));
> ```

ditto (please use `log_warning`)

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: 9 of 11 files reviewed, 11 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me93Y5S8URDBN3oS-EC:-MeO75a1DnPCIIplK6tH:b32w0x) (no related file):*

Yes, will do.
</blockquote></details>

OK

---
___
*[LibOS/shim/test/regression/pf_rename.c, line 43 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAnKHl3oxobEs2mygt:-MeO7APw-n-NDbM1xksy:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/LibOS/shim/test/regression/pf_rename.c#L43)):*

Please also `close(fd)` here (since we opened and read from it, but haven't yet closed).
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 67 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAoUhHAtUXzW1X-c57:-MeO7BFH7QhViPODWuXq:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L67)):*

Personal preference, but could we move this `*/` to the end of the previous line?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 742 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAoi7r6hsK8LGhnuBB:-MeO7D_99guRoj9i9s6q:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L742)):*

You should error out on not-allocated `new_path`, like this:
```
char* new_path = (char*)calloc(1, uri_size);
if (!new_path) {
    free(tmp);
    return -PAL_ERROR_NOMEM;
}
```
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 745 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAnmIQ8EiezroJKxQf:-MeO7ElH8dfICC-IQBuk:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L745)):*

We use `log_warning()` for all non-fatal errors, and this one is not fatal (because we don't kill the whole process, we just return the error to the application, and the app decides what to do). Please use `log_warning`.
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 746 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAow843sGnHAArPhYv:-MeO7Ga102OrCCHHejkW:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L746)):*

We also need to `free(new_path)`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 751 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAo0leApovOQNLkgTj:-MeO7Gvn1xrNMP16Z0Ek:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L751)):*

ditto (please use `log_warning`)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 752 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAp0pb4Uhdrw2zONaX:-MeO7HGl7PyXjjxiEcnJ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L752)):*

We also need to `free(new_path)`
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 761 at r2](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeAoGmt8HmeS8ymjF2W:-MeO7Hkd2TBMZESMqYyH:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f8e760c0b283a4bf5f44c1edf684b0ef5ccf8353/Pal/src/host/Linux-SGX/db_files.c#L761)):*

ditto (please use `log_warning`)
</blockquote></details>

Done.

dimakuv (changes requested):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, @mkow, and @svenkata9)

___
*[LibOS/shim/test/regression/pf_rename.c, line 33 at r3](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeOVdL1Bf2tQAwQsBzM:-MeOVdL260ScDKmRKF7q:bgsxx3n) ([raw file](https://github.com/oscarlab/graphene/blob/fd04449f213974028a16a6d5de6b0562edef01ec/LibOS/shim/test/regression/pf_rename.c#L33)):*
> ```C
> 
>     num_bytes = read(fd, buf, sizeof(buf));
>     close (fd);
> ```

There is a space in-between, please remove it
___
*[Pal/src/host/Linux-SGX/db_files.c, line 743 at r3](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeOVktP8MgDmoGv6_Ee:-MeOVktP8MgDmoGv6_Ef:b-od9g60) ([raw file](https://github.com/oscarlab/graphene/blob/fd04449f213974028a16a6d5de6b0562edef01ec/Pal/src/host/Linux-SGX/db_files.c#L743)):*
> ```C
> 
>         if (!new_path)
>         {
> ```

Please move this `{` to the previous line

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: 9 of 11 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, and @mkow)

___
*[LibOS/shim/test/regression/pf_rename.c, line 33 at r3](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeOVdL1Bf2tQAwQsBzM:-MeOZJ8s6_BNkIvfB-s6:b-c3lv5u) ([raw file](https://github.com/oscarlab/graphene/blob/fd04449f213974028a16a6d5de6b0562edef01ec/LibOS/shim/test/regression/pf_rename.c#L33)):*

There is a space in-between, please remove it
</blockquote></details>

Done. Accidentally had removed a bracket also - which was missed out in the run
___
*[Pal/src/host/Linux-SGX/db_files.c, line 743 at r3](https://reviewable.io/reviews/oscarlab/graphene/2521#-MeOVktP8MgDmoGv6_Ee:-MeOZ_4j5AHXtZjUKctY:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/fd04449f213974028a16a6d5de6b0562edef01ec/Pal/src/host/Linux-SGX/db_files.c#L743)):*

Please move this `{` to the previous line
</blockquote></details>

Done.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, and @mkow)

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, and @mkow)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me9CvEC0OUWJK0l1JmE:-MeQ713SC9ZInOvlQLbq:bcv7vhc) (no related file):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

We definitely can have the discussion to be on the same page. My thought process was the same as yours because the workload itself is triggering the rename during inside Graphene-SGX.
</blockquote></details>

I think based on today's discussion, we are good to proceed with the change

---

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, and @mkow)

___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 1347 at r1](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me4lVabffQf1J0Q_4mu-r1-1347:-MeQ7NLPDDOARYooY1dx:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/faf5348a8791e444c032e48a9083d8e4efdfb733/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L1347)):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Done.

dimakuv:
Jenkins, test this please

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv and @duanbing)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me93Y5S8URDBN3oS-EC:-MhvvRdQD4bANuDjr8Dd:b-896fix) (no related file):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

OK
</blockquote></details>

Done.

---

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: all files reviewed, 8 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, @mkow, and @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me93Y5S8URDBN3oS-EC:-Mi6p6IZ3tna4E68PFY2:b-6xxsa6) (no related file):*

Could you do this now?
</blockquote></details>

Dmitrii said he will do it during final rebase

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi3qjxP8fm9s0K7bmYA:-MiQjY4T-qAP6e6WuUSU:b-t6qmky) (no related file):*

@svenkata9: Could you squash and rebase this PR to the current master? It got pretty old and I'd like to review it against the current code.
</blockquote></details>

This is the reason that I have been asking to review the PR since 1.5 months

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6_Ynm5qI5oJuEvpt8:-Mi6pfQg7kvbyINUTt9D:bx4byft) (no related file):*

Is this expected to work if the file being renamed is already open? Something like:

```c
int fd = open("foo", ...);
rename("foo", "bar");
read(fd, ...);
```

If so, can you add a test?
</blockquote></details>

For native case, this flow works and so is it for pf case also. Should we add a test in that case (I can understand if it should not work, we should add a test case)?

---
___
*[Pal/src/host/Linux-SGX/db_files.c, line 761 at r4](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6e5TE-DJYHfL-2Y-z:-MiQjiWC4aecBocWcmFH:b7rlvci) ([raw file](https://github.com/oscarlab/graphene/blob/3f61a18f904bc16a34b4756c9c614d73d95369d4/Pal/src/host/Linux-SGX/db_files.c#L761)):*

This modifies `pf`, which (as I understand) is still going to be associated with the old path. What happens if somebody holds another `PAL_HANDLE` with this old path?
</blockquote></details>

I guess any file operation on that handle will cease to work, since the file has been renamed (and which should be the case for native also I believe)
___
*[Pal/src/host/Linux-SGX/db_files.c, line 772 at r4](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6hHxsAuvvzJFLMM9X:-MiQk9RfEfsYVLDscA14:b2k9bjt) ([raw file](https://github.com/oscarlab/graphene/blob/3f61a18f904bc16a34b4756c9c614d73d95369d4/Pal/src/host/Linux-SGX/db_files.c#L772)):*

What happens if this `ocall_rename` fails? I'm guessing the file is left in a unusable state...
</blockquote></details>

this is a tricky one. Yes, I guess the file will be unusable. One way to handle it is to restore the metadata back to what it was before if ocall_rename fails. Does that sound acceptable?
___
*[Pal/src/host/Linux-SGX/db_files.c, line 783 at r4](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6ey6kDSNVG4L5cfU3:-MiQjyxT4VEHWhU1Amx5:bemqr4d) ([raw file](https://github.com/oscarlab/graphene/blob/3f61a18f904bc16a34b4756c9c614d73d95369d4/Pal/src/host/Linux-SGX/db_files.c#L783)):*

At this point, the old `protected_file` object will no longer be associated by `handle`, and a new one starts being associated (because the association is by URI). Shouldn't you update the refcounts for these? Same as in `file_open` and `pf_file_close`. Otherwise, you will end up miscounting the references (and probably causing the refcount for new file to go below zero):

```c
file_open(&hdl, "foo");   // increases refcount for protectd_file "foo"
file_rename(hdl, "bar");  // creates protected_file "bar" with refcount 0
file_close(hdl);          // decreases refcount for protected_file "bar"
```

CC @dimakuv, I think we talked about these refcounts recently...
</blockquote></details>

Do you have any email thread on this/notes based on the discussion with @dimakuv?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, @mkow, @pwmarcz, and @svenkata9)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi3qjxP8fm9s0K7bmYA:-MiQmiK212-YU2pTN8PA:b20512z) (no related file):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

This is the reason that I have been asking to review the PR since 1.5 months
</blockquote></details>

Sorry for this, we just have a huge inflow of PRs and we aren't able to review all of them in a timely manner. But the rebase shouldn't be too complicated and you could use it to resolve Dmitrii's comment.

---
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 1355 at r4](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi3rMUO625-Uc7amsEU:-Mi3rMUP6GukFuNn-1yW:bpcqyik) ([raw file](https://github.com/oscarlab/graphene/blob/3f61a18f904bc16a34b4756c9c614d73d95369d4/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L1355)):*
> ```C
>     }
> 
>     if (new_path && strlen(new_path) > PATH_MAX_SIZE - 1) {
> ```

TMP: why here, not in ipf_rename_file?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @duanbing, @mkow, @pwmarcz, and @svenkata9)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Me93Y5S8URDBN3oS-EC:-MiQvl3p93aOn8R28c-b:b-63vl58) (no related file):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

Dmitrii said he will do it during final rebase
</blockquote></details>

Yes, we can do it during the final rebase. I guess what confused Michal is that you replied with `Done` at some point -- probably just misclicked on the Reviewable button.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6_Ynm5qI5oJuEvpt8:-MiQxI0i5j1PCbRcOu9d:b-gedz5i) (no related file):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

For native case, this flow works and so is it for pf case also. Should we add a test in that case (I can understand if it should not work, we should add a test case)?
</blockquote></details>

Actually, +1 to Pawel's comment. Could we add to your test a reverse rename, smth like this:
```
... your existing test logic ...
// don't close(fd); immediately
rename("pftmp/bar.txt", "pftmp/foo.txt");
lseek(fd, 0, SEEK_SET);   // rewind to beginning of the file
read(fd, buf, sizeof(buf));  // read the string again from this file
strncmp(input_text, buf, sizeof(input_text));  // check the read string again
close(fd); // now we can close this file finally
```

If this *fails*, then I would be fine with this error, but then we can add a TODO comment about this non-working case.

---
___
*[Pal/src/host/Linux-SGX/db_files.c, line 761 at r4](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6e5TE-DJYHfL-2Y-z:-MiQyZw94nSYhBQmzPJ4:b-jvrc9i) ([raw file](https://github.com/oscarlab/graphene/blob/3f61a18f904bc16a34b4756c9c614d73d95369d4/Pal/src/host/Linux-SGX/db_files.c#L761)):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

I guess any file operation on that handle will cease to work, since the file has been renamed (and which should be the case for native also I believe)
</blockquote></details>

That's not how native Linux works -- if the file was opened with the old name and then renamed, the opened file descriptor still operates correctly (one can do file operations on this file descriptor).

But this is probably how current Graphene files work... This needs some testing.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 772 at r4](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6hHxsAuvvzJFLMM9X:-MiQzQTVEuZaWiYA6UDX:b362he8) ([raw file](https://github.com/oscarlab/graphene/blob/3f61a18f904bc16a34b4756c9c614d73d95369d4/Pal/src/host/Linux-SGX/db_files.c#L772)):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

this is a tricky one. Yes, I guess the file will be unusable. One way to handle it is to restore the metadata back to what it was before if ocall_rename fails. Does that sound acceptable?
</blockquote></details>

If it is simple to restore the metadata back, then yes, restoring the PF metadata sounds reasonable here.

On the other hand, I would be fine with something like:
```
if (pf)
   log_warning("ocall_rename failed on an already-renamed protected file; file is unusable now");
```
___
*[Pal/src/host/Linux-SGX/db_files.c, line 783 at r4](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6ey6kDSNVG4L5cfU3:-MiR-ENy-pfazYjQmt97:b9ky56t) ([raw file](https://github.com/oscarlab/graphene/blob/3f61a18f904bc16a34b4756c9c614d73d95369d4/Pal/src/host/Linux-SGX/db_files.c#L783)):*
<details><summary><i>Previously, svenkata9 (Sankaranarayanan Venkatasubramanian) wrote…</i></summary><blockquote>

Do you have any email thread on this/notes based on the discussion with @dimakuv?
</blockquote></details>

I guess Pawel means this PR: https://github.com/oscarlab/graphene/pull/2372

Refcounting is totally broken in Protected Files, and that PR was my initial attempt to make it more reasonable. @pwmarcz To be honest, I think that refcounting is out of scope for this PR.

pwmarcz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, @mkow, and @svenkata9)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6_Ynm5qI5oJuEvpt8:-MiR9C3ACdpiKsNfNCJk:bptj5tc) (no related file):*
@dimakuv You might be interested in [my refactoring](https://github.com/oscarlab/graphene/pull/2646) that fixes various corner cases around rename, and adds a lot of tests :)

>For native case, this flow works and so is it for pf case also. Should we add a test in that case (I can understand if it should not work, we should add a test case)?

I agree with @dimakuv: it should work; if it works, it would be good to have a test, if not, then a TODO somewhere.

---
___
*[Pal/src/host/Linux-SGX/db_files.c, line 761 at r4](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6e5TE-DJYHfL-2Y-z:-MiRAmDH5ydQX2EBHdD3:b-72qdt8) ([raw file](https://github.com/oscarlab/graphene/blob/3f61a18f904bc16a34b4756c9c614d73d95369d4/Pal/src/host/Linux-SGX/db_files.c#L761)):*

That's not how native Linux works -- if the file was opened with the old name and then renamed, the opened file descriptor still operates correctly (one can do file operations on this file descriptor).

But this is probably how current Graphene files work... This needs some testing.
</blockquote></details>

Yeah, it's unfortunate that PAL handles store file path, and probably not easy to fix. If it doesn't work, then a TODO here is fine, I guess.

(By the way, this case works with normal PAL handles, because the path doesn't matter - only the file descriptor).
___
*[Pal/src/host/Linux-SGX/db_files.c, line 772 at r4](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6hHxsAuvvzJFLMM9X:-MiRAiO52EIk4asj4Hui:b-y42rs6) ([raw file](https://github.com/oscarlab/graphene/blob/3f61a18f904bc16a34b4756c9c614d73d95369d4/Pal/src/host/Linux-SGX/db_files.c#L772)):*

If it is simple to restore the metadata back, then yes, restoring the PF metadata sounds reasonable here.

On the other hand, I would be fine with something like:
```
if (pf)
   log_warning("ocall_rename failed on an already-renamed protected file; file is unusable now");
```
</blockquote></details>

Agreed, a warning should be OK too.
___
*[Pal/src/host/Linux-SGX/db_files.c, line 783 at r4](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6ey6kDSNVG4L5cfU3:-MiRB_Xg7Ng2hXpbtrse:bmexotm) ([raw file](https://github.com/oscarlab/graphene/blob/3f61a18f904bc16a34b4756c9c614d73d95369d4/Pal/src/host/Linux-SGX/db_files.c#L783)):*

I guess Pawel means this PR: https://github.com/oscarlab/graphene/pull/2372

Refcounting is totally broken in Protected Files, and that PR was my initial attempt to make it more reasonable. @pwmarcz To be honest, I think that refcounting is out of scope for this PR.
</blockquote></details>

@dimakuv Okay. But can you confirm that getting the refcounts wrong risks only leaking memory, not use-after-free or panic? If so, it should be OK to leave for now.

svenkata9 (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2521)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @dimakuv, @duanbing, @mkow, @pwmarcz, and @svenkata9)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6_Ynm5qI5oJuEvpt8:-MiffxXr7KOU-P3WTJKb:blzzbqq) (no related file):*

@dimakuv You might be interested in [my refactoring](https://github.com/oscarlab/graphene/pull/2646) that fixes various corner cases around rename, and adds a lot of tests :)

>For native case, this flow works and so is it for pf case also. Should we add a test in that case (I can understand if it should not work, we should add a test case)?

I agree with @dimakuv: it should work; if it works, it would be good to have a test, if not, then a TODO somewhere.
</blockquote></details>

I think it is not straightforward and does not always work. The reason is we use find_protected_file_handle which calls find_protected_file(const char* path). The latter uses HASH_FIND_STR(g_protected_files, path, pf) to return the pf. So, I see the right pf for which the fd turns out to be right and works, and sometimes the pf is incorrect because of two entries (one from the previously opened handle and the next one from the rename open).

Unless we associate handle and file name together and use it in the find_protected_file, we will get good and bad results. I probably will add a TODO

---
___
*[Pal/src/host/Linux-SGX/db_files.c, line 772 at r4](https://reviewable.io/reviews/oscarlab/graphene/2521#-Mi6hHxsAuvvzJFLMM9X:-MifheR6A_PWCWm8E92I:bdaxy6f) ([raw file](https://github.com/oscarlab/graphene/blob/3f61a18f904bc16a34b4756c9c614d73d95369d4/Pal/src/host/Linux-SGX/db_files.c#L772)):*

Agreed, a warning should be OK too.
</blockquote></details>

I tried updating metadata and it works (triggered a hardcoded ocall_rename failure). I will update the code to handle this case. File not being usable might not look good.

svenkata9:
Got too old. Merge conflicts. Will raise a new PR.

duanbing:
@svenkata9 ,  I see that 333 files changed, have you rebased the branch?

Issue #2517, opened 2021-07-07
RFC: interruptible POSIX locks


pwmarcz:
I'm trying to fix recently added POSIX locks (#2481) so that you can interrupt them, but I ran into a problem trying to do it with current IPC. I don't want to jump and change too much before I learn your opinion.

@boryspoplawski @mkow @dimakuv

## The problem

Graphene's current implementation of POSIX locks (`fcntl`) do not support interrupting the request:

```c
void handler(int signum) {
    printf("got signal %d\n", signum);
}
signal(SIGALRM, &handler);

fcntl(fd, F_SETLK, ...);
pid_t pid = fork();
if (pid == 0) {
    // child process:

    // trigger SIGALRM after 10 seconds
    alarm(10);

    // this should be interrupted:
    fcntl(fd, F_SETLKW, ...);
    ...
}
```

The `fcntl(F_SETLKW)` in the child process should wait for 10 seconds and then fail with EINTR (after the process receives SIGALRM). Under Graphene, it hangs indefinitely.

So far, I've seen this checked by LTP (test `fcntl16`), and I'm pretty sure it's used by `stress-ng` (see #2510). I'm not sure how much it's used in actual applications, but interrupting the `fcntl` call seems to be the only sensible way of trying to take a lock with timeout.

## How to interrupt an operation over IPC?

Interrupting the lock operation is easy in the single-process version. However, if we are in a remote process, the operation is requested over IPC, and it's not that easy to achieve the same semantics.

The current IPC implementation does something like this:

```c
lock_msg = { POSIX_LOCK_SET, .wait = true, ... };

// this waits for response, and retries on EINTR
ipc_send_message_and_get_response(lock_msg, &result);
return result;
```

As long as `ipc_send_message_and_get_response` keeps retrying, we have no way to return earlier. I've been thinking about some solutions:

1. **Stop waiting?**:  You could be tempted to modify the `ipc_send_message_and_get_response()` function to *not* wait (perhaps with some `retry = false` parameter). However, this does not actually cancel the operation, so the lock will be taken eventually. This is wrong - we want the `F_SETLKW` operation to fail without side effects.

2. **Stop waiting and issue another call?** We could send another IPC message to cancel the pending operation. Keep in mind that the operation might get finished before the "cancel" message is delivered, so we need to learn what the actual result is.
    ```c
    lock_msg = { POSIX_LOCK_SET, .wait = true, ... };

    // this returns -EINTR if the wait is interrupted
[...]

PR #2522, opened 2021-07-08
[LibOS] Make POSIX locks interruptible

### LINKED ISSUE #2510: Stress-ng test lockf hangs with Graphene
## Description of the problem
Description of lockf test
```
start  N workers that randomly lock and unlock regions of a file
#   using the POSIX lockf(3) locking mechanism. Each worker  creates
#   a  64  KB file and attempts to hold a maximum of 1024 concurrent
#   locks with a child process that also tries to hold 1024  concur‐
#   rent  locks.  Old  locks  are  unlocked in a first-in, first-out
#   basis.
```

Link to the lockf stress-ng test source code: https://github.com/ColinIanKing/stress-ng/blob/master/stress-lockf.c

This test when run under Graphene hangs for any number of stressors. I have attached the logs for lockf running with 8 stressors as well as 1 stressor.
[lockf_8stressor.log](https://github.com/oscarlab/graphene/files/6768646/lockf_8stressor.log)
[lockf_1stressor.log](https://github.com/oscarlab/graphene/files/6768647/lockf_1stressor.log)

## Steps to reproduce

Commands to run the lockf stressor
```
graphene-direct stress-ng --lockf 1 --timeout 60s 
graphene-direct stress-ng --lockf 8 --timeout 60s 
```

pwmarcz:
## Description of the changes 

This makes it possible for POSIX locks (`F_SETLKW`) to return `-EINTR` when interrupted, e.g. after receiving a signal from `alarm()`. See #2517 for design discussion. 

Fixes #2510.

## How to test this PR? 

* I added some test cases to `fcntl_lock`
* The `stress-ng` lockf test should no longer hang (see #2510)
* LTP `fcntl16` test is reenabled (the test depends on being able to interrupt a lock request, unfortunately it's racy and the interrupt is not always necessary)

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2522)

### DIFF
--- a/LibOS/shim/include/shim_fs_lock.h
+++ b/LibOS/shim/include/shim_fs_lock.h
@@ -111,7 +111,23 @@ int posix_lock_clear_pid(IDTYPE pid);
  * to add a lock (-EAGAIN, -ENOMEM etc.) will be sent in the response instead.
  */
 int posix_lock_set_from_ipc(const char* path, struct posix_lock* pl, bool wait, IDTYPE vmid,
-                            unsigned long seq);
+                            uint64_t seq);
+
+/*!
+ * \brief Cancel a lock request (IPC handler)
+ *
+ * \param path absolute path for a file
+ * \param vmid target process for IPC response
+ * \param seq sequence number for IPC response
+ *
+ * Cancels the request previously added by `posix_lock_set_from_ipc`: if the request with given
+ * `vmid` and `seq` parameters is still unresolved, it is removed and a response with -EINTR is
+ * sent. If the request has been resolved in the meantime, nothing happens.
+ *
+ * This function should be called if the wait for IPC response has been interrupted. It ensures that
+ * a response will be sent immediately.
+ */
+int posix_lock_cancel_from_ipc(const char* path, IDTYPE vmid, uint64_t seq);
 
 /*!
  * \brief Check for conflicting locks on a file (IPC handler)
--- a/LibOS/shim/include/shim_fs_lock.h
+++ b/LibOS/shim/include/shim_fs_lock.h
@@ -34,7 +34,6 @@ int init_fs_lock(void);
  * - The main process has to be able to look up the same file, so locking will not work for files in
  *   local-process-only filesystems (tmpfs).
  * - There is no deadlock detection (EDEADLK).
- * - The lock requests cannot be interrupted (EINTR).
  * - The locks work only on files that have a dentry (no pipes, sockets etc.)
  */
 
--- a/LibOS/shim/include/shim_ipc.h
+++ b/LibOS/shim/include/shim_ipc.h
@@ -33,6 +33,7 @@ enum {
     IPC_MSG_SYNC_CONFIRM_DOWNGRADE,
     IPC_MSG_SYNC_CONFIRM_CLOSE,
     IPC_MSG_POSIX_LOCK_SET,
+    IPC_MSG_POSIX_LOCK_CANCEL,
     IPC_MSG_POSIX_LOCK_GET,
     IPC_MSG_POSIX_LOCK_CLEAR_PID,
     IPC_MSG_CODE_BOUND,
@@ -117,6 +118,28 @@ int ipc_send_message(IDTYPE dest, struct shim_ipc_msg* msg);
  * discarded, but still awaited for.
  */
 int ipc_send_msg_and_get_response(IDTYPE dest, struct shim_ipc_msg* msg, void** resp);
+/*!
+ * \brief Send an IPC message and wait for a response (cancel on interrupt)
+ *
+ * \param dest vmid of the destination process
+ * \param msg message to send
+ * \param on_cancel callback to invoke when the wait is interrupted
+ * \param arg argument for the callback
+ * \param[out] resp upon successful return contains a pointer to the response
+ *
+ * A version of `ipc_send_msg_and_get_response` that allows canceling a request after the wait is
+ * interrupted. Intended for implementation of interruptible function calls.
+ *
+ * Behaves as `ipc_send_msg_and_get_response`, but after the wait for response is interrupted,
+ * invokes `on_cancel(dest, arg)`. If it succeeds, the wait is resumed. The callback is invoked only
+ * once: if the wait is interrupted again afterwards, the function keeps waiting.
+ *
+ * The \p on_cancel callback should return 0 on success, or a negative error code on failure. If
+ * the callback fails, the function does not resume the wait, but returns the error immediately.
+ */
+int ipc_send_msg_and_get_response_with_cancel(IDTYPE dest, struct shim_ipc_msg* msg,
+                                              int (*on_cancel)(IDTYPE, void*), void* arg,
+                                              void** resp);
 /*!
  * \brief Broadcast an IPC message
  *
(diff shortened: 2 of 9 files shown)


dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2522)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2522#-Me9J9kmCnnSNtrOGWQ9:-Me9J9kn1Ly13Y1zAVqB:b-ofgp2l) (no related file):*
Can you amend your commit to have the text from the PR description? Sounds very helpful in understanding what this commit resolves.

---
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2522#-Me9YAmI01dp0t5J1qxR:-Me9YAmI01dp0t5J1qxS:b90k7ok) (no related file):*
I found this change very readable. I definitely like this Option 4.

---
___
*[LibOS/shim/src/fs/shim_fs_lock.c, line 517 at r1](https://reviewable.io/reviews/oscarlab/graphene/2522#-Me9Ssf-9a0kfgdGUdoW:-Me9Ssf02H4iadT3nD9s:b-ebki6s) ([raw file](https://github.com/oscarlab/graphene/blob/f9d9631e174ab5bd10bf5d08e2e28832f71d6072/LibOS/shim/src/fs/shim_fs_lock.c#L517)):*
> ```C
>          * -EINTR. Either way, we delete the request.
>          */
>         ret = req->resolved ? req->result : -EINTR;
> ```

But `ret` could be something other than "interrupted" (if `DkEventWait` failed in a weird way). I would expect to return this error then, or have `die_or_inf_loop()` after the message "unexpected error from DkEventWait".
___
*[LibOS/shim/src/ipc/shim_ipc_fs_lock.c, line 54 at r1](https://reviewable.io/reviews/oscarlab/graphene/2522#-Me9X-VA1fvqsMkUsqXF:-Me9X-VA1fvqsMkUsqXG:b-6odcw8) ([raw file](https://github.com/oscarlab/graphene/blob/f9d9631e174ab5bd10bf5d08e2e28832f71d6072/LibOS/shim/src/ipc/shim_ipc_fs_lock.c#L54)):*
> ```C
>     void* data;
>     int ret = ipc_send_msg_and_get_response_with_cancel(
>         g_process_ipc_ids.leader_vmid, msg, &cancel_posix_lock, msg, &data);
> ```

I first wanted to suggest that only `F_SETLKW` uses this `with_cancel` version, but then I noticed that the man page allows `EINTR` even for `F_SETLK`/`F_GETLK`. So you are correct here.
___
*[LibOS/shim/test/regression/fcntl_lock.c, line 458 at r1](https://reviewable.io/reviews/oscarlab/graphene/2522#-Me9Xraf8z0ecDXlghHH:-Me9Xrag54HINzZasRhM:b-gfkx6) ([raw file](https://github.com/oscarlab/graphene/blob/f9d9631e174ab5bd10bf5d08e2e28832f71d6072/LibOS/shim/test/regression/fcntl_lock.c#L458)):*
> ```C
> 
>     if (pid == 0) {
>         lock(F_RDLCK, 0, 100);
> ```

Why not `F_WRLCK`? Just to cover more cases?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2522)* status: 8 of 9 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2522#-Me9J9kmCnnSNtrOGWQ9:-Me9apx019wSvCj76RY_:b-896fix) (no related file):*

Can you amend your commit to have the text from the PR description? Sounds very helpful in understanding what this commit resolves.
</blockquote></details>

Done.

---
___
*[LibOS/shim/src/fs/shim_fs_lock.c, line 517 at r1](https://reviewable.io/reviews/oscarlab/graphene/2522#-Me9Ssf-9a0kfgdGUdoW:-Me9au9a6xCkC922FO88:blegylc) ([raw file](https://github.com/oscarlab/graphene/blob/f9d9631e174ab5bd10bf5d08e2e28832f71d6072/LibOS/shim/src/fs/shim_fs_lock.c#L517)):*

But `ret` could be something other than "interrupted" (if `DkEventWait` failed in a weird way). I would expect to return this error then, or have `die_or_inf_loop()` after the message "unexpected error from DkEventWait".
</blockquote></details>

This is a race: if `DkEventWait` failed in a weird way, the lock operation still might have finished with success, in which case returning something else than 0 would be very confusing for the user. That's why I didn't want to return the error from `DkEventWait`.

Do you think `die_or_inf_loop()` is a good solution here? I did not want to do that either, because a failed wait shouldn't have any impact on *correctness* of the lock operation, so I don't think it should be a fatal error.

Another solution would be to return the error from `DkEventWait`, but *only* if the operation has not finished yet (i.e. `ret = req->resolved ? req->result : ret_from_wait`). I'm not sure if that's useful, though.
___
*[LibOS/shim/src/ipc/shim_ipc_fs_lock.c, line 54 at r1](https://reviewable.io/reviews/oscarlab/graphene/2522#-Me9X-VA1fvqsMkUsqXF:-Me9df7B6zU2OJKe6b9g:blg7pbk) ([raw file](https://github.com/oscarlab/graphene/blob/f9d9631e174ab5bd10bf5d08e2e28832f71d6072/LibOS/shim/src/ipc/shim_ipc_fs_lock.c#L54)):*

I first wanted to suggest that only `F_SETLKW` uses this `with_cancel` version, but then I noticed that the man page allows `EINTR` even for `F_SETLK`/`F_GETLK`. So you are correct here.
</blockquote></details>

Graphene won't fail with `EINTR` on these, though. We will send the cancel request, but it will not find any pending operation, so the original requests will proceed. The same will happen for `F_SETLKW` for a lock that can be granted immediately.
___
*[LibOS/shim/test/regression/fcntl_lock.c, line 458 at r1](https://reviewable.io/reviews/oscarlab/graphene/2522#-Me9Xraf8z0ecDXlghHH:-Me9eacI1m213e_H9Anq:b82vqg) ([raw file](https://github.com/oscarlab/graphene/blob/f9d9631e174ab5bd10bf5d08e2e28832f71d6072/LibOS/shim/test/regression/fcntl_lock.c#L458)):*

Why not `F_WRLCK`? Just to cover more cases?
</blockquote></details>

Oh, I can do `F_WRLCK`, same as in the other "eitr" test. Changed.

I'm not very consistent with these, though; some other tests check `F_RDLCK` vs. `F_WRLCK`.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2522)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

___
*[LibOS/shim/src/fs/shim_fs_lock.c, line 517 at r1](https://reviewable.io/reviews/oscarlab/graphene/2522#-Me9Ssf-9a0kfgdGUdoW:-MeAkB9X-IA1qG7UGTb5:b-88dhlt) ([raw file](https://github.com/oscarlab/graphene/blob/f9d9631e174ab5bd10bf5d08e2e28832f71d6072/LibOS/shim/src/fs/shim_fs_lock.c#L517)):*

This is a race: if `DkEventWait` failed in a weird way, the lock operation still might have finished with success, in which case returning something else than 0 would be very confusing for the user. That's why I didn't want to return the error from `DkEventWait`.

Do you think `die_or_inf_loop()` is a good solution here? I did not want to do that either, because a failed wait shouldn't have any impact on *correctness* of the lock operation, so I don't think it should be a fatal error.

Another solution would be to return the error from `DkEventWait`, but *only* if the operation has not finished yet (i.e. `ret = req->resolved ? req->result : ret_from_wait`). I'm not sure if that's useful, though.
</blockquote></details>

OK, understood your point. Given that it shouldn't happen in practice, and we have a log warning for this case, I'm resolving.

pwmarcz:
Closing as @boryspoplawski (https://github.com/oscarlab/graphene/issues/2517#issuecomment-877881232) wants not to use IPC responses for this. I'll be trying to rewrite this subsystem (and think how it ties with the sync engine).

PR #2546, opened 2021-07-14
[Pal/Linux-SGX] Restrict Untrusted RSP to point outside the enclave

dimakuv:
## Description of the changes 

Untrusted RSP value (`SSA[0].ursp`) is not sanitized by the SGX hardware upon EENTER/ERESUME and may be controlled by the attacker. To prevent any possibility of tampering with this value, we simply disallow it to point inside the SGX enclave.

## How to test this PR? 

All tests must pass.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2546)

### DIFF
--- a/Pal/src/host/Linux-SGX/enclave_entry.S
+++ b/Pal/src/host/Linux-SGX/enclave_entry.S
@@ -169,8 +169,23 @@ enclave_entry:
 1:
 
 	movq %gs:SGX_GPR, %rbx
-
 	movq %rdi, %rsi
+
+	# note that g_enclave_base/g_enclave_top are initialized only after the first ENCLAVE_START
+	# ecall, and below check always fails before this initialization is done -- this limitation is
+	# fine because we don't expect exceptions until the enclave is actually started
+	movq SGX_GPR_URSP(%rbx), %rdi
+	movq g_enclave_base(%rip), %rax
+	cmpq %rax, %rdi
+	jb 1f
+	movq g_enclave_top(%rip), %rax
+	cmpq %rax, %rdi
+	ja 1f
+
+	# detected malicious Untrusted RSP (g_enclave_base <= ursp && ursp <= g_enclave_top)
+	FAIL_LOOP
+1:
+
 	xorq %rdi, %rdi
 	movl SGX_GPR_EXITINFO(%rbx), %edi
 	testl $0x80000000, %edi
--- a/Pal/src/host/Linux-SGX/generated-offsets.c
+++ b/Pal/src/host/Linux-SGX/generated-offsets.c
@@ -50,6 +50,7 @@ __attribute__((__used__)) static void dummy(void) {
     OFFSET_T(SGX_GPR_R15, sgx_pal_gpr_t, r15);
     OFFSET_T(SGX_GPR_RFLAGS, sgx_pal_gpr_t, rflags);
     OFFSET_T(SGX_GPR_RIP, sgx_pal_gpr_t, rip);
+    OFFSET_T(SGX_GPR_URSP, sgx_pal_gpr_t, ursp);
     OFFSET_T(SGX_GPR_EXITINFO, sgx_pal_gpr_t, exitinfo);
     DEFINE(SGX_GPR_SIZE, sizeof(sgx_pal_gpr_t));
 


boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2546)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2546#-MeZR7LYAT1sTkNOjiie:-MeZR7LYAT1sTkNOjiif:b-1ufbe8) (no related file):*
`ursp` is sanitized on ecalls (https://github.com/oscarlab/graphene/blob/4cb98219d8e302055587a8952c1102415a72ba42/Pal/src/host/Linux-SGX/enclave_ecalls.c#L70) and the value from `ssa[0].ursp` seems not to be used anywhere? If that's the case why do we need this sanitization at all (or if that's just a sanity check, why not simply set that to some Linux x64 invalid pointer)?

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2546)* status: 1 of 2 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2546#-MeZR7LYAT1sTkNOjiie:-Me_Lv2B0dcKDcQMOtmX:b-8vfut5) (no related file):*

`ursp` is sanitized on ecalls (https://github.com/oscarlab/graphene/blob/4cb98219d8e302055587a8952c1102415a72ba42/Pal/src/host/Linux-SGX/enclave_ecalls.c#L70) and the value from `ssa[0].ursp` seems not to be used anywhere? If that's the case why do we need this sanitization at all (or if that's just a sanity check, why not simply set that to some Linux x64 invalid pointer)?
</blockquote></details>

We discussed with @boryspoplawski internally, and agreed that this fix is superfluous.

Let me try to summarize.

- The bug assumes having a non-{sanitized/updated} value in `ssa[0].rsp`. There are only two stack frames possible in Graphene (`NSSA = 2`), so `ssa[1]` is never even created.
- There is a single place in Graphene code where `ssa[0].rsp` is read and used as an offset to writing other values: https://github.com/oscarlab/graphene/blob/4cb98219d8e302055587a8952c1102415a72ba42/Pal/src/host/Linux-SGX/enclave_entry.S#L370. There are four flows that can lead to this line of code:
    - Very first ECALL on the enclave thread (process_start or thread_start): but then `SGX_READY_FOR_EXCEPTIONS == 0`  and this gate line stops the enclave: https://github.com/oscarlab/graphene/blob/4cb98219d8e302055587a8952c1102415a72ba42/Pal/src/host/Linux-SGX/enclave_entry.S#L198
    - Somewhere during enclave-code execution: then `ssa[0].rsp` is obviously trusted
    - Return from OCALL: then two cases after AEX:
        - `SGX_PRE_OCALL_STACK == 0`  -- then RSP was already restored to the previously-saved value and `ssa[0].rsp` contains this good value https://github.com/oscarlab/graphene/blob/4cb98219d8e302055587a8952c1102415a72ba42/Pal/src/host/Linux-SGX/enclave_entry.S#L671
        - `SGX_PRE_OCALL_STACK != 0`  -- then RSP is restored from `SGX_PRE_OCALL_STACK`; via the scratch register RSI: https://github.com/oscarlab/graphene/blob/4cb98219d8e302055587a8952c1102415a72ba42/Pal/src/host/Linux-SGX/enclave_entry.S#L327
    - `THREAD_RESET` ECALL: this only happens after `OCALL_EXIT` which implies `SGX_PRE_OCALL_STACK == 0`. Even if AEX happens (before line https://github.com/oscarlab/graphene/blob/4cb98219d8e302055587a8952c1102415a72ba42/Pal/src/host/Linux-SGX/enclave_entry.S#L124), we are still safe because of this gate line: https://github.com/oscarlab/graphene/blob/4cb98219d8e302055587a8952c1102415a72ba42/Pal/src/host/Linux-SGX/enclave_entry.S#L231

---

PR #2584, opened 2021-07-27
[LibOS] Rewrite string-based files

pwmarcz:
## Description of the changes 

This change decouples the two usages of "str" filesystem: "pseudo" filesystems, and `tmpfs`. The difference between them is that in "pseudo" filesystems, the file data is stored directly in handle.

Instead of `str_*` functions, there is a simpler set of `mem_file_*` functions that do not operate on handles or dentries, just on its own buffer.

The decoupling makes both use-cases simpler: there is no structure "inheritance" and no extra layer of reference counting.

## Next steps

* I converted file position from a pointer to `file_off_t`, so it brings us a step closer to pushing the file position upwards (so that it's not handled by individual filesystems).
* I got rid of the reference counting here, which unblocks fix to dentry lifecycle (I made a first attempt in #2499 but got blocked by this).
  * Instead of a bit field, the dentry state can be an enum; something like "new -> negative (file doesn't exist yet) -> valid -> deleted (after unlink)".
* The last filesystem I want to rewrite is `chroot`, but it turns out that it has some more advanced handling of deleted files, so I'll be trying to include that in dentry lifecycle.
  * To be more specific, the `chroot` filesystem allows you to have a handle to "old" deleted file, and a dentry to the "new" one; this is done using a `version` field. Instead, I want these files to be described by two separate dentries.

## How to test this PR? 

The existing tests should be enough.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 169 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfeTwvL2zxLRumQSbZ1:-MfeTwvL2zxLRumQSbZ2:bi7dg0) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L169)):*
> ```C
>     if ((uintptr_t)dent1 < (uintptr_t)dent2) {
>         unlock(&dent1->lock);
>         unlock(&dent2->lock);
> ```

Not the other way around?

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: all files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @mkow and @pwmarcz)

___
*[LibOS/shim/include/shim_fs_mem.h, line 19 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg3LWZEKIc1Nkjnnv5:-Mfg3LWZEKIc1Nkjnnv6:b-kh175k) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_fs_mem.h#L19)):*
> ```objc
>     char* buf;
>     file_off_t size;
>     file_off_t buf_size;
> ```

Shouldn't this be `size_t`? Because it is the size of the object in memory, not the size of the file.
___
*[LibOS/shim/include/shim_fs_mem.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg4HTZ4HD3u8mZoSkI:-Mfg4HTZ4HD3u8mZoSkJ:bq7o8du) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_fs_mem.h#L26)):*
> ```objc
> 
> /*
>  * The following operations correspond to the filesystem callbacks (see `shim_fs.h`). Note that the
> ```

Correspond? They do not in terms of their arguments...
___
*[LibOS/shim/include/shim_fs_mem.h, line 27 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg3cfT9YqRDNEeSQdL:-Mfg3cfT9YqRDNEeSQdM:b2rui28) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_fs_mem.h#L27)):*
> ```objc
> /*
>  * The following operations correspond to the filesystem callbacks (see `shim_fs.h`). Note that the
>  * caller has to pass the file position, and (in case of `read` and `write` update it after a
> ```

Forgot the `)`
___
*[LibOS/shim/include/shim_handle.h, line 34 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg5bTvCKvl1apSo2s1:-Mfg5bTw0j-pRuK-rmFk:bonvor1) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_handle.h#L34)):*
> ```objc
>     TYPE_FILE,       /* host files, used by `chroot` filesystem */
>     TYPE_DEV,        /* emulated devices, used by `dev` filesystem */
>     TYPE_STR,        /* string-based files (with data inside handle), handled by `pseudo_*`
> ```

Why do string-based files have data inside the handle and not inside the dentry? Is it just because it is simpler to write code? (I understand that with pseudo-files it is impossible to have several handles on the same dentry, so I understand that this solution is possible, just curious as to why have this distinction).
___
*[LibOS/shim/include/shim_handle.h, line 35 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg5ALn43lSDFdFOaSk:-Mfg5ALn43lSDFdFOaSl:bgxazhp) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_handle.h#L35)):*
> ```objc
>     TYPE_DEV,        /* emulated devices, used by `dev` filesystem */
>     TYPE_STR,        /* string-based files (with data inside handle), handled by `pseudo_*`
>                       * functions, used by several filesystems */
> ```

Is it true that it is "used by several filesystems"? I guess you mean sysfs, procfs?
___
*[LibOS/shim/include/shim_handle.h, line 181 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg4xEz91rb-nR24ch9:-Mfg4xEz91rb-nR24chA:b9c55j) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_handle.h#L181)):*
> ```objc
> struct shim_tmpfs_data {
>     struct shim_mem_file mem;
> };
> ```

This struct looks like an internal detail of `tmpfs/fs.c`. Maybe move it there?
___
*[LibOS/shim/src/fs/shim_fs_mem.c, line 10 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg7-0WE3KxSGgRn9pQ:-Mfg7-0WE3KxSGgRn9pR:b-410uxh) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/shim_fs_mem.c#L10)):*
> ```C
> #include "shim_fs_mem.h"
> 
> static int mem_file_resize(struct shim_mem_file* mem, file_off_t buf_size) {
> ```

Shouldn't `buf_size` be of type `size_t`?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 3 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgB5g40y7izP_NS25l:-MfgB5g508pi8K3dv1Wu:b-qoc0vs) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L3)):*
> ```C
> /* Copyright (C) 2021 Intel Corporation
>  *                    Li Xun <xun.li@intel.com>
> ```

Please add your name here
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 28 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgB_Ci4TtqRI9G8SeP:-MfgB_Ci4TtqRI9G8SeQ:b-biifte) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L28)):*
> ```C
> 
> /* Get data associated with dentry. This is created on demand, instead of during dentry validation,
>  * because dentries restored from checkpoint don't have the `data` field filled. */
> ```

I don't understand this second part, `dentries restored from checkpoint...` -- but we do not checkpoint `shim_tmpfs_data` during checkpointing. So does it mean that the child process will see the tmpfs dentry, but it will be empty? This is very weird semantics, why use it?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 42 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgCN0DBdgIwUek5Ivk:-MfgCN0DBdgIwUek5Ivl:bsmjos1) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L42)):*
> ```C
>     mem_file_init(&data->mem, /*data=*/NULL, /*size=*/0);
> 
>     *out_data = dent->data = data;
> ```

I highly dislike this pattern (my brain is wired to read it as `*out_data = dent->data == data`), please split into two lines.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 60 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgCicFEEA5TVQUE8pa:-MfgCicFEEA5TVQUE8pb:b-ylc6e0) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L60)):*
> ```C
>     if (!dent->parent) {
>         /* This is the root dentry, initialize it. */
>         dent->state |= DENTRY_ISDIRECTORY;
> ```

This should be removed after https://github.com/oscarlab/graphene/pull/2587 is merged.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 116 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgEoPg4TIPS5lhdBQQ:-MfgEoPg4TIPS5lhdBQR:bxv2fnt) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L116)):*
> ```C
> static int tmpfs_readdir(struct shim_dentry* dent, readdir_callback_t callback, void* arg) {
>     struct shim_dentry* child;
>     LISTP_FOR_EACH_ENTRY(child, &dent->children, siblings) {
> ```

Why don't you grab the dentry lock before this loop? Is it already taken by the caller? If yes, can we add an assert?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 131 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgEzo85V2Hs-Fk5MZl:-MfgEzo85V2Hs-Fk5MZm:bxv2fnt) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L131)):*
> ```C
>     if (dent->type == S_IFDIR) {
>         struct shim_dentry* child;
>         LISTP_FOR_EACH_ENTRY(child, &dent->children, siblings) {
> ```

Why don't you grab the dentry lock before this loop? Is it already taken by the caller? If yes, can we add an assert?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 140 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgFbeZ79XOFiVclQS1:-MfgFbeZ79XOFiVclQS2:b-jx604b) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L140)):*
> ```C
>     /* TODO: our `unlink` wipes the file data, even though there might be still file handles
>      * pointing to the file. A proper solution would keep the unlinked file until all handles are
>      * closed, but also allow creating a new file with the same name. */
> ```

This sounds very dangereous, we currently allow dangling pointers. Firstly, why can't we implement the proper solution yet? Is it because we would need to rewire the handle to `shim_tmpfs_data` in case of dentry deletion -- and this sounds like it needs a proper, non-trivial code changes?

Secondly, wouldn't it be better to have a memory leak here, instead of dangling pointers? Let's just not free this memory region. UPDATE: Actually, we don't have mmap for tmpfs, so how could the file handle point to the file after dentry deletion?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 169 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfeTwvL2zxLRumQSbZ1:-MfgG_T9AnHRy1BdHZOi:bovatqq) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L169)):*

Not the other way around?
</blockquote></details>

+1 to Michal.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 178 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgGjCf5QBlb_RMdbe1:-MfgGjCf5QBlb_RMdbe2:b-w8tpoa) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L178)):*
> ```C
> static int tmpfs_rename(struct shim_dentry* old, struct shim_dentry* new) {
>     lock_two_dentries(old, new);
>     new->data = old->data;
> ```

Is it possible that `new->data` already points to some data? If it's impossible, can we add an assert about it?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 238 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgHH273arG8UuWn1H2:-MfgHH273arG8UuWn1H3:b-1qn7z6) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L238)):*
> ```C
>     assert(hdl->type == TYPE_TMPFS);
> 
>     __UNUSED(hdl);
> ```

This can be removed.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 239 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgIX0BBCG2m-E9VbeZ:-MfgIX0IEjfbNZGsUfot:bnkj5lf) ([raw file](https://github.com/oscarlab/graphene/blob/6ce34fb73ddb0d480c0281aaf7c3c03ac6f1fc44/LibOS/shim/src/fs/tmpfs/fs.c#L239)):*
> ```C
> 
>     tmpfs_data->ctime = time / 1000000;
>     tmpfs_data->mtime = tmpfs_data->ctime;
> ```

We previously updated this create/modify timestamps on tmpfs operations. We don't want to do this anymore? Is it because they are considered useless?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 240 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgHhP7EQA4zyqs2KYr:-MfgHhP7EQA4zyqs2KYs:bm3bt5) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L240)):*
> ```C
>     __UNUSED(hdl);
> 
>     lock(&hdl->dentry->lock);
> ```

I'm getting confused again regarding the handle and its associated dentry... Can the handle change the dentry in the meantime? If yes, shouldn't you `lock(&hdl->lock)`?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: 5 of 13 files reviewed, 19 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[LibOS/shim/include/shim_fs_mem.h, line 19 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg3LWZEKIc1Nkjnnv5:-Mfg_Ban2d-BvbvSh02B:b9vjytb) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_fs_mem.h#L19)):*

Shouldn't this be `size_t`? Because it is the size of the object in memory, not the size of the file.
</blockquote></details>

Yeah, unfortunately it's awkward either way...  Updated.
___
*[LibOS/shim/include/shim_fs_mem.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg4HTZ4HD3u8mZoSkI:-MfgXNEm4Tn1m_ilWech:b-spyatt) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_fs_mem.h#L26)):*

Correspond? They do not in terms of their arguments...
</blockquote></details>

I reworded the comment. I still think the API is pretty self-explanatory, so adding proper Doxygen comments would be an overkill. Do you agree?
___
*[LibOS/shim/include/shim_fs_mem.h, line 27 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg3cfT9YqRDNEeSQdL:-MfgY08qE_URIdhM2ZjS:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_fs_mem.h#L27)):*

Forgot the `)`
</blockquote></details>

Fixed.
___
*[LibOS/shim/include/shim_handle.h, line 34 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg5bTvCKvl1apSo2s1:-Mfg_dXR8CpOk3JuWPhZ:blidwkz) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_handle.h#L34)):*

Why do string-based files have data inside the handle and not inside the dentry? Is it just because it is simpler to write code? (I understand that with pseudo-files it is impossible to have several handles on the same dentry, so I understand that this solution is possible, just curious as to why have this distinction).
</blockquote></details>

You can have several file handles on the same entry, say, if you open `/proc/cpuinfo` multiple times.

However, it's easier to store the data in the dentry because then it can be computed on-the-fly, and changed every time you open it again. 

From a (very) brief investigation, this is also [what Linux does](https://code.woboq.org/linux/linux/fs/proc/cpuinfo.c.html#cpuinfo_open): if you open a pseudo-file `/proc/cpuinfo`, it ends up calling a callback that computes the file content and associates it with the file handle.
___
*[LibOS/shim/include/shim_handle.h, line 35 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg5ALn43lSDFdFOaSk:-MfgbLno3jJKyYzxDKj2:b-e1yl7z) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_handle.h#L35)):*

Is it true that it is "used by several filesystems"? I guess you mean sysfs, procfs?
</blockquote></details>

Yes. I think just "handled by `pseudo_*` functions" will be clearer.
___
*[LibOS/shim/include/shim_handle.h, line 181 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg4xEz91rb-nR24ch9:-MfgbY2d-Qq7eRMG9vb_:b-qgwpf1) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_handle.h#L181)):*

This struct looks like an internal detail of `tmpfs/fs.c`. Maybe move it there?
</blockquote></details>

OK. I thought there was a precedent for keeping these `shim_*_data` definitions in `shim_handle.h` as well, but right now that's only for `chroot` (which I want to also change).
___
*[LibOS/shim/src/fs/shim_fs_mem.c, line 10 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg7-0WE3KxSGgRn9pQ:-Mfgblg03KpmDt_lvVlA:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/shim_fs_mem.c#L10)):*

Shouldn't `buf_size` be of type `size_t`?
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 3 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgB5g40y7izP_NS25l:-MfgbmrYDjTqeh76Wdro:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L3)):*

Please add your name here
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 28 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgB_Ci4TtqRI9G8SeP:-MfgbpZW6owN47tPyyd1:bl38ckc) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L28)):*

I don't understand this second part, `dentries restored from checkpoint...` -- but we do not checkpoint `shim_tmpfs_data` during checkpointing. So does it mean that the child process will see the tmpfs dentry, but it will be empty? This is very weird semantics, why use it?
</blockquote></details>

Well, this is the default behaviour of Graphene: dentries are checkpointed but the `data` field is wiped. It seems hard to do something else right now.

Longer term, the filesystem probably should have callbacks for checkpointing data. There is something like that for handles (`checkin` / `checkout`), but it's almost unused, and I'm afraid to build anything advanced on top of the current checkpointing system.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 42 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgCN0DBdgIwUek5Ivk:-Mfgd6ShDDcrFQHg3Y0d:b48u7lk) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L42)):*

I highly dislike this pattern (my brain is wired to read it as `*out_data = dent->data == data`), please split into two lines.
</blockquote></details>

OK, I'll keep that in mind. Done.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 116 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgEoPg4TIPS5lhdBQQ:-MfgdME5510NJCUwROCH:bfw07ao) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L116)):*

Why don't you grab the dentry lock before this loop? Is it already taken by the caller? If yes, can we add an assert?
</blockquote></details>

It is taken. I'm adding an assert.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 131 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgEzo85V2Hs-Fk5MZl:-Mfge4Fd9EHlJrqfYUMa:bapbf7p) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L131)):*

Why don't you grab the dentry lock before this loop? Is it already taken by the caller? If yes, can we add an assert?
</blockquote></details>

Unfortunately it's not taken. I'm taking the lock here, but this means the unlink implementation is racy :( I added a TODO.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 140 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgFbeZ79XOFiVclQS1:-MfgeNyXDGsVapbpvd3q:bqft0av) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L140)):*

This sounds very dangereous, we currently allow dangling pointers. Firstly, why can't we implement the proper solution yet? Is it because we would need to rewire the handle to `shim_tmpfs_data` in case of dentry deletion -- and this sounds like it needs a proper, non-trivial code changes?

Secondly, wouldn't it be better to have a memory leak here, instead of dangling pointers? Let's just not free this memory region. UPDATE: Actually, we don't have mmap for tmpfs, so how could the file handle point to the file after dentry deletion?
</blockquote></details>

Edited: `s/pointing to the file/associated with the dentry/`

Does that make more sense? There are no dangling pointers, it's just that after `unlink` the open file suddenly becomes empty.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 169 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfeTwvL2zxLRumQSbZ1:-MfgemAvBk_ib7dDd0fb:b-15s3ow) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L169)):*

+1 to Michal.
</blockquote></details>

Right! Fixed.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 178 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgGjCf5QBlb_RMdbe1:-Mfgez6f979Sy70dOV2T:b-xp8nac) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L178)):*

Is it possible that `new->data` already points to some data? If it's impossible, can we add an assert about it?
</blockquote></details>

Right, `new` can be an existing file, with associated data. I need to delete it here.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 238 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgHH273arG8UuWn1H2:-Mfgidxc2YGyK8NSsLIj:bxu51gv) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L238)):*

This can be removed.
</blockquote></details>

My mistake. Removed.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 239 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgIX0BBCG2m-E9VbeZ:-MfgtPJf9BpuLCPwIVa8:b-udzx4b) ([raw file](https://github.com/oscarlab/graphene/blob/6ce34fb73ddb0d480c0281aaf7c3c03ac6f1fc44/LibOS/shim/src/fs/tmpfs/fs.c#L239)):*

We previously updated this create/modify timestamps on tmpfs operations. We don't want to do this anymore? Is it because they are considered useless?
</blockquote></details>

Looks like this got more attention than I initially assumed: in the original PR for tmpfs there is some discussion about updating the times. I'll keep them.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 240 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgHhP7EQA4zyqs2KYr:-Mfgigpf4GagXAFX-zMP:bwcyibb) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L240)):*

I'm getting confused again regarding the handle and its associated dentry... Can the handle change the dentry in the meantime? If yes, shouldn't you `lock(&hdl->lock)`?
</blockquote></details>

`hdl->dentry` doesn't change during the dentry's lifetime. Then again, it's less confusing to take the lock (since this is the only place that doesn't do it). Updated.

dimakuv (changes requested):
> Reviewed 8 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[LibOS/shim/include/shim_fs_mem.h, line 26 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg4HTZ4HD3u8mZoSkI:-Mfh4rKs0GWFfvqzAx3S:b27e598) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_fs_mem.h#L26)):*

I reworded the comment. I still think the API is pretty self-explanatory, so adding proper Doxygen comments would be an overkill. Do you agree?
</blockquote></details>

LGTM. Yes, no need for detailed comments.
___
*[LibOS/shim/include/shim_handle.h, line 34 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg5bTvCKvl1apSo2s1:-Mfh669_54FnU4x4JQLT:b-ias1t3) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_handle.h#L34)):*
> You can have several file handles on the same entry, say, if you open /proc/cpuinfo multiple times.

Oops, true, my mistake.

But then doesn't it mean that we will have 2 copies of the same string data when we open `/proc/cpuinfo` twice? And what happens if we open e.g. `/dev/attestation/user_report_data` twice for write? We can end up with two disparate copies?

> However, it's easier to store the data in the dentry...

Is this a typo? Did you mean "handle" instead of "dentry"?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 28 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgB_Ci4TtqRI9G8SeP:-Mfh9KppBQ-uRC6Vh0mT:b-w5uzrd) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L28)):*

Well, this is the default behaviour of Graphene: dentries are checkpointed but the `data` field is wiped. It seems hard to do something else right now.

Longer term, the filesystem probably should have callbacks for checkpointing data. There is something like that for handles (`checkin` / `checkout`), but it's almost unused, and I'm afraid to build anything advanced on top of the current checkpointing system.
</blockquote></details>

Fair enough. But could you add a comment that this is "conforming to the current behavior of Graphene ..."?
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 140 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgFbeZ79XOFiVclQS1:-MfhCfPKDgHunVi4UYvB:b-q7ii9e) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L140)):*

Edited: `s/pointing to the file/associated with the dentry/`

Does that make more sense? There are no dangling pointers, it's just that after `unlink` the open file suddenly becomes empty.
</blockquote></details>

Yes, I understand what happens now.
___
*[LibOS/shim/test/fs/test_fs.py, line 316 at r2](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfhDX_r0w3nAaZGdVTe:-MfhDX_r0w3nAaZGdVTf:b-wm2vnv) ([raw file](https://github.com/oscarlab/graphene/blob/940cc4846ccccebd4131c9a319ad33c601064ee4/LibOS/shim/test/fs/test_fs.py#L316)):*
> ```Python
>         self.verify_copy(stdout, stderr, '/mounted/input', executable)
> 
>     def test_300_rename(self):
> ```

This test won't work on protected files right now: https://github.com/oscarlab/graphene/pull/2521

You may want to disable this test on protected files (I believe currently all tests from `test_fs.py` are inherited by `test_pf.py`).
___
*[LibOS/shim/test/fs/test_fs.py, line 321 at r2](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfhFzJfBhfB8BAe2NZU:-MfhFzJfBhfB8BAe2NZV:bw0hvll) ([raw file](https://github.com/oscarlab/graphene/blob/940cc4846ccccebd4131c9a319ad33c601064ee4/LibOS/shim/test/fs/test_fs.py#L321)):*
> ```Python
> 
>         stdout, stderr = self.run_binary(['rename', file1, file2])
>         self.assertIn('TEST OK', stdout)
> ```

This whole test doesn't seem to belong to `test/fs/`. Maybe better to move it under `test/regression/`?

The `test/fs/` tests are mainly to check copying of a bunch of files (created in `setUpClass()` above). But in this test, you just have two files.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: 6 of 18 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[LibOS/shim/include/shim_handle.h, line 34 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg5bTvCKvl1apSo2s1:-MfhGTdt2LUJsGv0IcX1:bt1c1df) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_handle.h#L34)):*
>But then doesn't it mean that we will have 2 copies of the same string data when we open `/proc/cpuinfo` twice?

Yes, you will have two copies of the same string data, each living as long as the handle is open.

>And what happens if we open e.g. /dev/attestation/user_report_data twice for write? We can end up with two disparate copies?

Yes. A `close` or `fsync` on a handle will cause the attestation driver to update the report data, but will not update what is visible in the other handle.

This is, apparently, *not* how writable pseudo-files work under Linux: I just did an experiment with `/proc/sys/kernel/hostname`, and changes to one handle are immediately visible in the other one.

>>However, it's easier to store the data in the dentry...

>Is this a typo? Did you mean "handle" instead of "dentry"?

Sorry, typo, I meant in the handle.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 28 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgB_Ci4TtqRI9G8SeP:-MfhHJFlC7p42Uopk7on:b-jy9lkr) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L28)):*

Fair enough. But could you add a comment that this is "conforming to the current behavior of Graphene ..."?
</blockquote></details>

Reworded to "Graphene's checkpointing system currently omits the `data` field", I hope that's clear enough.
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 60 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfgCicFEEA5TVQUE8pa:-MfhHRIz-XLkHwmJEBkJ:bxfu3u5) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/src/fs/tmpfs/fs.c#L60)):*

This should be removed after https://github.com/oscarlab/graphene/pull/2587 is merged.
</blockquote></details>

I rebased the branch, this is removed now.
___
*[LibOS/shim/test/fs/test_fs.py, line 321 at r2](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfhFzJfBhfB8BAe2NZU:-MfhI4FVBymzQ-qYWC-v:bsb8n7x) ([raw file](https://github.com/oscarlab/graphene/blob/940cc4846ccccebd4131c9a319ad33c601064ee4/LibOS/shim/test/fs/test_fs.py#L321)):*

This whole test doesn't seem to belong to `test/fs/`. Maybe better to move it under `test/regression/`?

The `test/fs/` tests are mainly to check copying of a bunch of files (created in `setUpClass()` above). But in this test, you just have two files.
</blockquote></details>

Okay, I thought `fs/` was a directory for filesystem tests in general, but then I see a lot of filesystem tests are in `regression/`. Sounds like it would be good to get rid of these `fs` tests, and also parametrize a bunch of `regression` tests by filesystem...

For now, I moved the rename test to `regression` and created two versions of it (for chroot and tmpfs).

dimakuv (changes requested):
> Reviewed 2 of 4 files at r3, 10 of 10 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow and @pwmarcz)

___
*[LibOS/shim/include/shim_handle.h, line 34 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg5bTvCKvl1apSo2s1:-MfhXJwSA07AAb_ZQwOy:bijgweg) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_handle.h#L34)):*
> This is, apparently, not how writable pseudo-files work under Linux: I just did an experiment with /proc/sys/kernel/hostname, and changes to one handle are immediately visible in the other one.

How do they work? Will it be hard to do a similar thing in our code?

Anyway, this is not a huge deal. We can just put a TODO comment somewhere in the code. I don't believe anyone opens the same pseudo-file twice for writing.
___
*[LibOS/shim/test/regression/rename.manifest.template, line 2 at r4](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfhW46R9-B9i1p93yWc:-MfhW46R9-B9i1p93yWd:b-76hmkt) ([raw file](https://github.com/oscarlab/graphene/blob/7607e63f56fce126e47e12d557142b0c4ea5612e/LibOS/shim/test/regression/rename.manifest.template#L2)):*
> ```
> loader.preload = "file:{{ graphene.libos }}"
> libos.entrypoint = "{{ env.PWD }}/rename"
> ```

Why not just `rename`?
___
*[LibOS/shim/test/regression/rename.manifest.template, line 9 at r4](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfhW1xM0-_9zbb93zWj:-MfhW1xM0-_9zbb93zWk:bbbsy77) ([raw file](https://github.com/oscarlab/graphene/blob/7607e63f56fce126e47e12d557142b0c4ea5612e/LibOS/shim/test/regression/rename.manifest.template#L9)):*
> ```
> fs.root.type = "chroot"
> fs.root.path = "/"
> fs.root.uri = "file:/"
> ```

Why mounting root?
___
*[LibOS/shim/test/regression/rename.manifest.template, line 20 at r4](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfhW5ve6ff25mpXlUI1:-MfhW5ve6ff25mpXlUI2:b2s2ria) ([raw file](https://github.com/oscarlab/graphene/blob/7607e63f56fce126e47e12d557142b0c4ea5612e/LibOS/shim/test/regression/rename.manifest.template#L20)):*
> ```
> 
> sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
> sgx.trusted_files.host_root_fs = "file:{{ env.PWD }}/host_root_fs"
> ```

What is this?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: 17 of 19 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/include/shim_handle.h, line 34 at r1](https://reviewable.io/reviews/oscarlab/graphene/2584#-Mfg5bTvCKvl1apSo2s1:-MfiVevw8UE0BNGgleiT:b-21oevs) ([raw file](https://github.com/oscarlab/graphene/blob/fb324d9b36fdc2a6ae80b20dd8d04112671c2e6d/LibOS/shim/include/shim_handle.h#L34)):*

> This is, apparently, not how writable pseudo-files work under Linux: I just did an experiment with /proc/sys/kernel/hostname, and changes to one handle are immediately visible in the other one.

How do they work? Will it be hard to do a similar thing in our code?

Anyway, this is not a huge deal. We can just put a TODO comment somewhere in the code. I don't believe anyone opens the same pseudo-file twice for writing.
</blockquote></details>

I added a TODO (to `shim_fs_pseudo.h`), but I also don't think it's very important.

As an aside: even if we stored the data per-dentry for both of these implementations (pseudo and tmpfs), I would still want them to be separate. These are different use-cases, and they were a pain to work with when they were coupled.
___
*[LibOS/shim/test/regression/rename.manifest.template, line 2 at r4](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfhW46R9-B9i1p93yWc:-MfiWlw_AEJp32EF1gCH:b-iwvfht) ([raw file](https://github.com/oscarlab/graphene/blob/7607e63f56fce126e47e12d557142b0c4ea5612e/LibOS/shim/test/regression/rename.manifest.template#L2)):*

Why not just `rename`?
</blockquote></details>

Looks like I copied the wrong template (`host_root_fs`). Fied.
___
*[LibOS/shim/test/regression/rename.manifest.template, line 9 at r4](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfhW1xM0-_9zbb93zWj:-MfiWmR8Er1FDrjA_1pj:bx1gsvk) ([raw file](https://github.com/oscarlab/graphene/blob/7607e63f56fce126e47e12d557142b0c4ea5612e/LibOS/shim/test/regression/rename.manifest.template#L9)):*

Why mounting root?
</blockquote></details>

Same as above. Fixed.
___
*[LibOS/shim/test/regression/rename.manifest.template, line 20 at r4](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfhW5ve6ff25mpXlUI1:-MfiX5Jv77KAYJSJsgMa:bx1gsvk) ([raw file](https://github.com/oscarlab/graphene/blob/7607e63f56fce126e47e12d557142b0c4ea5612e/LibOS/shim/test/regression/rename.manifest.template#L20)):*

What is this?
</blockquote></details>

Same as above. Fixed.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @mkow)

mkow (changes requested):
> Reviewed 4 of 8 files at r2, 2 of 4 files at r3, 9 of 10 files at r4, 1 of 2 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/src/fs/shim_fs_mem.c, line 11 at r6](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfmSvVv-DQrZLyyOx_g:-MfmSvVv-DQrZLyyOx_h:b-hrsl7m) ([raw file](https://github.com/oscarlab/graphene/blob/b27d390cd8b3b44736d15357dbfe0bfb68870881/LibOS/shim/src/fs/shim_fs_mem.c#L11)):*
> ```C
> 
> #define OVERFLOWS_FILE_OFF_T(n) __builtin_add_overflow_p((n), 0, (file_off_t)0)
> #define OVERFLOWS_SIZE_T(n) __builtin_add_overflow_p((n), 0, (size_t)0)
> ```

I like these macros :)
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 53 at r6](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfmYHmLEviEbSpemw6Z:-MfmYHmLEviEbSpemw6_:bfkxvp5) ([raw file](https://github.com/oscarlab/graphene/blob/b27d390cd8b3b44736d15357dbfe0bfb68870881/LibOS/shim/src/fs/tmpfs/fs.c#L53)):*
> ```C
>     mem_file_init(&data->mem, /*data=*/NULL, /*size=*/0);
> 
>     uint64_t time;
> ```

Could you add `_us` suffix to this? (to say which unit it uses)
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 241 at r6](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfmYBIL100wncPPJUJ3:-MfmYBIL100wncPPJUJ4:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/b27d390cd8b3b44736d15357dbfe0bfb68870881/LibOS/shim/src/fs/tmpfs/fs.c#L241)):*
> ```C
> 
> static int tmpfs_rename(struct shim_dentry* old, struct shim_dentry* new) {
>     uint64_t time;
> ```

ditto
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 305 at r6](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfmXmPY-FFtzozHINJr:-MfmXmPZ6xJRF7mVwDpL:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/b27d390cd8b3b44736d15357dbfe0bfb68870881/LibOS/shim/src/fs/tmpfs/fs.c#L305)):*
> ```C
>     assert(hdl->type == TYPE_TMPFS);
> 
>     uint64_t time;
> ```

ditto
___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 333 at r6](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfmY7Kj1aLViAjm7Ol9:-MfmY7Kj1aLViAjm7OlA:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/b27d390cd8b3b44736d15357dbfe0bfb68870881/LibOS/shim/src/fs/tmpfs/fs.c#L333)):*
> ```C
>     int ret;
> 
>     uint64_t time;
> ```

ditto

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: 18 of 19 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[LibOS/shim/src/fs/tmpfs/fs.c, line 53 at r6](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfmYHmLEviEbSpemw6Z:-MfmlX55-uae0KuYKy9w:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b27d390cd8b3b44736d15357dbfe0bfb68870881/LibOS/shim/src/fs/tmpfs/fs.c#L53)):*

Could you add `_us` suffix to this? (to say which unit it uses)
</blockquote></details>

Done.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/test/regression/rename.manifest.template, line 18 at r7](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfqBVxT-X9sZR3bZsOY:-MfqBVxT-X9sZR3bZsOZ:bkxdrny) ([raw file](https://github.com/oscarlab/graphene/blob/99c5dbef01405e1bde110ae97594ec8fd6882e03/LibOS/shim/test/regression/rename.manifest.template#L18)):*
> ```
> sgx.trusted_files.entrypoint = "file:{{ entrypoint }}"
> 
> sgx.allowed_files.tmp_dir = "file:tmp/"
> ```

What is this? I don't see how this is relevant. Accidental change?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/test/regression/rename.manifest.template, line 18 at r7](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfqBVxT-X9sZR3bZsOY:-MfqZ9cs30O-MEzbuQ3f:br80dm) ([raw file](https://github.com/oscarlab/graphene/blob/99c5dbef01405e1bde110ae97594ec8fd6882e03/LibOS/shim/test/regression/rename.manifest.template#L18)):*

What is this? I don't see how this is relevant. Accidental change?
</blockquote></details>

This test renames `tmp/file1` to `tmp/file2`, so it needs to be allowed to access them. It fails on SGX otherwise.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/test/regression/rename.manifest.template, line 18 at r7](https://reviewable.io/reviews/oscarlab/graphene/2584#-MfqBVxT-X9sZR3bZsOY:-Mfqd7nG9TvRsmVh3FjE:bvromu0) ([raw file](https://github.com/oscarlab/graphene/blob/99c5dbef01405e1bde110ae97594ec8fd6882e03/LibOS/shim/test/regression/rename.manifest.template#L18)):*

This test renames `tmp/file1` to `tmp/file2`, so it needs to be allowed to access them. It fails on SGX otherwise.
</blockquote></details>

Ah, we have two new tests in `test_libos.py`. Sorry, forgot about this. I thought that this test just works on `/mnt/tmpfs/`

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2584)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2585, opened 2021-07-27
[LibOS] Convert some `shim_qstr` instances to C strings

pwmarcz:
## Description of the changes 

This change gets rid of the following `shim_qstr` instances:

- `shim_dentry.name`
- `shim_mount.path`, `shim_dentry.uri`
- parameter to `follow_link` callback

The remaining ones are mostly `shim_handle.uri` and URIs used by pipe filesystem, but it's somewhat harder to change them, so I wanted to push this first.

## How to test this PR? 

This should be well-covered by existing tests.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)

### DIFF
--- a/LibOS/shim/include/shim_fs.h
+++ b/LibOS/shim/include/shim_fs.h
@@ -117,8 +117,9 @@ struct shim_dentry {
     int state; /* flags for managing state */
 
     /* File name, maximum of NAME_MAX characters. By convention, the root has an empty name. Does
-     * not change. */
-    struct shim_qstr name;
+     * not change. Length is kept for performance reasons. */
+    char* name;
+    size_t name_len;
 
     /* Mounted filesystem this dentry belongs to. Does not change. */
     struct shim_mount* mount;
@@ -193,7 +194,7 @@ struct shim_d_ops {
     int (*stat)(struct shim_dentry* dent, struct stat* buf);
 
     /* extracts the symlink name and saves in link */
-    int (*follow_link)(struct shim_dentry* dent, struct shim_qstr* link);
+    int (*follow_link)(struct shim_dentry* dent, char** out_target);
     /* set up symlink name to a dentry */
     int (*set_link)(struct shim_dentry* dent, const char* link);
 
@@ -243,8 +244,8 @@ struct shim_mount {
 
     struct shim_dentry* mount_point;
 
-    struct shim_qstr path;
-    struct shim_qstr uri;
+    char* path;
+    char* uri;
 
     struct shim_dentry* root;
 
@@ -595,10 +596,6 @@ int dentry_abs_path(struct shim_dentry* dent, char** path, size_t* size);
  * including the root), separated by `/`. A relative path never begins with `/`.
  */
 int dentry_rel_path(struct shim_dentry* dent, char** path, size_t* size);
-
-static inline const char* dentry_get_name(struct shim_dentry* dent) {
-    return qstrgetstr(&dent->name);
-}
 
 ino_t dentry_ino(struct shim_dentry* dent);
 
--- a/LibOS/shim/src/bookkeep/shim_handle.c
+++ b/LibOS/shim/src/bookkeep/shim_handle.c
@@ -448,8 +448,8 @@ int set_new_fd_handle_above_fd(FDTYPE fd, struct shim_handle* hdl, int fd_flags,
 static inline __attribute__((unused)) const char* __handle_name(struct shim_handle* hdl) {
     if (!qstrempty(&hdl->uri))
         return qstrgetstr(&hdl->uri);
-    if (hdl->dentry && !qstrempty(&hdl->dentry->name))
-        return qstrgetstr(&hdl->dentry->name);
+    if (hdl->dentry && hdl->dentry->name[0] != '\0')
+        return hdl->dentry->name;
     if (hdl->fs)
         return hdl->fs->name;
     return "(unknown)";
--- a/LibOS/shim/src/fs/pipe/fs.c
+++ b/LibOS/shim/src/fs/pipe/fs.c
@@ -173,7 +173,7 @@ static int fifo_open(struct shim_handle* hdl, struct shim_dentry* dent, int flag
          * one end (read or write) in our emulation, so we treat such FIFOs as read-only. This
          * covers most apps seen in the wild (in particular, LTP apps). */
         log_warning("FIFO (named pipe) '%s' cannot be opened in read-write mode in Graphene. "
-                    "Treating it as read-only.", qstrgetstr(&dent->mount->path));
+                    "Treating it as read-only.", dent->mount->path);
         flags = O_RDONLY;
     }
 
--- a/LibOS/shim/src/fs/proc/ipc-thread.c
+++ b/LibOS/shim/src/fs/proc/ipc-thread.c
@@ -39,8 +39,8 @@ bool proc_ipc_thread_pid_name_exists(struct shim_dentry* parent, const char* nam
 
 int proc_ipc_thread_follow_link(struct shim_dentry* dent, char** out_target) {
     assert(dent->parent);
-    const char* parent_name = qstrgetstr(&dent->parent->name);
-    const char* name = qstrgetstr(&dent->name);
+    const char* parent_name = dent->parent->name;
+    const char* name = dent->name;
 
     unsigned long pid;
     if (pseudo_parse_ulong(parent_name, IDTYPE_MAX, &pid) < 0)
--- a/LibOS/shim/src/fs/proc/thread.c
+++ b/LibOS/shim/src/fs/proc/thread.c
@@ -21,7 +21,7 @@ int proc_thread_follow_link(struct shim_dentry* dent, char** out_target) {
 
     lock(&g_process.fs_lock);
 
-    const char* name = qstrgetstr(&dent->name);
+    const char* name = dent->name;
     if (strcmp(name, "root") == 0) {
         dent = g_process.root;
         get_dentry(dent);
@@ -287,7 +287,7 @@ static char* describe_handle(struct shim_handle* hdl) {
 
 int proc_thread_fd_follow_link(struct shim_dentry* dent, char** out_target) {
     unsigned long fd;
-    if (pseudo_parse_ulong(qstrgetstr(&dent->name), FDTYPE_MAX, &fd) < 0)
+    if (pseudo_parse_ulong(dent->name, FDTYPE_MAX, &fd) < 0)
         return -ENOENT;
 
     struct shim_handle_map* handle_map = get_thread_handle_map(NULL);
--- a/LibOS/shim/src/fs/shim_dcache.c
+++ b/LibOS/shim/src/fs/shim_dcache.c
@@ -55,6 +55,8 @@ static struct shim_dentry* alloc_dentry(void) {
     return dent;
 }
 
+static void free_dentry(struct shim_dentry* dentry);
+
 int init_dcache(void) {
     if (!create_lock(&dcache_mgr_lock) || !create_lock(&g_dcache_lock)) {
         return -ENOMEM;
@@ -84,7 +86,14 @@ int init_dcache(void) {
     g_dentry_root->perm = PERM_rwx______;
     g_dentry_root->type = S_IFDIR;
 
-    qstrsetstr(&g_dentry_root->name, "", 0);
+    char* name = strdup("");
+    if (!name) {
+        free_dentry(g_dentry_root);
+        g_dentry_root = NULL;
+        return -ENOMEM;
+    }
+    g_dentry_root->name = name;
+    g_dentry_root->name_len = 0;
 
     return 0;
 }
@@ -108,7 +117,7 @@ static void free_dentry(struct shim_dentry* dent) {
         put_mount(dent->mount);
     }
 
-    qstrfree(&dent->name);
+    free(dent->name);
 
     if (dent->parent) {
         put_dentry(dent->parent);
@@ -173,10 +182,12 @@ struct shim_dentry* get_new_dentry(struct shim_mount* mount, struct shim_dentry*
     if (!dent)
         return NULL;
 
-    if (!qstrsetstr(&dent->name, name, name_len)) {
+    dent->name = alloc_substr(name, name_len);
+    if (!dent->name) {
         free_dentry(dent);
         return NULL;
     }
+    dent->name_len = name_len;
 
     if (parent && parent->nchildren >= DENTRY_MAX_CHILDREN) {
         log_warning("get_new_dentry: nchildren limit reached");
@@ -218,7 +229,7 @@ struct shim_dentry* lookup_dcache(struct shim_dentry* parent, const char* name, 
     struct shim_dentry* tmp;
     struct shim_dentry* dent;
     LISTP_FOR_EACH_ENTRY_SAFE(dent, tmp, &parent->children, siblings) {
-        if (qstrcmpstr(&dent->name, name, name_len) == 0) {
+        if (dent->name_len == name_len && memcmp(dent->name, name, dent->name_len) == 0) {
             get_dentry(dent);
             return dent;
         }
@@ -262,7 +273,7 @@ static size_t dentry_path_size(struct shim_dentry* dent, bool relative) {
         first = false;
 
         /* Add name */
-        size += dent->name.len;
+        size += dent->name_len;
 
         dent = up;
     }
@@ -301,10 +312,10 @@ static char* dentry_path_into_buf(struct shim_dentry* dent, bool relative, char*
         first = false;
 
         /* Add name */
-        if (pos < dent->name.len)
+        if (pos < dent->name_len)
             return NULL;
-        pos -= dent->name.len;
-        memcpy(&buf[pos], qstrgetstr(&dent->name), dent->name.len);
+        pos -= dent->name_len;
+        memcpy(&buf[pos], dent->name, dent->name_len);
 
         dent = up;
     }
@@ -400,7 +411,7 @@ static void dump_dentry(struct shim_dentry* dent, unsigned int level) {
     for (unsigned int i = 0; i < level; i++)
         buf_puts(&buf, "  ");
 
-    buf_puts(&buf, qstrgetstr(&dent->name));
+    buf_puts(&buf, dent->name);
     switch (dent->type) {
         case S_IFDIR: buf_puts(&buf, "/"); break;
         case S_IFLNK: buf_puts(&buf, " -> "); break;
@@ -491,7 +502,7 @@ BEGIN_CP_FUNC(dentry) {
         /* `fs_lock` is used only by process leader. */
         new_dent->fs_lock = NULL;
 
-        DO_CP_IN_MEMBER(qstr, new_dent, name);
+        DO_CP_MEMBER(str, dent, new_dent, name);
 
         if (new_dent->mount)
             DO_CP_MEMBER(mount, dent, new_dent, mount);
@@ -520,6 +531,7 @@ BEGIN_RS_FUNC(dentry) {
     __UNUSED(offset);
     struct shim_dentry* dent = (void*)(base + GET_CP_FUNC_ENTRY());
 
+    CP_REBASE(dent->name);
     CP_REBASE(dent->children);
     CP_REBASE(dent->siblings);
     CP_REBASE(dent->mount);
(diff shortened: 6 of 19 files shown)


dimakuv (changes requested):
> Reviewed 19 of 19 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/src/shim_checkpoint.c, line 227 at r1](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfc4tyT8-1EiTwjHjbK:-Mfc4tyU-kbewqBqph1H:b-iaslw9) ([raw file](https://github.com/oscarlab/graphene/blob/d71044fb554a1f26e9ee03e382ba75c42f79491d/LibOS/shim/src/shim_checkpoint.c#L227)):*
> ```C
>     char* new_str = (char*)base + ADD_CP_OFFSET(len + 1);
>     memcpy(new_str, str, len + 1);
>     *str_ptr = new_str;
> ```

This feels wrong. Here you're rewiring the parent's pointer to the newly created string somewhere in the checkpoint blob. You seem to be doing this to simplify your life: you don't need the restore function (`BEGIN_RS_FUNC`) anymore. But I'm pretty sure this leads to nasty side effects: you may have other pointers that point somewhere inside the original string, and now you have a discrepancy (because you have two objects). Also, I think the checkpoint blob is immediately removed after being sent to the child, so you have a dangling pointer. Also, I think this leaves you with a memory leak of the old string.

Hm... Or did I forget how this checkpointing magic works?
___
*[LibOS/shim/src/bookkeep/shim_handle.c, line 451 at r1](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfc5fq9Bqs-0bEh4TGf:-Mfc5fq9Bqs-0bEh4TGg:b-32mkyv) ([raw file](https://github.com/oscarlab/graphene/blob/d71044fb554a1f26e9ee03e382ba75c42f79491d/LibOS/shim/src/bookkeep/shim_handle.c#L451)):*
> ```C
>     if (!qstrempty(&hdl->uri))
>         return qstrgetstr(&hdl->uri);
>     if (hdl->dentry && hdl->dentry->name[0] != '\0')
> ```

Is this guaranteed that a dentry always has a non-NULL name? Otherwise you have a segfault here.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: 16 of 19 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/src/shim_checkpoint.c, line 227 at r1](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfc4tyT8-1EiTwjHjbK:-MfcEC4OEZyycJBtwxBO:baqw4nj) ([raw file](https://github.com/oscarlab/graphene/blob/d71044fb554a1f26e9ee03e382ba75c42f79491d/LibOS/shim/src/shim_checkpoint.c#L227)):*
Yeah, I think that's not how it should be (although that's hard to tell because the checkpoint system is super confusing and not documented at all...)

I tried to write a new version, please take a look. Here are my responses:

>Here you're rewiring the parent's pointer to the newly created string somewhere in the checkpoint blob. You seem to be doing this to simplify your life: you don't need the restore function (BEGIN_RS_FUNC) anymore.

Right. I think that instead of modifying `*obj` (which is the "source", in an old object) I should take it, save the content, and modify `*objp` which is the ("destination", for an already-copied object). Fixed.

However, I still maintain I don't need the restore function. The restore function seems to be for "inflating" the new object, e.g. adjusting its members. In our case, the new object is the string, and its content does not need any adjusting.

>But I'm pretty sure this leads to nasty side effects: you may have other pointers that point somewhere inside the original string, and now you have a discrepancy (because you have two objects).

Yes, that's what this `GET_FROM_CP_MAP` etc. seems to do: if there are multiple pointers to the same object, the object should get checkpointed only once. 

Fixed. However, I'm afraid I cannot do anything about pointers that point _somewhere inside_ the original string, just to the beginning of it. (We don't seem to have any such pointers, though).

>Also, I think the checkpoint blob is immediately removed after being sent to the child, so you have a dangling pointer. Also, I think this leaves you with a memory leak of the old string.

I think you assume that I modify the old object. I made sure to modify only the new object (this function is invoked by `DO_CP_IN_MEMBER`), becase that's what the code for `qstr` did. But, as mentioned above, the right way seems to be to modify only the "destination" pointer, so that should be fixed now.

---

By the way, for when we get to changing the checkpoint system: I think that this paradigm of "copy this data and adjust some pointers -> directly use the copied data" is misguided. It would be better to have explicit serialization/deserialization steps where you have to write (and then read) all the fields. The "checkpoint blob" would be something that you receive, deserialize, and then remove from memory. Do you see it that way?
___
*[LibOS/shim/src/bookkeep/shim_handle.c, line 451 at r1](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfc5fq9Bqs-0bEh4TGf:-MfcNSGF7kx2fSwOMjmi:bhc0u24) ([raw file](https://github.com/oscarlab/graphene/blob/d71044fb554a1f26e9ee03e382ba75c42f79491d/LibOS/shim/src/bookkeep/shim_handle.c#L451)):*

Is this guaranteed that a dentry always has a non-NULL name? Otherwise you have a segfault here.
</blockquote></details>

Yes, `shim_dentry.name` is never null.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/src/shim_checkpoint.c, line 227 at r1](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfc4tyT8-1EiTwjHjbK:-Mfd0kkWAWypEr8xZLca:b-ssd4nr) ([raw file](https://github.com/oscarlab/graphene/blob/d71044fb554a1f26e9ee03e382ba75c42f79491d/LibOS/shim/src/shim_checkpoint.c#L227)):*
Yes, the new version looks correct.

> However, I still maintain I don't need the restore function. The restore function seems to be for "inflating" the new object, e.g. adjusting its members. In our case, the new object is the string, and its content does not need any adjusting.

You are right, I didn't think that through. You already have the `CP_REBASE(dent->name)` statement to fix the pointer to the string in the child process, so nothing else to be done.

> Yes, that's what this GET_FROM_CP_MAP etc. seems to do: if there are multiple pointers to the same object, the object should get checkpointed only once.

You are right again. I actually don't know if different dentries can point to the same underlying name string (i.e., `dent1->name == dent2->name`. From what I remember, this is not possible in the code -- in this case you *could* get rid of this `if (!off) {`. But keeping it doesn't harm anyway. UPDATE: Well, you also use this code for `mount.path` and `mount.uri`, and probably will use it in other places, so keeping this safety check is good.

> However, I'm afraid I cannot do anything about pointers that point somewhere inside the original string, just to the beginning of it. (We don't seem to have any such pointers, though).

We seem to be thinking about different scenarios. I meant that in your original code, you modified the old object, and that would lead to discrepancies *in the parent*. But your comment seems to imply that you're worried about the "restore checkpoint" logic *in the child*. But you're right, we don't have such pointers during checkpointing.

> I think you assume that I modify the old object. I made sure to modify only the new object (this function is invoked by DO_CP_IN_MEMBER), becase that's what the code for qstr did. 

Yeah, I got confused by our beautiful checkpointing macros `DO_CP_MEMBER` vs `DO_CP_IN_MEMBER`.

---

>  The "checkpoint blob" would be something that you receive, deserialize, and then remove from memory.

Yes, this is the sane way of doing checkpointing. Whatever we have now is legacy (from ~2014), and I periodically marvel at its implementation. It's kinda beautiful in its own way. But I (and everyone who touched checkpointing code, just ask Borys) want this code rewritten asap.

Btw, the sane checkpointing implementation via proper serialization/deserialization will open the way for SCM_RIGHTS send syscall (to send FDs from parent enclave to child enclave -- Graphene currently can't support it).
___
*[LibOS/shim/src/shim_checkpoint.c, line 223 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfd0M53EkaVcXgfMMnu:-Mfd0M53EkaVcXgfMMnv:b-6035rr) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/src/shim_checkpoint.c#L223)):*
> ```C
>     assert(size == sizeof(char));
> 
>     char* str = obj;
> ```

You seem to not really use `str` below other than one place. Please either remove this `str` or use it instead of `obj`.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: 18 of 19 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/src/shim_checkpoint.c, line 223 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfd0M53EkaVcXgfMMnu:-MfhEUir1vX9LrZUgbVF:b-l6xzsb) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/src/shim_checkpoint.c#L223)):*

You seem to not really use `str` below other than one place. Please either remove this `str` or use it instead of `obj`.
</blockquote></details>

Done. I changed the code below to use `str`, I think that's the better choice becuase `strlen(obj)` looks too ugly.

boryspoplawski (changes requested):
> Reviewed 16 of 19 files at r1, 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: 18 of 19 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @pwmarcz)

___
*[LibOS/shim/include/shim_fs.h, line 127 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfh8u-NCRXYeIuHJEkC:-Mfh8u-ODJjqZ1suFviY:b-glqyai) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/include/shim_fs.h#L127)):*
> ```objc
>     /* File name, maximum of NAME_MAX characters. By convention, the root has an empty name. Does
>      * not change. */
>     char* name;
> ```

Length of this name seems to be used in some places, isn't it worth to stash it together with the pointer (maybe we use it just in couple of places and it's not worth)?
___
*[LibOS/shim/include/shim_fs.h, line 256 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfh9A-V2wrNpn-3_DKc:-Mfh9A-WD_Dyi8MI8HaL:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/include/shim_fs.h#L256)):*
> ```objc
> 
>     char* path;
>     char* uri;
> ```

ditto
___
*[LibOS/shim/src/shim_checkpoint.c, line 223 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfd0M53EkaVcXgfMMnu:-Mfh5Lb294stSS32abpS:b-5yp84s) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/src/shim_checkpoint.c#L223)):*

You seem to not really use `str` below other than one place. Please either remove this `str` or use it instead of `obj`.
</blockquote></details>

+1 to removing it.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 92 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfh743P8Jah_bLIVz9_:-Mfh743Q6PNCqfLtdMkw:b-7evh2d) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/src/fs/shim_dcache.c#L92)):*
> ```C
>     char* name = strdup("");
>     if (!name) {
>         free_dentry(g_dentry_root);
> ```

Shouldn't we also clear `g_dentry_root = NULL`?
Oh, this is init code, so it does not matter probably, but please do for sanity.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 231 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfh7iedDvKIpfuf2fJC:-Mfh7iee-orF6R14V03k:b-ycjk25) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/src/fs/shim_dcache.c#L231)):*
> ```C
>     LISTP_FOR_EACH_ENTRY_SAFE(dent, tmp, &parent->children, siblings) {
>         size_t dent_name_len = strlen(dent->name);
>         if (dent_name_len == name_len && memcmp(dent->name, name, name_len) == 0) {
> ```

Why not `!strcmp(a, b)`?
___
*[LibOS/shim/src/fs/shim_namei.c, line 138 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-MfhFrMj8CI09gUmdk9M:-MfhFrMk4q-CqvmyoyNp:b-cnfn4j) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/src/fs/shim_namei.c#L138)):*
> ```C
>                                 unsigned int link_depth) {
>     int ret;
>     char* target;
> ```

Please initialized this to `NULL`. You free it unconditionally (maybe `follow_link` initializes this in case of failures, I don't know, but explicitly setting is better at least for readability)

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: 13 of 19 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/include/shim_fs.h, line 127 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfh8u-NCRXYeIuHJEkC:-MfmGq2lBwfPlfs7wOOh:b7hihr9) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/include/shim_fs.h#L127)):*

Length of this name seems to be used in some places, isn't it worth to stash it together with the pointer (maybe we use it just in couple of places and it's not worth)?
</blockquote></details>

I think it's worth it, because file lookup right now traverses the whole directory. It needs to be fixed someday, but in the meantime, I don't want to make it even worse (especially since #2593 suggest this would be a problem for files in `/lib` and such)

I added `name_len` and converted the code to use it.
___
*[LibOS/shim/include/shim_fs.h, line 256 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfh9A-V2wrNpn-3_DKc:-MfmHIkiD2DJTN35PErQ:bj657gm) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/include/shim_fs.h#L256)):*

ditto
</blockquote></details>

Here, I don't believe it would do much: we use this to construct file URIs, so we need to traverse the whole URI anyway.
___
*[LibOS/shim/src/shim_checkpoint.c, line 223 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfd0M53EkaVcXgfMMnu:-MfmHr8FBkNtTXAgVO8I:b-93xexv) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/src/shim_checkpoint.c#L223)):*

+1 to removing it.
</blockquote></details>

`strlen(obj)` on a `void*` pointer looks weird to me, but I don't care that much. Changed.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 92 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfh743P8Jah_bLIVz9_:-MfmIYdKDtibXbpeXAsc:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/src/fs/shim_dcache.c#L92)):*

Shouldn't we also clear `g_dentry_root = NULL`?
Oh, this is init code, so it does not matter probably, but please do for sanity.
</blockquote></details>

Done.
___
*[LibOS/shim/src/fs/shim_dcache.c, line 231 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-Mfh7iedDvKIpfuf2fJC:-MfmIoKU3qP9Zo_PuX6l:bwwz2rw) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/src/fs/shim_dcache.c#L231)):*

Why not `!strcmp(a, b)`?
</blockquote></details>

Because `name` is not null-terminated (it's a string fragment from larger path).
___
*[LibOS/shim/src/fs/shim_namei.c, line 138 at r3](https://reviewable.io/reviews/oscarlab/graphene/2585#-MfhFrMj8CI09gUmdk9M:-MfmJ4YR2q3XUJ-9sRGj:bdf3nae) ([raw file](https://github.com/oscarlab/graphene/blob/21b5d53afe11d5d174eb042b406962e9ba180ae9/LibOS/shim/src/fs/shim_namei.c#L138)):*

Please initialized this to `NULL`. You free it unconditionally (maybe `follow_link` initializes this in case of failures, I don't know, but explicitly setting is better at least for readability)
</blockquote></details>

Right, it should be initialized. Fixed.

mkow (approved):
> Reviewed 13 of 19 files at r1, 6 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

boryspoplawski (approved):
> Reviewed 6 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

dimakuv (approved):
> Reviewed 5 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

dimakuv (approved):
> Reviewed 9 of 9 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

boryspoplawski (approved):
> Reviewed 9 of 9 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> Reviewed 9 of 9 files at r6.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2585)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv:
Jenkins, retest Jenkins-20.04 please (`sendfile09` LTP test timed out, I need to debug it)

PR #2586, opened 2021-07-27
[Pal/Linux-SGX] Suppress performing secret provisioning after execve

### LINKED ISSUE #2583: How many times does the graphene do remote attestation when the programe is multithreading like ppml ？
## Description of the problem
we try to count the times of getting  the wrap-key from provision-server when the programe configes the protectfile.  we want use the times for billing.  

We can't figure out the logic to get wrap key, It doesn't seem that programe only get  key one time when start.  For example ppml will get key three time from provision-server ,it means will do three times retmote attestation.

I kown when a programe is multithreading , it willl create multi enclave.  Does it mean every enclave will do one time remote attesstation and get the wrap-key from provision-server ?  So the ppml occurs three times attestation.

Could you give me some help ?

dimakuv:
## Description of the changes 

SGX Remote attestation and secret provisioning is performed via the `LD_PRELOAD = "libsecret_prov_attest.so"` trick. Previously, there was a bug in the `libsecret_prov_attest.so` library: its constructor was run before the `main()` code in each execve'd process. In multi-process apps that use execve syscall, this led to redundant remote attestation and secret provisioning requests to the Secret provisioning service.

Kudos to @yanzhichao who reported this bug. See https://github.com/oscarlab/graphene/issues/2583.

Fixes #2583.

## How to test this PR? 

I added a dummy `execve()` in the ra-tls-secret-prov example: run it, and without this PR you'll see two secre-provisioning requests. With this PR, there will be only one request.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2586)

### DIFF
--- a/Examples/ra-tls-secret-prov/src/secret_prov_pf_client.c
+++ b/Examples/ra-tls-secret-prov/src/secret_prov_pf_client.c
@@ -5,6 +5,7 @@
 #include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/wait.h>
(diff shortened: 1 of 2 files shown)


mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2586)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2586#-Mfc2z4m6ZWUKxFoG55D:-Mfc2z4m6ZWUKxFoG55E:bc87b20) (no related file):*
Ok, one thing I need to finally do: Disallow remote attestation outside of the root process (it's incorrect to do so in children - recall our discussions around loader rewrite).

---

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2586)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2601, opened 2021-07-31
[Docs] conf.py: Check the version of Sphinx to avoid invoking deprecated method

xymeng16:
This PR is related to building the documentation. 

In the setup() function of conf.py, app.add_stylesheet() is used to add a CSS file. But this only works for Sphinx under version 1.8 as Sphinx has renamed app.add_stylesheet() into app.add_css_file() since version 1.8. Hence I added a version check before invoking that to avoid that such an error happens.

## How to test?
Change the directory into Documentation/, and use `make html`.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2601)

### DIFF
--- a/Documentation/conf.py
+++ b/Documentation/conf.py
@@ -95,7 +95,11 @@ def generate_doxygen(app):
         subprocess.check_call(['doxygen', 'Doxyfile-{}'.format(p)])
 
 def setup(app):
-    app.add_stylesheet('css/graphene.css')
+    import sphinx
+    if sphinx.__version__ < '1.8': 
+        app.add_stylesheet('css/graphene.css')
+    else:
+        app.add_css_file('css/graphene.css')
     app.connect('builder-inited', generate_doxygen)
 
 breathe_domain_by_extension = {


mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2601)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @xymeng16)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2601#-Mfyl-Zd26qEhJRYEjLE:-Mfyl-Zd26qEhJRYEjLF:b57i9tw) (no related file):*
Hmm, but we have `sphinx==1.8.0` in our `requirements.txt`? So, I'm not sure if we should add IFs for other versions, I don't think we support them.
CC: @woju

---

xymeng16:
*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2601#-Mfyl-Zd26qEhJRYEjLE:-Mg0W3iy9FSYv-57aIlJ:b-ki9jh2) (no related file):*

Hmm, but we have `sphinx==1.8.0` in our `requirements.txt`? So, I'm not sure if we should add IFs for other versions, I don't think we support them.
CC: @woju
</blockquote></details>

Oh, thanks for your reply. I have the sphinx installed in my environment so I ignored the `requirements.txt`. So if applicable, please ignore and close this PR. 

(FYI, at least sphinx 4.1.2 and 4.0.0.b2 works based on my tests.)

---

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2601)* status: 0 of 1 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (3 more required), not enough approvals from different teams (2 more required, approved so far: ) (waiting on @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2601#-Mfyl-Zd26qEhJRYEjLE:-Mg2qImQ0nlXb8qmFDtw:b-ozxsxm) (no related file):*
<details><summary><i>Previously, xymeng16 (Xiangyi Meng) wrote…</i></summary><blockquote>

Oh, thanks for your reply. I have the sphinx installed in my environment so I ignored the `requirements.txt`. So if applicable, please ignore and close this PR. 

(FYI, at least sphinx 4.1.2 and 4.0.0.b2 works based on my tests.)
</blockquote></details>

No worries :)
Let's wait for @woju to comment on this, he's maintaining this part.

---

woju:
I'm unconvinced. We can't update sphinx at least until 22.04, because 20.04 still has sphinx 1.8 and we will need deb-packaged sphinx to build manpages in our debs (its not only for rtfd.io, where we could have greater latitude). When we'll be eventually upgrading, we'll just rename the function, without this conditional. So unless I miss something, this brings nothing but maintenance. If anything, I would prefer a comment `# after sphinx $VERSION, rename this to .add_css()`.

dimakuv:
Thanks for the PR, @xymeng16, but based on input from @mkow and @woju looks like we should close it.

Issue #2260, opened 2021-03-22
Incorrect remaining time is returned when nanosleep error code is set to EINTR


anjalirx-intel:
## Testfile
nanosleep02

## Description of the problem
This testcase verifies that nanosleep() will be successful to suspend the execution of a process, returns after the receipt of a signal and writes the remaining sleep time into the structure.
In this testcase the process sleeps for 1sec and then it get terminated by kill signal. The remaining time is incorrectly returned in Graphene OS

## Expected results
root@ubuntu-test-base:  ./nanosleep02
TINFO: Timeout per run is 0h 05m 00s
TPASS: nanosleep() returned -1, set errno to EINTR
TPASS: **nanosleep() slept for 1000092us, remaining time difference 80us**

## Actual results
root@ubuntu-test-base: ./pal_loader nanosleep02
TINFO: Timeout per run is 0h 05m 00s
TPASS: nanosleep() returned -1, set errno to EINTR
TFAIL: **nanosleep() remaining time 6996850us, expected 3999759us, diff 2997091us**
TBROK: Test haven't reported results!

anjalirx-intel:
@boryspoplawski The issue is fixed for Graphene Native but Graphene-SGX still fails

/home/intel/jenkins/workspace/local_ci_gsgx_apps_test/graphene/LibOS/shim/test/ltp/ltp_src/lib/tst_test.c:1250: TINFO: Timeout per run is 0h 05m 00s
/home/intel/jenkins/workspace/local_ci_gsgx_apps_test/graphene/LibOS/shim/test/ltp/ltp_src/testcases/kernel/syscalls/nanosleep/nanosleep02.c:57: TPASS: nanosleep() returned -1, set errno to EINTR
/home/intel/jenkins/workspace/local_ci_gsgx_apps_test/graphene/LibOS/shim/test/ltp/ltp_src/testcases/kernel/syscalls/nanosleep/nanosleep02.c:68: TFAIL: nanosleep() remaining time 5000009us, expected 3997585us, diff 1002424us
/home/intel/jenkins/workspace/local_ci_gsgx_apps_test/graphene/LibOS/shim/test/ltp/ltp_src/lib/tst_test.c:1067: TBROK: Test haven't reported results!

Summary:
passed   0
failed   0
skipped  0
warnings 0

boryspoplawski:
The original issue was fixed, the 2nd one is most likely unrelated and tracked in #2479

aneessahib:
[...]

PR #2602, opened 2021-08-02
[Pal] Make "syscall" instruction appear only once in PAL binary

### LINKED ISSUE #2532: [Pal/SGX] `EINTR` injection is broken
## Description of the problem

If an async signal happens while handling an ocall, `-EINTR` is injected and code returns back to the enclave as the actual host-level syscall was interrupted.
https://github.com/oscarlab/graphene/blob/4cb98219d8e302055587a8952c1102415a72ba42/Pal/src/host/Linux-SGX/sgx_exception.c#L184
If this happens after the syscall was actually done and before returning back to the enclave, `-EINTR` is injected even tho the syscall might have been successful.
Example problematic case (that we did already trigger on Jenlins) is `poll_closed_fd` from LibOS regression tests; for details see below.
Fixing this problem would probably require redesigning whole PAL SGX signal handling.

## Steps to reproduce

Run `poll_closed_fd` LibOS regression test a couple of times on current master (4cb98219d8e302055587a8952c1102415a72ba42 at the time of writing). Tested on Ubuntu18.04 in non-debug version (ofc with SGX).

## Expected results

Works every time.

## Actual results

Fails with `read error: Permission denied` from time to time. This happens when `SIGCHILD` interrupts read ocall after the actual syscall instruction in untrusted part. `-EINTR` gets injected even though some bytes were already read, the code issuing `ocall_read` has no idea about these bytes so they are lost, TLS session gets desynchronized (because this read happens on an encrypted pipe) and errors start to appear.

boryspoplawski:
## Description of the changes 
This is needed to properly handle signals interrupting ocalls on SGX PAL. It will probably also be useful in fixing the issue of signals appearing while in LibOS, but before issuing a host level syscall instruction (in which case the signal is blocked until the syscall returns or another signal appears).

Fixes #2532

## How to test this PR? 
Check #2532 for example code failing on current master.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/graphene/2602)


boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: 0 of 51 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg3Zm4OAM5mTaxhTxh5:-Mg3Zm4P2r7Xw0NTfrkx:b-8b0004) (no related file):*
Unfortunately I'm afraid that the current version of `do_syscall(long, ...)` is an UB: on each callsite arguments are not extended to the full register size. E.g. that could fail: `DO_SYSCALL(kill, -1, SIGTERM)` - `-1` is an `int` and will be passed via 32bit register (here `esi`), so the generated code will most likely be `mov esi, 0xffffffff`. If the kernel expects an `int` argument this might work, but if it expects `long` (64bit in this case), then this will fail as upper half of `rsi` contains zeros.
This can also be problematic on architectures that pass arguments in different places depending of the type of previous arguments (e.g. `f(u32 a0, u32 a1, u64 a2)` and args passed on stack: `a0` at offset `0`, `a1` at offset `4`(`sizeof(u32)`) and `a2` at offset `8` - now change all args to `u64`...).

I do not see any other solution that reintroducing an additional argument to `DO_SYSCALL` indicating the number of arguments and casting all necessary args to `long`, `uint64_t` or whatever (this code will be arch specific anyway).
Ideas?

---

dimakuv (changes requested):
> Reviewed 51 of 51 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: all files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg3Zm4OAM5mTaxhTxh5:-Mg5B0so2BOwKzH88taM:b-6vdw9f) (no related file):*

Unfortunately I'm afraid that the current version of `do_syscall(long, ...)` is an UB: on each callsite arguments are not extended to the full register size. E.g. that could fail: `DO_SYSCALL(kill, -1, SIGTERM)` - `-1` is an `int` and will be passed via 32bit register (here `esi`), so the generated code will most likely be `mov esi, 0xffffffff`. If the kernel expects an `int` argument this might work, but if it expects `long` (64bit in this case), then this will fail as upper half of `rsi` contains zeros.
This can also be problematic on architectures that pass arguments in different places depending of the type of previous arguments (e.g. `f(u32 a0, u32 a1, u64 a2)` and args passed on stack: `a0` at offset `0`, `a1` at offset `4`(`sizeof(u32)`) and `a2` at offset `8` - now change all args to `u64`...).

I do not see any other solution that reintroducing an additional argument to `DO_SYSCALL` indicating the number of arguments and casting all necessary args to `long`, `uint64_t` or whatever (this code will be arch specific anyway).
Ideas?
</blockquote></details>

What about this variadic-macro trick: https://stackoverflow.com/a/11763277/6325371? It basically overloads `DO_SYSCALL` macro with `DO_SYSCALL0`, `DO_SYSCALL1`, .. `DO_SYSCALL6` depending on the number of arguments provided. Looks very neat. Our use case also has the benefit of having only 0 to 6 arguments always, so this simplifies things.

And in each of these `DO_SYSCALL0`, `DO_SYSCALL1`, .. `DO_SYSCALL6` we cast the args before we call `do_syscall`.

---
___
*[Pal/src/host/Linux/db_files.c, line 161 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5D-Ow9X4tMz62DiDn:-Mg5D-Ow9X4tMz62DiDo:brd9qne) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux/db_files.c#L161)):*
> ```C
> 
>     if (IS_PTR_ERR(mem))
>         return PTR_TO_ERR(mem);
> ```

We also need to change from UNIX to PAL error code.
___
*[Pal/src/host/Linux/db_process.c, line 116 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5D_KmDkqz89g0QBOW:-Mg5D_KmDkqz89g0QBOX:bd6zi2o) ([raw file](https://github.com/oscarlab/graphene/blob/3182ca948c6712631a3ac9bf8983b8103351c9a9/Pal/src/host/Linux/db_process.c#L116)):*
> ```C
>  * NOTE: more tricks may be needed to prevent unexpected optimization for
>  * future compiler.
>  */
> ```

Why is this comment not relevant anymore? Is it because we call `do_syscall()` asm function immediately after vfork, and there is no "room" for compiler to rearrange something on stack before this call?
___
*[Pal/src/host/Linux/pal_linux.h, line 71 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5FkSaCwPFpo5ZKZG-:-Mg5FkSaCwPFpo5ZKZG0:b-hddku8) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux/pal_linux.h#L71)):*
> ```objc
>     })
> #else
> #define ARCH_MMAP(addr, len, prot, flags, fd, offset) \
> ```

Wouldn't it make more sense to move `ARCH_MMAP` to the arch-specific header? And since we don't have currently (and most probably never) the "old-style" x86 mmap, we just remove the `ifdef __i386__` version of mmap.

Also, renaming to smth like `DO_SYSCALL_MMAP` sounds better.
___
*[Pal/src/host/Linux-common/arch/x86_64/syscall.S, line 6 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5GtIh5pZdv-uiowUl:-Mg5GtIh5pZdv-uiowUm:biygif1) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-common/arch/x86_64/syscall.S#L6)):*
> ```
>  */
> 
> /* Oh boy. */
> ```

Why oh boy?
___
*[Pal/src/host/Linux-common/arch/x86_64/syscall.S, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5H4x01zOjerrudTpc:-Mg5H4x01zOjerrudTpd:b4evaa2) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-common/arch/x86_64/syscall.S#L9)):*
> ```
> #define __ASSEMBLY__
> #include <asm/unistd.h>
> #include <linux/sched.h>
> ```

Why do you need this one?
___
*[Pal/src/host/Linux-common/arch/x86_64/syscall.S, line 23 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5IZdI8yX6g-dU6hKY:-Mg5IZdI8yX6g-dU6hKZ:bolkdlr) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-common/arch/x86_64/syscall.S#L23)):*
> ```
>     mov %r9, %r8
>     mov 8(%rsp), %r9
>     syscall
> ```

r11 is not used by Linux syscalls? It is supposed to store `saved rflags`, but you don't shash rflags into r11 here, so r11 can be random.
___
*[Pal/src/host/Linux-common/arch/x86_64/syscall.S, line 81 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5KYtP4SB7j6k9Y_Ex:-Mg5KYtQ5GOOJkrNTGP6:b227gsp) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-common/arch/x86_64/syscall.S#L81)):*
> ```
> vfork:
>     .cfi_startproc
>     pop %rdx
> ```

What is this pop/push of rdx? I see that it is done also in Musl, Glibc... What is this for?
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 303 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5KqVN9jbWJHJkJ5R_:-Mg5KqVN9jbWJHJkJ5Ra:bfsrm35) ([raw file](https://github.com/oscarlab/graphene/blob/3182ca948c6712631a3ac9bf8983b8103351c9a9/Pal/src/host/Linux-SGX/pal_linux.h#L303)):*
> ```objc
> #define ARCH_VFORK()                                                                 \
>     (g_pal_enclave.pal_sec.in_gdb                                                    \
>          ? INLINE_SYSCALL(clone, 4, CLONE_VM | CLONE_VFORK | SIGCHLD, 0, NULL, NULL) \
> ```

I hope our GDB integration still works? I guess we have no idea why `SIGCHLD` was used for DEBUG?
___
*[Pal/src/host/Linux-SGX/sgx_enclave.c, line 208 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5LDCQ4PZCF7f78iPL:-Mg5LDCQ4PZCF7f78iPM:b-7ived9) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_enclave.c#L208)):*
> ```C
>     ms_ocall_fsync_t* ms = (ms_ocall_fsync_t*)pms;
>     ODEBUG(OCALL_FSYNC, ms);
>     return DO_SYSCALL_INTERRUPTIBLE(fsync, ms->ms_fd);
> ```

We could simplify like this almost all functions here (combine `return` with `DO_SYSCALL`).
___
*[Pal/src/host/Linux-SGX/sgx_entry.S, line 175 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5Ppjm48UqbxubxjyS:-Mg5PpjnAOVCXVLQ1xdH:bd6py0k) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_entry.S#L175)):*
> ```
> 	movq %rax, %rdi
> 	xor %esi, %esi
> 	lock xchgl %esi, %gs:PAL_TCB_URTS_LAST_ASYNC_EVENT
> ```

From what I understand, we can return both the "host-level syscall return value" (which may be a success) and the "last seen async event". So it is possible that both things happened: the syscall was finished, and there was an async event.

This is cool, but does our trusted-PAL code supports this already?
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 189 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5OYnD65qeqiapUX86:-Mg5OYnE3TMvopy7iVvx:b-x8kcr7) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_exception.c#L189)):*
> ```C
>             || rip == (uint64_t)&do_syscall_intr_after_check2) {
>         ucontext_set_ip(uc, (uint64_t)&do_syscall_intr_eintr);
>     }
> ```

So now we distinguish between two cases here:
1. Async signal happened during or right-before host-level syscall: in this case we emulate `-EINTR` on this syscall, without executing it.
2. Async signal happened anywhere else inside the untrusted-PAL code. We memorize this event in `last_async_event` but proceed with executing the host-level syscall.

So the difference from the old code is this case 2. This also seems to imply that an unfortunate timing may lead to an async signal being "postponed forever" until the host-level syscall is done or interrupted by something else. But I think we already have this assumption in the LibOS layer, so it doesn't really matter here. Is my understanding correct?
___
*[Pal/src/host/Linux-SGX/sgx_syscall.c, line 14 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5Nbbd3CG9--r92wj5:-Mg5Nbbd3CG9--r92wj6:bcadaul) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_syscall.c#L14)):*
> ```C
> 
> #define PAL_TCB_URTS_LAST_ASYNC_EVENT_OFFSET 0x58
> static_assert(offsetof(PAL_TCB_URTS, last_async_event) == PAL_TCB_URTS_LAST_ASYNC_EVENT_OFFSET,
> ```

Why such a weird way? Why not simply define the macro as:
```
#define PAL_TCB_URTS_LAST_ASYNC_EVENT_OFFSET offsetof(PAL_TCB_URTS, last_async_event) 
```
___
*[Pal/src/host/Linux-SGX/sgx_tls.h, line 108 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5Mp955GFZMQlnu6Rn:-Mg5Mp9610LQxc24RvHe:b-wz9mz7) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_tls.h#L108)):*
> ```objc
>     atomic_ulong async_signal_cnt; /* # of async signals, corresponds to # of SIGINT/SIGCONT/.. */
>     uint64_t profile_sample_time;  /* last time sgx_profile_sample() recorded a sample */
>     int32_t last_async_event;
> ```

Could you add a small comment as to *why* we want to track this?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: 44 of 51 files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg3Zm4OAM5mTaxhTxh5:-Mg5trZECnIMx1K3ovpc:b-km5xk) (no related file):*

What about this variadic-macro trick: https://stackoverflow.com/a/11763277/6325371? It basically overloads `DO_SYSCALL` macro with `DO_SYSCALL0`, `DO_SYSCALL1`, .. `DO_SYSCALL6` depending on the number of arguments provided. Looks very neat. Our use case also has the benefit of having only 0 to 6 arguments always, so this simplifies things.

And in each of these `DO_SYSCALL0`, `DO_SYSCALL1`, .. `DO_SYSCALL6` we cast the args before we call `do_syscall`.
</blockquote></details>

Damn, that's evil. But it would work.

---
___
*[Pal/src/host/Linux/db_files.c, line 161 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5D-Ow9X4tMz62DiDn:-Mg5u3Zf8G7J5RzCd7R5:b7f4sft) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux/db_files.c#L161)):*

We also need to change from UNIX to PAL error code.
</blockquote></details>

Opps, good catch.
___
*[Pal/src/host/Linux/db_process.c, line 116 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5D_KmDkqz89g0QBOW:-Mg5ujKnE_Vo3obfHeSA:bj4t473) ([raw file](https://github.com/oscarlab/graphene/blob/3182ca948c6712631a3ac9bf8983b8103351c9a9/Pal/src/host/Linux/db_process.c#L116)):*

Why is this comment not relevant anymore? Is it because we call `do_syscall()` asm function immediately after vfork, and there is no "room" for compiler to rearrange something on stack before this call?
</blockquote></details>

This comment was partially wrong and relied on specific compiler behavior. Now `vfork` is marked as `returns_twice`, which should give the compiler enough information. Note that the child process also ran a function `handle_set_cloexec` which was one big UB.
___
*[Pal/src/host/Linux/pal_linux.h, line 71 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5FkSaCwPFpo5ZKZG-:-Mg6ETJzAtsQms8d_vP4:b-mij2ty) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux/pal_linux.h#L71)):*

Wouldn't it make more sense to move `ARCH_MMAP` to the arch-specific header? And since we don't have currently (and most probably never) the "old-style" x86 mmap, we just remove the `ifdef __i386__` version of mmap.

Also, renaming to smth like `DO_SYSCALL_MMAP` sounds better.
</blockquote></details>

I just didn't want to stash even more changes to this PR, but you might be right that it's related and this PR is probably a good place to fix it.
Done. Since couple of popular archs (checked x86, arm and ppc) support a syscall with the same name (`mmap`) and same args, I just simply removed it.
___
*[Pal/src/host/Linux-common/arch/x86_64/syscall.S, line 6 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5GtIh5pZdv-uiowUl:-Mg6_YeL1l3hWBbVZDy-:brsjxls) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-common/arch/x86_64/syscall.S#L6)):*

Why oh boy?
</blockquote></details>

Need to manually define `__ASSEMBLY__` because these headers do not use a more standard `__ASSEMBLER__` (due to historical reasons).
Removed that comment.
___
*[Pal/src/host/Linux-common/arch/x86_64/syscall.S, line 9 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5H4x01zOjerrudTpc:-Mg6a3DFC-cUpaSUbI5k:bdi5189) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-common/arch/x86_64/syscall.S#L9)):*

Why do you need this one?
</blockquote></details>

Artifact from intermediate version which had clone flags hardcoded (until I realized we use clone in multiple places).
___
*[Pal/src/host/Linux-common/arch/x86_64/syscall.S, line 23 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5IZdI8yX6g-dU6hKY:-Mg6aF1Z6yycoz0O4URa:ble07mz) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-common/arch/x86_64/syscall.S#L23)):*

r11 is not used by Linux syscalls? It is supposed to store `saved rflags`, but you don't shash rflags into r11 here, so r11 can be random.
</blockquote></details>

I do not understand your comment. `r11` does not need to be preserved in function calls and it's not an argument to syscall, why would we want to do anything? The CPU stores flags in `r11`, but that's not important.
___
*[Pal/src/host/Linux-common/arch/x86_64/syscall.S, line 81 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5KYtP4SB7j6k9Y_Ex:-Mg6an1h7i2iu1wnuiLH:b-u92vlr) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-common/arch/x86_64/syscall.S#L81)):*

What is this pop/push of rdx? I see that it is done also in Musl, Glibc... What is this for?
</blockquote></details>

Because the child process will return here and most likely clobber the stack, including the return address from `vfork`, as it's going to call `execve` or `exit`. We stash the return address inside register because the child process wont affect those.
Picked `rdx` as `rax` is used by syscall, `rbx` needs to be preserved across function calls, `rcx` is used by syscall, and next is `rdx` :)
Updated the code a bit (jmp instead of push+ret).
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 303 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5KqVN9jbWJHJkJ5R_:-Mg6cMRjAONEY_Ta6XBb:bjzqg0w) ([raw file](https://github.com/oscarlab/graphene/blob/3182ca948c6712631a3ac9bf8983b8103351c9a9/Pal/src/host/Linux-SGX/pal_linux.h#L303)):*

I hope our GDB integration still works? I guess we have no idea why `SIGCHLD` was used for DEBUG?
</blockquote></details>

What do you mean? We use vfork syscall now (at least on x64) which is equivalent to `clone` with `SIGCHLD` (and some other flags). The question is why `SIGCHLD` was not set in non debug builds and I guess it was an optimization: back in the day we did not ignore `SIGCHILD` as we now do.
___
*[Pal/src/host/Linux-SGX/sgx_enclave.c, line 208 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5LDCQ4PZCF7f78iPL:-Mg6cyaY88V2e6dGaNdS:b-ft5bo1) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_enclave.c#L208)):*

We could simplify like this almost all functions here (combine `return` with `DO_SYSCALL`).
</blockquote></details>

We could do a lot of things with this file :) You want me to add them here? I could, no problem from my side.
___
*[Pal/src/host/Linux-SGX/sgx_entry.S, line 175 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5Ppjm48UqbxubxjyS:-Mg6dAH7-YBZ6CxJVMuC:bk0wuu7) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_entry.S#L175)):*

From what I understand, we can return both the "host-level syscall return value" (which may be a success) and the "last seen async event". So it is possible that both things happened: the syscall was finished, and there was an async event.

This is cool, but does our trusted-PAL code supports this already?
</blockquote></details>

Yes, it always supported it.
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 189 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5OYnD65qeqiapUX86:-Mg6dUzK6yFU5mcHOQ6a:b-gzkff3) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_exception.c#L189)):*

So now we distinguish between two cases here:
1. Async signal happened during or right-before host-level syscall: in this case we emulate `-EINTR` on this syscall, without executing it.
2. Async signal happened anywhere else inside the untrusted-PAL code. We memorize this event in `last_async_event` but proceed with executing the host-level syscall.

So the difference from the old code is this case 2. This also seems to imply that an unfortunate timing may lead to an async signal being "postponed forever" until the host-level syscall is done or interrupted by something else. But I think we already have this assumption in the LibOS layer, so it doesn't really matter here. Is my understanding correct?
</blockquote></details>

No. There are 3 cases:
1. Async signal happened while in enclave - nothing changes here, we just enter the enclave and let that code handle the event.
2a. Async signal happened outside of the enclave (in general) - we just memorize the event and report it when returning back from ocall. Old code did the same, but it also forcibly moved rip to ocall return code and injected `-EINTR`. Now we just save it and syscall executing functions checks that value and reports `-EINTR` instead of actually issuing the syscall (but only the `DO_SYSCALL_INTERRUPTIBLE` macro, in some places we want to issue the syscall unconditionally, which we can do with `DO_SYSCALL`).
2b. Special case of above: if the signal happened while checking whether the last_event is set (checking happens inside the `do_syscall_intr` function, just before `syscall` instruction), we forward `rip` behind that check and `syscall` instruction, where `-EINTR` is set. There are only two problematic instruction, both checked here.

This code should work in all cases, i.e. there are no lost signals and they always preempt blocking syscalls, even if such signals arrives before the actual syscall!
___
*[Pal/src/host/Linux-SGX/sgx_syscall.c, line 14 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5Nbbd3CG9--r92wj5:-Mg6fCQh3pj3ngWRM6DV:bnj7p8t) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_syscall.c#L14)):*

Why such a weird way? Why not simply define the macro as:
```
#define PAL_TCB_URTS_LAST_ASYNC_EVENT_OFFSET offsetof(PAL_TCB_URTS, last_async_event) 
```
</blockquote></details>

Because `offsetof` is evaulated at compile time, which happens after the preprocessor is run. Due to that we cannot `XSTRINGIFY` the result of `offsetof` and need to use an immediate value.
Added a comment.
___
*[Pal/src/host/Linux-SGX/sgx_tls.h, line 108 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5Mp955GFZMQlnu6Rn:-Mg6g2dN9X4N2ZMojPT0:b-duxc17) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_tls.h#L108)):*

Could you add a small comment as to *why* we want to track this?
</blockquote></details>

As always postponed adding comments to the last step and forgot about it ;)
Done.

aneessahib:
>>Async signal happened during or right-before host-level syscall: in this case we emulate -EINTR on this syscall, without executing it.

in case of futex waits- the application expects the remainder  of the time (total - interrupted time) to be sent back to the application. we have this logic after the host has returned from syscall, but it wont run If we EINTR without executing it. Which is probably why this fix is not working for #2260

boryspoplawski (commented):
What? We do execute it regardless of EINTRs. Besides this PR  did fix #2260 for me.

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602#-:-Mg70SGC4c8L3VQvZVGh:b-7cukvm)* status: 44 of 51 files reviewed, 14 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

dimakuv (dismissed):
> Reviewed 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux-common/arch/x86_64/syscall.S, line 23 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5IZdI8yX6g-dU6hKY:-Mg76bCe7pAuLsd3rRvV:biorfrw) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-common/arch/x86_64/syscall.S#L23)):*

I do not understand your comment. `r11` does not need to be preserved in function calls and it's not an argument to syscall, why would we want to do anything? The CPU stores flags in `r11`, but that's not important.
</blockquote></details>

Ok, my bad, I forgot that it is the `syscall` instruction itself who puts rflags into r11.
___
*[Pal/src/host/Linux-SGX/pal_linux.h, line 303 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5KqVN9jbWJHJkJ5R_:-Mg782SP9zXl8T0RbSL6:bomsr9q) ([raw file](https://github.com/oscarlab/graphene/blob/3182ca948c6712631a3ac9bf8983b8103351c9a9/Pal/src/host/Linux-SGX/pal_linux.h#L303)):*

What do you mean? We use vfork syscall now (at least on x64) which is equivalent to `clone` with `SIGCHLD` (and some other flags). The question is why `SIGCHLD` was not set in non debug builds and I guess it was an optimization: back in the day we did not ignore `SIGCHILD` as we now do.
</blockquote></details>

True, I just realized that previously vfork was done via `clone()` and now you switched to the actual `vfork()`.
___
*[Pal/src/host/Linux-SGX/sgx_enclave.c, line 208 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5LDCQ4PZCF7f78iPL:-Mg78MCT0S5coyz-8TIG:b96lpa8) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_enclave.c#L208)):*

We could do a lot of things with this file :) You want me to add them here? I could, no problem from my side.
</blockquote></details>

No, let's not bother.
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 189 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5OYnD65qeqiapUX86:-Mg7AKL0065j9NaxQeEC:b-1f1f2s) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_exception.c#L189)):*

No. There are 3 cases:
1. Async signal happened while in enclave - nothing changes here, we just enter the enclave and let that code handle the event.
2a. Async signal happened outside of the enclave (in general) - we just memorize the event and report it when returning back from ocall. Old code did the same, but it also forcibly moved rip to ocall return code and injected `-EINTR`. Now we just save it and syscall executing functions checks that value and reports `-EINTR` instead of actually issuing the syscall (but only the `DO_SYSCALL_INTERRUPTIBLE` macro, in some places we want to issue the syscall unconditionally, which we can do with `DO_SYSCALL`).
2b. Special case of above: if the signal happened while checking whether the last_event is set (checking happens inside the `do_syscall_intr` function, just before `syscall` instruction), we forward `rip` behind that check and `syscall` instruction, where `-EINTR` is set. There are only two problematic instruction, both checked here.

This code should work in all cases, i.e. there are no lost signals and they always preempt blocking syscalls, even if such signals arrives before the actual syscall!
</blockquote></details>

OK, you are right. I missed to combine the pieces in your PR. I was mainly interested in case 2a. But now I see that:
- An async signal happens, it lands in this exception handler which sets `last_async_event`
- Then the exception handler is done, Linux moves the thread to the normal execution context, which continues whenever it was interrupted
- If we were in the `do_syscall_intr()` prologue code (or somewhere before it), we'll notice `last_async_event` was set and skip the syscall
- If we were somewhere else (after `do_syscall_intr()`), we'll notice `last_async_event` in `sgx_entry.S` and set `rsi`

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: 50 of 51 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg3Zm4OAM5mTaxhTxh5:-Mg7G9FZ0pbXg1jsyzkT:b-896fix) (no related file):*

Damn, that's evil. But it would work.
</blockquote></details>

Done.

---
___
*[Pal/src/host/Linux-SGX/sgx_entry.S, line 175 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg7MrvB8YEuAuYo6cyF:-Mg7MrvB8YEuAuYo6cyG:b-1r3uxy) ([raw file](https://github.com/oscarlab/graphene/blob/41fbbdd5297055233fcd10544cac16587c9aa8cf/Pal/src/host/Linux-SGX/sgx_entry.S#L175)):*
> ```
> 	movq %rax, %rdi
> 	xor %esi, %esi
> 	lock xchgl %esi, %gs:PAL_TCB_URTS_LAST_ASYNC_EVENT
> ```

Note that a `lock` is implicit in `xchg` with one memory operand, but I just wanted to make this explicit.
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 189 at r1](https://reviewable.io/reviews/oscarlab/graphene/2602#-Mg5OYnD65qeqiapUX86:-Mg7HkgR8_NUp3IM7O8P:bd5pos8) ([raw file](https://github.com/oscarlab/graphene/blob/d245dff63aac750bff5eae6a188329d6a90c4ebe/Pal/src/host/Linux-SGX/sgx_exception.c#L189)):*

OK, you are right. I missed to combine the pieces in your PR. I was mainly interested in case 2a. But now I see that:
- An async signal happens, it lands in this exception handler which sets `last_async_event`
- Then the exception handler is done, Linux moves the thread to the normal execution context, which continues whenever it was interrupted
- If we were in the `do_syscall_intr()` prologue code (or somewhere before it), we'll notice `last_async_event` was set and skip the syscall
- If we were somewhere else (after `do_syscall_intr()`), we'll notice `last_async_event` in `sgx_entry.S` and set `rsi`
</blockquote></details>

Note that `last_async_event` will be returned regardless when the exception happens (even before the syscall).
This actually reminded me that there still is a small window when the even will be missed (well, actually reported on the next ocall): if it happens after `xchg %rsi, ...` instruction and before EENTER in `sgx_entry.S`. I don't think we should bother fixing that tho: only possible events are `PAL_INTERRUPTED` and `PAL_QUIT`. The former is merely an interrupt to make Graphene notice some state changes (so missing it is irrelevant when we return from ocall) and the latter rarely used.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 9 of 51 files at r1, 2 of 7 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Pal/include/host/Linux-common/syscall.h, line 27 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgBjfZ65ssf9-LSC4W_:-MgBjfZ65ssf9-LSC4Wa:b15e1k) ([raw file](https://github.com/oscarlab/graphene/blob/41fbbdd5297055233fcd10544cac16587c9aa8cf/Pal/include/host/Linux-common/syscall.h#L27)):*
> ```objc
> 
> #define DO_SYSCALL(name, args...) \
>     PICK_SYSCALL_MACRO(name, ##args, 6, 5, 4, 3, 2, 1, 0)(__NR_##name, ##args)
> ```

wow, that's a clever trick
___
*[Pal/include/host/Linux-common/syscall.h, line 30 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgBrH731z1b8UwD90kn:-MgBrH7448zgu6sOObyS:b94h6w6) ([raw file](https://github.com/oscarlab/graphene/blob/41fbbdd5297055233fcd10544cac16587c9aa8cf/Pal/include/host/Linux-common/syscall.h#L30)):*
> ```objc
> 
> #define IS_PTR_ERR(val) ((unsigned long)(val) >= -4095UL)
> #define PTR_TO_ERR(val) ((long)val)
> ```

`val` should be parenthesized

mkow (changes requested):
> Reviewed 34 of 51 files at r1, 5 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux-SGX/sgx_enclave.c, line 122 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgHjf5f38r9IaKlRKkm:-MgHjf5f38r9IaKlRKkn:b-yiz8ny) ([raw file](https://github.com/oscarlab/graphene/blob/41fbbdd5297055233fcd10544cac16587c9aa8cf/Pal/src/host/Linux-SGX/sgx_enclave.c#L122)):*
> ```C
>     ms_ocall_close_t* ms = (ms_ocall_close_t*)pms;
>     ODEBUG(OCALL_CLOSE, ms);
>     DO_SYSCALL(close, ms->ms_fd);
> ```

isn't `close()` also interruptible in some exotic cases?
___
*[Pal/src/host/Linux-SGX/sgx_enclave.c, line 188 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgHk7Xo5BQOtyX_hYW2:-MgHk7Xp0zYkM0XAw51M:b82jpkd) ([raw file](https://github.com/oscarlab/graphene/blob/41fbbdd5297055233fcd10544cac16587c9aa8cf/Pal/src/host/Linux-SGX/sgx_enclave.c#L188)):*
> ```
> O_NONBLOCK
> ```
this one is with `O_NONBLOCK`, so I guess it shouldn't be interruptible?
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 223 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgHmJkG0xpp78ptg6NU:-MgHmJkG0xpp78ptg6NV:b-w3gnkg) ([raw file](https://github.com/oscarlab/graphene/blob/3182ca948c6712631a3ac9bf8983b8103351c9a9/Pal/src/host/Linux-SGX/sgx_exception.c#L223)):*
> ```C
>         goto err;
> 
>     ret = set_signal_handler(SIGSYS, handle_sync_signal);
> ```

why it was here originally, and why can we remove it now?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/include/host/Linux-common/syscall.h, line 30 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgBrH731z1b8UwD90kn:-MgMWiyfCHf-wxLNC3wx:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/41fbbdd5297055233fcd10544cac16587c9aa8cf/Pal/include/host/Linux-common/syscall.h#L30)):*

`val` should be parenthesized
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/sgx_enclave.c, line 122 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgHjf5f38r9IaKlRKkm:-MgMWjxLCrSZL9lm_aTS:baspkf5) ([raw file](https://github.com/oscarlab/graphene/blob/41fbbdd5297055233fcd10544cac16587c9aa8cf/Pal/src/host/Linux-SGX/sgx_enclave.c#L122)):*

isn't `close()` also interruptible in some exotic cases?
</blockquote></details>

No. I mean it is, but Linux closes the fd anyway and only then returns `-EINTR`, but since you cannot repeat close on errors, I've decided to make it ignore signals.
___
*[Pal/src/host/Linux-SGX/sgx_enclave.c, line 188 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgHk7Xo5BQOtyX_hYW2:-MgMX6QJ3vGkfwiKNxS8:b-2p1no4) ([raw file](https://github.com/oscarlab/graphene/blob/41fbbdd5297055233fcd10544cac16587c9aa8cf/Pal/src/host/Linux-SGX/sgx_enclave.c#L188)):*

> ```
> O_NONBLOCK
> ```
this one is with `O_NONBLOCK`, so I guess it shouldn't be interruptible?
</blockquote></details>

It's rather about `F_SETFL`, but yes.
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 223 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgHmJkG0xpp78ptg6NU:-MgMXnfK5ASCo_2IUIod:b-tdttfx) ([raw file](https://github.com/oscarlab/graphene/blob/3182ca948c6712631a3ac9bf8983b8103351c9a9/Pal/src/host/Linux-SGX/sgx_exception.c#L223)):*

why it was here originally, and why can we remove it now?
</blockquote></details>

No idea why it was here. Why wouldn't we remove it, if it's useless?

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: 49 of 51 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 223 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgHmJkG0xpp78ptg6NU:-MgMY5bT05cFCAQkCU7d:b-tysqet) ([raw file](https://github.com/oscarlab/graphene/blob/3182ca948c6712631a3ac9bf8983b8103351c9a9/Pal/src/host/Linux-SGX/sgx_exception.c#L223)):*

No idea why it was here. Why wouldn't we remove it, if it's useless? 
</blockquote></details>

Wasn't there some kind of sandbox feature implemented a long time ago? Maybe it used seccomp...

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux-SGX/sgx_enclave.c, line 122 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgHjf5f38r9IaKlRKkm:-MgMYQEuCUjF6IbxupPZ:b4il27w) ([raw file](https://github.com/oscarlab/graphene/blob/41fbbdd5297055233fcd10544cac16587c9aa8cf/Pal/src/host/Linux-SGX/sgx_enclave.c#L122)):*

No. I mean it is, but Linux closes the fd anyway and only then returns `-EINTR`, but since you cannot repeat close on errors, I've decided to make it ignore signals.
</blockquote></details>

Could you say this in a comment here?
___
*[Pal/src/host/Linux-SGX/sgx_enclave.c, line 188 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgHk7Xo5BQOtyX_hYW2:-MgMYXSMBBm9Ct4u2jak:b-1ujfa8) ([raw file](https://github.com/oscarlab/graphene/blob/41fbbdd5297055233fcd10544cac16587c9aa8cf/Pal/src/host/Linux-SGX/sgx_enclave.c#L188)):*

It's rather about `F_SETFL`, but yes.
</blockquote></details>

Ah, true.
___
*[Pal/src/host/Linux-SGX/sgx_exception.c, line 223 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgHmJkG0xpp78ptg6NU:-MgMYZv_A6w-QFGm0ydr:b-1d5ldv) ([raw file](https://github.com/oscarlab/graphene/blob/3182ca948c6712631a3ac9bf8983b8103351c9a9/Pal/src/host/Linux-SGX/sgx_exception.c#L223)):*

Wasn't there some kind of sandbox feature implemented a long time ago? Maybe it used seccomp...
</blockquote></details>

Hmm, maybe...

boryspoplawski (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/host/Linux-SGX/sgx_enclave.c, line 122 at r3](https://reviewable.io/reviews/oscarlab/graphene/2602#-MgHjf5f38r9IaKlRKkm:-MgMwZJyBnUkoLd4GfpH:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/41fbbdd5297055233fcd10544cac16587c9aa8cf/Pal/src/host/Linux-SGX/sgx_enclave.c#L122)):*

Could you say this in a comment here?
</blockquote></details>

Done.

mkow (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2602)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2618, opened 2021-08-04
[Pal/{Linux,Linux-SGX}] Fix early log messages crashing PAL

pwmarcz:
## Description of the changes 

The logging subsystem (`log_error`) requires relocations to be applied before use, due to use of `log_level_to_prefix` array. This caused early calls to `log_error()` to crash.

To fix that, we apply relocations earlier in the startup process.

## Alternatives

- **Compile PAL so that no relocations are necessary** - I'm not sure if that's possible if PAL is position-independent? Apparently GCC has [-static-pie](https://stackoverflow.com/questions/62586808/how-are-relocations-supposed-to-work-in-static-pie-binaries), but it just shifts the work of applying these relocations to runtime (by calling a glibc function)...
- **Compile PAL so that it's at a static position** - probably it's not worth it just for that?
- **Make `log_*` code position-independent**. Maybe that's worth doing regardless of this fix, just in case? So that e.g. we can use log messages in relocation code.

   The problem is that the array (`static const char* log_level_to_prefix[]`) contains the right pointers only after applying relocations. The simplest fix would be to use a `switch` instead of a global array. Experimentally, this does produce position-independent code (although with `-O3` it still produces a sequence of conditions, not a lookup table).

  Alternatively, changing it to `const char log_level_to_prefix[][16]` produces a flat array.
- **Install stack canary **after** relocations** - under Linux host, we start by setting up the stack canary; this can be moved further down if the relocation code itself is marked as `__attribute_no_stack_protector`, but I'm not sure if that's a good idea

## How to test this PR? 

`log_error` calls just after `ELF_DYNAMIC_RELOCATE()` should work correctly, and there should be no calls _before_ it.

On master, calls to `log_error` before `ELF_DYNAMIC_RELOCATE()` produce a segfault under Linux, and print a message with no prefix under Linux-SGX (as the non-relocated pointers end up pointing to null bytes by accident).

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)

### DIFF
--- a/Pal/src/host/Linux-SGX/db_main.c
+++ b/Pal/src/host/Linux-SGX/db_main.c
@@ -520,6 +520,14 @@ noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char*
     /* Our arguments are coming directly from the urts. We are responsible to check them. */
     int ret;
 
+    /* Relocate PAL itself (note that this is required to run `log_error`) */
+    g_pal_map.l_addr = elf_machine_load_address();
+    g_pal_map.l_name = "libpal.so"; // to be overriden later
+    elf_get_dynamic_info((void*)g_pal_map.l_addr + elf_machine_dynamic(), g_pal_map.l_info,
+                         g_pal_map.l_addr);
+
+    ELF_DYNAMIC_RELOCATE(&g_pal_map);
+
     uint64_t start_time;
     ret = _DkSystemTimeQuery(&start_time);
     if (ret < 0) {
@@ -558,13 +566,8 @@ noreturn void pal_linux_main(char* uptr_libpal_uri, size_t libpal_uri_len, char*
     }
     libpal_path[libpal_uri_len] = '\0';
 
-    /* relocate PAL itself */
-    g_pal_map.l_addr = elf_machine_load_address();
+    /* Now that we have `libpal_path`, set name for PAL map */
     g_pal_map.l_name = libpal_path;
-    elf_get_dynamic_info((void*)g_pal_map.l_addr + elf_machine_dynamic(), g_pal_map.l_info,
-                         g_pal_map.l_addr);
-
-    ELF_DYNAMIC_RELOCATE(&g_pal_map);
 
     /*
      * We can't verify the following arguments from the urts. So we copy
--- a/Pal/src/host/Linux/db_main.c
+++ b/Pal/src/host/Linux/db_main.c
@@ -155,8 +155,19 @@ noreturn void pal_linux_main(void* initial_rsp, void* fini_callback) {
     dummy_tcb_for_stack_protector.common.self = &dummy_tcb_for_stack_protector.common;
     pal_tcb_set_stack_canary(&dummy_tcb_for_stack_protector.common, STACK_PROTECTOR_CANARY_DEFAULT);
     ret = pal_set_tcb(&dummy_tcb_for_stack_protector.common);
-    if (ret < 0)
-        INIT_FAIL(unix_to_pal_error(-ret), "pal_set_tcb() failed");
+    if (ret < 0) {
+        /* We failed to install a TCB (and haven't applied relocations yet), so no other code will
+         * work anyway */
+        DO_SYSCALL(exit_group, PAL_ERROR_DENIED);
+        die_or_inf_loop();
+    }
+
+    /* Relocate PAL itself (note that this is required to run `log_error`) */
+    g_pal_map.l_addr = elf_machine_load_address();
+    g_pal_map.l_name = "libpal.so"; // to be overriden later
+    elf_get_dynamic_info((void*)g_pal_map.l_addr + elf_machine_dynamic(), g_pal_map.l_info,
+                         g_pal_map.l_addr);
+    ELF_DYNAMIC_RELOCATE(&g_pal_map);
 
     uint64_t start_time;
     ret = _DkSystemTimeQuery(&start_time);
@@ -175,18 +186,14 @@ noreturn void pal_linux_main(void* initial_rsp, void* fini_callback) {
     if (argc < 4)
         print_usage_and_exit(argv[0]);  // may be NULL!
 
+    /* Now that we have `argv`, set name for PAL map */
+    g_pal_map.l_name = argv[0];
+
     // Are we the first in this Graphene's namespace?
     bool first_process = !strcmp(argv[2], "init");
     if (!first_process && strcmp(argv[2], "child")) {
         print_usage_and_exit(argv[0]);
     }
-
-    g_pal_map.l_addr = elf_machine_load_address();
-    g_pal_map.l_name = argv[0];
-    elf_get_dynamic_info((void*)g_pal_map.l_addr + elf_machine_dynamic(), g_pal_map.l_info,
-                         g_pal_map.l_addr);
-
-    ELF_DYNAMIC_RELOCATE(&g_pal_map);
 
     g_linux_state.host_environ = envp;
 
--- a/Pal/src/printf.c
+++ b/Pal/src/printf.c
@@ -6,6 +6,12 @@
 #include "pal.h"
 #include "pal_internal.h"
 
+/*
+ * NOTE: The logging subsystem cannot be used during early PAL startup, because the pointers in the
+ * `log_level_to_prefix` array need to be relocated first. If that becomes an issue (that is, if we
+ * want to use logging before or during relocation), this array can be converted to a switch.
+ */
+
 /* NOTE: We could add "pal" prefix to the below strings for more fine-grained log info */
 static const char* log_level_to_prefix[] = {
     [LOG_LEVEL_NONE]    = "",


boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2618#-MgJzBA6AfOg55DBK8t9:-MgJzBA6AfOg55DBK8tA:bnp5kk9) (no related file):*
I like this PR as-is. But just comments on alternatives:

> Compile PAL so that no relocations are necessary

This looks impossible to do. Some kind of runtime relocation routine is still needed, so no difference to what we do right now.

> Compile PAL so that it's at a static position

I don't like this one for several reasons. First, on Linux PAL, we may get into the problem with tools like Valgrind (IIRC, Valgrind wants to put the executable at a specific location? something like this). Second, on Linux-SGX PAL, we definitely want the trusted-PAL binary to be placed in different parts of the enclave memory (right now we place it at the top of enclave address space, and since enclave sizes vary, this near-the-top address also varies).

> Make log_* code position-independent.

This can bo done, but this only solves one particular problem (with log functions). The same issue can happen with some other functionality that we want to add later...

> Install stack canary after relocations

This is tricky. I am pretty sure we want to install the stack canary as the very first thing in the Linux PAL (in Linux-SGX PAL, we don't need to worry because the untrusted runtime already set up a dummy stack canary for us, so at least all function calls work, and installing the real stack canary can be pushed to the very end). Marking all functions before installing the stack canary as "no-stack-protector" is tedious, error-prone, and just not right.

---
___
*[Pal/src/host/Linux/db_main.c, line 161 at r1](https://reviewable.io/reviews/oscarlab/graphene/2618#-MgJyKSJCxC_bcU8FOBe:-MgJyKSJCxC_bcU8FOBf:b-1jq7bc) ([raw file](https://github.com/oscarlab/graphene/blob/f1ad322b8d7e2418d1b1ed1bdd6a5c2467e8fefd/Pal/src/host/Linux/db_main.c#L161)):*
> ```C
>         /* We failed to install a TCB (and haven't applied relocations yet), so no other code will
>          * work anyway */
>         int exitcode = unix_to_pal_error(-ret);
> ```

You don't need `unix_to_pal_error()`. We want to return the UNIX error code in the `exit_group` syscall. (Previously we had the convertion to the PAL error code because `INIT_FAIL()` assumes a PAL error code.)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: 1 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2618#-MgJzBA6AfOg55DBK8t9:-MgMCUuHAq3ZClK8AKuj:bgwyrll) (no related file):*
>IIRC, Valgrind wants to put the executable at a specific location? something like this

It's the opposite: if your code is position-independent, Valgrind will put it at an inconvenient location; you can make it static so that Valgrind will not mess with the position. But this is a bug in Valgrind, and definitely not a good justification for doing it on production.

>>Make `log_*` code position-independent.

>This can bo done, but this only solves one particular problem (with log functions). The same issue can happen with some other functionality that we want to add later...

Okay, I don't want to do this convertion without a good reason, but I'm adding a comment so that it's less confusing for the next person encountering this problem.

>Marking all functions before installing the stack canary as "no-stack-protector" is tedious, error-prone, and just not right.

Yeah, I had to annotate some random inline functions in `dynamic_link.h`, it felt very brittle.

---
___
*[Pal/src/host/Linux/db_main.c, line 161 at r1](https://reviewable.io/reviews/oscarlab/graphene/2618#-MgJyKSJCxC_bcU8FOBe:-MgMDncB5tVsGdBQ4pdQ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f1ad322b8d7e2418d1b1ed1bdd6a5c2467e8fefd/Pal/src/host/Linux/db_main.c#L161)):*

You don't need `unix_to_pal_error()`. We want to return the UNIX error code in the `exit_group` syscall. (Previously we had the convertion to the PAL error code because `INIT_FAIL()` assumes a PAL error code.)
</blockquote></details>

Done.

boryspoplawski (changes requested):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: 2 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[Pal/src/host/Linux/db_main.c, line 161 at r1](https://reviewable.io/reviews/oscarlab/graphene/2618#-MgJyKSJCxC_bcU8FOBe:-MgMV0kF9eyNEnmDwqaO:b-pfcs9j) ([raw file](https://github.com/oscarlab/graphene/blob/f1ad322b8d7e2418d1b1ed1bdd6a5c2467e8fefd/Pal/src/host/Linux/db_main.c#L161)):*

Done.
</blockquote></details>

I thought we had a convention of returning PAL error codes. This version is not ok, since it returns different codes in different places (PAL vs UNIX).

mkow (dismissed):
> Reviewed 1 of 2 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[Pal/src/host/Linux/db_main.c, line 161 at r1](https://reviewable.io/reviews/oscarlab/graphene/2618#-MgJyKSJCxC_bcU8FOBe:-MgM_518C3orTvIZ0asQ:b-29kcpo) ([raw file](https://github.com/oscarlab/graphene/blob/f1ad322b8d7e2418d1b1ed1bdd6a5c2467e8fefd/Pal/src/host/Linux/db_main.c#L161)):*

I thought we had a convention of returning PAL error codes. This version is not ok, since it returns different codes in different places (PAL vs UNIX).
</blockquote></details>

It does look like it. On the other hand, it means also annotating `unix_to_pal_error` (as well as `unix_to_pal_error_positive`) with `__attribute_no_stack_protector`... Should I?

boryspoplawski (changes requested):
> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/host/Linux/db_main.c, line 161 at r1](https://reviewable.io/reviews/oscarlab/graphene/2618#-MgJyKSJCxC_bcU8FOBe:-MgMm3v-8_OnYQsH3H1x:b-4xdg8d) ([raw file](https://github.com/oscarlab/graphene/blob/f1ad322b8d7e2418d1b1ed1bdd6a5c2467e8fefd/Pal/src/host/Linux/db_main.c#L161)):*

It does look like it. On the other hand, it means also annotating `unix_to_pal_error` (as well as `unix_to_pal_error_positive`) with `__attribute_no_stack_protector`... Should I?
</blockquote></details>

You know what? Just return `1` here, who cares.

dimakuv (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[Pal/src/host/Linux/db_main.c, line 161 at r1](https://reviewable.io/reviews/oscarlab/graphene/2618#-MgJyKSJCxC_bcU8FOBe:-MgPL2yi3qX02U-NyuUQ:bc14mhb) ([raw file](https://github.com/oscarlab/graphene/blob/f1ad322b8d7e2418d1b1ed1bdd6a5c2467e8fefd/Pal/src/host/Linux/db_main.c#L161)):*

You know what? Just return `1` here, who cares.
</blockquote></details>

Oops, true. Well, let's not hard-code `1` as Borys suggests, but at least an actual PAL error. Just do `INLINE_SYSCALL(exit_group, 1, PAL_ERROR_DENIED)`.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: 2 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[Pal/src/host/Linux/db_main.c, line 161 at r1](https://reviewable.io/reviews/oscarlab/graphene/2618#-MgJyKSJCxC_bcU8FOBe:-MgPW7KZ7CemyG3xkzEm:bbr3dvq) ([raw file](https://github.com/oscarlab/graphene/blob/f1ad322b8d7e2418d1b1ed1bdd6a5c2467e8fefd/Pal/src/host/Linux/db_main.c#L161)):*

Oops, true. Well, let's not hard-code `1` as Borys suggests, but at least an actual PAL error. Just do `INLINE_SYSCALL(exit_group, 1, PAL_ERROR_DENIED)`.
</blockquote></details>

Sounds good. Done.

dimakuv (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

pwmarcz (commented):
Rebased after f3e34953a4e9ea514f4f54e8f6139b2e0d17c860 changed `INLINE_SYSCALL` to `DO_SYSCALL`.

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618#-:-MgPpCas1LJZYkkhDCPP:b-fq6jbo)* status: 2 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski and @dimakuv)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @boryspoplawski)

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: all files reviewed, 1 unresolved discussion (waiting on @boryspoplawski)

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: :shipit: complete! all files reviewed, all discussions resolved

mkow (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2618)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2620, opened 2021-08-05
[LibOS] Relax check on UNIX domain socket's `addrlen` in `bind`

### LINKED ISSUE #2609: [LibOS] `bind` Unix socket, `addrlen` parameter checking is different from Linux, and causes error in Python
## Description of the problem
Graphene `shim_do_bind` function return  EINVAL for "addrlen is wrong" when `addrlen` is not same as `sizeof(struct sockaddr_un)`, which is 110.
https://github.com/oscarlab/graphene/blob/20156c6e23ea6570614197a97daf77dcd764153b/LibOS/shim/src/sys/shim_socket.c#L467-L469
But smaller size can work on Linux.

## Steps to reproduce
Copy example from `man bind` and change
```
 if (bind(sfd, (struct sockaddr *) &my_addr,  sizeof(struct sockaddr_un)) == -1)
```
to
```
 if (bind(sfd, (struct sockaddr *) &my_addr, strlen(MY_SOCK_PATH) + 4) == -1)
```
Compile and create Graphene manifest.
## Expected results
App with and without Graphene both can run without error.
## Actual results
App without Graphene can run without error.
App with Graphene returns `bind: Invalid argument`.

## Additional information
Bind Unix socket using Python can work on Linux, but fails with Graphene.
https://hg.python.org/cpython/file/3.6/Modules/socketmodule.c#l2700
```
res = bind(s->sock_fd, SAS2SA(&addrbuf), addrlen);
```
`addrlen` is calculated at
https://hg.python.org/cpython/file/3.6/Modules/socketmodule.c#l1581
```
*len_ret = path.len + offsetof(struct sockaddr_un, sun_path);
```

dimakuv:
## Description of the changes 

Previously, LibOS expected `addrlen` for UNIX domain sockets to be exactly the same as the size of `sockaddr_un` struct. This is not true in the common case: `addrlen` may be less, see `man unix`.

Fixes #2609.

## How to test this PR? 

Added a test case in our existing `unix.c` LibOS regression test.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)

### DIFF
--- a/LibOS/shim/src/sys/shim_socket.c
+++ b/LibOS/shim/src/sys/shim_socket.c
@@ -465,16 +465,31 @@ long shim_do_bind(int sockfd, struct sockaddr* addr, int _addrlen) {
     }
 
     if (sock->domain == AF_UNIX) {
-        if (addrlen != sizeof(struct sockaddr_un))
-            goto out;
-
-        struct sockaddr_un* saddr = (struct sockaddr_un*)addr;
-        char* spath               = saddr->sun_path;
-        struct shim_dentry* dent  = NULL;
-
-        if ((ret = path_lookupat(/*start=*/NULL, spath, LOOKUP_NO_FOLLOW | LOOKUP_CREATE, &dent)) < 0) {
+        if (addrlen < offsetof(struct sockaddr_un, sun_path) + 1 ||
+                addrlen > sizeof(struct sockaddr_un)) {
+            /* address length of the UNIX domain socket is typically calculated as
+             * `offsetof(struct sockaddr_un, sun_path) + strlen(sun_path) + 1`, see `man unix` */
             goto out;
         }
+
+        struct sockaddr_un* saddr = (struct sockaddr_un*)addr;
+        if (saddr->sun_path[0] == '\0') {
+            /* currently abstract UNIX domain sockets are not supported */
+            ret = -EOPNOTSUPP;
+            goto out;
+        }
+
+        size_t sun_path_size = addrlen - offsetof(struct sockaddr_un, sun_path);
+        if (strnlen(saddr->sun_path, sun_path_size) == sun_path_size) {
+            /* user-supplied UNIX domain name is not a NULL-terminating string */
+            goto out;
+        }
+
+        struct shim_dentry* dent = NULL;
+        ret = path_lookupat(/*start=*/NULL, saddr->sun_path, LOOKUP_NO_FOLLOW | LOOKUP_CREATE,
+                            &dent);
+        if (ret < 0)
+            goto out;
 
         if (!(dent->state & DENTRY_NEGATIVE)) {
             ret = -EADDRINUSE;
@@ -517,7 +532,7 @@ long shim_do_bind(int sockfd, struct sockaddr* addr, int _addrlen) {
     if (sock->domain == AF_UNIX) {
         struct shim_dentry* dent = sock->addr.un.dentry;
 
-        dent->state ^= DENTRY_NEGATIVE;
+        dent->state &= ~DENTRY_NEGATIVE;
         dent->state |= DENTRY_VALID;
         dent->fs   = &socket_builtin_fs;
         dent->type = S_IFSOCK;
@@ -733,16 +748,31 @@ long shim_do_connect(int sockfd, struct sockaddr* addr, int _addrlen) {
     }
 
     if (sock->domain == AF_UNIX) {
-        if (addrlen != sizeof(struct sockaddr_un))
-            goto out;
-
-        struct sockaddr_un* saddr = (struct sockaddr_un*)addr;
-        char* spath               = saddr->sun_path;
-        struct shim_dentry* dent  = NULL;
-
-        if ((ret = path_lookupat(/*start=*/NULL, spath, LOOKUP_CREATE | LOOKUP_FOLLOW, &dent)) < 0) {
+        if (addrlen < offsetof(struct sockaddr_un, sun_path) + 1 ||
+                addrlen > sizeof(struct sockaddr_un)) {
+            /* address length of the UNIX domain socket is typically calculated as
+             * `offsetof(struct sockaddr_un, sun_path) + strlen(sun_path) + 1`, see `man unix` */
             goto out;
         }
+
+        struct sockaddr_un* saddr = (struct sockaddr_un*)addr;
+        if (saddr->sun_path[0] == '\0') {
+            /* currently abstract UNIX domain sockets are not supported */
+            ret = -EOPNOTSUPP;
+            goto out;
+        }
+
+        size_t sun_path_size = addrlen - offsetof(struct sockaddr_un, sun_path);
+        if (strnlen(saddr->sun_path, sun_path_size) == sun_path_size) {
+            /* user-supplied UNIX domain name is not a NULL-terminating string */
+            goto out;
+        }
+
+        struct shim_dentry* dent = NULL;
+        ret = path_lookupat(/*start=*/NULL, saddr->sun_path, LOOKUP_FOLLOW | LOOKUP_CREATE,
+                            &dent);
+        if (ret < 0)
+            goto out;
 
         if (!(dent->state & DENTRY_NEGATIVE) && dent->type != S_IFSOCK) {
             ret = -ECONNREFUSED;
@@ -793,7 +823,7 @@ long shim_do_connect(int sockfd, struct sockaddr* addr, int _addrlen) {
     if (sock->domain == AF_UNIX) {
         struct shim_dentry* dent = sock->addr.un.dentry;
         lock(&dent->lock);
-        dent->state ^= DENTRY_NEGATIVE;
+        dent->state &= ~DENTRY_NEGATIVE;
         dent->state |= DENTRY_VALID;
         dent->fs   = &socket_builtin_fs;
         dent->type = S_IFSOCK;
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -900,6 +900,11 @@ class TC_80_Socket(RegressionTestCase):
         self.assertIn('[parent] TEST OK', stdout)
 
     def test_100_socket_unix(self):
+        if os.path.exists("dummy"):
+            os.remove("dummy")
+        if os.path.exists("u"):
+            os.remove("u")
+
         stdout, _ = self.run_binary(['unix'])
         self.assertIn('Data: This is packet 0', stdout)
         self.assertIn('Data: This is packet 1', stdout)
--- a/LibOS/shim/test/regression/unix.c
+++ b/LibOS/shim/test/regression/unix.c
@@ -7,6 +7,7 @@
 #include <err.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -47,8 +48,10 @@ static int server_dummy_socket(void) {
 
     address.sun_family = AF_UNIX;
     strncpy(address.sun_path, "dummy", sizeof(address.sun_path));
+    socklen_t minimal_address_size = offsetof(struct sockaddr_un, sun_path) +
+                                     strlen(address.sun_path) + 1;
 
-    if (bind(create_socket, (struct sockaddr*)&address, sizeof(address)) < 0) {
+    if (bind(create_socket, (struct sockaddr*)&address, minimal_address_size) < 0) {
         perror("bind");
         close(create_socket);
         exit(1);


boryspoplawski (approved):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

mkow (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required) (waiting on @dimakuv)

___
*[LibOS/shim/src/sys/shim_socket.c, line 470 at r1](https://reviewable.io/reviews/oscarlab/graphene/2620#-MgWUXI18wm6BLtDY6pw:-MgWUXI24vRB_R76kZ-P:bglzdw) ([raw file](https://github.com/oscarlab/graphene/blob/8fbcb995c05beec9f00a73169d4080f296af8d09/LibOS/shim/src/sys/shim_socket.c#L470)):*
> ```C
>         if (addrlen > sizeof(struct sockaddr_un)) {
>             /* address length of the UNIX domain socket is typically calculated as
>              * `offsetof(struct sockaddr_un, sun_path) + strlen(sun_path) + 1`, see `man unix` */
> ```

So, shouldn't we also check the minimum size? I.e. whether `addrlen >= offsetof(struct sockaddr_un, sun_path) + 1`.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow)

___
*[LibOS/shim/src/sys/shim_socket.c, line 470 at r1](https://reviewable.io/reviews/oscarlab/graphene/2620#-MgWUXI18wm6BLtDY6pw:-MgWfIV70pdP3JkDxY3S:b-ougsr6) ([raw file](https://github.com/oscarlab/graphene/blob/8fbcb995c05beec9f00a73169d4080f296af8d09/LibOS/shim/src/sys/shim_socket.c#L470)):*

So, shouldn't we also check the minimum size? I.e. whether `addrlen >= offsetof(struct sockaddr_un, sun_path) + 1`.
</blockquote></details>

We should.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: 2 of 3 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/src/sys/shim_socket.c, line 470 at r1](https://reviewable.io/reviews/oscarlab/graphene/2620#-MgWUXI18wm6BLtDY6pw:-MgdiBIX5DsTL8GDrCen:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/8fbcb995c05beec9f00a73169d4080f296af8d09/LibOS/shim/src/sys/shim_socket.c#L470)):*

We should.
</blockquote></details>

Done.

boryspoplawski (changes requested):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[LibOS/shim/src/sys/shim_socket.c, line 473 at r2](https://reviewable.io/reviews/oscarlab/graphene/2620#-MgeXRri-rsKiefTZ7cp:-MgeXRrj0Ag8q62o85Nw:b-cdh1rj) ([raw file](https://github.com/oscarlab/graphene/blob/ff2d6b3ee05fc5987cab2c62bf58917d20293582/LibOS/shim/src/sys/shim_socket.c#L473)):*
> ```C
>              * `offsetof(struct sockaddr_un, sun_path) + strlen(sun_path) + 1`, see `man unix` */
>             goto out;
>         }
> ```

We should also check whether there is a null byte in `sun_path` in first `addrlen - offsetof(struct sockaddr_un, sun_path)` bytes and fail if there isn't. Also we seem not to support abstract unix sockets (? I thought we did.) so we should also fail in case of `sun_path[0] == 0`.
(sorry, realized only now)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: 2 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/src/sys/shim_socket.c, line 473 at r2](https://reviewable.io/reviews/oscarlab/graphene/2620#-MgeXRri-rsKiefTZ7cp:-MgerN13BpwZKJn5nBRT:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/ff2d6b3ee05fc5987cab2c62bf58917d20293582/LibOS/shim/src/sys/shim_socket.c#L473)):*

We should also check whether there is a null byte in `sun_path` in first `addrlen - offsetof(struct sockaddr_un, sun_path)` bytes and fail if there isn't. Also we seem not to support abstract unix sockets (? I thought we did.) so we should also fail in case of `sun_path[0] == 0`.
(sorry, realized only now)
</blockquote></details>

Done.

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @mkow)

___
*[-- commits, line 2 at r3](https://reviewable.io/reviews/oscarlab/graphene/2620#-MgfaUPN5ZRpcxSJhkgm:-MgfbJ1iC-EHNmEiDlxa:bsvtm3h) ([raw file](https://github.com/oscarlab/graphene/blob/09864df70706a1bc1ad6ec12846d98043e000d93/-- commits#L2)):*

Now also in common (can be fixed at final rebase)
</blockquote></details>

Typo, I meant `connect`.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: 2 of 3 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[-- commits, line 2 at r3](https://reviewable.io/reviews/oscarlab/graphene/2620#-MgfaUPN5ZRpcxSJhkgm:-MgfuvwF4OLnxkgtfw42:bsvnfxv) ([raw file](https://github.com/oscarlab/graphene/blob/09864df70706a1bc1ad6ec12846d98043e000d93/-- commits#L2)):*

Typo, I meant `connect`.
</blockquote></details>

I'll fix during rebase.
___
*[LibOS/shim/src/sys/shim_socket.c, line 477 at r3](https://reviewable.io/reviews/oscarlab/graphene/2620#-MgfafUI2vMOX8cunJiz:-MgfvlYP6CrRmyLQGGcs:bxz1x5p) ([raw file](https://github.com/oscarlab/graphene/blob/09864df70706a1bc1ad6ec12846d98043e000d93/LibOS/shim/src/sys/shim_socket.c#L477)):*

We need to set the return error (default `EINVAL` is misleading). Maybe `ENOSYS` or `EOPNOTSUP`?
</blockquote></details>

Done. Decided for `EOPNOTSUP` because `ENOSYS` may be confused by novice users with "Oh, bind doesn't work at all"
___
*[LibOS/shim/src/sys/shim_socket.c, line 759 at r3](https://reviewable.io/reviews/oscarlab/graphene/2620#-Mgfb9FqEumj_cF__rYF:-MgfvyGy6V0tRPAcjTl_:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/09864df70706a1bc1ad6ec12846d98043e000d93/LibOS/shim/src/sys/shim_socket.c#L759)):*

ditto (error val)
</blockquote></details>

Done.

boryspoplawski (commented):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

mkow (changes requested):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: all files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @dimakuv)

___
*[LibOS/shim/src/sys/shim_socket.c, line 477 at r4](https://reviewable.io/reviews/oscarlab/graphene/2620#-MggI5nK-EFEoHxs2b4L:-MggI5nK-EFEoHxs2b4M:b-5q0u4x) ([raw file](https://github.com/oscarlab/graphene/blob/f6bbcea1f7d56fe271d40e8a98bbc3b7f4f0327f/LibOS/shim/src/sys/shim_socket.c#L477)):*
> ```C
>         struct sockaddr_un* saddr = (struct sockaddr_un*)addr;
>         if (saddr->sun_path[0] == '\0') {
>             /* currently do not support abstract UNIX domain sockets */
> ```

This sounds weird, because it's now in imperative mood. I'd either add "we" or change to "currently XYZ is not supported".
___
*[LibOS/shim/src/sys/shim_socket.c, line 760 at r4](https://reviewable.io/reviews/oscarlab/graphene/2620#-MggKFZ8EHtEB2A17qOA:-MggKFZ8EHtEB2A17qOB:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/f6bbcea1f7d56fe271d40e8a98bbc3b7f4f0327f/LibOS/shim/src/sys/shim_socket.c#L760)):*
> ```C
>         struct sockaddr_un* saddr = (struct sockaddr_un*)addr;
>         if (saddr->sun_path[0] == '\0') {
>             /* currently do not support abstract UNIX domain sockets */
> ```

ditto
___
*[LibOS/shim/src/sys/shim_socket.c, line 772 at r4](https://reviewable.io/reviews/oscarlab/graphene/2620#-MggPQh8AONsI3YRkrHQ:-MggPQh95gZ3w-ccDTaR:b-uhaf9k) ([raw file](https://github.com/oscarlab/graphene/blob/f6bbcea1f7d56fe271d40e8a98bbc3b7f4f0327f/LibOS/shim/src/sys/shim_socket.c#L772)):*
> ```C
> 
>         struct shim_dentry* dent = NULL;
>         ret = path_lookupat(/*start=*/NULL, saddr->sun_path, LOOKUP_NO_FOLLOW | LOOKUP_CREATE,
> ```

You introduced a bug here :) `LOOKUP_FOLLOW` was intended here, because `connect()` actually follows links. Also, why is `LOOKUP_CREATE` passed here?

Here goes a small example which checks this, we probably should check such stuff in our regression?
Note that this isn't cleaned up, so if you want to use it them please first re-read everything :)

```C
#define _GNU_SOURCE         
#include <err.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/syscall.h>
#include <sys/time.h>
#include <sys/mman.h>
#include <sys/socket.h>
#include <sys/types.h>
#include <sys/un.h>
#include <sys/wait.h>
#include <unistd.h>

int main(void) {
        struct sockaddr_un address;
        int socket_fd, ret;
        socklen_t address_length;

        socket_fd = socket(AF_UNIX, SOCK_STREAM, 0);
        if (socket_fd < 0) {
                printf("socket() failed\n");
                return 1;
        } 

        memset(&address, 0, sizeof(address));
        address.sun_family = AF_UNIX;
        snprintf(address.sun_path, sizeof(address.sun_path), "/tmp/dupalink");
        address_length = sizeof(address);
        ret = connect(socket_fd, &address, address_length);
        if (ret == -1) {
                err(1, "connect() failed");
        } else {
                printf("%d\n", ret);
        }
        write(socket_fd, "dupa\n", 5);

        return 0;
}
```
Setup:
```Bash
ln -s /tmp/dupa /tmp/dupalink
nc -l -U /tmp/dupa
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: 2 of 3 files reviewed, 4 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/src/sys/shim_socket.c, line 477 at r4](https://reviewable.io/reviews/oscarlab/graphene/2620#-MggI5nK-EFEoHxs2b4L:-MgiunyDE82mCtT85hWq:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f6bbcea1f7d56fe271d40e8a98bbc3b7f4f0327f/LibOS/shim/src/sys/shim_socket.c#L477)):*

This sounds weird, because it's now in imperative mood. I'd either add "we" or change to "currently XYZ is not supported".
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_socket.c, line 760 at r4](https://reviewable.io/reviews/oscarlab/graphene/2620#-MggKFZ8EHtEB2A17qOA:-MgiuoV0-rNW_NQhBtXp:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f6bbcea1f7d56fe271d40e8a98bbc3b7f4f0327f/LibOS/shim/src/sys/shim_socket.c#L760)):*

ditto
</blockquote></details>

Done.
___
*[LibOS/shim/src/sys/shim_socket.c, line 772 at r4](https://reviewable.io/reviews/oscarlab/graphene/2620#-MggPQh8AONsI3YRkrHQ:-Mgiv4Jw4HuXWidZu8Wh:b-tyitcm) ([raw file](https://github.com/oscarlab/graphene/blob/f6bbcea1f7d56fe271d40e8a98bbc3b7f4f0327f/LibOS/shim/src/sys/shim_socket.c#L772)):*
Done. But this comment opens up a big can of worms.

> Also, why is LOOKUP_CREATE passed here?

This is an interesting question. My investigation brings me to Chia-Che's "First official commit" (of course!) but I seem to understand the logic of having `LOOKUP_CREATE`. First, note that `LOOKUP_CREATE` doesn't really create a valid ready-to-use dentry, but instead returns a *negative* dentry which can be created: https://github.com/oscarlab/graphene/blob/45f5f12fc8c7c77387d0ba2900bb6274b8882c4e/LibOS/shim/include/shim_fs.h#L399

So, in this line, we ask Graphene "Could we technically create a file with this name in Graphene's virtual FS?" If the answer is no, then `ret < 0` and we just return an error. If the answer is yes, we double-check below that the dentry is either still negative (== the server didn't yet bind) or it is valid and a socket (== the server already bound). Otherwise, we think "This dentry is not a socket but rather some regular file, let's refuse connection".

Then, we do the actual connection via `DkStreamOpen()`. If the host's `connect()` fails for whatever reason, this is reflected here and we return with an error. Notice that the dentry is kept in the Graphene virtual FS, but the dentry is negative, so it doesn't really exist (more like a placeholder) and can be overwritten by anyone.

Finally, when the host's connection was successful, we mark the dentry as non-negative and valid: https://github.com/oscarlab/graphene/blob/09e226c5239638d4ada3455e8a02e2761cf5f43e/LibOS/shim/src/sys/shim_socket.c#L796

At this point, the UNIX socket domain was successfully connected at the host level, there is a corresponding and valid dentry in Graphene virtual FS, everything is fine.

So the whole logic is a bit perverse but probably that's the only way to be implemented. In particular:
1. Prepare the dentry but keep it negative
2. Try the actual host connection
3. Finalize the dentry

> Here goes a small example which checks this, we probably should check such stuff in our regression?

Thanks for the example, in hindsight connecting to a symlink makes sense. But as much as I want to have `dupa.c` in our tests, we have a problem: we do not allow UNIX domain sockets between a Graphene instance and the host (only between two Graphene instances). And since Graphene doesn't support creating symlinks inside its virtual FS, we can't really test this in a start-server-then-fork-client manner.

mkow (commented):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

___
*[LibOS/shim/src/sys/shim_socket.c, line 772 at r4](https://reviewable.io/reviews/oscarlab/graphene/2620#-MggPQh8AONsI3YRkrHQ:-MgjvexC3XkPCmzYIFtX:b-f98z5e) ([raw file](https://github.com/oscarlab/graphene/blob/f6bbcea1f7d56fe271d40e8a98bbc3b7f4f0327f/LibOS/shim/src/sys/shim_socket.c#L772)):*
> Finally, when the host's connection was successful, we mark the dentry as non-negative and valid: 

Waaait, but that's `connect()`, not `bind()`, why would it create any files? Or maybe I'm having a brainlag and I'm missing something?

Ad these lookup flags: I'm summoning @pwmarcz :)
Ad example: right :/

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/src/sys/shim_socket.c, line 772 at r4](https://reviewable.io/reviews/oscarlab/graphene/2620#-MggPQh8AONsI3YRkrHQ:-MgkJ4S7-e5i9M-AH9PH:b-fk91b6) ([raw file](https://github.com/oscarlab/graphene/blob/f6bbcea1f7d56fe271d40e8a98bbc3b7f4f0327f/LibOS/shim/src/sys/shim_socket.c#L772)):*

> Finally, when the host's connection was successful, we mark the dentry as non-negative and valid: 

Waaait, but that's `connect()`, not `bind()`, why would it create any files? Or maybe I'm having a brainlag and I'm missing something?

Ad these lookup flags: I'm summoning @pwmarcz :)
Ad example: right :/
</blockquote></details>

Regarding `LOOKUP_FOLLOW`: since there's no way to create a symlink in Graphene (we only use them in procfs etc.), I wouldn't worry too much until we add support it.

Regarding `LOOKUP_CREATE`: We probably fill a new dentry because the socket might have been created by another process, and we don't have a dentry for it. (A normal lookup will ask the underlying filesystem, which is probably `chroot`, but since sockets are emulated by Graphene, you will not find the socket).

So, if I'm right, Graphene will claim a file doesn't exist (e.g. `stat("socket") = -ENOENT`) until you call `connect("socket")`, after which it will become visible. That's not ideal, but probably real applications use `connect()` without checking for existence by other means first?

Using XOR is super fishy, though. Notice that the dentry can be either positive (and `type == S_IFSOCK`) or negative, so `dent->state ^= DENTRY_NEGATIVE` will flip it. That's probably not intended, and maybe now is a good time to fix it.

Also, this code is racy (we retrieve the dentry, then process it, even though it might have changed in the meantime). A correct way would be to do it all while holding `g_dcache_lock`. But then again, it's probably not worth fixing before setting up some consistent locking rules for FS, and before _that_ I want to introduce inodes.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/src/sys/shim_socket.c, line 772 at r4](https://reviewable.io/reviews/oscarlab/graphene/2620#-MggPQh8AONsI3YRkrHQ:-MglKbvR1CJQusoRkX3J:b-xjzkpb) ([raw file](https://github.com/oscarlab/graphene/blob/f6bbcea1f7d56fe271d40e8a98bbc3b7f4f0327f/LibOS/shim/src/sys/shim_socket.c#L772)):*

Regarding `LOOKUP_FOLLOW`: since there's no way to create a symlink in Graphene (we only use them in procfs etc.), I wouldn't worry too much until we add support it.

Regarding `LOOKUP_CREATE`: We probably fill a new dentry because the socket might have been created by another process, and we don't have a dentry for it. (A normal lookup will ask the underlying filesystem, which is probably `chroot`, but since sockets are emulated by Graphene, you will not find the socket).

So, if I'm right, Graphene will claim a file doesn't exist (e.g. `stat("socket") = -ENOENT`) until you call `connect("socket")`, after which it will become visible. That's not ideal, but probably real applications use `connect()` without checking for existence by other means first?

Using XOR is super fishy, though. Notice that the dentry can be either positive (and `type == S_IFSOCK`) or negative, so `dent->state ^= DENTRY_NEGATIVE` will flip it. That's probably not intended, and maybe now is a good time to fix it.

Also, this code is racy (we retrieve the dentry, then process it, even though it might have changed in the meantime). A correct way would be to do it all while holding `g_dcache_lock`. But then again, it's probably not worth fixing before setting up some consistent locking rules for FS, and before _that_ I want to introduce inodes.
</blockquote></details>

I agree on Pawel's assessment of `LOOKUP_CREATE`. Note that Graphene has an interesting virtualization of UNIX domain sockets:
- At Graphene virtual FS, the UNIX domain socket is created/found under the user-supplied path (e.g., `/myapp/mysocket`)
- At host real FS, the underlying actual UNIX domain socket is created/found under the mangled path (`/tmp/graphene/[ hash(/myapp/mysocket) ]`)

So I guess what Chia-Che did with this `LOOKUP_CREATE` trick originally is to circumvent the race between the client and the server: from the client's perspective there is not yet any "Graphene virtual file" even if the server pinged the client "I created this file, connect now". Recall that before (and still?) there was no synchronization of dentries ("Graphene virtual files") between Graphene processes. So yeah, a hack.

I will fix the XOR fishiness.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: 2 of 3 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @mkow)

___
*[LibOS/shim/src/sys/shim_socket.c, line 772 at r4](https://reviewable.io/reviews/oscarlab/graphene/2620#-MggPQh8AONsI3YRkrHQ:-MglPdDC5atSDh3s5vU2:b-faf04v) ([raw file](https://github.com/oscarlab/graphene/blob/f6bbcea1f7d56fe271d40e8a98bbc3b7f4f0327f/LibOS/shim/src/sys/shim_socket.c#L772)):*

I agree on Pawel's assessment of `LOOKUP_CREATE`. Note that Graphene has an interesting virtualization of UNIX domain sockets:
- At Graphene virtual FS, the UNIX domain socket is created/found under the user-supplied path (e.g., `/myapp/mysocket`)
- At host real FS, the underlying actual UNIX domain socket is created/found under the mangled path (`/tmp/graphene/[ hash(/myapp/mysocket) ]`)

So I guess what Chia-Che did with this `LOOKUP_CREATE` trick originally is to circumvent the race between the client and the server: from the client's perspective there is not yet any "Graphene virtual file" even if the server pinged the client "I created this file, connect now". Recall that before (and still?) there was no synchronization of dentries ("Graphene virtual files") between Graphene processes. So yeah, a hack.

I will fix the XOR fishiness.
</blockquote></details>

Done (XOR).

mkow (dismissed):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

boryspoplawski (commented):
> Reviewed 1 of 1 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2620)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #2625, opened 2021-08-05
[LibOS] Simplify RTLD code

pwmarcz:
## Description of the changes 

* Remove searchable ELF map list (we only need to access two maps: executable and interpreter)
* Refactor interpreter loading code
* Convert some globals to `g_*` naming scheme

This is a byproduct of upcoming FS refactoring.

## How to test this PR? 

Existing tests should be enough.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2625)

### DIFF
--- a/LibOS/shim/include/shim_internal.h
+++ b/LibOS/shim/include/shim_internal.h
@@ -342,11 +342,8 @@ extern void* __load_address_end;
 
 extern const char** migrated_envp;
 
-struct shim_handle;
-int init_brk_from_executable(struct shim_handle* exec);
 int init_brk_region(void* brk_region, size_t data_segment_size);
 void reset_brk(void);
-int init_loader(void);
 int init_rlimit(void);
 
 bool is_user_memory_readable(const void* addr, size_t size);
@@ -357,6 +354,8 @@ uint64_t get_rlimit_cur(int resource);
 void set_rlimit_cur(int resource, uint64_t rlim);
 
 int object_wait_with_retry(PAL_HANDLE handle);
+
+struct shim_handle;
 
 void _update_epolls(struct shim_handle* handle);
 void delete_from_epoll_handles(struct shim_handle* handle);
--- a/LibOS/shim/include/shim_utils.h
+++ b/LibOS/shim/include/shim_utils.h
@@ -141,11 +141,14 @@ void free(void* mem);
 void* malloc_copy(const void* mem, size_t size);
 
 /* ELF binary loading */
+struct link_map;
+int init_elf_objects(void);
 int check_elf_object(struct shim_handle* file);
-int load_elf_object(struct shim_handle* file);
-int load_elf_interp(struct shim_handle* exec);
-noreturn void execute_elf_object(struct shim_handle* exec, void* argp, elf_auxv_t* auxp);
-int remove_loaded_libraries(void);
+int load_elf_object(struct shim_handle* file, struct link_map** out_map);
+int load_elf_interp(struct link_map* exec_map);
+noreturn void execute_elf_object(struct link_map* exec_map, void* argp, elf_auxv_t* auxp);
+void remove_loaded_elf_objects(void);
+int init_brk_from_executable(struct link_map* exec_map);
 
 /* gdb debugging support */
 int init_r_debug(void);
--- a/LibOS/shim/src/bookkeep/shim_vma.c
+++ b/LibOS/shim/src/bookkeep/shim_vma.c
@@ -700,6 +700,11 @@ int bkeep_munmap(void* addr, size_t length, bool is_internal, void** tmp_vma_ptr
         free_vma(vma2);
     }
 
+    /*
+     * TODO: We call `remove_r_debug()` on the assumption that `addr` might be the beginning of a
+     * loaded ELF object. However, `remove_r_debug()` assumes that `addr` is the load base, while
+     * the first mapping of an ELF object might begin later than its load base.
+     */
     remove_r_debug(addr);
     return ret;
 }
--- a/LibOS/shim/src/shim_init.c
+++ b/LibOS/shim/src/shim_init.c
@@ -90,11 +90,10 @@ void* migrated_memory_end;
 
 const char** migrated_envp __attribute_migratable;
 
-/* library_paths is populated with LD_PRELOAD entries once during LibOS
- * initialization and is used in __load_interp_object() to search for ELF
- * program interpreter in specific paths. Once allocated, its memory is
- * never freed or updated. */
-char** library_paths = NULL;
+/* `g_library_paths` is populated with LD_PRELOAD entries once during LibOS initialization and is
+ * used in `__load_interp_object()` to search for ELF program interpreter in specific paths. Once
+ * allocated, its memory is never freed or updated. */
+char** g_library_paths = NULL;
 
 struct shim_lock __master_lock;
 bool lock_enabled;
@@ -324,7 +323,7 @@ int init_stack(const char** argv, const char** envp, const char*** out_argp,
 static int read_environs(const char** envp) {
     for (const char** e = envp; *e; e++) {
         if (strstartswith(*e, "LD_LIBRARY_PATH=")) {
-            /* populate library_paths with entries from LD_LIBRARY_PATH envvar */
+            /* populate `g_library_paths` with entries from LD_LIBRARY_PATH envvar */
             const char* s = *e + static_strlen("LD_LIBRARY_PATH=");
             size_t npaths = 2; // One for the first entry, one for the last NULL.
             for (const char* tmp = s; *tmp; tmp++)
@@ -352,8 +351,8 @@ static int read_environs(const char** envp) {
 
             paths[cnt] = NULL;
 
-            assert(!library_paths);
-            library_paths = paths;
+            assert(!g_library_paths);
+            g_library_paths = paths;
             return 0;
         }
     }
@@ -446,7 +445,7 @@ noreturn void* shim_init(int argc, void* args) {
     elf_auxv_t* new_auxv;
     RUN_INIT(init_stack, argv, envp, &new_argp, &new_auxv);
 
-    RUN_INIT(init_loader);
+    RUN_INIT(init_elf_objects);
     RUN_INIT(init_signal_handling);
     RUN_INIT(init_ipc_worker);
 
@@ -498,16 +497,10 @@ noreturn void* shim_init(int argc, void* args) {
 
     set_default_tls();
 
-    lock(&g_process.fs_lock);
-    struct shim_handle* exec = g_process.exec;
-    get_handle(exec);
-    unlock(&g_process.fs_lock);
-
-    if (exec) {
-        /* Passing ownership of `exec` to `execute_elf_object`. */
-        execute_elf_object(exec, new_argp, new_auxv);
-    }
-    process_exit(0, 0);
+    /* At this point, the exec map has been either copied from checkpoint, or initialized in
+     * `init_loader`. */
+    execute_elf_object(/*exec_map=*/NULL, new_argp, new_auxv);
+    /* UNREACHABLE */
 }
 
 static int get_256b_random_hex_string(char* buf, size_t size) {
--- a/LibOS/shim/src/shim_rtld.c
+++ b/LibOS/shim/src/shim_rtld.c
@@ -38,10 +38,7 @@
 #include "shim_vma.h"
 
 /*
- * Structure describing a loaded shared object. The `l_next' and `l_prev' members form a chain of
- * all the shared objects loaded at startup.
- *
- * Originally based on glibc link_map structure.
+ * Structure describing a loaded ELF object. Originally based on glibc link_map structure.
  */
 struct link_map {
     /* Base address shared object is loaded at. */
@@ -49,10 +46,6 @@ struct link_map {
 
     /* Object identifier: file path, or PAL URI if path is unavailable. */
     const char* l_name;
-
-    /* Chain of loaded objects. */
-    struct link_map* l_next;
-    struct link_map* l_prev;
 
     /* Pointer to program header table. */
     ElfW(Phdr)* l_phdr;
@@ -107,8 +100,8 @@ struct loadcmd {
     int prot;
 };
 
-static struct link_map* loaded_libraries = NULL;
-static struct link_map* interp_map = NULL;
+static struct link_map* g_exec_map = NULL;
+static struct link_map* g_interp_map = NULL;
 
 static int read_file_fragment(struct shim_handle* file, void* buf, size_t size, file_off_t offset);
 
@@ -311,7 +304,7 @@ static int execute_loadcmd(const struct loadcmd* c, ElfW(Addr) load_addr,
     return 0;
 }
 
-static struct link_map* __map_elf_object(struct shim_handle* file, ElfW(Ehdr)* ehdr) {
+static struct link_map* map_elf_object(struct shim_handle* file, ElfW(Ehdr)* ehdr) {
     ElfW(Phdr)* phdr = NULL;
     ElfW(Addr) interp_libname_vaddr = 0;
     struct loadcmd* loadcmds = NULL;
@@ -474,51 +467,12 @@ err:
     return NULL;
 }
 
-static inline struct link_map* __search_map_by_name(const char* name) {
-    struct link_map* l = loaded_libraries;
-    size_t len = strlen(name);
-
-    while (l) {
-        if (l->l_name && !memcmp(l->l_name, name, len + 1))
-            break;
-        l = l->l_next;
-    }
-
-    return l;
-}
-
-static inline struct link_map* __search_map_by_handle(struct shim_handle* file) {
-    struct link_map* l = loaded_libraries;
-
-    while (l) {
-        if (l->l_file == file)
-            break;
-        l = l->l_next;
-    }
-
-    return l;
-}
-
-static int __remove_elf_object(struct link_map* l) {
-    if (l->l_prev)
-        l->l_prev->l_next = l->l_next;
-    if (l->l_next)
-        l->l_next->l_prev = l->l_prev;
-
+static void remove_elf_object(struct link_map* l) {
     remove_r_debug((void*)l->l_addr);
-
-    if (loaded_libraries == l)
-        loaded_libraries = l->l_next;
-
-    if (interp_map == l)
-        interp_map = NULL;
-
     free(l);
-
-    return 0;
 }
 
-static int __check_elf_header(ElfW(Ehdr)* ehdr) {
+static int check_elf_header(ElfW(Ehdr)* ehdr) {
     const char* errstring __attribute__((unused));
 
 #if __ELF_NATIVE_CLASS == 32
@@ -612,12 +566,12 @@ static int read_file_fragment(struct shim_handle* file, void* buf, size_t size, 
     return 0;
 }
 
-static int __load_elf_header(struct shim_handle* file, ElfW(Ehdr)* ehdr) {
+static int load_elf_header(struct shim_handle* file, ElfW(Ehdr)* ehdr) {
     int ret = read_file_fragment(file, ehdr, sizeof(*ehdr), /*offset=*/0);
     if (ret < 0)
         return ret;
 
-    ret = __check_elf_header(ehdr);
+    ret = check_elf_header(ehdr);
     if (ret < 0)
         return ret;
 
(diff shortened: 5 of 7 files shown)


mkow (dismissed):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2625)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (changes requested):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2625)* status: all files reviewed, 13 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[LibOS/shim/include/shim_utils.h, line 149 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPIIwa2KBfTKCh52UB:-MgPIIwa2KBfTKCh52UC:bc70vcd) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/include/shim_utils.h#L149)):*
> ```objc
> int load_elf_interp(struct link_map* exec_map);
> noreturn void execute_elf_object(struct link_map* exec_map, void* argp, elf_auxv_t* auxp);
> void remove_loaded_libraries(void);
> ```

I very much dislike the old naming. Shouldn't this be `remove_loaded_elf_objects`? (See my other comments on `libraries` in the names.)
___
*[LibOS/shim/src/shim_init.c, line 448 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPIfN-9NB6j1TpX1ee:-MgPIfN-9NB6j1TpX1ef:b-94nl2e) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_init.c#L448)):*
> ```C
>     RUN_INIT(init_stack, argv, envp, &new_argp, &new_auxv);
> 
>     RUN_INIT(init_loader);
> ```

I would prefer to rename it to something more meaningful now. `init_exec_and_its_interp`?
___
*[LibOS/shim/src/shim_init.c, line 502 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPBAd40NaTCHpk2sj-:-MgPBAd56CoVroKm-cFU:bwi5n2v) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_init.c#L502)):*
> ```C
>     /* At this point, the exec map has been either copied from checkpoint, or initialized in
>      * `init_loader`. */
>     execute_elf_object(/*exec_map=*/NULL, new_argp, new_auxv);
> ```

Can you add a new line with `/* UNREACHABLE */`?
___
*[LibOS/shim/src/shim_rtld.c, line 42 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPC7gbEdaEBY7h3QXR:-MgPC7gbEdaEBY7h3QXS:b-x2bwsd) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L42)):*
> ```C
> /*
>  * Structure describing a loaded shared object. The `l_next' and `l_prev' members form a chain of
>  * all the shared objects loaded at startup.
> ```

Please remove the comment about `l_next/l_prev`
___
*[LibOS/shim/src/shim_rtld.c, line 106 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPHXv482vJTvpIWCoF:-MgPHXv482vJTvpIWCoG:b-tj461) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L106)):*
> ```C
> };
> 
> struct link_map* g_exec_map = NULL;
> ```

Why not static?
___
*[LibOS/shim/src/shim_rtld.c, line 636 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPDGExCqFY3eHu2UDG:-MgPDGExCqFY3eHu2UDH:b4y6ktd) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L636)):*
> ```C
>             filename_len = interp_name_len - i - 1;
>         }
>     }
> ```

Actually, why do we extract the filename from the full interpreter path? This is non-standard behavior, right? Is this purely a Graphene feature (to always use our `/lib/ld-linux-x86-64.so.2`)?
___
*[LibOS/shim/src/shim_rtld.c, line 661 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPJ-4V2wzxrhR6Ec6c:-MgPJ-4V2wzxrhR6Ec6d:bewo70m) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L661)):*
> ```C
> }
> 
> static int __load_interp_object(struct link_map* exec_map) {
> ```

Any reason to keep `__`?
___
*[LibOS/shim/src/shim_rtld.c, line 697 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPJ2LM1yWjwAIMEBQe:-MgPJ2LM1yWjwAIMEBQf:bc70vcd) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L697)):*
> ```C
> }
> 
> void remove_loaded_libraries(void) {
> ```

I very much dislike the old naming. Shouldn't this be `remove_loaded_elf_objects`? (See my other comments on `libraries` in the names.)
___
*[LibOS/shim/src/shim_rtld.c, line 768 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPFVCL1qzLFr06gqNS:-MgPFVCM6CXBbw5nENPS:b-r50snp) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L768)):*
> ```C
>         ret = load_elf_object(exec, &g_exec_map);
>         if (ret < 0) {
>             // TODO: Actually verify that the non-PIE-ness was the real cause of loading failure.
> ```

This would be cool to finally implement (as a separate PR). In general, we give the below confusing message even when the executable file doesn't exist or is a Bash/Perl script. This mightily confuses end users, who try `sgx.nonpie_binary` and still fail.

It would be better to assign proper return error codes inside `load_elf_object()` and print different messages based on these returned error codes. Not blocking, just something I remembered.
___
*[LibOS/shim/src/shim_rtld.c, line 799 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPJPQE6VRZwDdfWwI7:-MgPJPQE6VRZwDdfWwI8:bo2he0o) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L799)):*
> ```C
> }
> 
> int register_library(const char* name, unsigned long load_address) {
> ```

Why don't we have a counterpart function to this one? To stop tracking loaded libraries for debugging? Is it because they are removed during checkpoint-and-restore? Or this is simply not needed during a debug session? 
___
*[LibOS/shim/src/shim_rtld.c, line 821 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPHbWv2OIyHztu6bnd:-MgPHbWv2OIyHztu6bne:b-svsnhr) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L821)):*
> ```C
>     if (exec_map) {
>         /* If a new map is provided, it means we have cleared the existing one by calling
>          * `remove_loaded_libraries`. */
> ```

Maybe add a comment that this can happen on `execve` emulation?
___
*[LibOS/shim/src/shim_rtld.c, line 895 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPJjJa3TVEW_uDCFwL:-MgPJjJa3TVEW_uDCFwM:bu6weaq) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L895)):*
> ```C
> }
> 
> BEGIN_CP_FUNC(library) {
> ```

I dislike this `library` name, let's rename to `elf_object`. Or at least `link_map`. We never deal with libraries here, right? It is only ever the executable and its interpreter (if applicable).
___
*[LibOS/shim/src/shim_rtld.c, line 941 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPJu_vCp9MKPw0wzzr:-MgPJu_vCp9MKPw0wzzs:babddsf) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L941)):*
> ```C
> END_RS_FUNC(library)
> 
> BEGIN_CP_FUNC(loaded_libraries) {
> ```

I dislike this `loaded_libraries` name, let's rename to `exec_and_its_interp`?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2625)* status: 1 of 7 files reviewed, 12 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[LibOS/shim/include/shim_utils.h, line 149 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPIIwa2KBfTKCh52UB:-MgPiCcvC7afHfEi09m9:bskk5e3) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/include/shim_utils.h#L149)):*

I very much dislike the old naming. Shouldn't this be `remove_loaded_elf_objects`? (See my other comments on `libraries` in the names.)
</blockquote></details>

Renamed the way you suggested.
___
*[LibOS/shim/src/shim_init.c, line 448 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPIfN-9NB6j1TpX1ee:-MgPjDYJ1XwTNNp31_Aw:b-c14hfc) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_init.c#L448)):*

I would prefer to rename it to something more meaningful now. `init_exec_and_its_interp`?
</blockquote></details>

Changed to `init_elf_objects`, to match `loaded_elf_objects` etc.

(It's true that, by design, we track only the executable and interpreter ELFs, but I think there's no reason to expose that outside of the module).
___
*[LibOS/shim/src/shim_init.c, line 502 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPBAd40NaTCHpk2sj-:-MgPWgQn-zyfA3fz_opg:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_init.c#L502)):*

Can you add a new line with `/* UNREACHABLE */`?
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_rtld.c, line 42 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPC7gbEdaEBY7h3QXR:-MgPWmWvEW2Fvwc40iZZ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L42)):*

Please remove the comment about `l_next/l_prev`
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_rtld.c, line 106 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPHXv482vJTvpIWCoF:-MgPWofzAdIQRZwNtuuH:b96nlkc) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L106)):*

Why not static?
</blockquote></details>

Right, it should be.

(Initially I kept it exposed so that you can run `execute_elf_object(g_exec_map)`, but I decided `execute_elf_object(NULL)` is slightly less ugly).
___
*[LibOS/shim/src/shim_rtld.c, line 636 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPDGExCqFY3eHu2UDG:-MgPXVrzD4dQzQUOu5Hb:bcyob3u) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L636)):*

Actually, why do we extract the filename from the full interpreter path? This is non-standard behavior, right? Is this purely a Graphene feature (to always use our `/lib/ld-linux-x86-64.so.2`)?
</blockquote></details>

Yeah... I think 64-bit executables compiled with glibc always use `/lib64/ld-linux-x86-64.so.2` (the same path with `lib64`). We seem to also support `LD_LIBRARY_PATH` as override, which is again non-standard (it shouldn't change the interpreter).

I'm tempted to just change this code to use the interpreter path directly (and remove `LD_LIBRARY_PATH` parsing). @boryspoplawski @mkow any objections?

(I don't want to do it in this PR, though: because we currently mount Graphene libs under `/lib`, this is a breaking change, requires changing all manifests, etc. I can ticket it.)
___
*[LibOS/shim/src/shim_rtld.c, line 661 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPJ-4V2wzxrhR6Ec6c:-MgPad4q14AZ8bj3iQmL:bku8q6h) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L661)):*

Any reason to keep `__`?
</blockquote></details>

Actually I see we just rid of all `func / __func` pairs, so there is no reason to keep this naming scheme (looks like originally it was also used for static functions but I'm pretty sure we don't want that).

I removed `__` from remaining names in this file.
___
*[LibOS/shim/src/shim_rtld.c, line 697 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPJ2LM1yWjwAIMEBQe:-MgPj9vTAANojSERAMDP:bskk5e3) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L697)):*

I very much dislike the old naming. Shouldn't this be `remove_loaded_elf_objects`? (See my other comments on `libraries` in the names.)
</blockquote></details>

Renamed the way you suggested.
___
*[LibOS/shim/src/shim_rtld.c, line 768 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPFVCL1qzLFr06gqNS:-MgPbBg7A3DyGDM6aRuK:bqyqmn4) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L768)):*

This would be cool to finally implement (as a separate PR). In general, we give the below confusing message even when the executable file doesn't exist or is a Bash/Perl script. This mightily confuses end users, who try `sgx.nonpie_binary` and still fail.

It would be better to assign proper return error codes inside `load_elf_object()` and print different messages based on these returned error codes. Not blocking, just something I remembered.
</blockquote></details>

Acknowledged, although I don't want to go deeper right now.
___
*[LibOS/shim/src/shim_rtld.c, line 799 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPJPQE6VRZwDdfWwI7:-MgPcTDr65SEuRLcTIC5:bdk1hua) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L799)):*

Why don't we have a counterpart function to this one? To stop tracking loaded libraries for debugging? Is it because they are removed during checkpoint-and-restore? Or this is simply not needed during a debug session? 
</blockquote></details>

During `execve`, we unregister all libraries. But also we try to unregister a library during munmap, probably to cover the case of loading and unloading a library with `dlopen`.

(And I just realized it's probably not 100% correct because we look at `l_addr`... I added a TODO).

By the way, I know that `perf` tracks loaded libraries by tracking `mmap` and `munmap` operations (if a region is backed by a file, and is executable). Maybe we could do the same, and we wouldn't need `register_library` at all.
___
*[LibOS/shim/src/shim_rtld.c, line 821 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPHbWv2OIyHztu6bnd:-MgPhRqr3Vwgs441wK5i:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L821)):*

Maybe add a comment that this can happen on `execve` emulation?
</blockquote></details>

Done.
___
*[LibOS/shim/src/shim_rtld.c, line 895 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPJjJa3TVEW_uDCFwL:-MgPhmEEBBUFaoFOkKpv:b-ctflh5) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L895)):*

I dislike this `library` name, let's rename to `elf_object`. Or at least `link_map`. We never deal with libraries here, right? It is only ever the executable and its interpreter (if applicable).
</blockquote></details>

Renamed to `elf_object`.
___
*[LibOS/shim/src/shim_rtld.c, line 941 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPJu_vCp9MKPw0wzzr:-MgPi0PBCwFg1NYaNuuu:b-xmkixg) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L941)):*

I dislike this `loaded_libraries` name, let's rename to `exec_and_its_interp`?
</blockquote></details>

Renamed to `loaded_elf_objects`, for the same reason as the other comment (this is a public function called in `shim_clone`).

dimakuv (approved):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2625)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), "fixup! " found in commit messages' one-liners

___
*[LibOS/shim/src/shim_rtld.c, line 636 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPDGExCqFY3eHu2UDG:-MgQPsyx0zaxRjDIBPGL:b-tqaq6j) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L636)):*

Yeah... I think 64-bit executables compiled with glibc always use `/lib64/ld-linux-x86-64.so.2` (the same path with `lib64`). We seem to also support `LD_LIBRARY_PATH` as override, which is again non-standard (it shouldn't change the interpreter).

I'm tempted to just change this code to use the interpreter path directly (and remove `LD_LIBRARY_PATH` parsing). @boryspoplawski @mkow any objections?

(I don't want to do it in this PR, though: because we currently mount Graphene libs under `/lib`, this is a breaking change, requires changing all manifests, etc. I can ticket it.)
</blockquote></details>

OK, thanks for explanations. This is also my understanding that it's just some Graphene hack (probably to stash all dependent binaries under `<graphene>/Runtime/ -> /lib/`). Definitely for another PR.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2625)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required)

mkow (approved):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2625)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[LibOS/shim/src/shim_rtld.c, line 636 at r1](https://reviewable.io/reviews/oscarlab/graphene/2625#-MgPDGExCqFY3eHu2UDG:-MgQcZkvAv2qp8REs4IE:bwaix3b) ([raw file](https://github.com/oscarlab/graphene/blob/f197bb2140c44a1a55b09c967ccc34a446618b16/LibOS/shim/src/shim_rtld.c#L636)):*

OK, thanks for explanations. This is also my understanding that it's just some Graphene hack (probably to stash all dependent binaries under `<graphene>/Runtime/ -> /lib/`). Definitely for another PR.
</blockquote></details>

+1 to using the interp path directly. But can be done in a separate PR.

PR #2641, opened 2021-08-10
[Pal] Add `loader.env.[ENVIRON] = {passthrough=true}` manifest option

### LINKED ISSUE #2356: Allowlist for envp (& argv?)
## Description of the problem
Sometimes there may be cases in which it is desired to pass information from the host to the enclave, as long as it is handled in a secure way, as already discussed in #2347 .

However, when choosing the approach to pass the whole environment variables and sanitize them later in the application (e.g. in some kind of "premain"), [you can run into certain issues with the LD_ variables Graphene requires to correctly execute dynamically linked bianries](https://github.com/edgelesssys/marblerun/issues/158), thus potentially achieving unwanted code execution when a manifest with loose "pass-through" mounts/options was deployed.

## Suggested Solution

Given that the only other sane "easy" known way to actually pass through configuration or data from the host is by using "sgx.allowed_files", I would suggest to add certain allowed "envp" or "argv" (if this is handable in a reasonable way, argv seems more tricky to handle) arguments to be specifiable in the manifest, [similar to Occlum which allows to define certain "untrusted" env vars to pass through to the enclave.](https://github.com/occlum/occlum#configure-occlum)

## Thoughts
This is still not 100% free of misuse, however would at least prevent the case in which an user has to manually sanitize host envs or argv in their application (and potentially make mistakes on the way as it mixes host env & Graphene env vars), and would provide an easier way compared to `sgx.allowed_files` with a similar level of potential misuse.

I know from #2347 that @mkow discourages this kind of option, however I think there are definitely legit use cases to pass these kind of arguments from the host to the enclave, yet there seems to be a lack of other good/easy options so far. I think manually specifying the vars to pass through "untrusted" is still a better option than going with the shotgun-approach to allow all and sanitize later, or just don't and come up with clever workaround.

If there is another secure way for this use case, I think it requires at least some documentation then to show users how to handle this case correctly.

Feedback & Opinions appreciated!

dimakuv:
## Description of the changes 

It is now possible to define some envvars that are passed through from the host inside the Graphene instance (previously the only way to pass envvars was to set `loader.insecure__use_host_env`).

Fixes #2356.

## How to test this PR? 

A new LibOS regression test is added.

---
This change is [Reviewable](https://reviewable.io/reviews/gramineproject/graphene/2641)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -184,17 +184,16 @@ provided at runtime from an external (trusted) source.
 
 If the same variable is set in both, then ``loader.env.[ENVIRON]`` takes
 precedence. It is prohibited to specify both ``value`` and ``passthrough`` keys
-for the same environment variable. ``passthrough = false`` doesn't have any
-effect and is only specified here for completeness. If manifest option
-``insecure__use_host_env`` is specified, then ``passthrough`` manifest options
-have no effect (they are "consumed" by ``insecure__use_host_env``).
+for the same environment variable. If manifest option ``insecure__use_host_env``
+is specified, then ``passthrough = true`` manifest options have no effect (they
+are "consumed" by ``insecure__use_host_env``).
 
 .. note ::
    It is tempting to try to passthrough all environment variables using
    ``insecure__use_host_env`` and then disallow some of them using ``passthrough
    = false``. However, this deny list approach is intentionally prohibited.
-   Graphene completely ignores any ``passthrough`` manifest options when it
-   notices ``insecure__use_host_env``.
+   Graphene loudly fails if any ``passthrough = false`` manifest options are set
+   together with ``insecure__use_host_env``.
 
 Disabling ASLR
 ^^^^^^^^^^^^^^
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -156,10 +156,19 @@ both of the following options:
 ::
 
    loader.env.[ENVIRON] = "[VALUE]"
+   or
+   loader.env.[ENVIRON] = { value = "[VALUE]" }
+   or
+   loader.env.[ENVIRON] = { passthrough = [true|false] }
+
    loader.env_src_file = "file:file_with_serialized_envs"
 
-``loader.env.[ENVIRON]`` adds/overwrites a single environment variable and can
-be used multiple times to specify more than one variable.
+``loader.env.[ENVIRON]`` adds/overwrites/passes a single environment variable
+and can be used multiple times to specify more than one variable. To
+add/overwrite the environment variable, specify a TOML string (``"[VALUE]"``) or
+a TOML table with the key-value pair ``{ value = "[VALUE]" }``. To pass the
+environment variable from the host, specify a TOML table with the key-value pair
+``{ passthrough = true }``.
 
 ``loader.env_src_file`` allows to specify a URI to a file containing serialized
 environment, which can be generated using :file:`Tools/argv_serializer`. This
@@ -174,7 +183,18 @@ provided at runtime from an external (trusted) source.
    variables.
 
 If the same variable is set in both, then ``loader.env.[ENVIRON]`` takes
-precedence.
+precedence. It is prohibited to specify both ``value`` and ``passthrough`` keys
+for the same environment variable. ``passthrough = false`` doesn't have any
+effect and is only specified here for completeness. If manifest option
+``insecure__use_host_env`` is specified, then ``passthrough`` manifest options
+have no effect (they are "consumed" by ``insecure__use_host_env``).
+
+.. note ::
+   It is tempting to try to passthrough all environment variables using
+   ``insecure__use_host_env`` and then disallow some of them using ``passthrough
+   = false``. However, this deny list approach is intentionally prohibited.
+   Graphene completely ignores any ``passthrough`` manifest options when it
+   notices ``insecure__use_host_env``.
 
 Disabling ASLR
 ^^^^^^^^^^^^^^
--- a/LibOS/shim/test/regression/.gitignore
+++ b/LibOS/shim/test/regression/.gitignore
@@ -25,6 +25,7 @@
 /double_fork
 /env_from_file
 /env_from_host
+/env_passthrough
 /epoll_epollet
 /epoll_wait_timeout
 /eventfd
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -105,6 +105,7 @@ repo_manifests = \
 	device_passthrough.manifest \
 	env_from_file.manifest \
 	env_from_host.manifest \
+	env_passthrough.manifest \
 	file_check_policy_allow_all_but_log.manifest \
 	file_check_policy_strict.manifest \
 	multi_pthread_exitless.manifest
--- a/LibOS/shim/test/regression/env_passthrough.manifest.template
+++ b/LibOS/shim/test/regression/env_passthrough.manifest.template
@@ -1,0 +1,20 @@
+loader.preload = "file:{{ graphene.libos }}"
+loader.argv0_override = "bootstrap"
+libos.entrypoint = "bootstrap"
+
+loader.env.LD_LIBRARY_PATH = "/lib"
+
+loader.env.A = { passthrough = true }
+loader.env.B = { value = "OVERWRITTEN_VALUE" }
+loader.env.C = { passthrough = false }
+
+# loader.env.E = { passthrough = true, value = "THIS_IS_INCORRECT_SYNTAX" }
+
+fs.mount.lib.type = "chroot"
+fs.mount.lib.path = "/lib"
+fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
+
+sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
+sgx.trusted_files.bootstrap = "file:bootstrap"
+
+sgx.nonpie_binary = true
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -88,15 +88,33 @@ class TC_01_Bootstrap(RegressionTestCase):
         finally:
             os.remove('env_test_input')
 
+    def test_105_env_passthrough(self):
+        host_envs = {
+            'A': 'THIS_WILL_BE_PASSED',
+            'B': 'THIS_WILL_BE_OVERWRITTEN',
+            'C': 'THIS_SHOULDNT_BE_PASSED',
+            'D': 'THIS_SHOULDNT_BE_PASSED_TOO',
+        }
+        manifest_envs = {'LD_LIBRARY_PATH': '/lib'}
+        stdout, _ = self.run_binary(['env_passthrough'], env=host_envs)
+        self.assertIn('# of envs: %d\n' % (len(host_envs) - 2 + len(manifest_envs)), stdout)
+
+        # We don't enforce any specific order of envs, so we skip checking the index.
+        self.assertIn('] = LD_LIBRARY_PATH=/lib\n', stdout)
+        self.assertIn('] = A=THIS_WILL_BE_PASSED\n', stdout)
+        self.assertIn('] = B=OVERWRITTEN_VALUE\n', stdout)
+        self.assertNotIn('] = C=THIS_SHOULDNT_BE_PASSED\n', stdout)
+        self.assertNotIn('] = D=THIS_SHOULDNT_BE_PASSED_TOO\n', stdout)
+
     @unittest.skipUnless(HAS_SGX,
         'This test is only meaningful on SGX PAL because only SGX catches raw '
         'syscalls and redirects to Graphene\'s LibOS. If we will add seccomp to '
         'Linux PAL, then we should allow this test on Linux PAL as well.')
-    def test_105_basic_bootstrapping_static(self):
+    def test_106_basic_bootstrapping_static(self):
         stdout, _ = self.run_binary(['bootstrap_static'])
         self.assertIn('Hello world (bootstrap_static)!', stdout)
 
-    def test_106_basic_bootstrapping_pie(self):
+    def test_107_basic_bootstrapping_pie(self):
         stdout, _ = self.run_binary(['bootstrap_pie'])
         self.assertIn('User program started', stdout)
         self.assertIn('Local Address in Executable: 0x', stdout)
--- a/Pal/src/db_main.c
+++ b/Pal/src/db_main.c
@@ -64,24 +64,25 @@ static void load_libraries(void) {
     }
 }
 
-/* Finds `loader.env.{key}` in the manifest and returns its value in `val_out`/`passthrough_out`.
+/* Finds `loader.env.{key}` in the manifest and returns its value in `out_val`/`out_passthrough`.
  * Recognizes:
  *   - `loader.env.{key} = "val"`
  *   - `loader.env.{key} = { value = "val" }`
- *   - `loader.env.{key} = { passthrough = true|false }`, then `val_out = NULL`
- *   - `loader.env.{key}` doesn't exist, then returns 0 and `val_out = NULL`
+ *   - `loader.env.{key} = { passthrough = true|false }`, then `*out_val = NULL` and
+ *                                                             `*out_passthrough = true|false`
+ *   - `loader.env.{key}` doesn't exist, then returns 0 and `*out_val = NULL`
  *
- * For any other format, returns negative error code. The caller is responsible to free `val_out`.
+ * For any other format, returns negative error code. The caller is responsible to free `out_val`.
  */
-static int get_env_value_from_manifest(toml_table_t* toml_envs, const char* key, char** val_out,
-                                       bool* passthrough_out) {
+static int get_env_value_from_manifest(toml_table_t* toml_envs, const char* key, char** out_val,
+                                       bool* out_passthrough) {
     int ret;
 
-    assert(val_out);
-    assert(passthrough_out);
+    assert(out_val);
+    assert(out_passthrough);
 
-    *val_out = NULL;
-    *passthrough_out = false;
+    *out_val = NULL;
+    *out_passthrough = false;
 
     toml_table_t* toml_env_table = toml_table_in(toml_envs, key);
     if (!toml_env_table) {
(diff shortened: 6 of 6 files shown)


woju (commented):
> Reviewed 5 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2641)* status: 5 of 6 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel)

mkow (changes requested):
> Reviewed 2 of 6 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2641)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[Documentation/manifest-syntax.rst, line 197 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiWbVJ55Px23RufZnJt:-MiWbVJ64j8s4fpCr2Fw:ben7ncw) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Documentation/manifest-syntax.rst#L197)):*
> ```reStructuredText
>    = false``. However, this deny list approach is intentionally prohibited.
>    Graphene completely ignores any ``passthrough`` manifest options when it
>    notices ``insecure__use_host_env``.
> ```

Shouldn't we fail loudly in this case then? If you do what you described here, then users will just not notice that the blacklist was ignored.
___
*[Pal/src/db_main.c, line 71 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiWjQUnD8_ZAHBjVp1D:-MiWjQUnD8_ZAHBjVp1E:b-wdkp2l) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Pal/src/db_main.c#L71)):*
> ```C
>  *   - `loader.env.{key} = "val"`
>  *   - `loader.env.{key} = { value = "val" }`
>  *   - `loader.env.{key} = { passthrough = true|false }`, then `val_out = NULL`
> ```

not `*val_out`?
___
*[Pal/src/db_main.c, line 72 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiWjSm006leKpPSK0jh:-MiWjSm006leKpPSK0ji:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Pal/src/db_main.c#L72)):*
> ```C
>  *   - `loader.env.{key} = { value = "val" }`
>  *   - `loader.env.{key} = { passthrough = true|false }`, then `val_out = NULL`
>  *   - `loader.env.{key}` doesn't exist, then returns 0 and `val_out = NULL`
> ```

ditto
___
*[Pal/src/db_main.c, line 73 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiWjYqb2GK2Y6nLTloN:-MiWjYqb2GK2Y6nLTloO:b-scjyxx) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Pal/src/db_main.c#L73)):*
> ```C
>  *   - `loader.env.{key} = { passthrough = true|false }`, then `val_out = NULL`
>  *   - `loader.env.{key}` doesn't exist, then returns 0 and `val_out = NULL`
>  *
> ```

This table seems to be missing `passthrough_out` description altogether, is this intended or you just added it later and forgot to update? ;)
___
*[Pal/src/db_main.c, line 77 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiWfM1GE2rykXwMc3CA:-MiWfM1GE2rykXwMc3CB:b-8za76o) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Pal/src/db_main.c#L77)):*
> ```C
>  */
> static int get_env_value_from_manifest(toml_table_t* toml_envs, const char* key, char** val_out,
>                                        bool* passthrough_out) {
> ```

should be `out_` (i.e. prefix, not suffix; x2)
___
*[Pal/src/db_main.c, line 149 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiXZdb69xooBtjLua-g:-MiXZdb69xooBtjLua-h:bangkci) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Pal/src/db_main.c#L149)):*
> ```C
>     if (toml_envs_cnt == 0) {
>         /* no env entries found in the manifest */
>         return 0;
> ```

Do I read correctly, that you forward all the variables in this case? I don't like the signature and semantics of this whole function, especially the in-out argument, I almost missed this.
___
*[Pal/src/db_main.c, line 181 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiXTZyHA5loMshwByp4:-MiXTZyHA5loMshwByp5:bsis041) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Pal/src/db_main.c#L181)):*
> ```C
>         }
> 
>         /* at this point, we must passthrough envvar named `toml_env_key` from the host */
> ```

Please add a comment that this is O(host_env_cnt*passthrough_envs), but that shouldn't be a problem.
___
*[Pal/src/db_main.c, line 185 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiXUFrE0swy1QUc4UMz:-MiXUFrE0swy1QUc4UN-:b-a9l8lx) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Pal/src/db_main.c#L185)):*
> ```C
>             char* orig_env_key_end = strchr(*orig_env, '=');
> 
>             *orig_env_key_end = '\0';
> ```

You're temporarily destroying env of the current process, I don't like it. Why not just use `strstartswith` and then additionally check for `=`?
___
*[Pal/src/db_main.c, line 241 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiXWhiHDXQZleIn7cJf:-MiXWhiHDXQZleIn7cJg:b-hxvn13) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Pal/src/db_main.c#L241)):*
> ```C
>             return -PAL_ERROR_INVAL;
> 
>         *orig_env_key_end = '\0';
> ```

ditto, here you even leave the env destroyed in case of an error
___
*[Pal/src/db_main.c, line 553 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiXYDqq4O58khdm-bbJ:-MiXYDqq4O58khdm-bbK:bii1poi) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Pal/src/db_main.c#L553)):*
> ```C
>         }
>     } else {
>         ret = passthrough_envs_from_manifest(&environments);
> ```

We can't edit `environments`, it holds the current env for this very host process (it should actually be const, but for some reason isn't). Or maybe it actually doesn't edit it, but overwrites the pointer... No idea, this is too confusing, please refactor ;)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2641)* status: all files reviewed, 10 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @mkow)

___
*[Documentation/manifest-syntax.rst, line 197 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiWbVJ55Px23RufZnJt:-Mi_vyIE3Vyn1v7VbL3_:b819let) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Documentation/manifest-syntax.rst#L197)):*

Shouldn't we fail loudly in this case then? If you do what you described here, then users will just not notice that the blacklist was ignored.
</blockquote></details>

There was a reason why I chose to NOT fail when both `insecure__use_host_env` and `env.blabla` are specified. First, it is already possible to mix them together (then `env.blabla = "foo"` envs overwrite the insecure host envs) -- I definitely don't want to prohibit this legacy logic.

Second, it is very common to add `insecure__use_host_env = true` during debugging. Failing on a combination of `insecure__use_host_env` and `env.blabla` would mean that the debugging person needs to remove/comment out all `env.blabla` lines in the manifest, and then re-add/re-enable these lines again after debugging. I find it cumbersome.

What I can propose is: **always fail on encountering `env.blabla.passthrough = false`**. This setting is anyway meaningless and ignored by my current code. Does this sound reasonable?

mkow (changes requested):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2641)* status: all files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/db_main.c, line 185 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiXUFrE0swy1QUc4UMz:-Mib5MW6DPLsUzw-ePmz:bhj94q4) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Pal/src/db_main.c#L185)):*

Done. Used `alloc_substr` because we still need a properly NULL-terminated string (here and in other places).
</blockquote></details>

Why do we need this? You deallocate it right afterwards.
___
*[Pal/src/db_main.c, line 73 at r3](https://reviewable.io/reviews/oscarlab/graphene/2641#-Mib-t855TcrCS1XLive:-Mib-t868XmvuXXvWott:b-yxg57f) ([raw file](https://github.com/oscarlab/graphene/blob/bf2a3f0c85c28a44db2ea83a0210edfaff46582c/Pal/src/db_main.c#L73)):*
> ```C
>  *   - `loader.env.{key} = { passthrough = true|false }`, then `*out_val = NULL` and
>  *                                                             `*out_passthrough = true|false`
>  *   - `loader.env.{key}` doesn't exist, then returns 0 and `*out_val = NULL`
> ```

What about `*out_passthrough` in this case? In the previous case we need to read `*out_passthrough` if `*out_val = NULL`, so when this one happens we'd read uninitialized memory? (note that we can't (and probably don't need) differentiate both from the caller).
___
*[Pal/src/db_main.c, line 138 at r3](https://reviewable.io/reviews/oscarlab/graphene/2641#-Mib1hafA1M2i0VX6gLv:-Mib1hafA1M2i0VX6gLw:b-gw0ri6) ([raw file](https://github.com/oscarlab/graphene/blob/bf2a3f0c85c28a44db2ea83a0210edfaff46582c/Pal/src/db_main.c#L138)):*
> ```C
>     size_t idx = 0;
>     for (const char** orig_env = envp; *orig_env; orig_env++) {
>         new_envp[idx] = alloc_substr(*orig_env, strlen(*orig_env));
> ```

you can use `strdup()` instead
___
*[Pal/src/db_main.c, line 256 at r3](https://reviewable.io/reviews/oscarlab/graphene/2641#-Mib4kvE2XDhiH1uM5Ct:-Mib4kvE2XDhiH1uM5Cu:b-7aa24e) ([raw file](https://github.com/oscarlab/graphene/blob/bf2a3f0c85c28a44db2ea83a0210edfaff46582c/Pal/src/db_main.c#L256)):*
> ```C
> 
>         /* at this point, we must passthrough envvar named `toml_env_key` from the host; this is
>          * O(host_env_cnt * passthrough_envs), but happens only once during init so don't care */
> ```

`so don't care` -> `so we don't care` (sounds weird without this)
___
*[Pal/src/db_main.c, line 353 at r3](https://reviewable.io/reviews/oscarlab/graphene/2641#-Mib6XiEEKE7ZRgmGQpN:-Mib6XiEEKE7ZRgmGQpO:byh8lge) ([raw file](https://github.com/oscarlab/graphene/blob/bf2a3f0c85c28a44db2ea83a0210edfaff46582c/Pal/src/db_main.c#L353)):*
> ```C
>     for (const char** orig_env = envp; *orig_env; orig_env++) {
>         char* orig_env_key_end = strchr(*orig_env, '=');
>         assert(orig_env_key_end);
> ```

Can't this happen in runtime?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2641)* status: 5 of 6 files reviewed, 5 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[Pal/src/db_main.c, line 185 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiXUFrE0swy1QUc4UMz:-MieaxWR08QNpP28WCno:bxacpgx) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Pal/src/db_main.c#L185)):*

Why do we need this? You deallocate it right afterwards.
</blockquote></details>

I do `strcmp(env_key...)`  before `free(env_key)`. I find it easier to allocate a properly NULL-terminated string and compare it using `strcmp` rather than looking at corner cases of a non-NULL terminated string.
___
*[Pal/src/db_main.c, line 73 at r3](https://reviewable.io/reviews/oscarlab/graphene/2641#-Mib-t855TcrCS1XLive:-Miea62t5Ijj-ja1j1on:btkv6k5) ([raw file](https://github.com/oscarlab/graphene/blob/bf2a3f0c85c28a44db2ea83a0210edfaff46582c/Pal/src/db_main.c#L73)):*

What about `*out_passthrough` in this case? In the previous case we need to read `*out_passthrough` if `*out_val = NULL`, so when this one happens we'd read uninitialized memory? (note that we can't (and probably don't need) differentiate both from the caller).
</blockquote></details>

Done. I added a comment that `*out_passthrough = false` in this case (which is obvious from the code -- we set `*out_passthrough = false` in the very beginning). As you correctly mentioned, in the callers we don't differentiate between the cases 3 and 4 (both return 0 and  `*out_val = NULL`), because the callers already handle these two cases in a different manner.
___
*[Pal/src/db_main.c, line 138 at r3](https://reviewable.io/reviews/oscarlab/graphene/2641#-Mib1hafA1M2i0VX6gLv:-Mieaj4_EoNVJYn7cYP0:b-knl726) ([raw file](https://github.com/oscarlab/graphene/blob/bf2a3f0c85c28a44db2ea83a0210edfaff46582c/Pal/src/db_main.c#L138)):*

you can use `strdup()` instead
</blockquote></details>

Done. I wanted to have `alloc_substr()` everywhere for uniformity, but ok.
___
*[Pal/src/db_main.c, line 256 at r3](https://reviewable.io/reviews/oscarlab/graphene/2641#-Mib4kvE2XDhiH1uM5Ct:-MieaqsV3CzqdK7hge8l:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/bf2a3f0c85c28a44db2ea83a0210edfaff46582c/Pal/src/db_main.c#L256)):*

`so don't care` -> `so we don't care` (sounds weird without this)
</blockquote></details>

Done.
___
*[Pal/src/db_main.c, line 353 at r3](https://reviewable.io/reviews/oscarlab/graphene/2641#-Mib6XiEEKE7ZRgmGQpN:-MiebSwVBphuNQUuEeiV:b-6q4nbn) ([raw file](https://github.com/oscarlab/graphene/blob/bf2a3f0c85c28a44db2ea83a0210edfaff46582c/Pal/src/db_main.c#L353)):*

Can't this happen in runtime?
</blockquote></details>

No, it can't happen in runtime. Because about 20 lines above, we do the exact same for loop and have a proper `if (!orig_env_key_end)` check. So if the envvar doesn't have `=` in itself, this will be caught in that first for loop.

mkow (changes requested):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2641)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[Pal/src/db_main.c, line 185 at r2](https://reviewable.io/reviews/oscarlab/graphene/2641#-MiXUFrE0swy1QUc4UMz:-MifciVb3tnkpFLj4vKS:b-fryv2n) ([raw file](https://github.com/oscarlab/graphene/blob/95dfa0925a33a3a3c731af32a3724887a07b80eb/Pal/src/db_main.c#L185)):*

I do `strcmp(env_key...)`  before `free(env_key)`. I find it easier to allocate a properly NULL-terminated string and compare it using `strcmp` rather than looking at corner cases of a non-NULL terminated string.
</blockquote></details>

But why not `strncmp`?

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/graphene/2641)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/graphene/2641#-Mj-sR6wD0Y4-QL3bjd4:-Mj-sR6y5cmOuRISlZHN:b-qlu3v5) (no related file):*
I will simplify this PR to do only one pass over envvars.

---

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/gramineproject/graphene/2641)* status: all files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

*[a discussion](https://reviewable.io/reviews/gramineproject/graphene/2641#-Mj-sR6wD0Y4-QL3bjd4:-Mj-txLV7RquHvxKeC6L:b-jlkay2) (no related file):*

I will simplify this PR to do only one pass over envvars.
</blockquote></details>

[ Trying to publish again ]

---

dimakuv:
Jenkins, retest Docs-18.04 please (just testing)

dimakuv:
Jenkins, test Jenkins-18.04 please (just testing)

dimakuv:
Recreated as https://github.com/gramineproject/gramine/pull/56. Closing this one.

PR #2644, opened 2021-08-11
[Pal/Linux-SGX] Define SGX allowed/trusted/protected files as TOML arrays

### LINKED ISSUE #2593: SGX startup is slow due to quadratic TOML processing
## Description of the problem

Graphene-SGX startup is slow for manifests that have a lot of `trusted_files`.

## Steps to reproduce

On an Ubuntu 18.04 machine, current master branch (c602e5614db4b71e878a0b9ba3df816cee048a52). Try to run Python example with `graphene-sgx python -c "print('hello')"`.

## Expected results

This should be relatively quick.

## Actual results

The command takes 10 seconds:
```
real	0m10.816s
user	0m9.311s
sys	0m1.473s
```

It looks like `python.manifest.sgx` contains a lot of files (all of `/usr/lib/python3`, `/lib/x86_64-linux-gnu`, `/usr/lib/x86_64-linux-gnu`):

```
$ wc -l python.manifest.sgx
30049 python.manifest.sgx
$ grep /usr/lib/python3/ python.manifest.sgx | wc -l
10263
$ grep x86_64-linux-gnu python.manifest.sgx | wc -l
3621
```

Stopping in GDB shows that time is spent in [this loop](https://github.com/oscarlab/graphene/blob/c602e5614db4b71e878a0b9ba3df816cee048a52/Pal/src/host/Linux-SGX/enclave_framework.c#L750):

```c

    for (ssize_t i = 0; i < toml_trusted_files_cnt; i++) {
        const char* toml_trusted_file_key = toml_key_in(toml_trusted_files, i);
        assert(toml_trusted_file_key);
        toml_raw_t toml_trusted_file_raw = toml_raw_in(toml_trusted_files, toml_trusted_file_key);
        // ...
    }
```

It looks like `toml_raw_in` does a linear traversal of the whole `trusted_files` table.

dimakuv:
## Description of the changes 

This PR adds a new manifest syntax to define lists of SGX allowed, trusted, protected files. The previous syntax used TOML tables:

      sgx.trusted_files.file1 = "file:foo"
      sgx.trusted_files.file2 = "file:bar"

The new syntax uses TOML arrays:

      sgx.trusted_files = [ "file:foo", "file:bar" ]

The new syntax also allows to specify SHA256 hashes for a subset of trusted files (to skip hash generation during `graphene-sgx-sign`):

      [[sgx.trusted_files]]
      uri  = "file:trusted_testfile"
      sha256 = "c49a0aae384a14c8320f015ed5958d4402ba0726a31c4230cf772f76ff8aca2e"

The previous TOML-table syntax is still supported but deprecated. Graphene utility `graphene-sgx-sign` generates final SGX manifests using the new syntax, but `graphene-sgx` can still run old-syntax manifests.

All Graphene regression tests are updated to use the new syntax. **But all examples still use the old syntax** (to be fixed in next commits).

As a side effect, the TOML C library (`tomlc99`) is updated to the latest version -- it supports mixed TOML arrays.

Fixes #2593.

## How to test this PR? 

All tests should pass. Since I updated Graphene tests to the new syntax, and left the examples with the old syntax, both syntaxes will be tested in CI and both must succeed.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)

### DIFF
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -454,7 +454,10 @@ Allowed files
 
 ::
 
-    sgx.allowed_files.[identifier] = "[URI]"
+    sgx.allowed_files = [
+      "[URI]",
+      "[URI]",
+    ]
 
 This syntax specifies the files that are allowed to be created or loaded into
 the enclave unconditionally. In other words, allowed files can be opened for
@@ -471,16 +474,30 @@ Trusted files
 
 ::
 
-    sgx.trusted_files.[identifier] = "[URI]"
+    # entries can be strings
+    sgx.trusted_files = [
+      "[URI]",
+      "[URI]",
+    ]
+
+    # entries can also be tables
+    [[sgx.trusted_files]]
+    uri = "[URI]"
+    sha256 = "[HASH]"
 
 This syntax specifies the files to be cryptographically hashed at build time,
 and allowed to be accessed by the app in runtime only if their hashes match.
 This implies that trusted files can be only opened for reading (not for writing)
 and cannot be created if they do not exist already. The signer tool will
 automatically generate hashes of these files and add them to the SGX-specific
-manifest (``.manifest.sgx``). Marking files as trusted is especially useful for
-shared libraries: a |~| trusted library cannot be silently replaced by a
-malicious host because the hash verification will fail.
+manifest (``.manifest.sgx``). The manifest writer may also specify the hash for
+a file using the TOML-table syntax, in the field ``sha256``; in this case,
+hashing of the file will be skipped by the signer tool and the value in
+``sha256`` field will be used instead.
+
+Marking files as trusted is especially useful for shared libraries: a |~|
+trusted library cannot be silently replaced by a malicious host because the hash
+verification will fail.
 
 Protected files
 ^^^^^^^^^^^^^^^
@@ -488,7 +505,10 @@ Protected files
 ::
 
     sgx.protected_files_key = "[16-byte hex value]"
-    sgx.protected_files.[identifier] = "[URI]"
+    sgx.protected_files = [
+      "[URI]",
+      "[URI]",
+    ]
 
 This syntax specifies the files that are encrypted on disk and transparently
 decrypted when accessed by Graphene or by application running inside Graphene.
--- a/Documentation/manifest-syntax.rst
+++ b/Documentation/manifest-syntax.rst
@@ -693,3 +693,20 @@ Linux scheduler: the effective maximum is 250 samples per second.
 .. note::
    This option applies only to ``aex`` mode. In the ``ocall_*`` modes, currently
    all samples are taken.
+
+
+Deprecated options
+------------------
+
+Allowed/Trusted/Protected Files (deprecated schema)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+::
+
+    sgx.allowed_files.[identifier] = "[URI]"
+    sgx.trusted_files.[identifier] = "[URI]"
+    sgx.protected_files.[identifier] = "[URI]"
+
+These manifest options used the TOML-table schema that had a bogus
+``[identifier]`` key. This excessive TOML-table schema was replaced with a more
+appropriate TOML-array one.
--- a/LibOS/shim/test/fs/manifest.template
+++ b/LibOS/shim/test/fs/manifest.template
@@ -27,17 +27,21 @@ fs.mount.tmpfs.type = "tmpfs"
 fs.mount.tmpfs.path = "/mnt-tmpfs"
 fs.mount.tmpfs.uri = "file:dummy-unused-by-tmpfs-uri"
 
-sgx.trusted_files.entrypoint = "file:{{ entrypoint }}"
+sgx.nonpie_binary = true
+sgx.thread_num = 16
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.libgcc_s = "file:{{ arch_libdir }}/libgcc_s.so.1"
+sgx.allowed_files = [
+  "file:tmp/",
+]
 
-sgx.allowed_files.tmp_dir = "file:tmp/"
+sgx.trusted_files = [
+  "file:{{ entrypoint }}",
+  "file:{{ graphene.runtimedir() }}/",
+  "file:{{ arch_libdir }}/libgcc_s.so.1",
+]
 
 sgx.protected_files_key = "ffeeddccbbaa99887766554433221100"
-sgx.protected_files.input = "file:tmp/pf_input"
-sgx.protected_files.output = "file:tmp/pf_output"
-
-sgx.nonpie_binary = true
-
-sgx.thread_num = 16
+sgx.protected_files = [
+  "file:tmp/pf_input",
+  "file:tmp/pf_output",
+]
--- a/LibOS/shim/test/ltp/manifest.template
+++ b/LibOS/shim/test/ltp/manifest.template
@@ -28,17 +28,19 @@ fs.mount.tmp.uri = "file:/tmp"
 
 sys.brk.max_size = "32M"
 sys.stack.size = "4M"
-
 sgx.nonpie_binary = true
 
-sgx.trusted_files.entrypoint = "file:{{ entrypoint }}"
+sgx.allowed_files = [
+  "file:/tmp",
+]
 
-sgx.trusted_files.ld = "file:{{ graphene.runtimedir() }}/ld-linux-x86-64.so.2"
-sgx.trusted_files.libc = "file:{{ graphene.runtimedir() }}/libc.so.6"
-sgx.trusted_files.libdl = "file:{{ graphene.runtimedir() }}/libdl.so.2"
-sgx.trusted_files.libm = "file:{{ graphene.runtimedir() }}/libm.so.6"
-sgx.trusted_files.libpthread = "file:{{ graphene.runtimedir() }}/libpthread.so.0"
-sgx.trusted_files.librt = "file:{{ graphene.runtimedir() }}/librt.so.1"
-sgx.trusted_files.libstdbuf = "file:{{ coreutils_libdir }}/libstdbuf.so"
-
-sgx.allowed_files.tmp = "file:/tmp"
+sgx.trusted_files = [
+  "file:{{ entrypoint }}",
+  "file:{{ graphene.runtimedir() }}/ld-linux-x86-64.so.2",
+  "file:{{ graphene.runtimedir() }}/libc.so.6",
+  "file:{{ graphene.runtimedir() }}/libdl.so.2",
+  "file:{{ graphene.runtimedir() }}/libm.so.6",
+  "file:{{ graphene.runtimedir() }}/libpthread.so.0",
+  "file:{{ graphene.runtimedir() }}/librt.so.1",
+  "file:{{ coreutils_libdir }}/libstdbuf.so",
+]
--- a/LibOS/shim/test/regression/.gitignore
+++ b/LibOS/shim/test/regression/.gitignore
@@ -3,6 +3,7 @@
 
 /nonexisting_testfile
 /testfile
+/trusted_testfile
 
 /.cache
 /abort
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -197,5 +197,6 @@ clean-tmp:
 		__pycache__ \
 		libos-regression.xml \
 		testfile \
+		trusted_testfile \
 		tmp/* \
 		*.o
--- a/LibOS/shim/test/regression/argv_from_file.manifest.template
+++ b/LibOS/shim/test/regression/argv_from_file.manifest.template
@@ -9,9 +9,13 @@ fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
 
-sgx.allowed_files.argv = "file:argv_test_input"
-
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.bootstrap = "file:bootstrap"
-
 sgx.nonpie_binary = true
+
+sgx.allowed_files = [
+  "file:argv_test_input",
+]
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:bootstrap",
+]
--- a/LibOS/shim/test/regression/attestation.manifest.template
+++ b/LibOS/shim/test/regression/attestation.manifest.template
@@ -13,12 +13,13 @@ fs.mount.bin.type = "chroot"
 fs.mount.bin.path = "/bin"
 fs.mount.bin.uri = "file:/bin"
 
-# sgx-related
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.attestation = "file:attestation"
-
 sgx.nonpie_binary = true
-
 sgx.remote_attestation = true
+
 sgx.ra_client_spid = "{{ ra_client_spid }}"
 sgx.ra_client_linkable = {{ ra_client_linkable }}
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:attestation",
+]
--- a/LibOS/shim/test/regression/bootstrap_cpp.manifest.template
+++ b/LibOS/shim/test/regression/bootstrap_cpp.manifest.template
@@ -18,14 +18,14 @@ fs.mount.host_usr_lib.type = "chroot"
 fs.mount.host_usr_lib.path = "/usr/{{ arch_libdir }}"
 fs.mount.host_usr_lib.uri = "file:/usr/{{ arch_libdir }}"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.libgcc_s = "file:{{ arch_libdir }}/libgcc_s.so.1"
-sgx.trusted_files.libstdcxx = "file:/usr{{ arch_libdir }}/libstdc++.so.6"
-sgx.trusted_files.libunwindso = "file:/usr{{ arch_libdir }}/libunwind.so.8"
-sgx.trusted_files.liblzma = "file:{{ arch_libdir }}/liblzma.so.5"
-
-sgx.trusted_files.entrypoint = "file:bootstrap_cpp"
-
 sgx.thread_num = 8
-
 sgx.nonpie_binary = true
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:{{ arch_libdir }}/libgcc_s.so.1",
+  "file:/usr{{ arch_libdir }}/libstdc++.so.6",
+  "file:/usr{{ arch_libdir }}/libunwind.so.8",
+  "file:{{ arch_libdir }}/liblzma.so.5",
+  "file:bootstrap_cpp",
+]
--- a/LibOS/shim/test/regression/debug_log_file.manifest.template
+++ b/LibOS/shim/test/regression/debug_log_file.manifest.template
@@ -10,7 +10,9 @@ fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.bootstrap = "file:bootstrap"
-
 sgx.nonpie_binary = true
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:bootstrap",
+]
--- a/LibOS/shim/test/regression/debug_log_inline.manifest.template
+++ b/LibOS/shim/test/regression/debug_log_inline.manifest.template
@@ -9,7 +9,9 @@ fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.bootstrap = "file:bootstrap"
-
 sgx.nonpie_binary = true
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:bootstrap",
+]
--- a/LibOS/shim/test/regression/device_passthrough.manifest.template
+++ b/LibOS/shim/test/regression/device_passthrough.manifest.template
@@ -11,8 +11,9 @@ fs.mount.dev.type = "chroot"
 fs.mount.dev.path = "/dev/host-zero"
 fs.mount.dev.uri = "dev:/dev/zero"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-
-sgx.trusted_files.entrypoint = "file:device_passthrough"
-
 sgx.nonpie_binary = true
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:device_passthrough",
+]
--- a/LibOS/shim/test/regression/env_from_file.manifest.template
+++ b/LibOS/shim/test/regression/env_from_file.manifest.template
@@ -8,9 +8,12 @@ fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
 
+sgx.nonpie_binary = true
+
+# this tests the old syntax for allowed_files (TOML table)
 sgx.allowed_files.env = "file:env_test_input"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.bootstrap = "file:bootstrap"
-
-sgx.nonpie_binary = true
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:bootstrap",
+]
--- a/LibOS/shim/test/regression/env_from_host.manifest.template
+++ b/LibOS/shim/test/regression/env_from_host.manifest.template
@@ -9,7 +9,9 @@ fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.bootstrap = "file:bootstrap"
-
 sgx.nonpie_binary = true
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:bootstrap",
+]
--- a/LibOS/shim/test/regression/file_check_policy_allow_all_but_log.manifest.template
+++ b/LibOS/shim/test/regression/file_check_policy_allow_all_but_log.manifest.template
@@ -9,11 +9,19 @@ fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
 
+sgx.nonpie_binary = true
 sgx.file_check_policy = "allow_all_but_log"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.file_check_policy = "file:file_check_policy"
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+]
 
-sgx.trusted_files.test = "file:trusted_testfile"
+# below entry in sgx.trusted_files is to test TOML-table syntax without `sha256`
+[[sgx.trusted_files]]
+uri = "file:file_check_policy"
 
-sgx.nonpie_binary = true
+# below entry in sgx.trusted_files is for testing purposes (trusted_testfile has
+# hard-coded contents, so we can use pre-calculated SHA256 hash)
+[[sgx.trusted_files]]
+uri = "file:trusted_testfile"
+sha256 = "41dacdf1e6d0481d3b1ab1a91f93139db02b96f29cfdd3fb0b819ba1e33cafc4"
--- a/LibOS/shim/test/regression/file_check_policy_strict.manifest.template
+++ b/LibOS/shim/test/regression/file_check_policy_strict.manifest.template
@@ -9,11 +9,19 @@ fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
 
+sgx.nonpie_binary = true
 sgx.file_check_policy = "strict"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.file_check_policy = "file:file_check_policy"
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+]
 
-sgx.trusted_files.test = "file:trusted_testfile"
+# below entry in sgx.trusted_files is to test TOML-table syntax without `sha256`
+[[sgx.trusted_files]]
+uri = "file:file_check_policy"
 
-sgx.nonpie_binary = true
+# below entry in sgx.trusted_files is for testing purposes (trusted_testfile has
+# hard-coded contents, so we can use pre-calculated SHA256 hash)
+[[sgx.trusted_files]]
+uri = "file:trusted_testfile"
+sha256 = "41dacdf1e6d0481d3b1ab1a91f93139db02b96f29cfdd3fb0b819ba1e33cafc4"
--- a/LibOS/shim/test/regression/host_root_fs.manifest.template
+++ b/LibOS/shim/test/regression/host_root_fs.manifest.template
@@ -12,7 +12,9 @@ fs.mount.graphene_lib.type = "chroot"
 fs.mount.graphene_lib.path = "/lib"
 fs.mount.graphene_lib.uri = "file:{{ graphene.runtimedir() }}"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.host_root_fs = "file:{{ env.PWD }}/host_root_fs"
-
 sgx.nonpie_binary = true
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:{{ env.PWD }}/host_root_fs",
+]
--- a/LibOS/shim/test/regression/init_fail.manifest.template
+++ b/LibOS/shim/test/regression/init_fail.manifest.template
@@ -13,7 +13,9 @@ fs.mount.test.type = "chroot"
 fs.mount.test.path = "/test"
 fs.mount.test.uri = "file:I_DONT_EXIST"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.init_fail = "file:init_fail"
-
 sgx.nonpie_binary = true
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:init_fail",
+]
--- a/LibOS/shim/test/regression/init_fail2.manifest.template
+++ b/LibOS/shim/test/regression/init_fail2.manifest.template
@@ -8,11 +8,13 @@ fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.init_fail = "file:init_fail"
-
 sgx.nonpie_binary = true
 
 # this is an impossible combination of options, LibOS must fail very early in init process
 sgx.enclave_size = "256M"
 sys.brk.max_size = "512M"
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:init_fail",
+]
--- a/LibOS/shim/test/regression/large_mmap.manifest.template
+++ b/LibOS/shim/test/regression/large_mmap.manifest.template
@@ -12,11 +12,14 @@ fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.large_mmap = "file:large_mmap"
-
-sgx.allowed_files.testfile = "file:testfile"
-
 sgx.enclave_size = "8G"
-
 sgx.nonpie_binary = true
+
+sgx.allowed_files = [
+  "file:testfile",
+]
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:large_mmap",
+]
--- a/LibOS/shim/test/regression/manifest.template
+++ b/LibOS/shim/test/regression/manifest.template
@@ -22,17 +22,19 @@ fs.mount.bin.type = "chroot"
 fs.mount.bin.path = "/bin"
 fs.mount.bin.uri = "file:/bin"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.libgcc_s = "file:{{ arch_libdir }}/libgcc_s.so.1"
-sgx.trusted_files.libstdcxx = "file:/usr{{ arch_libdir }}/libstdc++.so.6"
-
-sgx.trusted_files.entrypoint = "file:{{ entrypoint }}"
-sgx.trusted_files.exec_victim = "file:exec_victim"
-
-sgx.allowed_files.tmp_dir = "file:tmp/"
-sgx.allowed_files.root = "file:root" # for getdents test
-sgx.allowed_files.testfile = "file:testfile" # for mmap_file test
-
 sgx.thread_num = 16
-
 sgx.nonpie_binary = true
+
+sgx.allowed_files = [
+  "file:tmp/",
+  "file:root", # for getdents test
+  "file:testfile", # for mmap_file test
+]
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:{{ arch_libdir }}/libgcc_s.so.1",
+  "file:/usr{{ arch_libdir }}/libstdc++.so.6",
+  "file:{{ entrypoint }}",
+  "file:exec_victim",
+]
--- a/LibOS/shim/test/regression/multi_pthread.manifest.template
+++ b/LibOS/shim/test/regression/multi_pthread.manifest.template
@@ -7,11 +7,13 @@ fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.multi_pthread = "file:multi_pthread"
-
 # app runs with 4 parallel threads + Graphene has couple internal threads
 sgx.thread_num = 8
 
 sgx.nonpie_binary = true
 sgx.enable_stats = true
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:multi_pthread",
+]
--- a/LibOS/shim/test/regression/multi_pthread_exitless.manifest.template
+++ b/LibOS/shim/test/regression/multi_pthread_exitless.manifest.template
@@ -7,12 +7,14 @@ fs.mount.lib.type = "chroot"
 fs.mount.lib.path = "/lib"
 fs.mount.lib.uri = "file:{{ graphene.runtimedir() }}"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.multi_pthread = "file:multi_pthread"
-
 # app runs with 4 parallel threads + Graphene has couple internal threads
 sgx.thread_num = 8
 sgx.rpc_thread_num = 8
 
 sgx.nonpie_binary = true
 sgx.enable_stats = true
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:multi_pthread",
+]
--- a/LibOS/shim/test/regression/openmp.manifest.template
+++ b/LibOS/shim/test/regression/openmp.manifest.template
@@ -26,10 +26,10 @@ fs.mount.usrlib.path = "/usrlib"
 fs.mount.usrlib.uri = "file:/usr/{{ arch_libdir }}"
 
 sgx.thread_num = 32
-
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.openmp = "file:openmp"
-
-sgx.trusted_files.libgomp_native = "file:/usr{{ arch_libdir }}/libgomp.so.1"
-
 sgx.nonpie_binary = true
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:openmp",
+  "file:/usr{{ arch_libdir }}/libgomp.so.1",
+]
--- a/LibOS/shim/test/regression/rename.manifest.template
+++ b/LibOS/shim/test/regression/rename.manifest.template
@@ -12,9 +12,13 @@ fs.mount.tmpfs.type = "tmpfs"
 fs.mount.tmpfs.path = "/mnt/tmpfs"
 fs.mount.tmpfs.uri = "file:dummy-unused-by-tmpfs-uri"
 
-sgx.trusted_files.runtime = "file:{{ graphene.runtimedir() }}/"
-sgx.trusted_files.entrypoint = "file:{{ entrypoint }}"
-
-sgx.allowed_files.tmp_dir = "file:tmp/"
-
 sgx.nonpie_binary = true
+
+sgx.allowed_files = [
+  "file:tmp/",
+]
+
+sgx.trusted_files = [
+  "file:{{ graphene.runtimedir() }}/",
+  "file:{{ entrypoint }}",
+]
--- a/LibOS/shim/test/regression/test_libos.py
+++ b/LibOS/shim/test/regression/test_libos.py
@@ -267,6 +267,15 @@ class TC_02_OpenMP(RegressionTestCase):
     'This test is only meaningful on SGX PAL because file-check-policy is '
     'only relevant to SGX.')
 class TC_03_FileCheckPolicy(RegressionTestCase):
+    @classmethod
+    def setUpClass(cls):
+        with open('trusted_testfile', 'w') as f:
+            f.write('trusted_testfile')
+
+    @classmethod
+    def tearDownClass(cls):
+        os.remove('trusted_testfile')
+
     def test_000_strict_success(self):
         stdout, _ = self.run_binary(['file_check_policy_strict', 'read', 'trusted_testfile'])
         self.assertIn('file_check_policy succeeded', stdout)
--- a/LibOS/shim/test/regression/trusted_testfile
+++ b/LibOS/shim/test/regression/trusted_testfile
@@ -1,1 +1,0 @@
-trusted_testfile
--- a/Pal/regression/manifest.template
+++ b/Pal/regression/manifest.template
@@ -7,7 +7,13 @@ loader.insecure__use_cmdline_argv = true
 loader.pal_internal_mem_size = "64M"
 
 fs.mount.root.uri = "file:"
-sgx.trusted_files.entrypoint = "file:{{ entrypoint }}"
+
 sgx.nonpie_binary = true # all tests are currently non-PIE unless overridden
 
-sgx.allowed_files.to_send_tmp = "file:to_send.tmp" # for SendHandle test
+sgx.allowed_files = [
+  "file:to_send.tmp", # for SendHandle test
+]
+
+sgx.trusted_files = [
+  "file:{{ entrypoint }}",
+]
--- a/Pal/src/host/Linux-SGX/enclave_framework.c
+++ b/Pal/src/host/Linux-SGX/enclave_framework.c
@@ -484,6 +484,11 @@ failed:
 static int register_file(const char* uri, const char* checksum_str, bool check_duplicates) {
     int ret;
 
+    if (checksum_str && strlen(checksum_str) != sizeof(sgx_file_hash_t) * 2) {
+        log_error("Hash (%s) of a trusted file %s is not a SHA256 hash", checksum_str, uri);
+        return -PAL_ERROR_INVAL;
+    }
+
     size_t uri_len = strlen(uri);
     if (uri_len >= URI_MAX) {
         log_error("Size of file exceeds maximum %dB: %s", URI_MAX, uri);
@@ -526,7 +531,7 @@ static int register_file(const char* uri, const char* checksum_str, bool check_d
         }
         new->size = attr.pending_size;
 
-        assert(strlen(checksum_str) >= sizeof(sgx_file_hash_t) * 2);
+        assert(strlen(checksum_str) == sizeof(sgx_file_hash_t) * 2);
         for (size_t i = 0; i < sizeof(sgx_file_hash_t); i++) {
             int8_t byte1 = hex2dec(checksum_str[i * 2]);
             int8_t byte2 = hex2dec(checksum_str[i * 2 + 1]);
@@ -598,17 +603,17 @@ static int init_trusted_files_from_toml_table(void) {
 
     toml_table_t* manifest_sgx = toml_table_in(g_pal_state.manifest_root, "sgx");
     if (!manifest_sgx)
-        return -PAL_ERROR_NOTDEFINED;
+        return 0;
 
     toml_table_t* toml_trusted_files = toml_table_in(manifest_sgx, "trusted_files");
     if (!toml_trusted_files)
-        return -PAL_ERROR_NOTDEFINED;
+        return 0;
 
     ssize_t toml_trusted_files_cnt = toml_table_nkval(toml_trusted_files);
     if (toml_trusted_files_cnt < 0)
         return -PAL_ERROR_DENIED;
     if (toml_trusted_files_cnt == 0)
-        return -PAL_ERROR_NOTDEFINED;
+        return 0;
 
     char* toml_trusted_file_str     = NULL;
     char* toml_trusted_checksum_key = NULL;
@@ -618,13 +623,20 @@ static int init_trusted_files_from_toml_table(void) {
         /* read sgx.trusted_file.<key> entry from manifest */
         const char* toml_trusted_file_key = toml_key_in(toml_trusted_files, i);
         assert(toml_trusted_file_key);
+
         toml_raw_t toml_trusted_file_raw = toml_raw_in(toml_trusted_files, toml_trusted_file_key);
-        assert(toml_trusted_file_raw);
+        if (!toml_trusted_file_raw) {
+            log_error("Invalid trusted file in manifest: '%s'", toml_trusted_file_key);
+            ret = -PAL_ERROR_INVAL;
+            goto out;
+        }
 
         ret = toml_rtos(toml_trusted_file_raw, &toml_trusted_file_str);
         if (ret < 0) {
-            log_error("Invalid trusted file in manifest: '%s'", toml_trusted_file_key);
-            continue;
+            log_error("Invalid trusted file in manifest: '%s' (not a string)",
+                      toml_trusted_file_key);
+            ret = -PAL_ERROR_INVAL;
+            goto out;
         }
 
         /* read corresponding sgx.trusted_checksum.<key> entry from manifest */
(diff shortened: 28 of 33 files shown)


woju (changes requested):
> Reviewed 22 of 29 files at r1.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 21 of 29 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @woju)

___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 792 at r1](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgpWsvp55mj7OPFyTvl:-MgpWsvtADLEcsBpLuin:bgm8m3u) ([raw file](https://github.com/oscarlab/graphene/blob/00a2b6bb6c4c345fec591bab7d3ca9a403890366/Pal/src/host/Linux-SGX/enclave_framework.c#L792)):*
> ```C
>         }
> 
>         toml_raw_t toml_trusted_hash_raw = toml_raw_in(toml_trusted_file, "hash");
> ```

Could we name this `"sha256"` instead of "hash"? 1) Explicit is better than implicit, 2) we may want to change the algorithm in the future (cf. SHA1 in git).
___
*[python/graphenelibos/sgx_sign.py, line 206 at r1](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgpY14p8uiK31XM1nBy:-MgpY14p8uiK31XM1nBz:b84s2tn) ([raw file](https://github.com/oscarlab/graphene/blob/00a2b6bb6c4c345fec591bab7d3ca9a403890366/python/graphenelibos/sgx_sign.py#L206)):*
> ```Python
>             if hash_ is None:
>                 hash_ = get_hash(target).hex()
>             targets[idx] = uri, target, hash_
> ```

Are you sure this works? You're modifying the collection, over which you're iterating (via `enumerate(targets)` above), and while this sometimes works (eg. with lists), sometimes this breaks (notably hash-based structures, like dicts and sets). If you're not sure, I'd suggest to duplicate either collection (like `list(enumerate(...))`).

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 21 of 29 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgqUhqr45sSByMcCB-8:-MgqUhr3ASSoH69zpYwr:b-cg3fxn) (no related file):*
Also, please fix documentation, at least `Documentation/manifest-syntax.rst` and `Pal/src/host/Linux-SGX/protected-files/README.rst`.

---

pwmarcz (changes requested):
> Reviewed 27 of 29 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv)

___
*[-- commits, line 28 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvNnDS4VfVZwbNdVnz:-MgvNnDS4VfVZwbNdVo-:bwel4h5) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/-- commits#L28)):*
Can this be a separate commit, before the main one?
___
*[common/src/toml.patch, line 26 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvOPgi-kHc_tfR3LLp:-MgvOPgi-kHc_tfR3LLq:b-80o59p) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/common/src/toml.patch#L26)):*
> ```Diff
>  			if (!arr) return -1;
>  		}
> -		if (arr->kind != 't')
> ```

What's happening here? Does that mean tomlc99 doesn't support mixed arrays in some cases, and you're patching it by removing the "array mismatch" error?

(Probably worth explaining in the commit message at least).
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 781 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvR2Mw9eHKnfKl0_am:-MgvR2Mw9eHKnfKl0_an:bql2b7) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/Pal/src/host/Linux-SGX/enclave_framework.c#L781)):*
> ```C
>     for (ssize_t i = 0; i < toml_trusted_files_cnt; i++) {
>         /* read `sgx.trusted_file = {uri = "file:foo", hash = "deadbeef"}` entry from manifest */
>         toml_table_t* toml_trusted_file = toml_table_at(toml_trusted_files, i);
> ```

What happens if the entry is not a TOML table (e.g. `trusted_files = [ "foo", "bar" ]`), do we get a NULL here? Can you handle it?
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 784 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvRNNLEMxZIjBG9BR_:-MgvRNNLEMxZIjBG9BRa:ba96pop) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/Pal/src/host/Linux-SGX/enclave_framework.c#L784)):*
> ```C
> 
>         toml_raw_t toml_trusted_uri_raw = toml_raw_in(toml_trusted_file, "uri");
>         assert(toml_trusted_uri_raw);
> ```

This can be null if there's no `uri` key, right? Is it a good idea to use asserts instead of checking the value and returning an error?

I assume you're doing this because the manifest has been produced and validated by our (Python?) tools. But if so, it would still be better to return an error instead of debug-mode `assert`...
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 518 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvSKVLAYCyCU0ft7f-:-MgvSKVLAYCyCU0ft7f0:b-lztvto) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/Pal/src/host/Linux-SGX/enclave_pf.c#L518)):*
> ```C
>         ret = register_protected_files_from_toml_array();
>         if (ret == -PAL_ERROR_NOTDEFINED) {
>             /* no allowed files were found, perfectly valid case */
> ```

Using an error code (`-PAL_ERROR_NOTDEFINED`) as a valid case looks a bit awkward. How about making these functions return the number of files parsed? Then you can call the other one if the number is 0.
___
*[python/graphenelibos/sgx_sign.py, line 167 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvSg402twWS5ImxvZT:-MgvSg402twWS5ImxvZU:bg4ue0o) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/python/graphenelibos/sgx_sign.py#L167)):*
> ```Python
> 
> def append_trusted_dir_or_file(targets, val, check_exist):
>     try:
> ```

This looks convoluted, the `except:` branch does no check that `val` is actually a string. I think it's much better to check the type directly:

```python
if isinstance(val, dict):
    uri = val['uri']
    hash = val['hash']
elif isinstance(val, str):
    uri = val
    hash = None
else:
    raise ValueError(f'unknown trusted file format: {val!r}')
```

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: all files reviewed, 9 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv and @woju)

___
*[python/graphenelibos/sgx_sign.py, line 206 at r1](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgpY14p8uiK31XM1nBy:-MgvUb9L9bnkEFFfnb8D:bxv6szq) ([raw file](https://github.com/oscarlab/graphene/blob/00a2b6bb6c4c345fec591bab7d3ca9a403890366/python/graphenelibos/sgx_sign.py#L206)):*

Are you sure this works? You're modifying the collection, over which you're iterating (via `enumerate(targets)` above), and while this sometimes works (eg. with lists), sometimes this breaks (notably hash-based structures, like dicts and sets). If you're not sure, I'd suggest to duplicate either collection (like `list(enumerate(...))`).
</blockquote></details>

If you want to modify it in place, you can use the index (also, modify only if it needs changing):

```python
for i in range(len(targets)):
     uri, target, hash_ = targets[i]
    if hash_ is None:
        hash_ = get_hash(target).hex()
        targets[i] = uri, target, hash_
```

If you don't mind copying the array, this looks more straightforward:

```python
hashed_targets = []
for (uri, target, hash_) in targets:
    if hash_ is None:
        hash_ = get_hash(target).hex()
    hashed_targets.append((uri, target, hash_))
```

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 23 of 31 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @pwmarcz and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgqUhqr45sSByMcCB-8:-MiMMRjd8EjjfDwqLniR:b-896fix) (no related file):*

Also, please fix documentation, at least `Documentation/manifest-syntax.rst` and `Pal/src/host/Linux-SGX/protected-files/README.rst`.
</blockquote></details>

Done.

---
___
*[-- commits, line 28 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvNnDS4VfVZwbNdVnz):*

Can this be a separate commit, before the main one?
</blockquote></details>

Done
___
*[common/src/toml.patch, line 26 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvOPgi-kHc_tfR3LLp:-MiLzZSS4__qTuo2qZ3o:b-kb5xtc) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/common/src/toml.patch#L26)):*

What's happening here? Does that mean tomlc99 doesn't support mixed arrays in some cases, and you're patching it by removing the "array mismatch" error?

(Probably worth explaining in the commit message at least).
</blockquote></details>

Done. I removed this hack. This was a leftover from my hacking, thanks for noticing.
___
*[common/src/toml.patch, line 91 at r3](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiMMcLA8RV1Wu5tSNPN:-MiMMcLA8RV1Wu5tSNPO:bk6byzk) ([raw file](https://github.com/oscarlab/graphene/blob/d4eb01e781a7bc078edb20d3d3df1f18836f0e20/common/src/toml.patch#L91)):*
> ```Diff
>  	*ret = strtoll(buf, &endp, base);
> -	return (errno || *endp) ? -1 : 0;
> +    return (*endp) ? -1 : 0;
> ```

Meh... Could we ignore this blooper? I don't want to recreate the patch file.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 792 at r1](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgpWsvp55mj7OPFyTvl:-MiMASXKB_UkW1ND-NDm:b-dgxz11) ([raw file](https://github.com/oscarlab/graphene/blob/00a2b6bb6c4c345fec591bab7d3ca9a403890366/Pal/src/host/Linux-SGX/enclave_framework.c#L792)):*

Could we name this `"sha256"` instead of "hash"? 1) Explicit is better than implicit, 2) we may want to change the algorithm in the future (cf. SHA1 in git).
</blockquote></details>

Done.

Note that I only changed it in this new TOML syntax and in this function: the old TOML syntax uses `sgx.trusted_checksum` (and is kept for legacy reasons) and the other C functions use `hash` words (because it is not important how we name it locally in our C code and because it would require a lot of unrelated changes).
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 781 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvR2Mw9eHKnfKl0_am:-MiM1H1P9tcC9KtC_XEy:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/Pal/src/host/Linux-SGX/enclave_framework.c#L781)):*

What happens if the entry is not a TOML table (e.g. `trusted_files = [ "foo", "bar" ]`), do we get a NULL here? Can you handle it?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 784 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvRNNLEMxZIjBG9BR_:-MiM1IBM7oNi-2YXtyrJ:bk36ero) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/Pal/src/host/Linux-SGX/enclave_framework.c#L784)):*

This can be null if there's no `uri` key, right? Is it a good idea to use asserts instead of checking the value and returning an error?

I assume you're doing this because the manifest has been produced and validated by our (Python?) tools. But if so, it would still be better to return an error instead of debug-mode `assert`...
</blockquote></details>

Done.

No, I did this because I was in vacation mode already that day...
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 518 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvSKVLAYCyCU0ft7f-:-MiM8xE89nLpJccHLMU0:b6vtfne) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/Pal/src/host/Linux-SGX/enclave_pf.c#L518)):*

Using an error code (`-PAL_ERROR_NOTDEFINED`) as a valid case looks a bit awkward. How about making these functions return the number of files parsed? Then you can call the other one if the number is 0.
</blockquote></details>

Done.

Actually did some more refactoring to make allowed / trusted / protected files logic similar and added more error checking.
___
*[python/graphenelibos/sgx_sign.py, line 206 at r1](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgpY14p8uiK31XM1nBy:-MiMCw-50lUu9ql6toF-:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/00a2b6bb6c4c345fec591bab7d3ca9a403890366/python/graphenelibos/sgx_sign.py#L206)):*

If you want to modify it in place, you can use the index (also, modify only if it needs changing):

```python
for i in range(len(targets)):
     uri, target, hash_ = targets[i]
    if hash_ is None:
        hash_ = get_hash(target).hex()
        targets[i] = uri, target, hash_
```

If you don't mind copying the array, this looks more straightforward:

```python
hashed_targets = []
for (uri, target, hash_) in targets:
    if hash_ is None:
        hash_ = get_hash(target).hex()
    hashed_targets.append((uri, target, hash_))
```
</blockquote></details>

Done.
___
*[python/graphenelibos/sgx_sign.py, line 167 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvSg402twWS5ImxvZT:-MiMBDf33vfmlhnaGBAJ:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/python/graphenelibos/sgx_sign.py#L167)):*

This looks convoluted, the `except:` branch does no check that `val` is actually a string. I think it's much better to check the type directly:

```python
if isinstance(val, dict):
    uri = val['uri']
    hash = val['hash']
elif isinstance(val, str):
    uri = val
    hash = None
else:
    raise ValueError(f'unknown trusted file format: {val!r}')
```
</blockquote></details>

Done.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 23 of 31 files reviewed, 10 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz and @woju)

___
*[python/graphenelibos/sgx_sign.py, line 206 at r1](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgpY14p8uiK31XM1nBy:-MiMPyDiDkQSWMao2g6d:b-yln9k0) ([raw file](https://github.com/oscarlab/graphene/blob/00a2b6bb6c4c345fec591bab7d3ca9a403890366/python/graphenelibos/sgx_sign.py#L206)):*

Done.
</blockquote></details>

Done. Used the approach with duplication (new collection). The first approach made Pylint unhappy:
```
C:204, 8: Consider using enumerate instead of iterating with range and len (consider-using-enumerate)
```

woju (commented):
> Reviewed 4 of 8 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 28 of 31 files reviewed, 7 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgqUhqr45sSByMcCB-8:-MiMehZ66JkLcE5Mo5Ko:b-ucfkn4) (no related file):*

Done.
</blockquote></details>

Thanks.

---
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 792 at r1](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgpWsvp55mj7OPFyTvl:-MiMectQ49B7wG0vWi-p:b-eignk6) ([raw file](https://github.com/oscarlab/graphene/blob/00a2b6bb6c4c345fec591bab7d3ca9a403890366/Pal/src/host/Linux-SGX/enclave_framework.c#L792)):*

Done.

Note that I only changed it in this new TOML syntax and in this function: the old TOML syntax uses `sgx.trusted_checksum` (and is kept for legacy reasons) and the other C functions use `hash` words (because it is not important how we name it locally in our C code and because it would require a lot of unrelated changes).
</blockquote></details>

Yes, certainly. Thanks.
___
*[python/graphenelibos/sgx_sign.py, line 206 at r1](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgpY14p8uiK31XM1nBy:-MgvfYgtBMi24Xw1lxj7:bakbg9l) ([raw file](https://github.com/oscarlab/graphene/blob/00a2b6bb6c4c345fec591bab7d3ca9a403890366/python/graphenelibos/sgx_sign.py#L206)):*

Done. Used the approach with duplication (new collection). The first approach made Pylint unhappy:
```
C:204, 8: Consider using enumerate instead of iterating with range and len (consider-using-enumerate)
```
</blockquote></details>

Seems bogus, if you'd like, add to ignored in `.pylintrc`. Resolving here.

mkow (changes requested):
> Reviewed 21 of 29 files at r1, 2 of 2 files at r2, 7 of 8 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: all files reviewed, 30 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[common/src/toml.patch, line 91 at r3](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiMMcLA8RV1Wu5tSNPN:-MiNniy11bO3KPqQ6Po_:brf7jod) ([raw file](https://github.com/oscarlab/graphene/blob/d4eb01e781a7bc078edb20d3d3df1f18836f0e20/common/src/toml.patch#L91)):*

Meh... Could we ignore this blooper? I don't want to recreate the patch file.
</blockquote></details>

Can't you edit this line manually in the .patch file? ;)
___
*[Documentation/manifest-syntax.rst, line 457 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNVwRL-1lwRkASsbI_:-MiNVwRL-1lwRkASsbIa:b-gp46at) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L457)):*
> ```reStructuredText
> ::
> 
>     sgx.allowed_files = [ "[URI]", "[URI]" ]
> ```

I'd use a nicer formatting here, as people will most likely just copy paste it, plus they'll have many more files. E.g.:

```TOML
sgx.allowed_files = [
  "[URI]",
  "[URI]",
]
```
___
*[Documentation/manifest-syntax.rst, line 459 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNVfSZ7Ne0Gso6Cmf7:-MiNVfS_CAxsZR82sox2:b-db55d6) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L459)):*
> ```reStructuredText
>     sgx.allowed_files = [ "[URI]", "[URI]" ]
>     # Or, deprecated TOML-table syntax:
>     # sgx.allowed_files.[identifier] = "[URI]"
> ```

I wouldn't list the old syntax at all and just purge it from any documentation + warn if used. And then stop supporting it at some point.
___
*[Documentation/manifest-syntax.rst, line 476 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNWD7j69-BzTdRerx4:-MiNWD7j69-BzTdRerx5:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L476)):*
> ```reStructuredText
> ::
> 
>     sgx.trusted_files = [ "[URI]", "[URI]" ]  # entries can be strings
> ```

ditto
___
*[Documentation/manifest-syntax.rst, line 478 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNWH-2BirBp1QQtXYW:-MiNWH-31IFLwFr2-WtO:b-12k6rl) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L478)):*
> ```reStructuredText
>     sgx.trusted_files = [ "[URI]", "[URI]" ]  # entries can be strings
> 
>     [[sgx.trusted_files]]                     # entries can be tables
> ```

`entries can be tables` -> `entries can also be tables`
___
*[Documentation/manifest-syntax.rst, line 491 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNWPUTAcLbXB4htsQ0:-MiNWPUTAcLbXB4htsQ1:bas8cv4) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L491)):*
> ```reStructuredText
> automatically generate hashes of these files and add them to the SGX-specific
> manifest (``.manifest.sgx``). The manifest writer may also specify the hash for
> a file using the TOML-table syntax, in the field ``sha256``.
> ```

Please mention explicitly that hashing this file will be skipped by `graphene-sgx-sign` tool and this value will be used instead.
___
*[Documentation/manifest-syntax.rst, line 504 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNWcbE2XMlssnppwJx:-MiNWcbE2XMlssnppwJy:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L504)):*
> ```reStructuredText
>     sgx.protected_files_key = "[16-byte hex value]"
> 
>     sgx.protected_files = [ "[URI]", "[URI]" ]
> ```

ditto
___
*[Documentation/manifest-syntax.rst, line 506 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNWdBS4Bc7zkJg3Buu:-MiNWdBS4Bc7zkJg3Buv:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L506)):*
> ```reStructuredText
>     sgx.protected_files = [ "[URI]", "[URI]" ]
>     # Or, deprecated TOML-table syntax:
>     # sgx.protected_files.[identifier] = "[URI]"
> ```

ditto
___
*[LibOS/shim/test/fs/manifest.template, line 46 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNX1U3De1AhxCw7hkj:-MiNX1U3De1AhxCw7hkk:bdqsb27) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/LibOS/shim/test/fs/manifest.template#L46)):*
> ```
> sgx.protected_files = [
>   "file:tmp/pf_input",
>   "file:tmp/pf_output"
> ```

I prefer putting the comma after the last element if it's listed one-per-line. The diff is then nicer if you add/reshuffle the lines, and you can also move them easier when experimenting with code (no need to fix the missing comma).
ditto for all other places
___
*[LibOS/shim/test/regression/rename.manifest.template, line 22 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNZ_Gu3Oo0Q-a1ERT_:-MiNZ_Gu3Oo0Q-a1ERTa:b-i252fe) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/LibOS/shim/test/regression/rename.manifest.template#L22)):*
> ```
> ]
> 
> sgx.allowed_files = [ "file:tmp/" ]
> ```

In other manifests you listed allowed files first, could you do the same here?
___
*[Pal/regression/manifest.template, line 17 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNZg7X1bvzol--6kRe:-MiNZg7X1bvzol--6kRf:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/regression/manifest.template#L17)):*
> ```
> ]
> 
> sgx.allowed_files = [
> ```

ditto
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 560 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNa7Tf4jt6IcNgH33t:-MiNa7Tg3e77u17q30Na:b-hqzcs7) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L560)):*
> ```C
>     int ret;
> 
>     if (checksum_str && strlen(checksum_str) < sizeof(sgx_file_hash_t) * 2) {
> ```

not `!=`?
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 752 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNfBD4A753rojUv9n5:-MiNfBD4A753rojUv9n6:b-doo955) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L752)):*
> ```C
>               "= [\"file1\", ..]'.");
> 
>     ret = (int)toml_trusted_files_cnt;
> ```

could we return this via argument? also, is it really used?
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 777 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNgpqJBaFGXl28EH-e:-MiNgpqJBaFGXl28EH-f:bm0wn7g) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L777)):*
> ```C
>         return 0;
> 
>     char* toml_trusted_uri_str  = NULL;
> ```

double space
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 824 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNiJHPDUkpcPn_Vbfc:-MiNiJHPDUkpcPn_Vbfd:bow0zw0) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L824)):*
> ```C
>         free(toml_trusted_uri_str);
>         free(toml_trusted_sha256_str);
>         toml_trusted_uri_str  = NULL;
> ```

doublespace
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 828 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNigkLDI--_KkTMbx5:-MiNigkM2Oc097ZnYZpw:b-rcypfz) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L828)):*
> ```C
>     }
> 
>     ret = (int)toml_trusted_files_cnt;
> ```

ditto (ret via arg)
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 884 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNjJ2k73Z7-GSzwJr6:-MiNjJ2l2BkxqBD4bhxd:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L884)):*
> ```C
>               "= [\"file1\", ..]'.");
> 
>     ret = (int)toml_allowed_files_cnt;
> ```

ditto
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 932 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNjn375Iur3N0RjKK8:-MiNjn375Iur3N0RjKK9:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L932)):*
> ```C
>     }
> 
>     ret = (int)toml_allowed_files_cnt;
> ```

ditto
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 953 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNk2ED9eGoI0553a1E:-MiNk2ED9eGoI0553a1F:bundr4g) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L953)):*
> ```C
>     }
> 
>     log_debug("Found %d trusted files in the manifest", ret);
> ```

is this actually useful for anything?
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 972 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNk4Uj4nfqeBuDMG3L:-MiNk4Uj4nfqeBuDMG3M:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L972)):*
> ```C
>     }
> 
>     log_debug("Found %d allowed files in the manifest", ret);
> ```

ditto
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 518 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvSKVLAYCyCU0ft7f-:-MiNkzRE28oZf1lAvmUt:b-ki6zcv) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/Pal/src/host/Linux-SGX/enclave_pf.c#L518)):*

Done.

Actually did some more refactoring to make allowed / trusted / protected files logic similar and added more error checking.
</blockquote></details>

OTOH: We try to not mix values and error codes in return values and I think we got rid of a lot of such places recently. But I agree that a magic error code is hacky, so maybe just return by an arg?
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 480 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNkWAI6d_Wam9btfvt:-MiNkWAI6d_Wam9btfvu:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_pf.c#L480)):*
> ```C
>               "= [\"file1\", ..]'.");
> 
>     ret = (int)toml_pfs_cnt;
> ```

ditto
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 533 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNlWep-jOo08injHHn:-MiNlWep-jOo08injHHo:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_pf.c#L533)):*
> ```C
>     }
> 
>     ret = (int)toml_pfs_cnt;
> ```

ditto
___
*[python/graphenelibos/sgx_sign.py, line 170 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNmLFd7aeEgDbuW67_:-MiNmLFd7aeEgDbuW67a:b-wp3x5g) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/python/graphenelibos/sgx_sign.py#L170)):*
> ```Python
>         # trusted file is specified as TOML table `{uri = "file:foo", sha256 = "deadbeef"}`
>         uri_ = val['uri']
>         hash_ = val['sha256']
> ```

Won't this throw if only the URI was provided? Is this a valid case actually, or maybe we don't support it? If not, why?
___
*[python/graphenelibos/sgx_sign.py, line 178 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNmEQm5G5_VXD3L1Fk:-MiNmEQm5G5_VXD3L1Fl:b-ewhk8h) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/python/graphenelibos/sgx_sign.py#L178)):*
> ```Python
>         raise ValueError(f'Unknown trusted file format: {val!r}')
> 
>     path = Path(resolve_uri(uri_, check_exist))
> ```

Shouldn't we skip checking the existence if the hash was manually provided? Also, could you test this in LibOS regression? (that we don't require this file to be present when `graphene-sgx-sign` runs)

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 3 of 35 files reviewed, 30 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @mkow, @pwmarcz, and @woju)

___
*[common/src/toml.patch, line 91 at r3](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiMMcLA8RV1Wu5tSNPN:-MiPyGYt8ledq_TM0CUC:b-rcysse) ([raw file](https://github.com/oscarlab/graphene/blob/d4eb01e781a7bc078edb20d3d3df1f18836f0e20/common/src/toml.patch#L91)):*

Can't you edit this line manually in the .patch file? ;)
</blockquote></details>

Done. Technically, the hash should have changed but who cares.
___
*[Documentation/manifest-syntax.rst, line 457 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNVwRL-1lwRkASsbI_:-MiPydPiAv9O43n92Cyr:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L457)):*

I'd use a nicer formatting here, as people will most likely just copy paste it, plus they'll have many more files. E.g.:

```TOML
sgx.allowed_files = [
  "[URI]",
  "[URI]",
]
```
</blockquote></details>

Done.
___
*[Documentation/manifest-syntax.rst, line 459 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNVfSZ7Ne0Gso6Cmf7:-MiPydp18tKreAUqdTVs:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L459)):*

I wouldn't list the old syntax at all and just purge it from any documentation + warn if used. And then stop supporting it at some point.
</blockquote></details>

Done.
___
*[Documentation/manifest-syntax.rst, line 476 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNWD7j69-BzTdRerx4:-MiQ-E7kEEQxyl-F2gOG:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L476)):*

ditto
</blockquote></details>

Done.
___
*[Documentation/manifest-syntax.rst, line 478 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNWH-2BirBp1QQtXYW:-MiQ-EXi6vi96IUN3g9G:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L478)):*

`entries can be tables` -> `entries can also be tables`
</blockquote></details>

Done.
___
*[Documentation/manifest-syntax.rst, line 491 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNWPUTAcLbXB4htsQ0:-MiQ-Uaw0eofs_RK38-O:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L491)):*

Please mention explicitly that hashing this file will be skipped by `graphene-sgx-sign` tool and this value will be used instead.
</blockquote></details>

Done.
___
*[Documentation/manifest-syntax.rst, line 504 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNWcbE2XMlssnppwJx:-MiQ-a0x1QUkhP3K7Cut:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L504)):*

ditto
</blockquote></details>

Done.
___
*[Documentation/manifest-syntax.rst, line 506 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNWdBS4Bc7zkJg3Buu:-MiQ-aKC5zDm-GrQrTO5:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L506)):*

ditto
</blockquote></details>

Done.
___
*[LibOS/shim/test/fs/manifest.template, line 46 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNX1U3De1AhxCw7hkj:-MiQ-osy94VUHHVRmswy:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/LibOS/shim/test/fs/manifest.template#L46)):*

I prefer putting the comma after the last element if it's listed one-per-line. The diff is then nicer if you add/reshuffle the lines, and you can also move them easier when experimenting with code (no need to fix the missing comma).
ditto for all other places
</blockquote></details>

Done.
___
*[LibOS/shim/test/regression/rename.manifest.template, line 22 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNZ_Gu3Oo0Q-a1ERT_:-MiQ0dhVAeHklEfU30Ny:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/LibOS/shim/test/regression/rename.manifest.template#L22)):*

In other manifests you listed allowed files first, could you do the same here?
</blockquote></details>

Done.
___
*[Pal/regression/manifest.template, line 17 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNZg7X1bvzol--6kRe:-MiQ0ej54s_SNRsuWmQt:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/regression/manifest.template#L17)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 560 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNa7Tf4jt6IcNgH33t:-MiQ1N8s6uKPYCJZKcdm:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L560)):*

not `!=`?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 752 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNfBD4A753rojUv9n5:-MiQ2aA0444dCiF6W4N1:bvnddf6) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L752)):*

could we return this via argument? also, is it really used?
</blockquote></details>

Done. Actually, there is no sense to use either TOML-table syntax or TOML-array syntax. I now use both, and thus the return value is irrelevant.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 777 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNgpqJBaFGXl28EH-e:-MiQ2pIY3E6bCWFulYtL:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L777)):*

double space
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 824 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNiJHPDUkpcPn_Vbfc:-MiQ2wfwA3wKk0nOBZLA:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L824)):*

doublespace
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 828 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNigkLDI--_KkTMbx5:-MiQ2xiF0FVJCaK2ebFA:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L828)):*

ditto (ret via arg)
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 884 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNjJ2k73Z7-GSzwJr6:-MiQ2ykk6yst2dvNhv0x:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L884)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 932 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNjn375Iur3N0RjKK8:-MiQ2zdC0JATRa1VJTIF:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L932)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 953 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNk2ED9eGoI0553a1E:-MiQ3-OQApo4ktBAMVc2:bf2yiqf) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L953)):*

is this actually useful for anything?
</blockquote></details>

Done. No, it was not, removed.
___
*[Pal/src/host/Linux-SGX/enclave_framework.c, line 972 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNk4Uj4nfqeBuDMG3L:-MiQ30zS8aGZ9jNJUbDS:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_framework.c#L972)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 518 at r2](https://reviewable.io/reviews/oscarlab/graphene/2644#-MgvSKVLAYCyCU0ft7f-:-MiQ3OyA9AUa1_OgnZmb:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/12f7b3d5e8ee057c725b7c4e5df388ace2374c0d/Pal/src/host/Linux-SGX/enclave_pf.c#L518)):*

OTOH: We try to not mix values and error codes in return values and I think we got rid of a lot of such places recently. But I agree that a magic error code is hacky, so maybe just return by an arg?
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 480 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNkWAI6d_Wam9btfvt:-MiQ38o75AwGgf12kA9W:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_pf.c#L480)):*

ditto
</blockquote></details>

Done.
___
*[Pal/src/host/Linux-SGX/enclave_pf.c, line 533 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNlWep-jOo08injHHn:-MiQ39X49XSfkZW3Elsw:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Pal/src/host/Linux-SGX/enclave_pf.c#L533)):*

ditto
</blockquote></details>

Done.
___
*[python/graphenelibos/sgx_sign.py, line 170 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNmLFd7aeEgDbuW67_:-MiQ5jTo6txwns4jiFfS:b-kn0koz) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/python/graphenelibos/sgx_sign.py#L170)):*

Won't this throw if only the URI was provided? Is this a valid case actually, or maybe we don't support it? If not, why?
</blockquote></details>

Done. Good catch...
___
*[python/graphenelibos/sgx_sign.py, line 178 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNmEQm5G5_VXD3L1Fk:-MiQEshb1ZquMsHflXVc:b-f0fnyc) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/python/graphenelibos/sgx_sign.py#L178)):*

Shouldn't we skip checking the existence if the hash was manually provided? Also, could you test this in LibOS regression? (that we don't require this file to be present when `graphene-sgx-sign` runs)
</blockquote></details>

Done. Good catch. Changed the tests logic to create `trusted_testfile` on the fly, before running the corresponding tests.

mkow (approved):
> Reviewed 32 of 32 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: all files reviewed, 6 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[common/src/toml.patch, line 91 at r3](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiMMcLA8RV1Wu5tSNPN:-MiQV9fk1BrVTPI-rLT-:b3o30rf) ([raw file](https://github.com/oscarlab/graphene/blob/d4eb01e781a7bc078edb20d3d3df1f18836f0e20/common/src/toml.patch#L91)):*

Done. Technically, the hash should have changed but who cares.
</blockquote></details>

Oh, true :) Whatever.

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: all files reviewed, 7 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @dimakuv and @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 459 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNVfSZ7Ne0Gso6Cmf7:-MiQw0c_0BacOcv5yW_c:b-obrjwq) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L459)):*

Done.
</blockquote></details>

I disagree, we shoudn't just remove the documentation of any supported schema. If you don't want it here, please add "Deprecated schema" (or sth like that) section somewhere near the end.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: all files reviewed, 7 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz and @woju)

___
*[Documentation/manifest-syntax.rst, line 459 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNVfSZ7Ne0Gso6Cmf7:-MiQyFZX9ay5yjMmKAjB:bybc33m) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L459)):*

I disagree, we shoudn't just remove the documentation of any supported schema. If you don't want it here, please add "Deprecated schema" (or sth like that) section somewhere near the end.
</blockquote></details>

What's the point of keeping it? It only makes the docs longer and I don't see any case where someone should actually read about it.
It will cease to be supported soon.

pwmarcz (approved):
> Reviewed 1 of 8 files at r3, 32 of 32 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[common/src/toml.patch, line 91 at r3](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiMMcLA8RV1Wu5tSNPN:-MiN91qO-CCWw4U0pHht:bcuikc0) ([raw file](https://github.com/oscarlab/graphene/blob/d4eb01e781a7bc078edb20d3d3df1f18836f0e20/common/src/toml.patch#L91)):*

Oh, true :) Whatever.
</blockquote></details>

+1 for not caring too much about the hashes, editing the patches manually is too convenient (Emacs even adjusts the line numbers and counts if necessary :) ).
___
*[Documentation/manifest-syntax.rst, line 459 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNVfSZ7Ne0Gso6Cmf7:-MiR52XS22vRNLNt6Brc:b-khwp1g) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L459)):*

What's the point of keeping it? It only makes the docs longer and I don't see any case where someone should actually read about it.
It will cease to be supported soon.
</blockquote></details>

Release notes?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[Documentation/manifest-syntax.rst, line 459 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNVfSZ7Ne0Gso6Cmf7:-MiRHMuZ3wpLCM3jWJ45:balhmtn) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L459)):*

Release notes?
</blockquote></details>

@pwmarcz: You mean adding this only to release notes? This is something which will definitely go there, I think this part is out of question :) (we always have "breaking changes" section)

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 459 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNVfSZ7Ne0Gso6Cmf7:-MiRRS5I4-VOSgFFPXH5:b8v5dmq) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L459)):*

@pwmarcz: You mean adding this only to release notes? This is something which will definitely go there, I think this part is out of question :) (we always have "breaking changes" section)
</blockquote></details>

This is orthogonal. Yes, this obviously should be mentioned in release notes, but I think as long we support this, we should have this documented. We can remove this when we'll stop supporting it.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[Documentation/manifest-syntax.rst, line 459 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNVfSZ7Ne0Gso6Cmf7:-MiRV3YTA2_LS1DKwq9U:b-qj7hdt) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L459)):*

This is orthogonal. Yes, this obviously should be mentioned in release notes, but I think as long we support this, we should have this documented. We can remove this when we'll stop supporting it.
</blockquote></details>

IMO it's just taking space in the docs and no one should ever need to use it anymore. So, I vote for removing it.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 34 of 35 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow, @pwmarcz, and @woju)

___
*[Documentation/manifest-syntax.rst, line 459 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNVfSZ7Ne0Gso6Cmf7:-MiRwYNj7EQ3oeyuL0-3:b-et8wlb) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L459)):*

IMO it's just taking space in the docs and no one should ever need to use it anymore. So, I vote for removing it.
</blockquote></details>

Done. Added the `Deprecated syntax` section.

mkow (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 34 of 35 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv, @mkow, @pwmarcz, and @woju)

___
*[Documentation/manifest-syntax.rst, line 459 at r4](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiNVfSZ7Ne0Gso6Cmf7:-MiRwxcGAYNAph712YWK:b-jy2fsg) ([raw file](https://github.com/oscarlab/graphene/blob/6dc8090271c920742c63494db3c5c5a80074c4fd/Documentation/manifest-syntax.rst#L459)):*

Done. Added the `Deprecated syntax` section.
</blockquote></details>

For the record: we did a small vote on this and an additional section won.
___
*[Documentation/manifest-syntax.rst, line 717 at r6](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiRy5cx6mAFypvgeJ-u:-MiRy5cx6mAFypvgeJ-v:bhvelh2) ([raw file](https://github.com/oscarlab/graphene/blob/f05ab35c1101477e1fe5512ae82291734df25e7e/Documentation/manifest-syntax.rst#L717)):*
> ```reStructuredText
> listed below, with short descriptions.
> 
> Executable (removed)
> ```

What's the point of including all of these? They don't work anymore and no one is using them.

dimakuv (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 34 of 35 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow, @pwmarcz, and @woju)

___
*[Documentation/manifest-syntax.rst, line 717 at r6](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiRy5cx6mAFypvgeJ-u:-MiS-amI8AIFtJgkBvMw:b4fahnn) ([raw file](https://github.com/oscarlab/graphene/blob/f05ab35c1101477e1fe5512ae82291734df25e7e/Documentation/manifest-syntax.rst#L717)):*

What's the point of including all of these? They don't work anymore and no one is using them.
</blockquote></details>

Before I fixup: do you want me to remove all the `(removed)` manifest options?

What about `Program Break Size (renamed)`? Remove or keep?

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 34 of 35 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @mkow, @pwmarcz, and @woju)

___
*[Documentation/manifest-syntax.rst, line 717 at r6](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiRy5cx6mAFypvgeJ-u:-MiS0uGq7pey8fc4u1XJ:b-df5ofp) ([raw file](https://github.com/oscarlab/graphene/blob/f05ab35c1101477e1fe5512ae82291734df25e7e/Documentation/manifest-syntax.rst#L717)):*

Before I fixup: do you want me to remove all the `(removed)` manifest options?

What about `Program Break Size (renamed)`? Remove or keep?
</blockquote></details>

I vote for removing everything which doesn't currently work (and I think that was @woju's original proposal, to document only things that still work).

woju (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 34 of 35 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel) (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 698 at r6](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiS1msk1PfOoWHae57T:-MiS1msl2REZFY1yif0y:b-938pd1) ([raw file](https://github.com/oscarlab/graphene/blob/f05ab35c1101477e1fe5512ae82291734df25e7e/Documentation/manifest-syntax.rst#L698)):*
> ```reStructuredText
> 
> 
> Deprecated syntax
> ```

This is "schema", not syntax. Syntax is TOML, we define schema inside.
___
*[Documentation/manifest-syntax.rst, line 717 at r6](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiRy5cx6mAFypvgeJ-u:-MiS1XTZ1fp84P_9dyvg:b-r4n7h8) ([raw file](https://github.com/oscarlab/graphene/blob/f05ab35c1101477e1fe5512ae82291734df25e7e/Documentation/manifest-syntax.rst#L717)):*

I vote for removing everything which doesn't currently work (and I think that was @woju's original proposal, to document only things that still work).
</blockquote></details>

Right, we shouldn't document things that don't actually work. That's really for release notes and git log only.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 34 of 35 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 702 at r7](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiS59FF6-VxsRKJQgx7:-MiS59FF6-VxsRKJQgx8:b-mcclgu) ([raw file](https://github.com/oscarlab/graphene/blob/ea71d4f4d168377c302b5f6eca91e5978cb5d5a6/Documentation/manifest-syntax.rst#L702)):*
> ```reStructuredText
> 
> Before October 2020, manifests were not written in the TOML syntax but instead
> in an ad-hoc INI file-like syntax. Some of the differences include:
> ```

What about this section? It also lists things which are long-gone, I see no point in documenting them.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: 34 of 35 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: Intel), "fixup! " found in commit messages' one-liners (waiting on @dimakuv, @mkow, and @pwmarcz)

___
*[Documentation/manifest-syntax.rst, line 717 at r6](https://reviewable.io/reviews/oscarlab/graphene/2644#-MiRy5cx6mAFypvgeJ-u:-MiSArLb3872HdFEdC_R:bm7ycao) ([raw file](https://github.com/oscarlab/graphene/blob/f05ab35c1101477e1fe5512ae82291734df25e7e/Documentation/manifest-syntax.rst#L717)):*

Right, we shouldn't document things that don't actually work. That's really for release notes and git log only.
</blockquote></details>

+1, let's just document what works (even if it's about to be removed soon).

mkow (approved):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

woju (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

mkow (approved):
> Reviewed 6 of 6 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: :shipit: complete! all files reviewed, all discussions resolved

woju (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2644)* status: :shipit: complete! all files reviewed, all discussions resolved

dimakuv:
Jenkins, retest Jenkins-Debug-20.04 please (`test_libos.TC_50_GDB.test_000_gdb_backtrace` failed with `AssertionError: timeout (10 s) expired`)

PR #2657, opened 2021-08-19
Prepare for UBSan and Clang

pwmarcz:
This is in preparation to enable UBSan (and Clang at the same time, because Clang's checks are much stronger).

## Description of the changes 

See commit messages.

Note that these are just bugfixes in existing code. I will submit UBSan integration (and Jenkins configuration) in next PR.

## How to test this PR? 

The current tests should ensure there are no regressions. I also ran this code on a [draft branch](https://github.com/oscarlab/graphene/pull/2653) and it passes UBSan.

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)

### DIFF
--- a/LICENSE.addendum.txt
+++ b/LICENSE.addendum.txt
@@ -20,9 +20,7 @@ A number of files taken from other C libraries:
  ** Pal/include/elf/elf.h
  ** Pal/include/sysdeps/generic/ldsodefs.h
  ** common/src/network/hton.c
- ** common/src/string/atoi.c
  ** common/src/string/strchr.c
- ** common/src/string/strlen.c
  ** Pal/src/dynamic_link.h, do-rel.h, dl-machine-x86_64.h
  ** Pal/include/arch/x86_64/Linux/ - whole directory - for signal API definitions
  ** We also build a patched glibc
--- a/LibOS/shim/src/bookkeep/shim_signal.c
+++ b/LibOS/shim/src/bookkeep/shim_signal.c
@@ -422,7 +422,7 @@ bool is_user_string_readable(const char* addr) {
         return true;
     }
 
-    const char* next_page_addr = ALIGN_UP_PTR(addr + 1, PAGE_SIZE);
+    const char* next_page_addr = (const char*)ALIGN_UP((uintptr_t)addr + 1, PAGE_SIZE);
     assert(next_page_addr != addr);
 
     /* `next_page_addr` could wrap around which by itself might not be an error, let `access_ok`
--- a/LibOS/shim/src/fs/shim_fs_mem.c
+++ b/LibOS/shim/src/fs/shim_fs_mem.c
@@ -7,8 +7,11 @@
 #include "shim_fs.h"
 #include "shim_fs_mem.h"
 
-#define OVERFLOWS_FILE_OFF_T(n) __builtin_add_overflow_p((n), 0, (file_off_t)0)
-#define OVERFLOWS_SIZE_T(n) __builtin_add_overflow_p((n), 0, (size_t)0)
+#define OVERFLOWS(type, val)                        \
+    ({                                              \
+        type __dummy;                               \
+        __builtin_add_overflow((val), 0, &__dummy); \
+    })
 
 static int mem_file_resize(struct shim_mem_file* mem, size_t buf_size) {
     char* buf = malloc(buf_size);
@@ -23,7 +26,7 @@ static int mem_file_resize(struct shim_mem_file* mem, size_t buf_size) {
 }
 
 void mem_file_init(struct shim_mem_file* mem, char* data, size_t size) {
-    assert(!OVERFLOWS_FILE_OFF_T(size));
+    assert(!OVERFLOWS(file_off_t, size));
 
     mem->buf = data;
     mem->buf_size = size;
@@ -55,7 +58,7 @@ ssize_t mem_file_write(struct shim_mem_file* mem, file_off_t pos_start, const vo
     if (__builtin_add_overflow(pos_start, size, &pos_end))
         return -EFBIG;
 
-    if (OVERFLOWS_SIZE_T(pos_end))
+    if (OVERFLOWS(size_t, pos_end))
         return -EFBIG;
 
     if (size > 0) {
@@ -81,7 +84,7 @@ ssize_t mem_file_write(struct shim_mem_file* mem, file_off_t pos_start, const vo
 int mem_file_truncate(struct shim_mem_file* mem, file_off_t size) {
     assert(size >= 0);
 
-    if (OVERFLOWS_SIZE_T(size))
+    if (OVERFLOWS(size_t, size))
         return -EFBIG;
 
     int ret = mem_file_resize(mem, size);
--- a/LibOS/shim/src/ipc/shim_ipc_fs_lock.c
+++ b/LibOS/shim/src/ipc/shim_ipc_fs_lock.c
@@ -27,7 +27,11 @@ int ipc_posix_lock_set(const char* path, struct posix_lock* pl, bool wait) {
     struct shim_ipc_msg* msg = __alloca(total_msg_size);
     init_ipc_msg(msg, IPC_MSG_POSIX_LOCK_SET, total_msg_size);
     memcpy(msg->data, &msgin, sizeof(msgin));
-    memcpy(((struct shim_ipc_posix_lock*)&msg->data)->path, path, path_len + 1);
+
+    /* Copy path after message (`msg->data` is unaligned, so we have to compute the offset
+     * manually) */
+    char* path_ptr = (char*)&msg->data + offsetof(struct shim_ipc_posix_lock, path);
+    memcpy(path_ptr, path, path_len + 1);
 
     void* data;
     int ret = ipc_send_msg_and_get_response(g_process_ipc_ids.leader_vmid, msg, &data);
@@ -63,7 +67,11 @@ int ipc_posix_lock_get(const char* path, struct posix_lock* pl, struct posix_loc
     struct shim_ipc_msg* msg = __alloca(total_msg_size);
     init_ipc_msg(msg, IPC_MSG_POSIX_LOCK_GET, total_msg_size);
     memcpy(msg->data, &msgin, sizeof(msgin));
-    memcpy(((struct shim_ipc_posix_lock*)&msg->data)->path, path, path_len + 1);
+
+    /* Copy path after message (`msg->data` is unaligned, so we have to compute the offset
+     * manually) */
+    char* path_ptr = (char*)&msg->data + offsetof(struct shim_ipc_posix_lock, path);
+    memcpy(path_ptr, path, path_len + 1);
 
     void* data;
     int ret = ipc_send_msg_and_get_response(g_process_ipc_ids.leader_vmid, msg, &data);
--- a/LibOS/shim/test/regression/Makefile
+++ b/LibOS/shim/test/regression/Makefile
@@ -135,7 +135,6 @@ CFLAGS-debug = -g3
 CFLAGS-debug_regs-x86_64 = -g3
 CFLAGS-exec_same = -pthread
 CFLAGS-shared_object = -fPIC -pie
-CFLAGS-openmp = -fopenmp
 CFLAGS-multi_pthread = -pthread
 CFLAGS-exit_group = -pthread
 CFLAGS-abort_multithread = -pthread
@@ -156,6 +155,12 @@ LDLIBS-attestation += ../../../../common/src/crypto/mbedtls/install/lib/libmbedc
 
 CFLAGS-fp_multithread += -pthread -fno-builtin  # see comment in the test's source
 LDLIBS-fp_multithread += -lm
+
+ifeq ($(findstring clang,$(CC)),clang)
+CFLAGS-openmp = -fopenmp=libgomp
+else
+CFLAGS-openmp = -fopenmp
+endif
 
 proc_common: proc_common.o dump.o
 	$(call cmd,cmulti)
--- a/Pal/include/do-rel.h
+++ b/Pal/include/do-rel.h
@@ -25,6 +25,9 @@ static void __attribute_unused elf_dynamic_do_rel(struct link_map* l, ElfW(Addr)
     if (!l->l_info[DT_SYMTAB])
         return;
 
+    if (relsize == 0)
+        return;
+
     ElfW(Rel)* r = (void*)reladdr, *end = (void*)(reladdr + relsize);
     ElfW(Sym)* symtab    = (void*)D_PTR(l->l_info[DT_SYMTAB]);
     ElfW(Word) nrelative = l->l_info[RELCOUNT_IDX] ? l->l_info[RELCOUNT_IDX]->d_un.d_val : 0;
--- a/Pal/src/host/Linux-SGX/protected-files/protected_files.c
+++ b/Pal/src/host/Linux-SGX/protected-files/protected_files.c
@@ -1006,11 +1006,14 @@ static file_node_t* ipf_append_data_node(pf_context_t* pf) {
         return NULL;
     }
 
+    uint64_t node_number, physical_node_number;
+    get_node_numbers(pf->offset, NULL, &node_number, NULL, &physical_node_number);
+
     new_file_data_node->type = FILE_DATA_NODE_TYPE;
     new_file_data_node->new_node = true;
     new_file_data_node->parent = file_mht_node;
-    get_node_numbers(pf->offset, NULL, &new_file_data_node->node_number, NULL,
-                     &new_file_data_node->physical_node_number);
+    new_file_data_node->node_number = node_number;
+    new_file_data_node->physical_node_number = physical_node_number;
 
     if (!lruc_add(pf->cache, new_file_data_node->physical_node_number, new_file_data_node)) {
         free(new_file_data_node);
--- a/Pal/src/host/Linux/db_memory.c
+++ b/Pal/src/host/Linux/db_memory.c
@@ -162,17 +162,22 @@ unsigned long _DkMemoryAvailableQuota(void) {
 
 /* Expects `line` to be in the same format as "/proc/self/maps" entries, i.e. starting with
  * "hexadecimalnumber-hexadecimalnumber", e.g. "1fe3-87cc ...". */
-static void parse_line(const char* line, uintptr_t* start_ptr, uintptr_t* end_ptr) {
-    char* next = NULL;
-    *start_ptr = (uintptr_t)strtol(line, &next, 16);
-    assert(next && next[0] == '-');
-    *end_ptr = (uintptr_t)strtol(next + 1, NULL, 16);
+static int parse_line(const char* line, uintptr_t* start_ptr, uintptr_t* end_ptr) {
+    const char* next = NULL;
+    unsigned long val;
 
-#define TEST(x) assert((uintptr_t)(x) == (uintptr_t)strtol(#x, NULL, 16))
-    /* If this assert fails that probably means `strtol` implementation has changed and the two
-     * above need to be changed to `strtoul` (which we do not implement at the moment). */
-    TEST(0xffff000011112222ul); // arbitrary number which is negative when cast to long
-#undef TEST
+    if (str_to_ulong(line, 16, &val, &next) < 0)
+        return -PAL_ERROR_INVAL;
+    *start_ptr = val;
+
+    assert(next);
+    if (next[0] != '-')
+        return -PAL_ERROR_INVAL;
+
+    if (str_to_ulong(next + 1, 16, &val, &next) < 0)
+        return -PAL_ERROR_INVAL;
+    *end_ptr = val;
+    return 0;
 }
 
 /* This function is very fragile w.r.t. "/proc/self/maps" file format. */
@@ -212,9 +217,13 @@ int get_vdso_and_vvar_ranges(uintptr_t* vdso_start, uintptr_t* vdso_end, uintptr
         *line_end = '\0';
 
         if (!memcmp(vdso_str, line_end - vdso_str_len, vdso_str_len)) {
-            parse_line(buf, vdso_start, vdso_end);
+            ret = parse_line(buf, vdso_start, vdso_end);
+            if (ret < 0)
+                goto out;
         } else if (!memcmp(vvar_str, line_end - vvar_str_len, vvar_str_len)) {
-            parse_line(buf, vvar_start, vvar_end);
+            ret = parse_line(buf, vvar_start, vvar_end);
+            if (ret < 0)
+                goto out;
         }
 
         size_t new_size = 0;
--- a/common/src/string/atoi.c
+++ b/common/src/string/atoi.c
@@ -1,23 +1,6 @@
-/* Copyright (C) 1991, 1997 Free Software Foundation, Inc.
-   This file is part of the GNU C Library.
-
-   The GNU C Library is free software; you can redistribute it and/or
-   modify it under the terms of the GNU Lesser General Public
-   License as published by the Free Software Foundation; either
-   version 2.1 of the License, or (at your option) any later version.
-
-   The GNU C Library is distributed in the hope that it will be useful,
-   but WITHOUT ANY WARRANTY; without even the implied warranty of
-   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-   Lesser General Public License for more details.
-
-   You should have received a copy of the GNU Lesser General Public
-   License along with the GNU C Library; if not, write to the Free
-   Software Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA
-   02111-1307 USA. */
-
 /* Copyright (C) 2021 Intel Corporation
  *                    Vijay Dhanraj <vijay.dhanraj@intel.com>
+ *                    Paweł Marczewski <pawel@invisiblethingslab.com>
  */
 
 #include <limits.h>
@@ -25,9 +8,9 @@
 
 #include "api.h"
 
-long strtol(const char* s, char** endptr, int base) {
-    int neg  = 0;
-    long val = 0;
+static void begin_number(const char* str, int base, const char** out_s, int* out_base,
+                         int* out_sign) {
+    const char* s = str;
 
     // gobble initial whitespace
     while (*s == ' ' || *s == '\t') {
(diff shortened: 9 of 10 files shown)


pwmarcz:
Jenkins, retest Jenkins-20.04, please (timeout in LTP, seems unrelated)

pwmarcz (commented):
>Jenkins, retest Jenkins-20.04, please (timeout in LTP, seems unrelated)

Weird, looks like that didn't trigger anything...

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657#-:-Mhnvm-24egzCOv_e6Ax:brxo5uu)* status: 0 of 9 files reviewed, all discussions resolved, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[common/src/string/strlen.c, line 10 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhTsnc97DFIjpmDg9T-:-MhTsnc97DFIjpmDg9T0:b-ea59dx) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/common/src/string/strlen.c#L10)):*
> ```C
> /* Find the length of S, but scan at most MAXLEN characters.  If no
>    '\0' terminator is found in that many characters, return MAXLEN.  */
> size_t strnlen(const char* str, size_t maxlen) {
> ```

This is going to be slower

mkow (changes requested):
> Weird, looks like that didn't trigger anything...

Heh, you put a comma before "please" which our primitive regexes didn't foresee :)

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657#-:-MhpA-U4DD0g6qA23iIN:beq1ij7)* status: 0 of 9 files reviewed, 2 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[common/src/string/strlen.c, line 3 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhTthQI9RrwK1Lu9yB8:-MhTthQI9RrwK1Lu9yB9:b23jk3r) ([raw file](https://github.com/oscarlab/graphene/blob/398727edcee3f8e29c050d2cfdcc7a79d7387e5d/common/src/string/strlen.c#L3)):*
> ```C
> /* Find the length of STRING, but scan at most MAXLEN characters.
>    Copyright (C) 1991,1993,1997,2000,2001,2005 Free Software Foundation, Inc.
>    Contributed by Jakub Jelinek <jakub@redhat.com>.
> ```

Please update LICENSE.addendum.txt
___
*[LibOS/shim/src/fs/shim_fs_mem.c, line 12 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhTtUThBBAAxWGU7L03:-MhTtUThBBAAxWGU7L04:b-kpnmmf) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/LibOS/shim/src/fs/shim_fs_mem.c#L12)):*
> ```C
> #define OVERFLOWS(type, val)                        \
>     ({                                              \
>         file_off_t __dummy;                         \
> ```

`file_off_t` -> `type`

mkow:
Jenkins, retest Jenkins-20.04 please (this version of the trigger should work :) )

mkow (changes requested):
> Reviewed 9 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: all files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[common/src/string/atoi.c, line 128 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpMp06B4iLVLqoXuCI:-MhpMp06B4iLVLqoXuCJ:beiytxt) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/common/src/string/atoi.c#L128)):*
> ```C
>     }
> 
>     if (s == str)
> ```

unrelated, but why is this check here, but not in `strtol()`?
___
*[common/src/string/atoi.c, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpNdafEkIBqgyCk6uV:-MhpNdafEkIBqgyCk6uW:b-e3zr7q) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/common/src/string/atoi.c#L165)):*
> ```C
> 
>         __builtin_mul_overflow(value, 10, &value);
>         __builtin_add_overflow(value, digit * sign, &value);
> ```

Why not just do the arithmetic normally? These functions are assuming the input is correct.

Also, I'm not sure how the out value is defined on overflow in these macros? The documentation doesn't specify this :/
___
*[common/src/string/strlen.c, line 10 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhTsnc97DFIjpmDg9T-:-MhpF1sWAmX9T-ZUXBYm:b6giatn) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/common/src/string/strlen.c#L10)):*

This is going to be slower 
</blockquote></details>

This is used only internally and usually for short strings, so it shouldn't really matter IMO.
___
*[Pal/include/do-rel.h, line 28 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpIhsDEAjIQhfp2JWX:-MhpIhsEAPEmtx6cYtWm:bvlduo2) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/Pal/include/do-rel.h#L28)):*
> ```objc
>         return;
> 
>     if (!reladdr)
> ```

Actually, can't we have a valid relocation at zero?
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 1016 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpKh3n7KNjgC2hpqMP:-MhpKh3o507jw7s1H7WD:b-64zirp) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L1016)):*
> ```C
>     new_file_data_node->parent = file_mht_node;
>     new_file_data_node->node_number = node_number;
>     new_file_data_node->physical_node_number = physical_node_number;
> ```

I don't understand this change. If the old code had an unaligned access, then the new one also has it? I.e. if `&new_file_data_node->node_number` is unaligned, then you can't assign to it (or do anything else with it).

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: 6 of 10 files reviewed, 6 unresolved discussions, not enough approvals from maintainers (2 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @mkow)

___
*[common/src/string/atoi.c, line 128 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpMp06B4iLVLqoXuCI:-Mhr1mtNE34rAF7ESf_U:b-fsn638) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/common/src/string/atoi.c#L128)):*
Because `strtol` emulates the standard behavior, which is:

>If  there  were  no digits at all, strtol() stores the original value of nptr in *endptr (and returns 0).

And `str_to_ulong` is our slightly nicer version that doesn't accept "empty" numbers.
___
*[common/src/string/atoi.c, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpNdafEkIBqgyCk6uV:-Mhr2YzLEbbOm3uwyjrD:bjupx8h) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/common/src/string/atoi.c#L165)):*
> Why not just do the arithmetic normally? These functions are assuming the input is correct.

Good point. And if the input is not correct, it's better to trigger UBSan here. Changed to normal arithmetic.

>Also, I'm not sure how the out value is defined on overflow in these macros? The documentation doesn't specify this :/

It's defined as performing the operation at infinite precision and then downcasting. I'm guessing for normal 2s complement arithmetic it should be equivalent to 2s complement signed overflow.
___
*[common/src/string/strlen.c, line 3 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhTthQI9RrwK1Lu9yB8:-Mhr4GWx3WF35Ks2CFfJ:bpby6tf) ([raw file](https://github.com/oscarlab/graphene/blob/398727edcee3f8e29c050d2cfdcc7a79d7387e5d/common/src/string/strlen.c#L3)):*

Please update LICENSE.addendum.txt
</blockquote></details>

`atoi.c` also looks like it has no glibc code anymore. I removed both from `LICENSE.addendum.txt` and removed the license header from `atoi.c`.
___
*[LibOS/shim/src/fs/shim_fs_mem.c, line 12 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhTtUThBBAAxWGU7L03:-Mhr4nHP0xu9aAoq60Jj:bt2vr1b) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/LibOS/shim/src/fs/shim_fs_mem.c#L12)):*

`file_off_t` -> `type`
</blockquote></details>

Fixed.
___
*[Pal/include/do-rel.h, line 28 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpIhsDEAjIQhfp2JWX:-Mhr4ybZEel4sF0vcKrF:b-klqqb) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/Pal/include/do-rel.h#L28)):*

Actually, can't we have a valid relocation at zero?
</blockquote></details>

I don't think it can (this is an array of relocation records, not relocation targets), but I don't know enough to rule it out.

I changed the condition to `relsize == 0`, that case should be a no-op anyway (because `r == relative == end`, so the loops do nothing).
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 1016 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpKh3n7KNjgC2hpqMP:-MhrAF6d4DAvDwd8JfSL:b-7jpkfr) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L1016)):*

I don't understand this change. If the old code had an unaligned access, then the new one also has it? I.e. if `&new_file_data_node->node_number` is unaligned, then you can't assign to it (or do anything else with it).
</blockquote></details>

I believe unaligned access to packed struct members (not taking a pointer) is safe. Unfortunately I can't find any guarantees in the documentation, but here's what I found:

* Packed structures are not part of the standard, so the behavior is compiler-specific anyway.
* GCC and Clang's warnings (`-Waddress-for-packed-member`) complain about taking a pointer, but there's no warning about direct access.
  * See also [the discussion about adding this warning](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51628), where it seems to be accepted that:
    >When a program accesses a misaligned member of a packed struct, the compiler generates whatever code is necessary to read or write the correct value.
* UBSan also complains about unaligned pointers, but not unaligned member access.
* GCC and Clang are aware the access is unaligned in such cases, and generate the right code on platforms where it's needed (I checked ARM and RISC-V on Godbolt)
  * It looks like *specifically for ARM on GCC>=9.3* the compiler generates unaligned access (and you can override that with `-mno-unaligned-access`).That's because apparently [unaligned access on ARM is permitted, used to be slower, but isn't slower anymore](https://patchwork.ozlabs.org/project/uboot/patch/4FF16CF2.9090808@kmckk.co.jp/).

Tell me if you have any evidence to the contrary.

But anyway, the alternative is to rewrite the whole protected-files subsystem, isn't it?

mkow (dismissed):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[common/src/string/atoi.c, line 128 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpMp06B4iLVLqoXuCI:-MhrP1Dm5Gt5gUDadWUQ:b8339qs) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/common/src/string/atoi.c#L128)):*

Because `strtol` emulates the standard behavior, which is:

>If  there  were  no digits at all, strtol() stores the original value of nptr in *endptr (and returns 0).

And `str_to_ulong` is our slightly nicer version that doesn't accept "empty" numbers.
</blockquote></details>

Ah, right, I missed that this one is our custom ;)
___
*[common/src/string/atoi.c, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpNdafEkIBqgyCk6uV:-MhrPBj_6ltEDyWELhoh:b-gv9cel) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/common/src/string/atoi.c#L165)):*

> Why not just do the arithmetic normally? These functions are assuming the input is correct.

Good point. And if the input is not correct, it's better to trigger UBSan here. Changed to normal arithmetic.

>Also, I'm not sure how the out value is defined on overflow in these macros? The documentation doesn't specify this :/

It's defined as performing the operation at infinite precision and then downcasting. I'm guessing for normal 2s complement arithmetic it should be equivalent to 2s complement signed overflow.
</blockquote></details>

Yeah, but downcasting is not defined for signed overflow case and 2s complement is not guaranteed?
___
*[Pal/include/do-rel.h, line 28 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpIhsDEAjIQhfp2JWX:-MhrPafV4bKMrFGRT-d5:b-ezjh3d) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/Pal/include/do-rel.h#L28)):*

I don't think it can (this is an array of relocation records, not relocation targets), but I don't know enough to rule it out.

I changed the condition to `relsize == 0`, that case should be a no-op anyway (because `r == relative == end`, so the loops do nothing).
</blockquote></details>

That's better :)
___
*[Pal/src/host/Linux-SGX/protected-files/protected_files.c, line 1016 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpKh3n7KNjgC2hpqMP:-MhrTEn5EFUSYGJt2Z7v:b-lwhpfm) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/Pal/src/host/Linux-SGX/protected-files/protected_files.c#L1016)):*

I believe unaligned access to packed struct members (not taking a pointer) is safe. Unfortunately I can't find any guarantees in the documentation, but here's what I found:

* Packed structures are not part of the standard, so the behavior is compiler-specific anyway.
* GCC and Clang's warnings (`-Waddress-for-packed-member`) complain about taking a pointer, but there's no warning about direct access.
  * See also [the discussion about adding this warning](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51628), where it seems to be accepted that:
    >When a program accesses a misaligned member of a packed struct, the compiler generates whatever code is necessary to read or write the correct value.
* UBSan also complains about unaligned pointers, but not unaligned member access.
* GCC and Clang are aware the access is unaligned in such cases, and generate the right code on platforms where it's needed (I checked ARM and RISC-V on Godbolt)
  * It looks like *specifically for ARM on GCC>=9.3* the compiler generates unaligned access (and you can override that with `-mno-unaligned-access`).That's because apparently [unaligned access on ARM is permitted, used to be slower, but isn't slower anymore](https://patchwork.ozlabs.org/project/uboot/patch/4FF16CF2.9090808@kmckk.co.jp/).

Tell me if you have any evidence to the contrary.

But anyway, the alternative is to rewrite the whole protected-files subsystem, isn't it?
</blockquote></details>

That's quite an interesting problem, I didn't consider that this may be somehow specified for non-pointer accesses, but you may be right, as the compiler knows in all such cases that the access is unaligned and it can emit proper instructions for it. The unsolvable case is only with pointers, which can be passed to functions which are allowed to assume that they are aligned.

I also checked some more RISCy archs and GCC splits the writes/reads into single-byte accesses there, so it seems to be aware of this problem and handles it properly.

So, it seems to be legal, although not explicitly allowed in the documentation.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[common/src/string/atoi.c, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpNdafEkIBqgyCk6uV:-Mhrd4jXCOQksBND95q9:b-kwb80b) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/common/src/string/atoi.c#L165)):*

Yeah, but downcasting is not defined for signed overflow case and 2s complement is not guaranteed?
</blockquote></details>

Oh. So probably there's no guarantees about the result, but at least the compiler doesn't treat it as UB.

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: all files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[common/src/string/atoi.c, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpNdafEkIBqgyCk6uV:-MhsQxwM4XTgFB78ur88:b-4sjsxw) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/common/src/string/atoi.c#L165)):*
Which contradicts their own statement, hehe :)

> these built-in functions have fully defined behavior for all argument values.

boryspoplawski (changes requested):
> Reviewed 6 of 9 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[common/src/string/strlen.c, line 8 at r2](https://reviewable.io/reviews/oscarlab/graphene/2657#-MiCEW356gsWkMFFBSRe:-MiCEW361_C4Y-Bzc8vZ:bfwmgot) ([raw file](https://github.com/oscarlab/graphene/blob/e012c1184127d9b025bf634cc8b4bde85c63cec8/common/src/string/strlen.c#L8)):*
> ```C
> #include "api.h"
> 
> /* Find the length of S, but scan at most MAXLEN characters.  If no
> ```

Please change to our code guidelines and use actual arg names (`str`, `maxlen`).
___
*[LibOS/shim/src/ipc/shim_ipc_fs_lock.c, line 27 at r2](https://reviewable.io/reviews/oscarlab/graphene/2657#-MiCtlBf5-gXNwAYQRRU:-MiCtlBgAssw9uPt7sJo:brj9n12) ([raw file](https://github.com/oscarlab/graphene/blob/e012c1184127d9b025bf634cc8b4bde85c63cec8/LibOS/shim/src/ipc/shim_ipc_fs_lock.c#L27)):*
> ```C
>     size_t path_len = strlen(path);
>     size_t total_msg_size = get_ipc_msg_size(sizeof(msgin) + path_len + 1);
>     struct shim_ipc_msg* msg = __alloca(total_msg_size);
> ```

Not blocking cause it's unrelated, but we should use `malloc` here.
___
*[LibOS/shim/src/ipc/shim_ipc_fs_lock.c, line 67 at r2](https://reviewable.io/reviews/oscarlab/graphene/2657#-MiCtqK44up-ssQ2itKs:-MiCtqK44up-ssQ2itKt:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/e012c1184127d9b025bf634cc8b4bde85c63cec8/LibOS/shim/src/ipc/shim_ipc_fs_lock.c#L67)):*
> ```C
>     size_t path_len = strlen(path);
>     size_t total_msg_size = get_ipc_msg_size(sizeof(msgin) + path_len + 1);
>     struct shim_ipc_msg* msg = __alloca(total_msg_size);
> ```

ditto
___
*[Pal/src/host/Linux/db_memory.c, line 222 at r2](https://reviewable.io/reviews/oscarlab/graphene/2657#-MiCwDcl8d_xLSHuHdxt:-MiCwDcl8d_xLSHuHdxu:b-ypft4m) ([raw file](https://github.com/oscarlab/graphene/blob/e012c1184127d9b025bf634cc8b4bde85c63cec8/Pal/src/host/Linux/db_memory.c#L222)):*
> ```C
>             ret = parse_line(buf, vdso_start, vdso_end);
>             if (ret < 0)
>                 goto out;
> ```

`unix_to_pal_error` missing

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: all files reviewed, 2 unresolved discussions, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[common/src/string/atoi.c, line 165 at r1](https://reviewable.io/reviews/oscarlab/graphene/2657#-MhpNdafEkIBqgyCk6uV:-MiLCNPw7DcbOtFgsRBg:b-2mvtln) ([raw file](https://github.com/oscarlab/graphene/blob/d5b585794ea7b351254d90d7436d7bbd91ba0587/common/src/string/atoi.c#L165)):*

Which contradicts their own statement, hehe :)

> these built-in functions have fully defined behavior for all argument values.
</blockquote></details>

I read "fully defined behavior" as something like "we won't tell you what it is, but we won't assume it never happens and optimize something out". And it's *probably* deterministic...
___
*[common/src/string/strlen.c, line 8 at r2](https://reviewable.io/reviews/oscarlab/graphene/2657#-MiCEW356gsWkMFFBSRe:-MiLDBpPE89mn5-vTgbx:b-uj6w85) ([raw file](https://github.com/oscarlab/graphene/blob/e012c1184127d9b025bf634cc8b4bde85c63cec8/common/src/string/strlen.c#L8)):*

Please change to our code guidelines and use actual arg names (`str`, `maxlen`).
</blockquote></details>

Actually I'd rather just remove this comment. This is just standard behavior, and we don't include documentation other standard functions (`strlen`, `strcmp` etc.)
___
*[LibOS/shim/src/ipc/shim_ipc_fs_lock.c, line 27 at r2](https://reviewable.io/reviews/oscarlab/graphene/2657#-MiCtlBf5-gXNwAYQRRU:-MiLDvf154JI4c64fVLg:b-aq3v0j) ([raw file](https://github.com/oscarlab/graphene/blob/e012c1184127d9b025bf634cc8b4bde85c63cec8/LibOS/shim/src/ipc/shim_ipc_fs_lock.c#L27)):*

Not blocking cause it's unrelated, but we should use `malloc` here.
</blockquote></details>

Acknowledged, but I don't want to fix it now, especially since all of the IPC code uses alloca for the messages.
___
*[Pal/src/host/Linux/db_memory.c, line 222 at r2](https://reviewable.io/reviews/oscarlab/graphene/2657#-MiCwDcl8d_xLSHuHdxt:-MiLFavRBBdjVnfTBVpD:b-b41tke) ([raw file](https://github.com/oscarlab/graphene/blob/e012c1184127d9b025bf634cc8b4bde85c63cec8/Pal/src/host/Linux/db_memory.c#L222)):*

`unix_to_pal_error` missing
</blockquote></details>

Oh, I see we use PAL errors in this code, including helper functions (`read_proc_meminfo`). I changed `parse_line` to also return PAL errors.

dimakuv (changes requested):
> Reviewed 4 of 9 files at r1, 4 of 4 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @pwmarcz)

___
*[LibOS/shim/test/regression/Makefile, line 160 at r3](https://reviewable.io/reviews/oscarlab/graphene/2657#-MiLkXUCAGmKzNJqQXwP:-MiLkXUCAGmKzNJqQXwQ:bilyyv0) ([raw file](https://github.com/oscarlab/graphene/blob/5f2e82df75fab557614370271c6ab931618e82f5/LibOS/shim/test/regression/Makefile#L160)):*
> ```Makefile
> 
> ifeq ($(findstring clang,$(CC)),clang)
> CFLAGS-openmp = -fopenmp=libgomp
> ```

In the commit message you write: `Make sure we still use our patched libgomp.` What does it mean? What does Clang try to use by default?

boryspoplawski (dismissed):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[LibOS/shim/src/ipc/shim_ipc_fs_lock.c, line 27 at r2](https://reviewable.io/reviews/oscarlab/graphene/2657#-MiCtlBf5-gXNwAYQRRU:-MiLlTvY7w4z8jUM7F7z:b-u278kk) ([raw file](https://github.com/oscarlab/graphene/blob/e012c1184127d9b025bf634cc8b4bde85c63cec8/LibOS/shim/src/ipc/shim_ipc_fs_lock.c#L27)):*

Acknowledged, but I don't want to fix it now, especially since all of the IPC code uses alloca for the messages.
</blockquote></details>

That's not true, most of the IPC code uses `malloc` when sending, only IPC worker thread callbacks use `alloca`.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @dimakuv)

___
*[LibOS/shim/src/ipc/shim_ipc_fs_lock.c, line 27 at r2](https://reviewable.io/reviews/oscarlab/graphene/2657#-MiCtlBf5-gXNwAYQRRU:-MiLovxi3cJgxjqJ_Us3:b-3mg8g0) ([raw file](https://github.com/oscarlab/graphene/blob/e012c1184127d9b025bf634cc8b4bde85c63cec8/LibOS/shim/src/ipc/shim_ipc_fs_lock.c#L27)):*

That's not true, most of the IPC code uses `malloc` when sending, only IPC worker thread callbacks use `alloca`.
</blockquote></details>

Oh, I see. I grepped for `total_msg_size` but there's also code that uses `msg_size`.
___
*[LibOS/shim/test/regression/Makefile, line 160 at r3](https://reviewable.io/reviews/oscarlab/graphene/2657#-MiLkXUCAGmKzNJqQXwP:-MiLp96L3x4kLtj8D1MJ:b991wvq) ([raw file](https://github.com/oscarlab/graphene/blob/5f2e82df75fab557614370271c6ab931618e82f5/LibOS/shim/test/regression/Makefile#L160)):*

In the commit message you write: `Make sure we still use our patched libgomp.` What does it mean? What does Clang try to use by default?
</blockquote></details>

It uses `libomp.so` which comes from LLVM.

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: all files reviewed, all discussions resolved, "fixup! " found in commit messages' one-liners

boryspoplawski (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

mkow (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2657)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #2106, opened 2021-01-21
Integration with sanitizers


mkow:
## Description of the problem

It would be good to enable some sanitizers in CI (e.g. AddressSanitizer from Clang) to allow for better and more reliable detection of low-level bugs.

The tricky part is that we're running without stdlib and manage memory on our own (especially on SGX), so most likely we'll need some hacks to get ASan working.

Blocked by #1794 (Clang support).

Previous work: @yamahata and @stefanberger ran GCC's UBSan (see #724, #871, #2089, #2094, #2097, #2101, #2102, #2103). It's much less powerful than ASan, but still something! I think we can try to add it to CI as a first step, to have at least something.

mkow:
Yeah, this definitely will require substantial efforts, but also give us quite huge gains.

PR #2101, opened 2021-01-19
[Pal,LibOS] Enable UBSAN via 'make UBSAN=1'

stefanberger:
## Description of the changes 

This PR enables a build with UBSAN enable by using 'make USBAN=1' as shown below. Some test cases that do not work under UBSAN are deactivated when USBAN is used. The regression tests on the PAL and LibOS level are passing now. LTP, however, has several failing tests.

## How to test this PR? 

```
make clean
make USBAN=1 DEBUG=1 -j32
pushd Pal/regression
make USBAN=1 DEBUG=1 regression -j32
popd
pushd LibOS/shim/test/regression
make USBAN=1 DEBUG=1 regression -j32
popd
```

---
This change is [Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2101)

PR #2662, opened 2021-08-24
UBSan support

pwmarcz:
Based on #2657 which needs to be merged first.

Tracking issue: #2106.

## Description of the changes 

This enables UndefinedBehaviorSanitizer in more or less all of our code (Common, LibOS, PAL, outer PAL). See `ubsan.c` for the main part of implementation; everything else is glue.

I wanted to have a regression test that checks if UBSan still works. This requires triggering some code inside LibOS. The initial idea was to define our own syscall, but it looks pretty hacky (inventing our own syscall number, making sure it doesn't collide with anything, making sure the syscall table is the right size, etc.)

Instead, I think we can use the same mechanism as we use for `register_library`: make this a more general LibOS entry vector. I think the same mechanism can be used for other unit tests that we want to run inside Graphene (e.g. testing filesystem or IPC code).

### Why implement our own hooks?

The previous attempt (#2101) worked by importing the real `libubsan` library. However, that pulls in glibc and other libraries, and causes LibOS to be linked against these, which is probably not what we want and might cause weird side effects (since e.g. we have our own memory allocation). Also, this approach won't work anymore because we decided to remove dynamic linking support from LibOS.

So instead I'm implementing our own hooks. They're quite simple: display a message, along with source position (fortunately the compiler provides file name and line number). To see more information (e.g. stack trace), you have to use gdb (you can break on `shim_abort` / `pal_abort`).

(There's an even cheaper option, `-fsanitize-trap`, that emits trap instructions and doesn't require any hooks. But as you can see, the hooks can be rather simple).

## How to test this PR? 

UBSan is not enabled on Jenkins yet, but you can compile the code yourself (see `building.rst`) and play with it.


pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: 0 of 26 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[.ci/linux-direct-sanitizers.jenkinsfile, line 1 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMQ1YED-q0G76jWj5Y:-MiMQ1YFADAHqXe9L864:b-qntv2o) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/.ci/linux-direct-sanitizers.jenkinsfile#L1)):*
> ```
> node('nonsgx_slave') {
> ```

@woju Can we have these 2 configurations enabled in Jenkins? I remember we talked about this together with @mkow.

They use Clang, Ubuntu 20.04, DEBUG, and additionally have UBSan enabled (and possibly will have other sanitizers, such as ASan, enabled in the future).

What's important is we wil *finally* have SGX in debug configuration, as well as compilation using Clang - both of these break occasionally because they're not in CI yet.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: 0 of 26 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[.ci/linux-direct-sanitizers.jenkinsfile, line 1 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMQ1YED-q0G76jWj5Y:-MiMSahj2Pj2iT2JiUzJ:bvx89fk) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/.ci/linux-direct-sanitizers.jenkinsfile#L1)):*

@woju Can we have these 2 configurations enabled in Jenkins? I remember we talked about this together with @mkow.

They use Clang, Ubuntu 20.04, DEBUG, and additionally have UBSan enabled (and possibly will have other sanitizers, such as ASan, enabled in the future).

What's important is we wil *finally* have SGX in debug configuration, as well as compilation using Clang - both of these break occasionally because they're not in CI yet.
</blockquote></details>

@mkow I called these `linux-direct-sanitizers` and `linux-sgx-sanitizers`, are you OK with these names or can you suggest other ones? (@woju wants to have a confirmation)

mkow (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: 0 of 26 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL)

___
*[.ci/linux-direct-sanitizers.jenkinsfile, line 1 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMQ1YED-q0G76jWj5Y:-MiMTNtc09nGAv12uLmJ:brxp9sd) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/.ci/linux-direct-sanitizers.jenkinsfile#L1)):*

@mkow I called these `linux-direct-sanitizers` and `linux-sgx-sanitizers`, are you OK with these names or can you suggest other ones? (@woju wants to have a confirmation)
</blockquote></details>

Sounds good!

woju (changes requested):
> Reviewed 13 of 26 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: 13 of 26 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @pwmarcz)

___
*[meson.build, line 36 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMYlzoEJAHy88v2ElQ:-MiMYlzoEJAHy88v2ElR:b62qd52) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/meson.build#L36)):*
> ```Meson
> skeleton = get_option('skeleton') == 'enabled'
> 
> ubsan = get_option('ubsan') == 'enabled'
> ```

Meson has builtin option `b_sanitize`: https://mesonbuild.com/Builtin-options.html#base-options, why didn't you use that one?

If you don't want the base option, please at least check if it wasn't set and `error()` out if it was.
___
*[.ci/linux-direct-sanitizers.jenkinsfile, line 1 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMQ1YED-q0G76jWj5Y:-MiMclWLAf_7wjpte5n8:bfsdfww) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/.ci/linux-direct-sanitizers.jenkinsfile#L1)):*

Sounds good!
</blockquote></details>

Please ping me when you'd want this pipeline added to Jenkins.
___
*[Documentation/building.rst, line 179 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMblmR9AnCG5jnAemX:-MiMblmR9AnCG5jnAemY:bv5ytqb) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/Documentation/building.rst#L179)):*
> ```reStructuredText
>    Debug builds (:command:`DEBUG=1`, :command:`DEBUGOPT=1`) and UBSan builds
>    (:command:`UBSAN=1`) are not suitable for production.
> ```

Instead this collective warning, could you write that each time in respective option? The warning would be closer to whatever user reads, and also there's less risk that new not-for-prod options would not be added here.
___
*[Scripts/regression.py, line 67 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMaiVK60McQ3NiXASW:-MiMaiVK60McQ3NiXASX:b4kk6vb) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/Scripts/regression.py#L67)):*
> ```Python
>         return self.run_binary(args, prefix=prefix, env=env, **kwds)
> 
>     def run_binary(self, args, *, timeout=None, prefix=None, expect_fail=False, **kwds):
> ```

There's already `expect_returncode` context manager, isn't that sufficient?

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: 11 of 26 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[meson.build, line 36 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMYlzoEJAHy88v2ElQ:-MiMwa0N7u7LOBjSC09i:b-5zjy7n) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/meson.build#L36)):*

Meson has builtin option `b_sanitize`: https://mesonbuild.com/Builtin-options.html#base-options, why didn't you use that one?

If you don't want the base option, please at least check if it wasn't set and `error()` out if it was.
</blockquote></details>

I think that whatever sanitizer integration we build is going to be rather custom, so I'd prefer not to use Meson-level flags that possibly mess with our configuration.

I added an `error()` here.
___
*[Documentation/building.rst, line 179 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMblmR9AnCG5jnAemX:-MiMxIPv5VlrKDguZS8a:b-f3hta7) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/Documentation/building.rst#L179)):*

Instead this collective warning, could you write that each time in respective option? The warning would be closer to whatever user reads, and also there's less risk that new not-for-prod options would not be added here.
</blockquote></details>

Okay. Done.
___
*[Scripts/regression.py, line 67 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMaiVK60McQ3NiXASW:-MiMxS4UB5E6qhCeyPMy:bignodm) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/Scripts/regression.py#L67)):*

There's already `expect_returncode` context manager, isn't that sufficient?
</blockquote></details>

`expect_returncode` loses the output, and I want to check exactly what happens.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: 10 of 26 files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[Scripts/regression.py, line 67 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMaiVK60McQ3NiXASW:-MiN8daX0hu63EGBPe6D:b-hp3s1t) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/Scripts/regression.py#L67)):*

`expect_returncode` loses the output, and I want to check exactly what happens.
</blockquote></details>

As discussed offline: I can just use `e.stderr` on exception. Change reverted.

woju (changes requested):
> Reviewed 1 of 26 files at r1, 1 of 4 files at r3, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: 14 of 26 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz and @woju)

___
*[Pal/src/meson.build, line 4 at r4](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiN9fHL9Rpt7qIU1ag9:-MiN9fHL9Rpt7qIU1agA:b-i38jkp) ([raw file](https://github.com/oscarlab/graphene/blob/b6e5d6c8977580d9cc424355ce38e91688e97005/Pal/src/meson.build#L4)):*
> ```Meson
>     cflags_custom_stack_protector,
>     '-DIN_PAL',
> ] + cflags_sanitizers
> ```

```
cflags_pal_common = [
    cflags_custom_stack_protector,
    cflags_sanitizers,
    '-DIN_PAL',
]
```
(lists gets flattened)
___
*[Scripts/regression.py, line 67 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMaiVK60McQ3NiXASW:-MiN7L71EqICfXAGbi3E:b2nhpa7) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/Scripts/regression.py#L67)):*

As discussed offline: I can just use `e.stderr` on exception. Change reverted.
</blockquote></details>

Thx

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: 13 of 26 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

___
*[Pal/src/meson.build, line 4 at r4](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiN9fHL9Rpt7qIU1ag9:-MiNAzTo0qmEhm4C2Zkf:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/b6e5d6c8977580d9cc424355ce38e91688e97005/Pal/src/meson.build#L4)):*

```
cflags_pal_common = [
    cflags_custom_stack_protector,
    cflags_sanitizers,
    '-DIN_PAL',
]
```
(lists gets flattened)
</blockquote></details>

Done.

woju (commented):
> Reviewed 1 of 26 files at r1, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: 17 of 26 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @woju)

boryspoplawski (changes requested):
> Reviewed 19 of 26 files at r1, 1 of 2 files at r2, 2 of 4 files at r3, 1 of 2 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, 19 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMYrMa3k-GzrLdIDG7:-MiMYrMa3k-GzrLdIDG8:bzica8f) (no related file):*
Out of curiosity: have you tried using UBSan with gcc?

---
___
*[meson_options.txt, line 10 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMWQpw8MhupyF5PGa0:-MiMWQpw8MhupyF5PGa1:bn75skn) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/meson_options.txt#L10)):*
> ```Meson
> 
> option('ubsan', type: 'combo', choices: ['disabled', 'enabled'],
>     description: 'Enable undefined behavior sanitizer (supported in LibOS)')
> ```

Not in all of the Graphene (e.g. PAL)?
___
*[.ci/ubuntu18.04.dockerfile, line 10 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMXc6cBnWJ5F6BJ2RQ:-MiMXc6cBnWJ5F6BJ2RR:bkd7r0r) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/.ci/ubuntu18.04.dockerfile#L10)):*
> ```Dockerfile
>     bison \
>     build-essential \
>     clang \
> ```

We seem not to be using clang on any 18.04 pipelines, so maybe do not add it here (for faster CI run times)? Just an idea, not blocking.
___
*[common/src/ubsan.c, line 21 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMZCBdCk1iETybGo4P:-MiMZCBdCk1iETybGo4Q:b-xjj0gu) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/common/src/ubsan.c#L21)):*
> ```C
>  */
> 
> #ifdef UBSAN
> ```

Wouldn't it be better to if-y the makefile/meson? I.e. build this file and link it if `UBSAN` is set, instead of unconditionally compile + link, but sometimes make the source empty?
___
*[common/src/ubsan.c, line 31 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMZhRb3X3bCAll5wuP:-MiMZhRb3X3bCAll5wuQ:b1whbip) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/common/src/ubsan.c#L31)):*
> ```C
> /* Type definitions (adapted from libubsan) */
> 
> struct TypeDescriptor;
> ```

Please use Graphene coding style (ditto for the whole file).
___
*[common/src/ubsan.c, line 31 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNeO6s7a1eR7K1yb-g:-MiNeO6t0vP7fvGFxZ3e:bjlbphg) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L31)):*
> ```C
> /* Type definitions (adapted from libubsan) */
> 
> struct TypeDescriptor;
> ```

Is there really no public API and we have to manually define these?
___
*[common/src/ubsan.c, line 42 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNiIRdD8gjQoxEq91X:-MiNiIReDByXf8AlAvyM:bonxd3p) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L42)):*
> ```C
>     struct SourceLocation Loc;
>     const struct TypeDescriptor* Type;
>     unsigned char LogAlignment;
> ```

https://wiki.osdev.org/Undefined_Behavior_Sanitization shows that this is `uintptr_t`. Which one is bugged?
Update: ok, it has `alignment` (without log), probably different version. How can we be sure that these structs match what's expected by UBSan?
___
*[common/src/ubsan.c, line 43 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNiPO4DH5dNU74359_:-MiNiPO4DH5dNU74359a:b-koeana) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L43)):*
> ```C
>     const struct TypeDescriptor* Type;
>     unsigned char LogAlignment;
>     unsigned char TypeCheckKind;
> ```

Wouldn't `uint8_t` be better?
___
*[common/src/ubsan.c, line 63 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNkQRv7TnDIi5SY7wQ:-MiNkQRv7TnDIi5SY7wR:bjhuj37) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L63)):*
> ```C
> #define HANDLER_ABORT(name) __ubsan_handle_##name##_abort
> 
> #define __UBSAN_SIMPLE_HANDLER(name, fmt, params, ...) \
> ```

K&R params?
Where did you get this macro from?
___
*[common/src/ubsan.c, line 66 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNihN9Drw4jc4BWbqT:-MiNihN9Drw4jc4BWbqU:b-40xiaw) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L66)):*
> ```C
>     void HANDLER(name) params;                         \
>     void HANDLER(name) params {                        \
>         log_error("ubsan: " fmt, ##__VA_ARGS__); \
> ```

Misaligned `\`.
___
*[common/src/ubsan.c, line 103 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNl8i69pNXnwvA_jZv:-MiNl8i69pNXnwvA_jZw:b228yug) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L103)):*
> ```C
>                        "pointer overflow: applying offset to 0x%lx produced 0x%lx")
> UBSAN_SIMPLE_HANDLER_1(load_invalid_value,
>                        "load of invalid value: %ld")
> ```

What does this mean?
___
*[common/src/ubsan.c, line 107 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNlAhQBy1AhAetCByO:-MiNlAhQBy1AhAetCByP:bblov99) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L107)):*
> ```C
>                        "__builtin_unreachable")
> UBSAN_SIMPLE_HANDLER_2(shift_out_of_bounds,
>                        "shift out of bounds: %ld, %ld")
> ```

Would be nice to know which value is shifted by which.
___
*[common/src/ubsan.c, line 139 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNjrXJ0apZbOC9f-KZ:-MiNjrXJ0apZbOC9f-K_:b-yzkngz) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L139)):*
> ```C
>     if (!Pointer) {
>         log_error("ubsan: null pointer dereference");
>     } else if (Pointer & ((1 << Data->LogAlignment) - 1)) {
> ```

`1ul`, otherwise if `LogAlignment` is `>= 32` this is undefined.
Update: ok, this can never be that big, but still there is no reason not to change it.
___
*[common/src/ubsan.c, line 141 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNk3Em-cviYd_v3I_F:-MiNk3Em-cviYd_v3I_G:b-7ropln) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L141)):*
> ```C
>     } else if (Pointer & ((1 << Data->LogAlignment) - 1)) {
>         log_error("ubsan: misaligned address %p for type with alignment %u", (void*)Pointer,
>                   1 << Data->LogAlignment);
> ```

ditto
___
*[Documentation/building.rst, line 198 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNnGe196VF3FiYP38c:-MiNnGe196VF3FiYP38d:b-h2kt0o) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/Documentation/building.rst#L198)):*
> ```reStructuredText
>   UBSAN=1` and :command:`meson -Dubsan=enabled`. This causes Graphene to abort
>   when undefined behavior is detected (and display information about source
>   line).
> ```

I guess we can also build debug version with UBSan. Would be nice to have some info about it.
___
*[LibOS/shim/include/shim_entry.h, line 43 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNvMRF5Sw2bMEATT5m:-MiNvMRF5Sw2bMEATT5n:b-n8w21z) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/shim_entry.h#L43)):*
> ```objc
>  * Used by Graphene's tests.
>  */
> int run_test(const char* test_name);
> ```

Why are `run_test` and `register_library` public interface now? They are always called via `handle_call`.
___
*[LibOS/shim/include/arch/x86_64/shim_entry_api.h, line 22 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNsSZm0t_sOKz0q4Fp:-MiNsSZnApQCSXeE3GP3:b-7r1pjb) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/arch/x86_64/shim_entry_api.h#L22)):*
> ```objc
> 
> /* Custom call numbers */
> #define SHIM_CALL_REGISTER_LIBRARY    1
> ```

enum?
___
*[LibOS/shim/include/arch/x86_64/shim_entry_api.h, line 49 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNoaC08zATOU5i5dAr:-MiNoaC08zATOU5i5dAs:b-jf6k4u) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/arch/x86_64/shim_entry_api.h#L49)):*
> ```objc
> #undef SHIM_STR
> 
> static inline int shim_call(int number, unsigned long arg1, unsigned long arg2, unsigned long arg3,
> ```

Why 4 arguments?
___
*[LibOS/shim/include/arch/x86_64/shim_entry_api.h, line 60 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNtu8LAaJX2qeKQQw5:-MiNtu8LAaJX2qeKQQw6:b-ud4u5m) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/arch/x86_64/shim_entry_api.h#L60)):*
> ```objc
> }
> 
> static inline int shim_register_library(const char* name, unsigned long load_address) {
> ```

Actually, why the public interfaces are called `shim_X` and internal just `X`?
___
*[LibOS/shim/include/arch/x86_64/shim_entry_api.h, line 60 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNuf9PCgRvWD-0uQqU:-MiNuf9PCgRvWD-0uQqV:bplxvfq) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/arch/x86_64/shim_entry_api.h#L60)):*
> ```objc
> }
> 
> static inline int shim_register_library(const char* name, unsigned long load_address) {
> ```

These 2 functions are not arch specific and could be in a non-arch specific header.
___
*[LibOS/shim/src/shim_call.c, line 57 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNxEBB7XSnKtCwszvh:-MiNxEBB7XSnKtCwszvi:bfye6ka) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/src/shim_call.c#L57)):*
> ```C
>         ret = run_test_undefined();
>     } else {
>         ret = -EINVAL;
> ```

Please add a `log_*` call here.

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: 18 of 27 files reviewed, 19 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @woju)

*[a discussion](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMYrMa3k-GzrLdIDG7:-MiQOTAl81dcXGDU4O7w:bvhqqxt) (no related file):*

Out of curiosity: have you tried using UBSan with gcc?
</blockquote></details>

Yes. This branch should compile with both gcc and clang, and they both follow roughly the same API for hooks. Gcc has less checks, though: most of the problems fixed by #2657 were found after I switched to clang.

---
___
*[meson_options.txt, line 10 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMWQpw8MhupyF5PGa0:-MiQP3FY4APAL86fzPna:b-ciy8va) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/meson_options.txt#L10)):*

Not in all of the Graphene (e.g. PAL)?
</blockquote></details>

Right, this is outdated. I removed the "supported in LibOS" part.
___
*[common/src/ubsan.c, line 21 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMZCBdCk1iETybGo4P:-MiQPwnJ84ykdhKUR7Mu:b9r9jvt) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/common/src/ubsan.c#L21)):*

Wouldn't it be better to if-y the makefile/meson? I.e. build this file and link it if `UBSAN` is set, instead of unconditionally compile + link, but sometimes make the source empty?
</blockquote></details>

Can we revisit this when we convert everything to Meson? I'd rather not add another mechanism to builds, and it's a similar case to `DEBUG`: we also have whole modules that are compiled conditionally.
___
*[common/src/ubsan.c, line 31 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMZhRb3X3bCAll5wuP:-MiQSDIU7xTJXUixCoIN:b-2hjufs) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/common/src/ubsan.c#L31)):*

Please use Graphene coding style (ditto for the whole file).
</blockquote></details>

OK. I wanted to match the structures as defined by UBSan but I don't mind changing the naming convention.
___
*[common/src/ubsan.c, line 31 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNeO6s7a1eR7K1yb-g:-MiQTi5G3dedGVc9wBO7:b44jiwj) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L31)):*

Is there really no public API and we have to manually define these?
</blockquote></details>

As far as I know, [this is the only API](https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/ubsan/ubsan_handlers.h) but it's a header file written in C++.
___
*[common/src/ubsan.c, line 42 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNiIRdD8gjQoxEq91X:-MiQU9jT2UqmUsCN91Qi:b-yhwe2n) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L42)):*

https://wiki.osdev.org/Undefined_Behavior_Sanitization shows that this is `uintptr_t`. Which one is bugged?
Update: ok, it has `alignment` (without log), probably different version. How can we be sure that these structs match what's expected by UBSan?
</blockquote></details>

It looks like UBSan is versioned by the function names. The page you linked defines the `type_mismatch` handler, this is for `type_mismatch_v1`.

(We also define `type_mismatch`, but we don't parse the extra info - GCC uses it, Clang doesn't anymore).
___
*[common/src/ubsan.c, line 43 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNiPO4DH5dNU74359_:-MiQUhyi2aHwqNN5BkvB:b-ho5wpv) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L43)):*

Wouldn't `uint8_t` be better?
</blockquote></details>

Agreed. Done.
___
*[common/src/ubsan.c, line 63 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNkQRv7TnDIi5SY7wQ:-MiQUk4mDFOijN_72A67:b-y16l50) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L63)):*
>K&R params?

What do you mean? The `void HANDLER(name) params;`? This is so that the compiler doesn't complain about defining a function that wasn't previously declared. I don't think there's value in declaring these separately.

>Where did you get this macro from?

I wrote it myself.

LLVM doesn't use so many macros, but their handlers are much more custom.
___
*[common/src/ubsan.c, line 66 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNihN9Drw4jc4BWbqT:-MiQVp_Q0CE_6TTirbGv:bj3eoea) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L66)):*

Misaligned `\`.
</blockquote></details>

Fixed, thanks.
___
*[common/src/ubsan.c, line 103 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNl8i69pNXnwvA_jZv:-MiQW9FT6wQ-h0YvrmMm:b-y7q5a3) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L103)):*

What does this mean?
</blockquote></details>

This is used by `-fsanitize=bool` ("Load of a bool value which is neither true nor false") and `-fsanitize=enum` ("Load of a value of an enumerated type which is not in the range of representable values for that enumerated type"). I edited the message.
___
*[common/src/ubsan.c, line 107 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNlAhQBy1AhAetCByO:-MiQWVhjDxpSRBRoTPCu:byfvnqa) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L107)):*

Would be nice to know which value is shifted by which.
</blockquote></details>

It's "shift of LHS by RHS". I edited the message
___
*[common/src/ubsan.c, line 139 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNjrXJ0apZbOC9f-KZ:-MiQWoCB2k_cO2g8ns_g:b-qnuxww) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L139)):*

`1ul`, otherwise if `LogAlignment` is `>= 32` this is undefined.
Update: ok, this can never be that big, but still there is no reason not to change it.
</blockquote></details>

I'll go ahead and do what LLVM does, which is `(uintptr_t)1`.
___
*[common/src/ubsan.c, line 141 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNk3Em-cviYd_v3I_F:-MiQi9EADPIno4McibRE:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L141)):*

ditto
</blockquote></details>

Done.
___
*[Documentation/building.rst, line 198 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNnGe196VF3FiYP38c:-MiQZFarAxVi_dLNuNhC:bihhzgg) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/Documentation/building.rst#L198)):*

I guess we can also build debug version with UBSan. Would be nice to have some info about it.
</blockquote></details>

I added some info, I can't think of a better way to put it.
___
*[LibOS/shim/include/shim_entry.h, line 43 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNvMRF5Sw2bMEATT5m:-MiQ_mcx57D3ONpPl7kL:bep9l2d) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/shim_entry.h#L43)):*

Why are `run_test` and `register_library` public interface now? They are always called via `handle_call`.
</blockquote></details>

Good point. I'm making `run_test` static, and moving the declaration of `register_library` together with other ELF-related functions (it's defined in `shim_rtld.c`).
___
*[LibOS/shim/include/arch/x86_64/shim_entry_api.h, line 22 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNsSZm0t_sOKz0q4Fp:-MiQaSfj9gBB_r7QkjvA:b2nhktp) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/arch/x86_64/shim_entry_api.h#L22)):*

enum?
</blockquote></details>

OK.
___
*[LibOS/shim/include/arch/x86_64/shim_entry_api.h, line 49 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNoaC08zATOU5i5dAr:-MiQaj6aDVZVsuCGB2rs:b-3zleq9) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/arch/x86_64/shim_entry_api.h#L49)):*

Why 4 arguments?
</blockquote></details>

Right, no need to add more until we need them. Changed to 2 args.
___
*[LibOS/shim/include/arch/x86_64/shim_entry_api.h, line 60 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNtu8LAaJX2qeKQQw5:-MiQcPqr7WO_YOZw1iiK:bhkzirw) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/arch/x86_64/shim_entry_api.h#L60)):*

Actually, why the public interfaces are called `shim_X` and internal just `X`?
</blockquote></details>

Because they need to be called something different (we include this file in LibOS as well), and `shim_*` makes it easier to identify them in external code.
___
*[LibOS/shim/include/arch/x86_64/shim_entry_api.h, line 60 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNuf9PCgRvWD-0uQqU:-MiQdyeu9awOGtzVcqCM:bce15wo) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/arch/x86_64/shim_entry_api.h#L60)):*

These 2 functions are not arch specific and could be in a non-arch specific header.
</blockquote></details>

That's right, but it's used (by symlinking) in patches (currently: two versions of glibc, and gcc/libgomp, another patch for glibc in review), so they would all need to symlink two files. I'm not sure if it's worth doing just for this.
___
*[LibOS/shim/src/shim_call.c, line 57 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNxEBB7XSnKtCwszvh:-MiQfgBw6QWr9mInAA0Z:b-896fix) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/src/shim_call.c#L57)):*

Please add a `log_*` call here.
</blockquote></details>

Done.
___
*[.ci/ubuntu18.04.dockerfile, line 10 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMXc6cBnWJ5F6BJ2RQ:-MiQPcZ9E4OtCFP9qr0g:b4spi1v) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/.ci/ubuntu18.04.dockerfile#L10)):*

We seem not to be using clang on any 18.04 pipelines, so maybe do not add it here (for faster CI run times)? Just an idea, not blocking.
</blockquote></details>

OK, I don't mind it either way. Done.

boryspoplawski (changes requested):
> Reviewed 8 of 9 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, 3 unresolved discussions, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz and @woju)

___
*[common/src/ubsan.c, line 21 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMZCBdCk1iETybGo4P:-MiRMMJ51VXGohy_GhP3:b88fnog) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/common/src/ubsan.c#L21)):*

Can we revisit this when we convert everything to Meson? I'd rather not add another mechanism to builds, and it's a similar case to `DEBUG`: we also have whole modules that are compiled conditionally.
</blockquote></details>

We can, but we probably won't. How about:
```
diff --git a/common/src/Makefile b/common/src/Makefile
index 231f5921..8f6f445f 100644
--- a/common/src/Makefile
+++ b/common/src/Makefile
@@ -112,8 +112,11 @@ objs += \
        string/strstr.o \
        string/toml_utils.o \
        string/utils.o \
-       toml.o \
-       ubsan.o
+       toml.o
+
+ifeq ($(UBSAN),1)
+objs += ubsan.o
+endif
 
 $(addprefix $(target),crypto/adapters/mbedtls_adapter.o): crypto/mbedtls/library/aes.c
 
```
Or either force @woju to make a written statement here that he will fix it when rewriting PAL SGX to meson :)
___
*[common/src/ubsan.c, line 31 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNeO6s7a1eR7K1yb-g:-MiRVAiT6olrJbhpR9Cv:b32w6w) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L31)):*

As far as I know, [this is the only API](https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/ubsan/ubsan_handlers.h) but it's a header file written in C++.
</blockquote></details>

Uh
___
*[common/src/ubsan.c, line 63 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNkQRv7TnDIi5SY7wQ:-MiRYhGG-rlGesvO9t59:b-o1hjiu) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/common/src/ubsan.c#L63)):*

>K&R params?

What do you mean? The `void HANDLER(name) params;`? This is so that the compiler doesn't complain about defining a function that wasn't previously declared. I don't think there's value in declaring these separately.

>Where did you get this macro from?

I wrote it myself.

LLVM doesn't use so many macros, but their handlers are much more custom.
</blockquote></details>

I missed `(` in macro invocation.
This looks ugly and is hard to follow, but I don't have idea other than de-macroing these 4 `UBSAN_SIMPLE_HANDLER_x` defines, so unblocking.
___
*[LibOS/shim/include/arch/x86_64/shim_entry_api.h, line 60 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNtu8LAaJX2qeKQQw5:-MiRXRzy1_yu6ey5mZhz:b-2wvtnz) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/arch/x86_64/shim_entry_api.h#L60)):*

Because they need to be called something different (we include this file in LibOS as well), and `shim_*` makes it easier to identify them in external code.
</blockquote></details>

Then maybe name external ones  like this `graphene_register_library`? `shim_` is not a good prefix, it's not Graphene specific and previously we used it only internally.

dimakuv (approved):
> Reviewed 12 of 26 files at r1, 1 of 2 files at r2, 1 of 2 files at r5, 3 of 3 files at r6, 7 of 9 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, 3 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz and @woju)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @woju)

___
*[common/src/ubsan.c, line 21 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMZCBdCk1iETybGo4P:-MiRmIgMCbnoL_cvcBdu:b-f3hta7) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/common/src/ubsan.c#L21)):*

We can, but we probably won't. How about:
```
diff --git a/common/src/Makefile b/common/src/Makefile
index 231f5921..8f6f445f 100644
--- a/common/src/Makefile
+++ b/common/src/Makefile
@@ -112,8 +112,11 @@ objs += \
        string/strstr.o \
        string/toml_utils.o \
        string/utils.o \
-       toml.o \
-       ubsan.o
+       toml.o
+
+ifeq ($(UBSAN),1)
+objs += ubsan.o
+endif
 
 $(addprefix $(target),crypto/adapters/mbedtls_adapter.o): crypto/mbedtls/library/aes.c
 
```
Or either force @woju to make a written statement here that he will fix it when rewriting PAL SGX to meson :)
</blockquote></details>

Okay. Done.
___
*[LibOS/shim/include/arch/x86_64/shim_entry_api.h, line 60 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNtu8LAaJX2qeKQQw5:-MiRoMqj6q1yGNP0VI4x:b-bq8277) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/arch/x86_64/shim_entry_api.h#L60)):*

Then maybe name external ones  like this `graphene_register_library`? `shim_` is not a good prefix, it's not Graphene specific and previously we used it only internally.
</blockquote></details>

We used `shim_register_library` as an external function for some time now, and included `<shim_*.h>` files in patched code since the beginning of this project.

I would rather not rename it all (the header file, functions, constants) to `graphene_*` since the whole project is changing name, so the names will become _more_ obsolete than if I didn't touch them at all.

How about I create a ticket saying "Rename external APIs to `gramine_*`" to be done after we finish the move? We could also rename `SYSCALLDB` to something like `GRAMINE_SYSCALL`.

(I also considered `libos_*` but it's also pretty generic).

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski and @woju)

___
*[.ci/linux-direct-sanitizers.jenkinsfile, line 1 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMQ1YED-q0G76jWj5Y:-MiRra_oCg1BlQvSulmT:bqem481) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/.ci/linux-direct-sanitizers.jenkinsfile#L1)):*

Please ping me when you'd want this pipeline added to Jenkins.
</blockquote></details>

@woju Can you do it now?

(There were no big problems during reviews so far, and [I triggered a separate build](https://github.com/oscarlab/graphene/pull/2653) to make sure Jenkins will pass on this branch).

boryspoplawski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, 2 unresolved discussions, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz and @woju)

___
*[common/src/ubsan.c, line 21 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMZCBdCk1iETybGo4P:-MiRs1up0NNlocxR1Wuw:bfhqmln) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/common/src/ubsan.c#L21)):*

Okay. Done.
</blockquote></details>

Not done?
___
*[LibOS/shim/include/arch/x86_64/shim_entry_api.h, line 60 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNtu8LAaJX2qeKQQw5:-MiRsCy6D-uCTm1HrI64:bxwo3vf) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/arch/x86_64/shim_entry_api.h#L60)):*
> We used shim_register_library as an external function for some time now

Yeah, I've missed this before.
> and included <shim_*.h> files in patched code since the beginning of this project.

Filenames are separate issue.

Ok, please create such an issue, but using `Graphene` name.

boryspoplawski (commented):
> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski)

___
*[common/src/ubsan.c, line 21 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMZCBdCk1iETybGo4P:-MiS7I8m5A1nvHHyx28R:bf06eg1) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/common/src/ubsan.c#L21)):*

Not done?
</blockquote></details>

Sorry, didn't push before.
___
*[LibOS/shim/include/arch/x86_64/shim_entry_api.h, line 60 at r6](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiNtu8LAaJX2qeKQQw5:-MiS8tg1BExQbFH3VT7s:b-sy8egh) ([raw file](https://github.com/oscarlab/graphene/blob/035bedcf55fb58241ebf66e0d73e5d201215511c/LibOS/shim/include/arch/x86_64/shim_entry_api.h#L60)):*

> We used shim_register_library as an external function for some time now

Yeah, I've missed this before.
> and included <shim_*.h> files in patched code since the beginning of this project.

Filenames are separate issue.

Ok, please create such an issue, but using `Graphene` name.
</blockquote></details>

Ticketed in #2669.

boryspoplawski (dismissed):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

mkow (changes requested):
> Reviewed 12 of 26 files at r1, 1 of 2 files at r2, 2 of 2 files at r5, 3 of 3 files at r6, 6 of 9 files at r7, 1 of 1 files at r8, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, 1 unresolved discussion, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

___
*[common/src/ubsan.c, line 110 at r9](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiSOV-IB4bLk-yCV8K7:-MiSOV-JEsczmAf1prkt:boo7db7) ([raw file](https://github.com/oscarlab/graphene/blob/5e2f955ca0284d857ac806f0a875891c1b82c1ec/common/src/ubsan.c#L110)):*
> ```
> bound bound
> ```

I guess this wasn't intended? ;)

dimakuv (dismissed):
> Reviewed 1 of 26 files at r1, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, 1 unresolved discussion, "fixup! " found in commit messages' one-liners (waiting on @pwmarcz)

pwmarcz (commented):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: 26 of 27 files reviewed, 1 unresolved discussion, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners (waiting on @boryspoplawski, @dimakuv, and @mkow)

___
*[common/src/ubsan.c, line 110 at r9](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiSOV-IB4bLk-yCV8K7:-MiWPPff1VpIsVDKqzY7:b7wmcfz) ([raw file](https://github.com/oscarlab/graphene/blob/5e2f955ca0284d857ac806f0a875891c1b82c1ec/common/src/ubsan.c#L110)):*

> ```
> bound bound
> ```

I guess this wasn't intended? ;)
</blockquote></details>

These tend to be hard to notice :) Thanks, fixed.

mkow (dismissed):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[common/src/ubsan.c, line 110 at r9](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiSOV-IB4bLk-yCV8K7:-MiWY_FO4meojtOZLe9w:b32vhw) ([raw file](https://github.com/oscarlab/graphene/blob/5e2f955ca0284d857ac806f0a875891c1b82c1ec/common/src/ubsan.c#L110)):*

These tend to be hard to notice :) Thanks, fixed.
</blockquote></details>

:)

woju:
Jenkins, test Jenkins-Direct-Sanitizers, please

woju:
Jenkins, test Jenkins-SGX-Sanitizers, please

woju:
Jenkins, test Jenkins-SGX-Sanitizers please

woju:
Jenkins, test Jenkins-Direct-Sanitizers please

woju:
Jenkins, test this please

woju:
Jenkins, test Jenkins-SGX-Sanitizers please

woju:
Jenkins, test this please

woju (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL), "fixup! " found in commit messages' one-liners

___
*[.ci/linux-direct-sanitizers.jenkinsfile, line 1 at r2](https://reviewable.io/reviews/oscarlab/graphene/2662#-MiMQ1YED-q0G76jWj5Y:-MiWoiCl13L2CATow9IZ:b-2jasg7) ([raw file](https://github.com/oscarlab/graphene/blob/25f88b64cb76c196576360485f0f975105632bd1/.ci/linux-direct-sanitizers.jenkinsfile#L1)):*

@woju Can you do it now?

(There were no big problems during reviews so far, and [I triggered a separate build](https://github.com/oscarlab/graphene/pull/2653) to make sure Jenkins will pass on this branch).
</blockquote></details>

Done. I triggered the final retest, if passes, it'll be OK. Sorry for the noise.

pwmarcz (commented):
I rebased the code but there seem to be problems after Jenkins got upgraded to newer version (?)

> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662#-:-MiWvLzu5xaxJG4Wa44Z:bne4m7r)* status: 23 of 27 files reviewed, all discussions resolved, not enough approvals from maintainers (1 more required), not enough approvals from different teams (1 more required, approved so far: ITL) (waiting on @boryspoplawski, @dimakuv, @mkow, and @woju)

mkow (approved):
> Reviewed 4 of 4 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: all files reviewed, all discussions resolved, not enough approvals from different teams (1 more required, approved so far: ITL)

dimakuv (approved):
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: :shipit: complete! all files reviewed, all discussions resolved

woju:
Jenkins, test Jenkins-18.04 please

woju (approved):
> Reviewed 4 of 4 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/oscarlab/graphene/2662)* status: :shipit: complete! all files reviewed, all discussions resolved

